PR #60, opened 2025-03-05
feat(react): support lynx ssr

hzy:
In this PR, we add support for Lynx SSR by adding two call-by-native api `ssrEncode` and `ssrHydrate`.

## Checklist

### DIFF
--- a/packages/react/runtime/src/lynx/calledByNative.ts
+++ b/packages/react/runtime/src/lynx/calledByNative.ts
@@ -38,7 +38,7 @@ function ssrEncode() {
 function ssrHydrate(info: string) {
   const nativePage = __GetPageElement();
   if (!nativePage) {
-    throw 'SSR Hydration Failed! Please check if the SSR content loaded successfully!';
+    throw new Error('SSR Hydration Failed! Please check if the SSR content loaded successfully!');
   }
 
   const refsMap = __GetTemplateParts(nativePage);
--- a/packages/react/runtime/src/lynx/calledByNative.ts
+++ b/packages/react/runtime/src/lynx/calledByNative.ts
@@ -13,8 +13,51 @@ import { renderMainThread } from '../lifecycle/render.js';
 import { hydrate } from '../hydrate.js';
 import { markTiming, PerformanceTimingKeys, setPipeline } from './performance.js';
 import { __pendingListUpdates } from '../list.js';
+import { ssrHydrateByOpcodes } from '../opcodes.js';
+
+function ssrEncode() {
+  const { __opcodes } = __root;
+  delete __root.__opcodes;
+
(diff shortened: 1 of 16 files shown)


packages/react/runtime/src/lynx/calledByNative.ts
```diff
@@ -23,9 +66,13 @@ function injectCalledByNative(): void {
       return null;
     },
     removeComponents: function(): void {},
+    ...(__ENABLE_SSR__ ? { ssrEncode, ssrHydrate } : {}),
   };
 
   Object.assign(globalThis, calledByNative);
+  Object.assign(globalThis, {
```
Yradex, line 73: Is it possible to assign this only when `__FIRST_SCREEN_SYNC_TIMING__ == 'jsReady'`?

packages/react/runtime/src/lynx/calledByNative.ts
```diff
@@ -13,8 +13,51 @@ import { renderMainThread } from '../lifecycle/render.js';
...
+  __root_values && __root.setAttribute('values', __root_values);
+  ssrHydrateByOpcodes(__opcodes, __root as SnapshotInstance, refsMap);
+
+  (__root as SnapshotInstance).__elements = [nativePage];
+  (__root as SnapshotInstance).__element_root = nativePage;
+}
 
 function injectCalledByNative(): void {
+  if (process.env['NODE_ENV'] !== 'test') {
+    if (__FIRST_SCREEN_SYNC_TIMING__ !== 'jsReady' && __ENABLE_SSR__) {
```
Yradex, line 56: Is it possible to check this at compile time?

packages/react/runtime/src/lynx/calledByNative.ts
```diff
@@ -13,8 +13,51 @@ import { renderMainThread } from '../lifecycle/render.js';
...
+    return JSON.stringify({ __opcodes, __root_values: __root.__values });
+  } finally {
+    SnapshotInstance.prototype.toJSON = oldToJSON;
+  }
+}
+
+function ssrHydrate(info: string) {
+  const nativePage = __GetPageElement();
+  if (!nativePage) {
+    throw 'SSR Hydration Failed! Please check if the SSR content loaded successfully!';
```
colinaaa: I'm not convinced that throwing an error is the best approach. While we lack a call stack here (as `ssrHydrate` is invoked natively), there might still be scenarios where someone expects an error to be thrown.

colinaaa:
```suggestion
    throw new Error('SSR Hydration Failed! Please check if the SSR content loaded successfully!');
```

colinaaa (approved):

Yradex (approved):

PR #64, opened 2025-03-05
feat(testing-library): add @lynx-js/lynx-dom

upupming:
## Checklist

### DIFF
--- a/packages/testing-library/lynx-dom/src/index.ts
+++ b/packages/testing-library/lynx-dom/src/index.ts
@@ -1,0 +1,399 @@
+import EventEmitter from 'events';
+import { createGlobalThis, LynxGlobalThis } from './lynx/GlobalThis';
+import { initElementTree } from './lynx/nativeMethod';
+import { VirtualConsole } from './lynx/virtual-console';
+export type { LynxFiberElement } from './lynx/nativeMethod';
+
+type ElementTree = ReturnType<typeof initElementTree>;
+type FilterUnderscoreKeys<T> = {
+  [K in keyof T]: K extends `__${string}` ? K : never;
+}[keyof T];
+type PickUnderscoreKeys<T> = Pick<T, FilterUnderscoreKeys<T>>;
+type ElementTreeGlobals = PickUnderscoreKeys<ElementTree>;
+
+declare global {
+  var lynxDOM: LynxDOM;
+  var elementTree: ElementTree;
+  var __JS__: boolean;
+  var __LEPUS__: boolean;
+
+  namespace lynxCoreInject {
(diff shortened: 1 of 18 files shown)


packages/testing-library/lynx-dom/src/index.ts
```diff
@@ -0,0 +1,401 @@
...
+  const JsContext = ee;
+
+  function __LoadLepusChunk(
+    chunkName: string,
+    options,
+  ) {
+    if (process.env.DEBUG) {
+      console.log('__LoadLepusChunk', chunkName, options);
+    }
+    debugger;
```
colinaaa: Is this `debugger` intended to be added?

packages/testing-library/lynx-dom/src/index.ts
```diff
@@ -0,0 +1,401 @@
...
+
+  target.console.profile = () => {};
+  target.console.profileEnd = () => {};
+
+  target.__LoadLepusChunk = __LoadLepusChunk;
+
+  globalThis.onInjectMainThreadGlobals?.(target);
+}
+
+const BLACKLIST_GLOBALS = [
```
colinaaa: Please use `FORBIDDEN_LIST_GLOBALS` or `IGNORE_LIST_GLOBALS`.

packages/testing-library/lynx-dom/src/index.ts
```diff
@@ -0,0 +1,393 @@
...
+  };
+
+  target.__LoadLepusChunk = __LoadLepusChunk;
+
+  globalThis.onInjectBackgroundThreadGlobals?.(target);
+}
+
+export type ThreadType = 'background' | 'main';
+
+export class LynxDOM {
```
Huxpro, line 323: Let's rename this to LynxToplevel or something. We shouldn't confuse our setup with Web DOM furthermore.

Additionally, I encourage you to add more documentation-comments whenever you declared a new concept. 

I wouldn't even concern about you use AI to generate such doc comments as a starting point, so you can understand how AI read through your code.

upupming: Good catch, since it is used by vitest/jest to setup their environment:

- https://jestjs.io/docs/configuration#testenvironment-string
- https://vitest.dev/guide/environment

I would like to name it `LynxRuntime` (just like `edge-runtime`) or `LynxEnvorinment` (means can be used as environment), let me use `LynxRuntime` first, please comment it you have any suggestions~

upupming:
After using JSDOM to implement element API, we reduced the PR size, please review https://github.com/lynx-family/lynx-stack/pull/74 directly.

PR #74, opened 2025-03-05
feat(testing-library): add react lynx testing library

upupming:
## Summary

In this PR, we implemented the ReactLynx Testing Library, a simple and complete ReactLynx unit testing library that encourages good testing practices. We are hugely inspired by [react-testing-library](https://github.com/testing-library/react-testing-library).

Example usage:

```js
import '@testing-library/jest-dom';
import { test, expect } from 'vitest';
import { render } from '@lynx-js/react/testing-library';

test('renders options.wrapper around node', async () => {
  const WrapperComponent = ({ children }) => (
    <view data-testid='wrapper'>{children}</view>
  );
  const Comp = () => {
    return <view data-testid='inner' style='background-color: yellow;' />;
  };
  const { container, getByTestId } = render(<Comp />, {
    wrapper: WrapperComponent,
  });
  expect(getByTestId('wrapper')).toBeInTheDocument();
  expect(container.firstChild).toMatchInlineSnapshot(`
    <view
      data-testid="wrapper"
    >
      <view
        data-testid="inner"
        style="background-color: yellow;"
      />
    </view>
  `);
});
```

## Checklist

- [x] Tests updated.
- [x] Documentation updated. See also https://github.com/lynx-family/lynx-website/pull/117


BronzonTech-Cloud (approved):
Great 😃 , I was also facing the await issues ! 
Thanks bro 🤝🏽
It remains some linter error but it's great

upupming:
I think all problems we discussed before are resolved, and the PR is ready to be reviewed!

packages/testing-library/lynx-environment/src/lynx/ElementAPI.ts
```diff
@@ -0,0 +1,382 @@
...
+      this.root = page;
+      lynxEnv.jsdom.window.document.body.appendChild(page);
+      return page;
+    }
+
+    __CreateRawText(text: string): LynxElement {
+      const element = lynxEnv.jsdom.window.document.createTextNode(
+        text,
+      ) as unknown as LynxElement;
+      element.$$uiSign = uiSignNext++;
```
hzy: Seems to be duplicate with `__CreateElement` ?

packages/testing-library/react-lynx-testing-library/src/__tests__/events.test.jsx
```diff
@@ -0,0 +1,165 @@
...
+              },
+            }
+          `);
+          expect(ref.current.constructor.name).toMatchInlineSnapshot(
+            `"NodesRef"`,
+          );
+          const element = __GetElementByUniqueId(
+            Number(ref.current._nodeSelectToken.identifier),
+          );
+          expect(element).toMatchInlineSnapshot(`<view />`);
```
hzy: Seems `ref` and `event` are completely omitted in snapshot?

Can we somehow make it in snapshot? For easier testing...

upupming: One can simply use `element.props` to do asserts on `ref` and `event`, let me write a document for this:

```js
expect(element).toMatchInlineSnapshot(`<view />`);
expect(element.props).toMatchInlineSnapshot(`
  {
    "event": {
      "bindEvent:tap": [Function],
    },
    "has-react-ref": true,
  }
`)
```

upupming: @hzy Added examples in our docs: https://github.com/lynx-family/lynx-website/pull/117

pnpm-workspace.yaml
```diff
@@ -11,6 +11,10 @@ packages:
   - packages/web-platform/*
   - packages/webpack/*
   - packages/third-party/*
+  - packages/testing-library/*
+  - packages/testing-library/examples/*
+  - "!packages/testing-library/examples"
+  - .github/actions/eco-ci
```
colinaaa: This seems duplicated.

.dprint.jsonc
```diff
@@ -55,6 +55,8 @@
     "**/expected/**",
     "**/rspack-expected/**",
     "packages/**/test/**/hotCases/**",
+
+    "packages/testing-library/react-lynx-testing-library/tsconfig.json",
```
colinaaa: Why do we need this? (instead of just format it)

packages/testing-library/react-lynx-testing-library/package.json
```diff
@@ -0,0 +1,59 @@
+{
+  "name": "@lynx-js/react-lynx-testing-library",
+  "version": "0.0.0",
+  "description": "",
+  "keywords": [],
+  "license": "ISC",
```
colinaaa:
```suggestion
  "license": "Apache-2.0",
```

packages/testing-library/lynx-environment/src/lynx/ElementPAPI.ts
```diff
@@ -0,0 +1,470 @@
...
+    __UpdateListActions(
+      _list: LynxElement,
+      _removals: number[],
+      _insertions: number[],
+      _moveFrom: number[],
+      _moveTo: number[],
+      _updateFrom: number[],
+      _updateTo: number[],
+    ) {}
+
```
hzy: I think this can be removed?

This `__UpdateListActions` is already removed in Lynx Engine.

packages/react/testing-library/tsup.config.ts
```diff
@@ -0,0 +1,27 @@
+import { Options } from 'tsup';
+import { fixImportsPlugin } from 'esbuild-fix-imports-plugin';
+
+const options: Options = {
```
colinaaa: Why are we having a `tsup.config.ts` without `scripts.build` in `package.json`?

packages/react/testing-library/src/vitest.config.js
```diff
@@ -0,0 +1,100 @@
...
+    plugins: [
+      transformReactLynxPlugin(),
+      react({ jsxImportSource: runtimePkgName }),
+    ],
+    test: {
+      environment: require.resolve(
+        '@lynx-js/lynx-environment/env/vitest',
+      ),
+      globals: true,
+      setupFiles: [path.posix.join(__dirname, 'vitest-global-setup')],
```
colinaaa: I don't think this would work on Windows. Since `__dirname` would contain `C:\path\to\node_modules`

packages/react/testing-library/README.md
```diff
@@ -0,0 +1,76 @@
...
+import { createVitestConfig } from '@lynx-js/react/testing-library/vitest-config';
+
+const defaultConfig = createVitestConfig();
+const config = defineConfig({
+  test: {
+    // ...
+  },
+});
+
+export default defineConfig(mergeConfig(defaultConfig, config));
```
colinaaa: The `defineConfig` here seems useless
```suggestion
export default mergeConfig(defaultConfig, config);
```

packages/react/testing-library/src/__tests__/act.test.jsx
```diff
@@ -0,0 +1,290 @@
+import '@testing-library/jest-dom';
```
colinaaa, line 1: We may need a `.npmignore` to avoid publishing files in `__tests__`

packages/react/package.json
```diff
@@ -151,17 +164,33 @@
...
     "@lynx-js/types": "^3.2.1",
     "@microsoft/api-extractor": "catalog:",
-    "@types/react": "^18.3.20"
+    "@testing-library/dom": "^10.4.0",
+    "@testing-library/jest-dom": "^6.6.3",
+    "@types/react": "^18.3.20",
+    "@vitejs/plugin-react": "4.3.4"
   },
   "peerDependencies": {
+    "@lynx-js/lynx-environment": "*",
```
colinaaa: I don't think it's correct to add these packages to `peerDependencies`. This would require to manually install all these packages.

I think we could either:

1. Install them as `dependencies`
2. Bundle the testing library (i.e.: with Rslib)

packages/testing-library/lynx-environment/package.json
```diff
@@ -0,0 +1,40 @@
+{
+  "name": "@lynx-js/lynx-environment",
```
colinaaa: I’m not a fan of this package name. It doesn’t sound like a testing package, and having two "lynx" feels redundant. What about `@lynx-js/jest-environment` or `@lynx-js/vitest-environment` instead?

packages/testing-library/lynx-environment/rslib.config.ts
```diff
@@ -0,0 +1,28 @@
...
+      dts: true,
+    },
+    {
+      bundle: false,
+      format: 'cjs',
+      syntax: 'es2021',
+    },
+  ],
+  output: {
+    sourceMap: true,
```
colinaaa: Why do we need sourceMap?

packages/testing-library/lynx-environment/package.json
upupming: Let me call it `@lynx-js/test-environment` now, it only supports `vitest` now, but we will support jest too in the future.

packages/react/runtime/src/lynx/tt.ts
```diff
@@ -1,72 +1,29 @@
...
 import { reloadBackground } from '../lifecycle/reload.js';
 import { renderBackground } from '../lifecycle/render.js';
-import { CHILDREN, COMPONENT, DIFF, DIFFED, FORCE } from '../renderToOpcodes/constants.js';
+import { CHILDREN } from '../renderToOpcodes/constants.js';
 import { __root } from '../root.js';
 import { globalRefsToSet, updateBackgroundRefs } from '../snapshot/ref.js';
 import { backgroundSnapshotInstanceManager } from '../snapshot.js';
 import { destroyWorklet } from '../worklet/destroy.js';
-
-export function runWithForce(cb: () => void): void {
```
upupming, line 22: We extract this to a single file because it will cause a circular reference in the vitest test.

packages/react/runtime/src/lynx/tt.ts
```diff
@@ -110,7 +67,7 @@ function onLifecycleEvent([type, data]: [string, any]) {
   }
 }
 
-async function onLifecycleEventImpl(type: string, data: any): Promise<void> {
+function onLifecycleEventImpl(type: string, data: any): void {
```
upupming, line 70: We reduce an extra async function here to avoid using waitSchedule in test.

packages/react/package.json
upupming: We prebundled `@lynx-js/test-environment` and `@testing-library/dom`

And we use `VitestPackageInstaller` to ensure user installs jsdom when run testing

packages/react/testing-library/src/pure.jsx
```diff
@@ -0,0 +1,128 @@
...
+      flushDelayedLifecycleEvents();
+    });
+  }
+
+  return {
+    container: lynxEnv.mainThread.elementTree.root,
+    unmount: cleanup,
+    rerender: (rerenderUi) => {
+      lynxEnv.resetLynxEnv();
+      return render(wrapUiIfNeeded(rerenderUi), {});
```
hzy: should the `rerender` inherits the options of its `render`?
```suggestion
      return render(wrapUiIfNeeded(rerenderUi), {
        {
          queries,
          wrapper: WrapperComponent,
          enableMainThread,
          enableBackgroundThread,
        }
      });
```

packages/react/testing-library/src/fire-event.ts
```diff
@@ -0,0 +1,184 @@
+// @ts-nocheck
+import { fireEvent as domFireEvent, createEvent } from '@testing-library/dom';
+
+function getElement(elemOrNodesRef) {
+  if (elemOrNodesRef?.constructor?.name === 'NodesRef') {
```
hzy: can this be replaced with `instanceof`?

packages/react/testing-library/src/fire-event.ts
```diff
@@ -0,0 +1,184 @@
...
+  },
+  scrolltoloweredge: {
+    defaultInit: {},
+  },
+  scrolltonormalstate: {
+    defaultInit: {},
+  },
+};
+
+Object.keys(eventMap).forEach((key) => {
```
hzy, line 152: Question: what if an event is not in this list?

packages/react/testing-library/loaders/jsx-loader.js
```diff
@@ -0,0 +1,49 @@
...
+    worklet: {
+      filename: relativePath,
+      runtimePkg: `${runtimePkgName}/internal`,
+      target: 'MIXED',
+    },
+    refresh: false,
+    cssScope: false,
+  });
+
+  if (result.errors.length > 0) {
```
hzy, line 43: should we output the warnings rather than just discard it?

packages/react/testing-library/src/__tests__/ref.test.jsx
```diff
@@ -0,0 +1,35 @@
+import { createRef, Component } from '@lynx-js/react';
+import { render } from '..';
+import { expect } from 'vitest';
+
+describe('component ref', () => {
```
hzy, line 6: Should we add a test for ref of element like `<view/>`?

packages/react/testing-library/src/pure.jsx
```diff
@@ -0,0 +1,128 @@
...
+    ...getQueriesForElement(lynxEnv.mainThread.elementTree.root, queries),
+  };
+}
+
+export function cleanup() {
+  const isMainThread = !__MAIN_THREAD__;
+
+  // Ensure componentWillUnmount is called
+  globalThis.lynxEnv.switchToBackgroundThread();
+  preactRender(null, __root);
```
hzy, line 88: For now the ref of element is triggered from main-thread.

packages/react/testing-library/src/__tests__/ref.test.jsx
```diff
@@ -0,0 +1,35 @@
...
+        return this.props.show && (
+          <view>
+            <Child ref={ref1} />
+            <Child ref={ref2} />
+          </view>
+        );
+      }
+    }
+
+    render(<Comp show />);
```
hzy: Should we add a test for unref?

packages/react/testing-library/src/fire-event.ts
upupming: For example a event named `catchcustomevent={handler}`. User can still use `fireEvent(element, event)` to trigger it, but no `fireEvent.customevent()` will be available like `fireEvent.tap`.

I have added a test case in `packages/react/testing-library/src/__tests__/events.test.jsx`.

packages/react/testing-library/src/__tests__/ref.test.jsx
```diff
@@ -0,0 +1,363 @@
...
+      NodesRef {
+        "_nodeSelectToken": {
+          "identifier": "3",
+          "type": 2,
+        },
+        "_selectorQuery": {},
+      }
+    `);
+    ref1.mockClear();
+    ref2.current = null;
```
hzy: Maybe this should done by framework?
```suggestion
```

hzy (approved):

colinaaa (approved):

PR #153, opened 2025-03-10
fix(webpack/react): shake `useImperativeHandle`

colinaaa:
## Summary

We currently only have Component instance on the background thread.

So we can safely remove the callback of `useImperativeHandle` on the main-thread.

## Checklist


Huxpro (approved):
Pre-stamped! Don't forget to mention this addition in the <https://lynxjs.org/react/thinking-in-reactlynx.html#rule-1>. We can categorize this with `ref` as they are basically siblings.

PR #173, opened 2025-03-11
feat(web): add onNapiLoaderCall to mock value handler in lynx-core

Sherry-hue:
## Summary

feat: add two prop of lynx-view about `napiLoader`:

- `napiModulesMap`: [optional] the napiModule which is called in lynx-core. key is module-name, value is esm url.

- `onNapiModulesCall`: [optional] the NapiModule value handler.

**Warning:** This is the internal implementation of `@lynx-js/lynx-core`. In most cases, this API is not required for projects.

1. The `napiModulesMap` value should be a esm url which export default a function with two parameters:

- `NapiModules`: oriented `napiModulesMap`, which you can use to call other Napi-Modules

- `NapiModulesCall`: trigger `onNapiModulesCall`

example:

```js
const color_environment = URL.createObjectURL(
  new Blob(
    [`export default function(NapiModules, NapiModulesCall) {
  return {
    getColor() {
      NapiModules.color_methods.getColor({ color: 'green' }, color => {
        console.log(color);
      });
    },
    ColorEngine: class ColorEngine {
      getColor(name) {
        NapiModules.color_methods.getColor({ color: 'green' }, color => {
          console.log(color);
        });
      }
    },
  };
};`],
    { type: 'text/javascript' },
  ),
);

const color_methods = URL.createObjectURL(
  new Blob(
    [`export default function(NapiModules, NapiModulesCall) {
  return {
    async getColor(data, callback) {
      const color = await NapiModulesCall('getColor', data);
      callback(color);
    },
  };
};`],
    { type: 'text/javascript' },
  ),
);

lynxView.napiModuleMap = {
  'color_environment': color_environment,
  'color_methods': color_methods,
};
```

2. The `onNapiModulesCall` function has three parameters:

- `name`: the first parameter of `NapiModulesCall`, the function name
- `data`: the second parameter of `NapiModulesCall`, data
- `moduleName`: the module-name of the called napi-module

```js
lynxView.onNapiModulesCall = (name, data, moduleName) => {
  if (name === 'getColor' && moduleName === 'color_methods') {
    return data.color;
  }
};
```

## Checklist


packages/web-platform/web-constants/src/types/MainThreadStartConfigs.ts
```diff
@@ -13,4 +14,5 @@ export interface MainThreadStartConfigs {
   entryId: string;
   browserConfig: BrowserConfig;
   nativeModulesUrl?: string;
+  napiLoaderCall?: NapiLoaderCall;
```
PupilTong: napiModuleMap?

packages/web-platform/web-worker-runtime/src/backgroundThread/background-apis/createNapiLoader.ts
```diff
@@ -0,0 +1,56 @@
...
+
+import { type NapiLoaderCall, type NativeApp } from '@lynx-js/web-constants';
+
+export const createNapiLoader = async (
+  napiLoaderUrl: NapiLoaderCall,
+  nativeApp: NativeApp,
+) => {
+  const napiLoaderCall: Record<string, any> = {};
+  for (
+    const [moduleName, moduleString] of Object.entries(
```
PupilTong: The name is really weird.
Type `NapiLoaderCall` is a set of urls for napi modules.
`napiLoaderUrl` is singular but it's a set.

packages/web-platform/web-core/src/uiThread/startUIThread.ts
```diff
@@ -39,6 +40,7 @@ export function startUIThread(
   },
   overrideTagMap: Record<string, string> = {},
   nativeModulesUrl: string | undefined,
+  napiLoaderCall?: NapiLoaderCall,
```
PupilTong: remove the `?` and fill a default value by createLynxView?

packages/web-platform/web-worker-runtime/src/backgroundThread/background-apis/startBackgroundThread.ts
```diff
@@ -38,6 +39,10 @@ export function startBackgroundThread(
         markTimingInternal,
         customNativeModules,
       });
+
+      config.napiLoaderCall
+        && await createNapiLoader(config.napiLoaderCall, nativeApp);
```
PupilTong: const napiLoader = await createNapiLoader(config.napiLoaderCall, nativeApp);

then assign it to globalThis

packages/web-platform/web-worker-runtime/src/backgroundThread/background-apis/createNapiLoader.ts
```diff
@@ -0,0 +1,56 @@
...
+    nativeApp.nativeModuleProxy,
+  );
+  (globalThis as any)['napiLoaderOnRT' + nativeApp.id] = {
+    load(moduleName: string) {
+      return napiLoaderCall[moduleName];
+    },
+  };
+};
+
+function recursiveFunctionCallBinder(
```
PupilTong: This could make it hard to understand the difference of NativeModule and napi module.

Could you please create a new callback to ui thread instead of using the NativeModulesCall?

Also, We should allow napi modules to pass https://developer.mozilla.org/en-US/docs/Web/API/Web_Workers_API/Transferable_objects cross threads.

Sherry-hue: add napiModuleMap, onNapiModulesCall to distinguish NativeModules.

I will create a new PR to implement the napiModule use transfer

packages/web-platform/web-core/src/apis/LynxView.ts
```diff
@@ -181,6 +185,30 @@ export class LynxView extends HTMLElement {
...
+    this.#napiModuleMap = map;
+  }
+
+  #onNapiModulesCall?: NapiModulesCall;
+  /**
+   * @param
+   * @property
+   */
+  get onNapiModulesCall(): NapiModulesCall | undefined {
+    return this.#onNativeModulesCall;
```
PupilTong: onNativeModulesCall?

PupilTong (approved):

PR #174, opened 2025-03-11
feat(rspeedy): support lazyCompilation

gaoachao:
## Summary

implement #7

## Checklist

### DIFF
--- a/packages/rspeedy/core/src/config/rsbuild/lazyCompilation.ts
+++ b/packages/rspeedy/core/src/config/rsbuild/lazyCompilation.ts
@@ -1,0 +1,55 @@
+// Copyright 2024 The Lynx Authors. All rights reserved.
+// Licensed under the Apache License Version 2.0 that can be found in the
+// LICENSE file in the root directory of this source tree.
+import { createRequire } from 'node:module'
+import { fileURLToPath } from 'node:url'
+
+import type { Rspack } from '@rsbuild/core'
+
+import { getIp } from '../../utils/getIp.js'
+import type { LazyCompilationOptions } from '../dev/lazyCompilation.js'
+
+const __filename = fileURLToPath(import.meta.url)
+const require = createRequire(__filename)
+
+export function toRsbuildLazyCompilation(
+  options: boolean | LazyCompilationOptions | undefined,
+  hostname: string | undefined,
+):
+  | boolean
+  | Rspack.LazyCompilationOptions
+{
+  if (!options) return false
+
+  const defaultOptions: Rspack.LazyCompilationOptions = {
+    test: module => {
+      const HMRModules = [
+        '@rspack/core/hot/dev-server.js',
+        'webpack-dev-transport/lib/client/index.js',
+      ]
+
+      const isHMRImports = HMRModules.some(m =>
+        module?.nameForCondition()?.endsWith(m)
+      )
+
+      const isBackground = module?.layer === 'react:background'
+
+      return isBackground && !isHMRImports
+    },
+    backend: {
+      client: require.resolve(
+        '../../../hot/lazy-compilation-fetch.js',
+      ),
+      listen: {
+        host: hostname ?? getIp(),
+      },
+    },
+  }
+
+  if (options === true) return defaultOptions
+
+  return {
+    ...options,
+    ...defaultOptions,
+  }
+}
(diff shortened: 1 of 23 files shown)


packages/rspeedy/core/src/config/rsbuild/lazyCompilation.ts
```diff
@@ -0,0 +1,55 @@
...
+        host: hostname ?? getIp(),
+      },
+    },
+  }
+
+  if (options === true) return defaultOptions
+
+  return {
+    ...options,
+    ...defaultOptions,
```
colinaaa, line 48: This doesn't seem correct. Should this be

```js
return {
  ...defaultOptions,
  ...options,
}
```

packages/rspeedy/core/src/utils/getIp.ts
```diff
@@ -0,0 +1,62 @@
...
+    })
+
+  if (!host) {
+    throw new Error(`No valid IP found`)
+  }
+
+  return host
+}
+
+const ip = await findIp('v4')
```
colinaaa: Is it possible to avoid invoking `findIp` at the top level? This way, we could execute `rspeedy build` in a network-free environment.

It cannot work now since you are importing `utils/getIp.ts` in `config/rsbuild/lazyCompilation.ts` which would be imported to when running `rspeedy build`. So another way to solve this is to lazy import lazy compilation-related only when development. (e.g.: in `dev.plugin.ts`)

packages/webpack/css-extract-webpack-plugin/src/CssExtractRspackPlugin.ts
```diff
@@ -166,27 +179,48 @@ class CssExtractRspackPluginImpl {
...
+        );
+
         compilation.hooks.runtimeModule.tap(
           this.name,
           (runtimeModule, chunk) => {
             if (runtimeModule.name === 'require_chunk_loading') {
               const asyncChunks = Array.from(chunk.getAllAsyncChunks())
+                .filter(c =>
+                  c.id
+                    !== '_react_background_packages_react_refresh_dist_index_js'
```
colinaaa: Why do we have to hard-code this here?

gaoachao: The `_react_background_packages_react_refresh_dist_index_js` is a lazy chunk without CSS, which should not be included in `cssHotUpdatePath` or `lynxLazyChunkIds` to prevent unnecessary fetch requests for `.hot-update.json` file during HMR. However, we could manage it using a constant for better coding practices and maintainability.

packages/webpack/template-webpack-plugin/src/LynxAsyncChunksRuntimeModule.ts
```diff
@@ -24,25 +26,39 @@ export function createLynxAsyncChunksRuntimeModule(
       const chunk = this.chunk!;
       const compilation = this.compilation!;
 
-      return `// lynx async chunks ids
+      return `
```
upupming, line 29: Is this runtime module dev only, will is increase prod size?

gaoachao: It is not dev only. However, in production mode, `__webpack_require__.lynx_aci` will only be injected when lazy bundles exist in the project. This behavior is unavoidable.

upupming:
Sames not working on dynamic import, the required path is `http://xxx/.rspeedy/async/_react_background_src_a_ts/_react_background_src_a_ts.efbcd5cceed67828.css.hot-update.json`, but the actual path is `a.js.xxx.bundle`:



```js
useEffect(async () => {
    const a = await import('./a.js');
    console.log('a', a);
    console.info('Hello, ReactLynx');
  }, []);
```

image

image

gaoachao:
> Sames not working on dynamic import, the required path is `http://xxx/.rspeedy/async/_react_background_src_a_ts/_react_background_src_a_ts.efbcd5cceed67828.css.hot-update.json`, but the actual path is `a.js.xxx.bundle`:
> 
> ```js
> useEffect(async () => {
>     const a = await import('./a.js');
>     console.log('a', a);
>     console.info('Hello, ReactLynx');
>   }, []);
> ```
> 
> image image

The `lynx.bundle` architecture prevents us from enabling the `imports` option in `lazyCompilation` settings, which would have resolved this non-critical error.

PR #175, opened 2025-03-11
fix: gesture config not being processed correctly

f0rdream:
## Summary

Gesture may have configs like `minDistance` fro `PanGesture`, but it is not correctly processed. Now we add it when processing gesture object.

## Checklist

### DIFF
--- a/.changeset/tangy-pugs-peel.md
+++ b/.changeset/tangy-pugs-peel.md
@@ -1,0 +1,11 @@
+---
+'@lynx-js/react': patch
+---
+
+fix: gesture config not processed correctly
+
+```typescript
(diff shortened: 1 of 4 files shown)


.changeset/tangy-pugs-peel.md
```diff
@@ -0,0 +1,5 @@
+---
+'@lynx-js/react': patch
+---
+
+Fix gesture not being processed correctly
```
Yradex: can you provide a better description so our users can understand what it fixes?

f0rdream: Sure, changeset updated

Yradex (approved):

PR #181, opened 2025-03-12
fix: avoid override user defined fetch function

upupming:
## Checklist


packages/webpack/runtime-wrapper-webpack-plugin/src/RuntimeWrapperWebpackPlugin.ts
```diff
@@ -229,8 +229,8 @@ lynx = lynx || {};
 lynx.targetSdkVersion=lynx.targetSdkVersion||${
       JSON.stringify(targetSdkVersion)
     };
-${overrideRuntimePromise ? `var Promise = lynx.Promise;` : ''}
-var fetch = lynx.fetch;
+${overrideRuntimePromise ? `g.Promise = g.Promise || lynx.Promise;` : ''}
+g.fetch = g.fetch || lynx.fetch;
```
upupming: It will pollute globalThis when using SharedContext

gaoachao (dismissed):

PR #194, opened 2025-03-13
feat(web):  `createRpcEndpoint` adds a new parameter: `hasReturnTransfer`.

Sherry-hue:
## Summary

feat: `createRpcEndpoint` adds a new parameter: `hasReturnTransfer`.

When `isSync`: false, `hasReturn`: true, you can add `transfer` to the callback postMessage created.

At this time, the return value structure of register-handler is changed: `{ data: unknown; transfer: Transferable[]; } | Promise<{ data: unknown; transfer: Transferable[];}>`.

## Checklist

### DIFF
--- a/.changeset/forty-cobras-think.md
+++ b/.changeset/forty-cobras-think.md
@@ -1,0 +1,10 @@
+---
+"@lynx-js/web-worker-rpc": patch
+"@lynx-js/web-constants": patch
+---
+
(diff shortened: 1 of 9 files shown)


PupilTong (changes requested):
The types should be improved.

The handler type of `registerHandler` should be different with the endpoints's handler which do not need to transfer.

`createCall` has the same issue.

PupilTong (dismissed):

PupilTong (approved):

PR #198, opened 2025-03-13
feat: use shadowroot to isolate one lynx-view

PupilTong:
Before this commit, we have been detecting if current browser supports the `@scope` rule.
This allows us to scope one lynx-view's styles.

After this commit we always create a shadowroot to scope then.

Also for the new shadowroot pattern, we add a new **attribute** `inject-head-links`.
By default, we will iterate all `<link rel="stylesheet">` in the `<head>`, and use `@import url()` to import them inside the shadowroot.
Developers could add a `inject-head-links="false"` to disable this behavior.
#51

### DIFF
--- a/packages/web-platform/web-core/src/apis/LynxView.ts
+++ b/packages/web-platform/web-core/src/apis/LynxView.ts
@@ -16,6 +16,7 @@ import {
   type NativeModulesCall,
   type UpdateDataType,
 } from '@lynx-js/web-constants';
+import { inShadowRootStyles } from './inShadowRootStyles.js';
 
 /**
  * Based on our experiences, these elements are almost used in all lynx cards.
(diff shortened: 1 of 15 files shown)


packages/web-platform/web-core/src/apis/LynxView.ts
```diff
@@ -410,7 +415,24 @@ export class LynxView extends HTMLElement {
...
+            this.attachShadow({ mode: 'open' });
+          }
+          const styleElement = document.createElement('style');
+          this.shadowRoot!.append(styleElement);
+          const styleSheet = styleElement.sheet!;
+          styleSheet.insertRule(inShadowRootStyles);
+          const injectHeadLinks =
+            this.getAttribute('inject-head-links') !== 'false';
+          if (injectHeadLinks) {
+            document.head.querySelectorAll('link[rel="stylesheet"]').forEach(
```
Sherry-hue, line 428: Multiple lynx-views on a page need to repeatedly copy docuemnt.header-style to them.
Will this cause other problems, such as performance?

packages/web-platform/web-core/src/apis/LynxView.ts
```diff
@@ -16,6 +16,7 @@ import {
   type NativeModulesCall,
   type UpdateDataType,
 } from '@lynx-js/web-constants';
+import { inShadowRootStyles } from './inShadowRootStyles.js';
```
Sherry-hue, line 19: Is there no need to propose a separate file?

packages/web-platform/web-core/src/apis/LynxView.ts
PupilTong: It's not a copy. We only use the @import rule to make references to them

packages/web-platform/web-core/src/apis/LynxView.ts
PupilTong: > Is there no need to propose a separate file?

It's like a seprated css file.

Sherry-hue (approved):

PR #199, opened 2025-03-13
add(feature#111): adding tunnel feature to plugin-qrcode

ManojPanda3:
## Summary

This pull request resolves issue #111, which addresses a feature to add in the `rspeedy cli` component. This change modifies `/home/manoj/code/contributes/lynx-stack/packages/rspeedy/plugin-qrcode`  to have a option for tunnel activation and deactivation in the loop function at `shortcuts.ts` and also add a `tunnel.ts` which has types and logic to create an tunnel. The core logic change to give a option in cli tool which on select toggle the tunnel on by changing `tunnel.isOpen` state to be true and `tunnel.url` to the ngrok tunnel url

## Checklist
- [x] Tests updated (added a vitest test file named as `tunnel.test.ts` to cover the corrected logic.)
### Before activating tunnel
Before activating tunnel
### After activating tunnel
After activating tunnel
### Preview in LynxExplorer
Preview in LynxExplorer

### DIFF
--- a/.changeset/ready-numbers-type.md
+++ b/.changeset/ready-numbers-type.md
@@ -1,0 +1,5 @@
+---
+"@lynx-js/rspeedy": patch
+---
+
+Bump Rsdoctor v1.0.0-rc.0
--- a/packages/rspeedy/core/package.json
+++ b/packages/rspeedy/core/package.json
@@ -59,7 +59,7 @@
     "@microsoft/api-extractor": "catalog:",
     "@rsbuild/core": "catalog:",
     "@rsbuild/plugin-css-minimizer": "1.0.2",
-    "@rsdoctor/rspack-plugin": "^0.4.13",
+    "@rsdoctor/rspack-plugin": "1.0.0-rc.0",
     "chokidar": "^4.0.3",
     "commander": "^13.1.0",
     "exit-hook": "^4.0.0",
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -253,8 +253,8 @@ importers:
         specifier: 1.0.2
         version: 1.0.2(@rsbuild/core@1.2.16)(webpack@5.98.0)
       '@rsdoctor/rspack-plugin':
-        specifier: ^0.4.13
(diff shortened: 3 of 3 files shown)


colinaaa (changes requested):
Thank you for your contribution!

However, I don't think most of the user would need this (especially the `ngrok` is a paid service).

So I wonder if this could be implemented as a standalone plugin instead of a new feature of the QR code plugin.

We actually have similar plugins like this internally. You may use the `registerConsoleShortcuts` function exported from `@lynx-js/qrcode-rsbuild-plugin/shortcuts`. Here is an example:

```js
export function pluginNgrok(options): RsbuildPlugin {
  return {
    name: 'lynx:rsbuild:ngrok',
    pre: ['lynx:rsbuild:api'],
    setup(api) {
      api.onAfterStartProdServer(async ({ environments, port }) => {
        await main(environments['lynx'], port)
      })

      api.onDevCompileDone(async ({ isFirstCompile, stats, environments }) => {
        if (!api.context.devServer) return

        if (stats.hasErrors()) return

        if (!isFirstCompile) {
          return
        }

        await main(environments['lynx'], api.context.devServer.port)
      })

      async function main(
        environmentContext: EnvironmentContext | undefined,
        port: number,
      ) {
        if (!environmentContext) {
          // Not lynx environment, skip print QRCode
          return
        }

        const ngrok = await connectNgrokTunnel(tunnelConfig)

        const unregister = await registerConsoleShortcuts({
          api,
          entries: Object.keys(environmentContext.entry),
          schema: options.schema,
          port,
          customShortcuts: {
            o: {
              value: 'c',
              label: 'Connect to ngrok',
            },
          },
        })

        api.onCloseDevServer(() => {
          unregister()
          ngrok.stop()
        })
      }
    },
  }
}
```

Or you may just implement a ngrok Rsbuild plugin that does not change the QR code that Rspeedy generated. Only the tunnel is created.

Let me know if you have any further question.

ManojPanda3:
Thanks for the suggestion ,
I will create an plugin for it.

PR #204, opened 2025-03-14
feat(react): allow some `<list-item/>`s to be deferred

hzy:
## Checklist


colinaaa:
It would be fantastic to have a documentation PR prepared for the new `<List />` component in the lynx-website :)

.changeset/smart-lies-glow.md
```diff
@@ -0,0 +1,33 @@
+---
+"@lynx-js/react": patch
+---
+
+New built-in component `<List/>`, which provide a virtualized list view. Unlike the element `<list/>`, `<List/>` render each item on-demand, which is more efficient for large list.
```
Yradex:
```suggestion
New built-in component `<List/>`, which provides a virtualized list view. Unlike the element `<list/>`, `<List/>` renders each item on-demand, which is more efficient for large lists.
```

.changeset/smart-lies-glow.md
```diff
@@ -0,0 +1,33 @@
...
+        </text>
+      </view>
+    );
+  }}
+  keyExtractor={(item) => `${item.x}`}
+  getItemProps={(item) => ({
+    style: 'height: 120rpx;',
+  })}
+  getItemLayout={(item) => ({
+    'full-span': item.x === 10,
```
Yradex: Is it OK to use `'full-span'` here, since it has been commented in `ListItemLayoutProps`?

packages/react/components/src/List.tsx
```diff
@@ -0,0 +1,150 @@
...
+      override state = {
+        isReady: false,
+      };
+
+      onComponentAtIndex = (data: ListItemComponentAtIndexEventData) => {
+        const { listID, childCtxId } = data;
+        this.setState({ isReady: true }, () => {
+          // @ts-expect-error it's a native method
+          // eslint-disable-next-line @typescript-eslint/no-unsafe-call, @typescript-eslint/no-unsafe-member-access
+          lynx.getNativeApp().callLepusMethod('rLynxOnListItemReady', { listID, childCtxId });
```
Yradex: Should `rLynxOnListItemReady` be added into `LifecycleConstant` as well?

hzy:
bump

.changeset/smart-lies-glow.md
```diff
@@ -0,0 +1,18 @@
+---
+"@lynx-js/react": minor
```
colinaaa, line 2: We need to modify `peerDependencies` of `@lynx-js/react-rsbuild-plugin` and `@lynx-js/react-webpack-plugin` and add a changeset for them.

packages/react/types/react.d.ts
```diff
@@ -142,7 +142,9 @@ declare module '@lynx-js/types' {
      * we can use typeof item as the reuse-identifier.
      * By default, the developer does not provide a reuse-identifier, which is determined by the framework at the compile stage.
      */
-    'reuse-identifier'?: Key;
+    'reuse-identifier'?: Key | null | undefined;
+
+    'defer'?: boolean | undefined;
```
colinaaa, line 165: Do we need some comments on `defer`?

packages/react/components/src/DeferredListItem.tsx
```diff
@@ -0,0 +1,45 @@
...
+export interface DeferredListItem {
+  defer?: boolean;
+  renderListItem: (children: ReactNode | undefined) => JSX.Element;
+  renderChildren: () => ReactNode;
+}
+
+export const DeferredListItem: FC<DeferredListItem> = ({ defer, renderListItem, renderChildren }) => {
+  const __cloneElement = __LEPUS__ ? _cloneElementLepus : _cloneElement;
+
+  const initialDeferred = useRef(defer);
```
colinaaa: I didn't see any test for setting `defer` to value other than `true`. E.g.:

- `<list-item defer={false}>`
- `<list-item defer={defer}>`

But it looks like we do support these patterns.

Is this intentional, or should we only focus on using `defer={true}`?

packages/react/transform/src/swc_plugin_list/mod.rs
```diff
@@ -0,0 +1,307 @@
...
+
+fn jsx_list_item_deferred(n: &JSXElement) -> bool {
+  if !jsx_is_list_item(n) {
+    return false;
+  }
+
+  return n.opening.attrs.iter().any(|attr| {
+    if let JSXAttrOrSpread::JSXAttr(attr) = attr {
+      if let JSXAttrName::Ident(ident) = &attr.name {
+        ident.sym == "defer"
```
colinaaa, line 45: Do we still need to apply transformations when `defer={false}` is explicitly set?

I'm considering making `defer={true}` the default value in a future version.

packages/react/runtime/src/list.ts
```diff
@@ -263,13 +265,58 @@ export function componentAtIndexFactory(
...
+
+      let p: Promise<number>;
+      return (p = new Promise<number>((resolve) => {
+        Object.defineProperty(childCtx.__extraProps, 'isReady', {
+          set(isReady) {
+            if (isReady === 1) {
+              delete childCtx.__extraProps!['isReady'];
+              childCtx.__extraProps!['isReady'] = 1;
+
+              Promise.resolve().then(() => {
```
colinaaa: We could save some time & memory by creating a resolved Promise and re-use it.

```js
const resolvedPromise = Promise.resolve();
  const componentAtChildCtx = () => {
              resolvedPromise.then(() => {
  }
```

packages/react/testing-library/src/__tests__/list.test.jsx
```diff
@@ -546,3 +543,133 @@ describe('list', () => {
...
+              <view style='height: 50rpx; width: 600rpx; background-color: #fff; border-bottom: 1px solid #ccc;'>
+                <text style='padding: 10rpx;'>Item {index + 1}</text>
+              </view>
+            </list-item>
+          ))}
+        </list>
+      );
+    }
+
+    console.log(App.toString());
```
colinaaa:
```suggestion
```

packages/react/runtime/src/list.ts
```diff
@@ -263,13 +265,58 @@ export function componentAtIndexFactory(
...
+      __OnLifecycleEvent([LifecycleConstant.publishEvent, {
+        handlerName: `${childCtx.__id}:__extraProps:onComponentAtIndex`,
+        data: {
+          listID,
+          childCtxId: childCtx.__id,
+        },
+      }]);
+
+      let p: Promise<number>;
+      return (p = new Promise<number>((resolve) => {
```
colinaaa, line 86: Is is correct to return a `Promise` in `componentAtIndex`? It seems like the Lynx Engine would cast it to a number:

https://github.com/lynx-family/lynx/blob/e795287f127b7ccd5719fb4d17186d5f80c63751/core/renderer/dom/fiber/list_element.cc#L167-L175

packages/react/runtime/src/list.ts
```diff
@@ -263,13 +265,58 @@ export function componentAtIndexFactory(
...
+          },
+          get() {
+            return p;
+          },
+        });
+      }));
+    } else if (maybePromise<number>(childCtx.__extraProps?.['isReady'])) {
+      // throw new Error('componentAtIndex called on a preparing list-item');
+      return childCtx.__extraProps['isReady'].then((uiSign) => {
+        const root = childCtx.__elements![0]!;
```
colinaaa: Is the flush in `.then` truly necessary?

It appears a flush might occur once `childCtx.__extraProps['isReady']` is resolved.

Would the only difference be in the `operationID`?

packages/react/runtime/src/list.ts
```diff
@@ -366,25 +420,68 @@ export function componentAtIndexFactory(
...
+        readyUiSigns.push(u);
+      } else {
+        unreadyOperationIDs.push(operationID);
+        unreadyPromises.push(u);
+      }
     });
+
+    if (unreadyPromises.length > 0) {
+      Promise.all(unreadyPromises).then((uiSigns) => {
+        __FlushElementTree(list, {
```
colinaaa, line 266: ditto here. It looks like a flush would be made in `componentAtChildCtx`.

Should we move all `__FlushElementTree` into `componentAtIndex` and `componentAtIndexes` instead of `componentAtChildCtx`?

packages/react/components/src/DeferredListItem.tsx
```diff
@@ -0,0 +1,45 @@
...
+  const onGetDomRef = useRef<RefCallback<SnapshotInstance>>((ctx) => {
```
colinaaa: nitpick: it's weird to use `useRef` here :)

Can we use `useCallback` instead to avoid the type assertions `ctx!.__extraProps`?

packages/react/runtime/src/list.ts
```diff
@@ -263,13 +265,58 @@ export function componentAtIndexFactory(
...
     const platformInfo = childCtx.__listItemPlatformInfo || {};
 
+    // The lifecycle of this `__extraProps.isReady`:
+    //   0 -> Promise<number> -> 1
+    // 0: The initial state, the list-item is not ready yet, we will send a event to background
+    //    when `componentAtIndex` is called on it
+    // Promise<number>: A promise that will be resolved when the list-item is ready
+    // 1: The list-item is ready, we can use it to render the list
+    if (childCtx.__extraProps?.['isReady'] === 0) {
+      __OnLifecycleEvent([LifecycleConstant.publishEvent, {
```
Yradex, line 80: There is a risk that this event could be sent before `LifecycleConstant.firstScreen`.

packages/react/components/src/DeferredListItem.tsx
```diff
@@ -0,0 +1,45 @@
...
+import { cloneElement as _cloneElementLepus } from '@lynx-js/react/lepus';
+
+export interface DeferredListItem {
+  defer?: boolean;
+  renderListItem: (children: ReactNode | undefined) => JSX.Element;
+  renderChildren: () => ReactNode;
+}
+
+export const DeferredListItem: FC<DeferredListItem> = ({ defer, renderListItem, renderChildren }) => {
+  const __cloneElement = __LEPUS__ ? _cloneElementLepus : _cloneElement;
```
upupming:
```suggestion
  const __cloneElement = __MAIN_THREAD__ ? _cloneElementMainThread : _cloneElement;
```

packages/react/components/src/DeferredListItem.tsx
```diff
@@ -0,0 +1,45 @@
+// Copyright 2025 The Lynx Authors. All rights reserved.
+// Licensed under the Apache License Version 2.0 that can be found in the
+// LICENSE file in the root directory of this source tree.
+
+import type { FC, ReactNode, RefCallback } from 'react';
+
+import { cloneElement as _cloneElement, useRef, useState } from '@lynx-js/react';
+import type { SnapshotInstance } from '@lynx-js/react/internal';
+import { cloneElement as _cloneElementLepus } from '@lynx-js/react/lepus';
```
upupming:
```suggestion
import { cloneElement as _cloneElementMainThread } from '@lynx-js/react/lepus';
```

packages/react/transform/src/swc_plugin_list/mod.rs
```diff
@@ -0,0 +1,307 @@
...
+  test!(
+    module,
+    Syntax::Es(EsSyntax {
+      jsx: true,
+      ..Default::default()
+    }),
+    |t| visit_mut_pass(ListVisitor::new(Some(t.comments.clone()))),
+    should_transform_list_item_deferred_basic,
+    r#"
+    <list-item defer key="1"></list-item>;
```
upupming: We should add `item-key` here as it is a required field of `<list-item>` (same below)
```suggestion
    <list-item defer item-key="1" key="1"></list-item>;
```

packages/react/transform/tests/__swc_snapshots__/src/swc_plugin_list/mod.rs/should_transform_list_item_deferred_with_children_with_snapshot.js
```diff
@@ -0,0 +1,40 @@
...
+    return [
+        el
+    ];
+}, null, null, undefined, globDynamicComponentEntry, null);
+<ReactLynxRuntimeComponents.DeferredListItem renderListItem={(__c)=><__snapshot_da39a_test_1 values={[
+        noop
+    ]}>{__c}</__snapshot_da39a_test_1>} renderChildren={()=><>
+      <__snapshot_da39a_test_2/>
+      <__snapshot_da39a_test_3/>
+      <__snapshot_da39a_test_4/>
```
upupming: I remember static elements will be compressed to a single snapshot, why we cannot compress them here after deferred list transformation?

packages/react/runtime/__test__/preact.test.jsx
```diff
@@ -610,20 +610,24 @@ describe('document - background', () => {
...
+          0="bar"
         />
       </root>
     `);
     delete BackgroundSnapshotInstance.prototype.toJSON;
   });
+
+  it('extraProps - should set __extraProps', () => {
+    render(cloneElement(<view />, { x: 1 }), scratch);
+  });
```
upupming: Is this test missed anything?

packages/react/types/react.d.ts
hzy: Absolutely yes

packages/react/runtime/src/list.ts
hzy: Yes, i've removed this branch in the next commit, it should be unreachable as said by @DwwWxx

packages/react/runtime/src/list.ts
hzy: Good catch, I will change it to undefined

packages/react/runtime/lepus/index.js
```diff
@@ -41,3 +41,47 @@ export function createElement(type, props, children) {
...
+/**
+ * Clones the given VNode, optionally adding attributes/props and replacing its
+ * children.
+ * @param {VNode} vnode The virtual DOM element to clone
+ * @param {object} props Attributes/props to add when cloning
+ * @param {Array<ComponentChildren>} rest Any additional arguments will be used
+ * as replacement children.
+ * @returns {VNode}
+ */
+export function cloneElement(vnode, props, children) {
```
colinaaa, line 54: We may reduce some bytes to use `createElement` in `cloneElement`:

```js
export function cloneElement(vnode, props, ...rest) {
	return createElement(
		vnode.type,
		assign(
			vnode.ref ? { ref: vnode.ref } : {},
			vnode.key ? { key: vnode.key } : {},
			vnode.props,
			props
		),
		...rest
	);
}
```

> https://github.com/preactjs/preact/pull/4814/commits/37d531f55078bb5b37837baf3561dee7947a1f8f

> [!NOTE]
> Preact did not merge this change since it would increase the **compressed size**.
> But I think it would be OK for us since we are using AOT bytecode in main-thread.

.changeset/smart-lies-glow.md
hzy: done!

packages/react/runtime/__test__/preact.test.jsx
hzy: done!

packages/react/transform/src/lib.rs
```diff
@@ -603,7 +609,7 @@ fn transform_react_lynx_inner(
       compat_plugin,
       worklet_plugin,
       css_scope_plugin,
-      snapshot_plugin,
+      (list_plugin, snapshot_plugin),
```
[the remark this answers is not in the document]
hzy: when we have a lot plugins in the chain, swc's marco could not handle that many args, using tuple is REQUIRED.

packages/react/transform/tests/__swc_snapshots__/src/swc_plugin_list/mod.rs/should_transform_list_item_deferred_with_children_with_snapshot.js
hzy: if we compressed it into a single snapshot, a wrapper will be created, like:

```
<wrapper>
    <__snapshot_da39a_test_2/>
    <__snapshot_da39a_test_3/>
    <__snapshot_da39a_test_4/>
</wrapper>
```

I prefer not adding a wrapper :D

packages/react/runtime/lepus/index.js
hzy: Great! I will try this on a separate PR and compare the size

.changeset/smart-lies-glow.md
```diff
@@ -0,0 +1,18 @@
...
+---
+
+Allow some `<list-item/>`s to be defer and render at background thread.
```
Yradex:
```suggestion
Allow some `<list-item/>`s to be deferred and rendered in the background thread.
```

.changeset/smart-lies-glow.md
```diff
@@ -0,0 +1,18 @@
...
+
+Using the following syntax:
```
Yradex:
```suggestion
Use the following syntax:
```

.changeset/smart-lies-glow.md
```diff
@@ -0,0 +1,18 @@
...
+```diff
+<list>
+-  <list-item item-key="...">
++  <list-item item-key="..." defer>
+      <SomeHeavyComponent />
+  </list-item>
+</list>
+```
+
+You should render your heavyweight components with `defer` attribute to avoid blocking the main thread.
```
Yradex:
```suggestion
You should render your heavyweight components with the `defer` attribute to avoid blocking the main thread.
```

packages/react/runtime/src/backgroundSnapshot.ts
```diff
@@ -242,15 +243,27 @@ export class BackgroundSnapshotInstance {
       return;
     }
 
-    // old path (`<__snapshot_xxxx_xxxx __0={} __1={} />` or `this.setAttribute(0, xxx)`)
-    // is reserved as slow path
-    const index = typeof key === 'string' ? Number(key.slice(2)) : key;
-    (this.__values ??= [])[index] = value;
+    if (typeof key === 'string') {
+      (this.__extraProps ??= {})[key] = value;
+      __globalSnapshotPatch?.push(
```
colinaaa: It seems like this can be simplified:

```js
if (typeof key === "string") {
  (this.__extraProps ??= {})[key] = value;
} else {
  // old path (`this.setAttribute(0, xxx)`)
  // is reserved as slow path
  (this.__values ??= [])[key] = value;
}
__globalSnapshotPatch?.push(
  SnapshotOperation.SetAttribute,
  this.__id,
  key,
  value
);
if (__PROFILE__) {
  console.profileEnd();
}
```

Yradex: Is it possible to merge this logic with the one on line 263?

packages/react/testing-library/src/__tests__/act.test.jsx
```diff
@@ -157,7 +158,7 @@ test('fireEvent triggers useEffect calls', async () => {
       [
         "rLynxChange",
         {
-          "data": "{"patchList":[{"snapshotPatch":[0,"__Card__:__snapshot_e8d0a_test_4",2,0,null,3,4,3,[0],1,2,3,null,4,2,[1],1,-1,2,null],"id":2}]}",
+          "data": "{"patchList":[{"snapshotPatch":[0,"__Card__:__snapshot_e8d0a_test_4",2,4,2,[1],0,null,3,4,3,[0],1,2,3,null,1,-1,2,null],"id":2}]}",
```
colinaaa, line 161: Why does this change?

It looks like a `SetAttributes()` (`4,2,[1])` happens earlier than before.

packages/react/runtime/src/list.ts
```diff
@@ -53,13 +56,55 @@ export function componentAtIndexFactory(
...
+              });
+            }
+          },
+          get() {
+            return p;
+          },
+        });
+      }));
+    } else if (maybePromise<number>(childCtx.__extraProps?.['isReady'])) {
+      throw new Error('componentAtIndex called on a pending deferred list-item');
```
Yradex:
```suggestion
      throw new Error('componentAtIndex was called on a pending deferred list item');
```

packages/react/runtime/src/backgroundSnapshot.ts
```diff
@@ -433,23 +465,7 @@ export function hydrate(
...
-                node.__values = undefined;
-                node.setAttribute('values', values);
-              }
-              __globalSnapshotPatch!.push(
-                SnapshotOperation.InsertBefore,
-                before.id,
-                node.__id,
-                target?.id,
-              );
+              helper2([node], before.id, target?.id);
```
colinaaa, line 458: The `helper2` function sets the attribute **before** recursing into `childNodes`, whereas the current version sets the attribute **after** recursing into `childNodes`.

```ts
  const helper2 = (afters: BackgroundSnapshotInstance[], parentId: number) => {
    for (const child of afters) {
      const id = child.__id;
      __globalSnapshotPatch!.push(SnapshotOperation.CreateElement, child.type, id);
      const values = child.__values;
      if (values) {
        child.__values = undefined;
        child.setAttribute('values', values);
      }
      helper2(child.childNodes, id);
      __globalSnapshotPatch!.push(SnapshotOperation.InsertBefore, parentId, id, undefined);
    }
  };
```

Is this OK for us? It seems like this would make the patch below a little different.

colinaaa (dismissed):

packages/react/runtime/src/snapshot.ts
```diff
@@ -165,10 +165,17 @@ export const backgroundSnapshotInstanceManager: {
       return null;
     }
     const spreadKey = res[2];
-    if (spreadKey) {
-      return (ctx.__values![expIndex] as { [spreadKey]: unknown })[spreadKey];
+    if (res[1] === '__extraProps') {
+      if (spreadKey) {
+        return ctx.__extraProps![spreadKey];
+      }
+      throw new Error('unreachable');
```
Yradex, line 172: Get a better error message, or just let it crash?

packages/react/runtime/src/backgroundSnapshot.ts
```diff
@@ -242,15 +243,27 @@ export class BackgroundSnapshotInstance {
...
+      );
+      if (__PROFILE__) {
+        console.profileEnd();
+      }
+      return;
+    }
 
+    // old path (`this.setAttribute(0, xxx)`)
+    // is reserved as slow path
+    (this.__values ??= [])[key] = value;
```
colinaaa:
```suggestion
    if (typeof key === 'string') {
      (this.__extraProps ??= {})[key] = value;
    } else {
      // old path (`this.setAttribute(0, xxx)`)
      // is reserved as slow path
      (this.__values ??= [])[key] = value;
    }
```

packages/react/types/react.d.ts
```diff
@@ -142,7 +142,27 @@ declare module '@lynx-js/types' {
...
+     *
+     * Render `<list-item />` with heavyweight component on background thread.
+     *
+     * ```tsx
+     * <list-item defer item-key="heavy-item">
+     *  <HeavyComponent />
+     * </list-item>
+     * ```
+     */
+    'defer'?: boolean | undefined;
```
Yradex:
```suggestion
    'reuse-identifier'?: Key | null | undefined;

    /**
     * Makes the `<list-item />` deferred, meaning it will not be eagerly rendered
     * on the main thread like a non-deferred `<list-item />`.
     * Instead, it will be rendered on the background thread when necessary.
     *
     * As a result, rendering becomes asynchronous,
     * which means that during extremely fast scrolling, some `<list-item />`s may not have been rendered yet—these will usually appear blank.
     *
     * @example
     *
     * Render a `<list-item />` containing a heavyweight component on the background thread.
     *
     * ```tsx
     * <list-item defer item-key="heavy-item">
     *   <HeavyComponent />
     * </list-item>
     * ```
     */
    'defer'?: boolean | undefined;
```

Yradex (dismissed):

packages/react/runtime/src/snapshot.ts
hzy: it should be unreachable

Yradex: If it's unreachable, why did you check `if (spreadKey)` in the first place?

Yradex (approved):

Issue #230, opened 2025-03-17
[Feature]: enable loadEnv in Rspeedy


zoolsher:
### What problem does this feature solve?

https://rsbuild.dev/api/javascript-api/core#load-env-variables

### What does the proposed API of configuration look like?

```js
export default {
    loadEnv: true, // default is true
}
```

PR #233, opened 2025-03-17
feat(rspeedy): support loadEnv config

zoolsher:
1. use rsbuild loadEnv API to load .env file
2. solve issue #230

## Checklist


packages/rspeedy/core/test/cli/dev.test.ts
```diff
@@ -53,6 +54,27 @@ describe('CLI - dev', () => {
...
+
+    const program = new Command('test')
+    await dev.call(
+      program,
+      join(fixturesRoot, 'project-with-env'),
+      {},
+    )
+
+    // make sure the output contains variable BAR
+    expect(core.logger.debug).toBeCalledWith(
```
colinaaa: I don't believe asserting on `logger.debug` is the most effective way to test, as it focuses on the implementation rather than the behavior.

I would prefer using `rsbuild.initConfigs()` and check if the environment variables are injected.

colinaaa (changes requested):

packages/rspeedy/core/test/cli/dev.test.ts
zoolsher: I updated the test case with `initConfigs`

.changeset/busy-snails-do.md
```diff
@@ -0,0 +1,5 @@
+---
+"@lynx-js/rspeedy": major
```
colinaaa:
```suggestion
"@lynx-js/rspeedy": patch
```

We don't need a `major` (which would bump to v1.0.0) for this change

packages/rspeedy/core/src/config/index.ts
```diff
@@ -43,6 +43,14 @@ export interface Config {
    */
   provider?: RsbuildConfig['provider']
 
+  /**
+   * Rspeedy automatically loads the .env file by default, utilizing the [Rsbuild API](https://rsbuild.dev/api/javascript-api/core#load-env-variables).
+   * You can use the environment variables defined in the .env file within your code by accessing them via `import.meta.env.FOO` or `process.env.Foo`.
+   * @see https://rsbuild.dev/guide/advanced/env-vars#env-file
+   * @defaultValue true
+   */
+  loadEnv?: boolean | undefined
```
colinaaa: Rsbuild does not have this (it has the `loadEnv` option in the [`CreateRsbuildOptions`](https://rsbuild.dev/api/javascript-api/core#options)).

Maybe we could do the same thing here.

1. Add `loadEnv` to `CreateRspeedyOptions`.
2. Enable `loadEnv: true` in Rspeedy CLI (e.g.: `dev.ts`, `build.ts`, etc.)

.changeset/busy-snails-do.md
```diff
@@ -0,0 +1,5 @@
...
+---
+
+support load .env file by default
```
colinaaa:
```suggestion
Support load `.env` file by default
```

colinaaa (changes requested):
There are also some TODOs related to loadEnv (it's fine to address them in future PRs).

1. Add documentation at https://github.com/lynx-family/lynx-stack/tree/main/website/docs/en/guide
2. Add CLI flags `--env-dir` and `--env-mode` (see #304)

packages/rspeedy/core/src/config/index.ts
zoolsher: I see. I thought this might be a break change for current projects, so I added this flag. It gives a way to switch back to the original behavior if needed.

colinaaa: This would only be a breaking change when there are `.env` files and the environment variables are used in the project without `source.define`. But it will be an error in Lynx. So I'm fine with enabling by default without opt-out.

PupilTong (approved):

colinaaa (approved):

packages/rspeedy/core/src/config/index.ts
colinaaa: Rsbuild v1.3.3 adds a `--no-env` CLI flag which could opt-out `loadEnv`. See: https://github.com/web-infra-dev/rsbuild/pull/4947. We could add that, too.

PR #238, opened 2025-03-18
feat: enable tree-shaking of `enum` in `@lynx-js/react` lib

upupming:
## Checklist


packages/react/rslib.config.ts
```diff
@@ -0,0 +1,26 @@
...
+    lib: ['runtime', 'components', 'worklet-runtime'].map(folder => {
+      return {
+        source: {
+          entry: {
+            [folder]: `./${folder}/src/**`,
+          },
+        },
+        output: {
+          distPath: {
+            root: `./${folder}/lib`,
```
colinaaa, line 15: I'm uncertain if we can output to `lib/`, as `tsc` might also generate files there.

packages/react/package.json
```diff
@@ -145,7 +145,9 @@
     "README.zh.md"
   ],
   "scripts": {
-    "api-extractor": "api-extractor run --verbose"
+    "api-extractor": "api-extractor run --verbose",
+    "build": "rslib build",
```
colinaaa, line 162: Add a `turbo.json` so that this build would not depends on `//#build` and `^build`.

Also, it could be cached if the content is not changed.

packages/react/rslib.config.ts
```diff
@@ -0,0 +1,26 @@
+import type { RslibConfig } from '@rslib/core';
+import { defineConfig } from '@rslib/core';
+
+const config: RslibConfig = defineConfig(
+  {
+    lib: ['runtime', 'components', 'worklet-runtime'].map(folder => {
```
colinaaa: The `worklet-runtime` also has a build script, which bundle the runtime into `worklet-runtime/dist/dev.js` and `worklet-runtime/dist/main.js` from `src/`.

So I don't think we need a bundleless build for it

packages/react/runtime/src/lynx/performance.ts
```diff
@@ -80,9 +85,23 @@ function setPipeline(pipeline: PipelineOptions | undefined): void {
...
+    lynx.performance?._markTiming?.(globalPipelineOptions.pipelineID, MainThreadPerformanceTimingKeys[timestampKey]);
+  }
+}
+function markTiming(
+  timestampKey: BackgroundThreadPerformanceTimingKeys,
+  force?: boolean,
+): void {
   if (globalPipelineOptions && (force || globalPipelineOptions.needTimestamps)) {
-    lynx.performance?._markTiming?.(globalPipelineOptions.pipelineID, PerformanceTimingKeys[timestampKey]);
+    lynx.performance?._markTiming?.(
```
colinaaa: This seems duplicated with `markMainThreadTiming`.

Can we use the same implementation but with different type interface instead?

E.g.:

```js
import { markTimingImpl } from './impl'

export function markMainThreadTiming(
  timestampKey: MainThreadPerformanceTimingKeys,
  force?: boolean
) {
  return markTimingImpl(MainThreadPerformanceTimingKeys, timestampKey, force)
}

export enum MainThreadPerformanceTimingKeys {
 /** keys */
}
```

```js
import 'background-only' // <-- This is important!
import { markTimingImpl } from './impl'

export function markBackgroundTiming(
  timestampKey: BackgroundPerformanceTimingKeys,
  force?: boolean
) {
  return markTimingImpl(BackgroundPerformanceTimingKeys, timestampKey, force)
}

export enum BackgroundPerformanceTimingKeys {
 /** keys */
}
```

colinaaa (commented):
Can we utilize `'background only'` directive and `import 'background-only'`?

packages/react/rslib.config.ts
upupming: yes, we replace the tsc output with rslib; tsc no longer generate lib/*.js now.

packages/react/rslib.config.ts
upupming: `worklet-runtime` has both `lib` (tsc) and `dist` (esbuild), we change`lib` to use rslib to build.

packages/react/runtime/src/lifecycle/patch/commit.ts
```diff
@@ -173,6 +176,7 @@ async function commitToMainThread(): Promise<void> {
 }
 
 function commitPatchUpdate(patchList: PatchList, patchOptions: Omit<PatchOptions, 'reloadVersion'>): Promise<void> {
+  'background-only';
```
upupming, line 186: @colinaaa I have added `'background-only'` here too.

packages/react/runtime/src/lynx/performance/background.ts
```diff
@@ -0,0 +1,104 @@
+import 'background-only';
```
upupming, line 1: @colinaaa Now this file is background-only.

packages/react/rslib.config.ts
colinaaa: It looks like `@lynx-js/react#build` is currently depends on `@lynx-js/react-worklet-runtime#build`.

https://github.com/lynx-family/lynx-stack/blob/bb1b9a4301e3597668dc7aad38dd9fb2c6a65e8b/packages/react/turbo.json#L5-L11

The execution sequence will be as follows:

1. `//#build` using `tsc --build`
2. `@lynx-js/react-worklet-runtime#build` using `esbuild`
3. `@lynx-js/react#build` using `rslib`

Running `rslib build` on worklet-runtime appears to have no effects.

examples/react/turbo.json
```diff
@@ -0,0 +1,12 @@
+{
+  "$schema": "https://turbo.build/schema.json",
+  "extends": ["//"],
+  "tasks": {
+    "build": {
+      "dependsOn": [
+        "^build",
+        "@lynx-js/react#build"
```
colinaaa: I don't think we need this since `^build` would contain `@lynx-js/react#build`
```suggestion
```

packages/react/runtime/src/lynx/performance/main-thread.ts
```diff
@@ -0,0 +1,28 @@
...
+  parse_changes_end,
+  patch_changes_start,
+  patch_changes_end,
+}
+
+function markMainThreadTiming(
+  timestampKey: MainThreadPerformanceTimingKeys,
+  force?: boolean,
+): void {
+  if (globalPipelineOptions && (force || globalPipelineOptions.needTimestamps)) {
```
colinaaa: Why can't we use `markTimingImpl` here?

packages/webpack/react-webpack-plugin/test/remove-background-only-loader.js
```diff
@@ -0,0 +1,7 @@
+// Copyright 2024 The Lynx Authors. All rights reserved.
+// Licensed under the Apache License Version 2.0 that can be found in the
+// LICENSE file in the root directory of this source tree.
+
+export default function removeBackgroundOnly(source) {
+  return source.replace(/import\s+['"]background-only['"];\n?/g, '');
```
colinaaa: Instead of removing `background-only`, can we simple add `background-only` as the dependencies of `@lynx-js/react`?

packages/react/package.json
colinaaa, line 162: Perhaps we could move `scripts.build` to each sub-workspace (e.g.: `runtime`, `components`, `worklet-runtime`).

upupming: I have split `worklet-runtime` to its own package.json, `components` is not a package, so I keep it here

packages/webpack/react-webpack-plugin/test/remove-background-only-loader.js
upupming: `import 'background-only'` is included in `packages/react/runtime/lib/lynx/performance/background.js`, so add `background-only` as a dep of `react-webpack-plugin` package does not help.

colinaaa: Yeah, I mean "as the dependencies of `@lynx-js/react`"

upupming: Sorry, I got it, it works great! I have updated to use a `background-only` package as `devDep` of `@lynx-js/react`.

packages/react/package.json
```diff
@@ -145,15 +145,18 @@
...
   },
   "dependencies": {
     "preact": "npm:@hongzhiyuan/preact@10.24.0-319c684e"
   },
   "devDependencies": {
     "@lynx-js/types": "^3.2.0",
     "@microsoft/api-extractor": "catalog:",
-    "@types/react": "^18.3.20"
+    "@types/react": "^18.3.20",
+    "background-only": "workspace:*"
```
colinaaa: This should be included in `dependencies`; otherwise, `@lynx-js/react` won't function properly without `@lynx-js/react-rsbuild-plugin`.

> Note that it is actually published at NPM. See: [`background-only`](https://www.npmjs.com/background-only)

packages/react/worklet-runtime/turbo.json
```diff
@@ -9,7 +9,8 @@
         "src"
```
colinaaa: Should we consider including `rslib.config.ts` in the inputs?

packages/react/rslib.config.ts
colinaaa: It looks like we have to use `rslib` to build `runtime` in development.

Can we:

1. Still use `tsc` in development (both `.js` and `.d.ts`), output to `lib/`.
2. Use `rslib` when publishing (both `.js` and `.d.ts`), output to `dist/` (we can set `package.json#publishConfig.exports`). Just like plugin-react: https://github.com/lynx-family/lynx-stack/blob/3ff882976dff7ccc9bef28d53a53f36c177c48d7/packages/rspeedy/plugin-react/package.json#L80-L81

In this case, we could have better DX inside the lynx-stack repo (with only `tsc --build --watch` to watch all changes).

upupming: - I have removed `emitDeclarationOnly: true` config from `tsconfig.json`, so now we can use `tsc --build --watch` to build runtime when developing locally.

- For `package.json#publishConfig.exports`, as our `exports` object in `packages/react/package.json` is too large. if we override it in `package.json#publishConfig.exports`, `exports` will double and the package.json may be hard to read and maintain. So not sure if we should override it. Currently, I put both `tsc` and `rsbuild` output in the `lib` to avoid this. It works well because `rslib` will overwrite the `lib` files when we run `turbo build`.

gaoachao (dismissed):

packages/react/rslib.config.ts
colinaaa: > * For `package.json#publishConfig.exports`, as our `exports` object in `packages/react/package.json` is too large. if we override it in `package.json#publishConfig.exports`, `exports` will double and the package.json may be hard to read and maintain. So not sure if we should override it.

That makes sense. Managing two large exports is challenging to maintain.

> Currently, I put both `tsc` and `rsbuild` output in the `lib` to avoid this. It works well because `rslib` will overwrite the `lib` files when we run `turbo build`.

I'm uncertain if it will continue to work once the `turbo build` of `@lynx-js/react` no longer relies on `//#build`. This is because there would still be a cache hit for `rslib build`, while `tsc --build` would not, as caching is being bypassed in that case.

It seems to be working now since `@lynx-js/test-environment` relies on `//#build`.

---

In this case, I think it would be better to avoid the race condition. We may improve our DX using something like the source build plugin: https://github.com/rspack-contrib/rsbuild-plugin-source-build

packages/rspeedy/plugin-react-alias/src/index.ts
```diff
@@ -82,6 +92,7 @@ export function pluginReactAlias(options: Options): RsbuildPlugin {
                   .set('@lynx-js/react/lepus$', reactLepus.mainThread)
                   .set('@lynx-js/react/lepus/jsx-runtime', jsxRuntime.mainThread)
                   .set('@lynx-js/react/lepus/jsx-dev-runtime', jsxDevRuntime.mainThread)
+                 .set('background-only$', backgroundOnly.mainThread)
```
colinaaa, line 95: I think we already have aliases for `background-only$` at https://github.com/lynx-family/lynx-stack/blob/4f74a12e5664a0bbc9d650a17568be00e62e0865/packages/rspeedy/plugin-react/src/backgroundOnly.ts#L42-L46

Why are we adding it again?

PR #239, opened 2025-03-18
fix(react/runtime): batch multiple patches for main thread communication

Yradex:
## Summary
This PR enhances the React runtime by implementing a batching mechanism that accumulates multiple patch updates and sends them to the main thread in a single render pass. This optimization reduces redundant layout operations, improves rendering performance, and prevents screen flickering.

## Core Improvements
- **Implemented batch rendering**: Added batching mechanism to accumulate patches and send them together in one render cycle
- **Optimized commit process**: Modified the commit hook to postpone sending updates until all patches are ready

## Implementation Details
- **Added patch accumulation**: Created a storage array and functions to collect multiple patches before committing
- **Timing API changes**: Adjusted the timing events to fire once per render cycle instead of for each individual update
- **Refactored render functions**: Introduced `renderBackground` function to standardize rendering across the codebase

## Testing
- Updated all relevant test files for compatibility with the new batching mechanism
- Added new test case to verify multiple updates are merged correctly

## Impact
This change represents a minor version update to the `@lynx-js/react` package. While it maintains the public API, it includes a breaking change in the timing behavior that affects applications relying on the previous timing API behavior.

### Dependency Updates
- Updated peer dependencies in `@lynx-js/react-rsbuild-plugin` and `@lynx-js/react-webpack-plugin` to support the new react version
- Added compatibility changeset for dependent plugins

## Checklist

### DIFF
--- a/packages/react/runtime/src/lifecycle/patch/commit.ts
+++ b/packages/react/runtime/src/lifecycle/patch/commit.ts
@@ -30,6 +30,8 @@ let nextCommitTaskId = 1;
 
 let globalBackgroundSnapshotInstancesToRemove: number[] = [];
 
+let patchesToCommit: Patch[] = [];
+
 interface Patch {
   id: number;
   snapshotPatch?: SnapshotPatch;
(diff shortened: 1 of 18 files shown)


packages/react/runtime/src/lifecycle/patch/commit.ts
```diff
@@ -48,21 +50,36 @@ interface PatchOptions {
 }
 
 function replaceCommitHook(): void {
+  // use our own `options.debounceRendering` to insert a timing flag before render
+  type DebounceRendering = (f: () => void) => void;
+  const injectDebounceRendering = (debounceRendering: DebounceRendering): DebounceRendering => {
+    return (f: () => void) => {
+      debounceRendering(() => {
+        f();
```
hzy, line 58: TODO: Can we mark `vdom_diff_start` before calling this function? To avoid hooking `options[DIFF]`...

hzy (approved):

colinaaa (approved):

PR #240, opened 2025-03-18
feat(web): list supports list-type: flow

Sherry-hue:
## Summary

feat: The list element supports list-type with **flow**.

It supports all attributes and events under single, and also provides `full-span`.

For detailed usage, please refer to the official website.

## Checklist


packages/web-platform/web-elements/src/XList/XListAttributes.ts
```diff
@@ -28,6 +30,14 @@ export class XListAttributes
     (v) => `${parseFloat(v)}px`,
   );
 
+  @registerAttributeHandler('span-count', true)
+  @registerAttributeHandler('column-count', true)
+  #handlerCount = bindToStyle(
+    () => this.#dom,
+    '--list-item-span-count',
+    (v) => `${parseFloat(v)}`,
```
PupilTong, line 38: Could we detect browser compatibility and use https://developer.mozilla.org/en-US/docs/Web/CSS/attr ?

PupilTong (approved):

packages/web-platform/web-elements/src/XList/XListAttributes.ts
Sherry-hue: If necessary, a new PR will be created to solve it, because there are other places using var

PR #248, opened 2025-03-19
chore: add window var to define callback argument in `background.js`

luhc228:
## Summary

The `window` is `undefined` in Lynx. It's useful to distinguish between Lynx and the Web.

## Checklist

### DIFF
--- a/.changeset/loud-signs-see.md
+++ b/.changeset/loud-signs-see.md
@@ -1,0 +1,11 @@
+---
+"@lynx-js/runtime-wrapper-webpack-plugin": patch
+---
+
+Add `window` variable to callback argument in `background.js` and the `window` is `undefined` in Lynx. Sometimes it's useful to distinguish between Lynx and the Web.
+
+```js
+define('background.js', (..., window) => {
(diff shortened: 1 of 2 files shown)


.changeset/loud-signs-see.md
```diff
@@ -0,0 +1,11 @@
+---
+"@lynx-js/runtime-wrapper-webpack-plugin": patch
+---
+
+Add `window` variable to callback argument in `background.js` and the `window` is `undefined` in Lynx. Sometimes it's useful to distinguish between Lynx and the Web.
+
+```js
+define('background.js', (..., window) => {
+  console.log(window); // `undefined` in Lynx
```
colinaaa, line 9: I'm okay with this change, but could you explain the difference between using `if (typeof window === 'undefined')` and this?

colinaaa (approved):

.changeset/loud-signs-see.md
luhc228: Some users may write the code like below previously:
```
if (!window) {
  // do something
}
```
We may need to be compatible with their code.

PR #251, opened 2025-03-19
fix: delay remove of intermediate debugging assets to afterEmit

upupming:
## Checklist


packages/webpack/web-webpack-plugin/src/index.ts
```diff
@@ -57,6 +60,31 @@ export class WebWebpackPlugin {
           });
           return encodeOptions;
         });
+        compiler.hooks.afterEmit.tap(WebWebpackPlugin.name, (compilation) => {
+          this.assets2Delete.forEach((asset) => {
```
colinaaa, line 64: It seems that the `assets2Delete` is never cleaned up, which could lead to memory leaks.

packages/webpack/web-webpack-plugin/src/index.ts
```diff
@@ -57,6 +60,31 @@ export class WebWebpackPlugin {
...
+            compilation.deleteAsset(asset);
+            if (compilation.outputOptions.path) {
+              const files2Delete = [
+                asset,
+              ];
+              if (asset.endsWith('.js')) {
+                files2Delete.push(asset + '.map');
+              }
+              for (const file of files2Delete) {
+                fs.rmSync(
```
colinaaa, line 74: Emitting the asset to disk first and then manually deleting it seems inefficient.

colinaaa (changes requested):
I don't believe this is the ideal way to resolve the issue. If you would like to consume the source-map in other plugin, maybe try to read source map from memory instead of expecting it to be generated on disk.

packages/webpack/web-webpack-plugin/src/index.ts
upupming: Yes, actually the right way to consume our sourcemap is to read source map from memory using `compiler`.

But the plugin user used now are using fs to find all sourcemap files.

image

PR #257, opened 2025-03-20
feat(web): list add sticky && gap attribute

Sherry-hue:
## Summary

feat: 1. list adds support for the `sticky` attribute. Now sticky-offset, sticky-top, and sticky-bottom will only take effect when `sticky` is `true`.

2. Added support for `list-main-axis-gap`, `list-cross-axis-gap`.

## Checklist


PupilTong (approved):
Great Job!

PR #266, opened 2025-03-20
refactor: code clean

PupilTong:
remove duplicated overrideLynxTagToHTMLTagMap


.changeset/flat-llamas-start.md
```diff
@@ -0,0 +1,5 @@
+---
+"@lynx-js/web-core": patch
+---
+
+refactor: code clean
```
colinaaa, line 5: nitpick: I believe an empty changeset should suffice.

PR #282, opened 2025-03-21
feat(rspeedy/core): apply Rsdoctor `linter.rules['ecma-version-check']`

colinaaa:
## Summary

Apply a default rule option which will only warn syntax beyond ES2019.

This is a following up to #186.

## Checklist

- [x] **Tests updated** (or not required).


gaoachao (approved):
LGTM!

PR #326, opened 2025-03-25
fix(react/runtime): fix stack underflow issue when running on PrimJS

Yradex:
## Summary

fix: https://github.com/lynx-family/lynx/issues/474

## Checklist


colinaaa (approved):

packages/react/runtime/src/lifecycle/patch/commit.ts
hzy, line 1: In what scenarios will this issue occur?

colinaaa: In production :) I think it is related with transforming to ES2015

PR #331, opened 2025-03-25
feat(web): supports list with waterfall

Sherry-hue:
## Summary

feat(web): Web Platform supports list with waterfall

## Checklist


packages/web-platform/web-elements/src/XList/XListWaterfall.ts
```diff
@@ -0,0 +1,124 @@
...
+
+        this.#childrenObserver = new MutationObserver((mutationList) => {
+          const mutation = mutationList?.[0]!;
+
+          if (mutation.type === 'childList') {
+            // let the inserted list-item rendered, otherwise its width and height are not determined yet.
+            // requestAnimationFrame is called before inserted list-item is rendered, so setTimeout is used here
+            setTimeout(() => {
+              this.#layoutListItem(spanCount);
+            });
```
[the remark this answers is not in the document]
Sherry-hue: no，#childrenObserver should only observe the x-list children-list always

.changeset/thick-cameras-smash.md
```diff
@@ -0,0 +1,5 @@
+---
+"@lynx-js/web-elements": minor
```
PupilTong: patch

packages/web-platform/web-elements/src/XList/XListWaterfall.ts
```diff
@@ -0,0 +1,229 @@
...
+    if (this.#dom.getAttribute('list-type') === 'waterfall') {
+      const spanCount = parseFloat(
+        this.#dom.getAttribute('span-count')
+          || this.#dom.getAttribute('column-count')
+          || '',
+      ) || 1;
+      const scrollOrientation = this.#dom.getAttribute('scroll-orientation')
+        || 'vertical';
+
+      requestAnimationFrame(() => this.#createWaterfallContainer());
```
PupilTong: Could you please explain this? Why do we create the slot after a animation frame?

packages/web-platform/web-elements/src/XList/XListWaterfall.ts
```diff
@@ -0,0 +1,229 @@
...
+  };
+
+  constructor(dom: XList) {
+    this.#dom = dom;
+  }
+
+  #resizeObserverInit = (spanCount: number, isScrollVertical: boolean) => {
+    this.#resizeObserver?.disconnect();
+    this.#resizeObserver = new ResizeObserver(() => {
+      requestAnimationFrame(() =>
```
PupilTong: why do we wait a animation frame(Macro task) rather than a micro task?

PupilTong (dismissed):
LGTM! Great Job

PupilTong:
@Sherry-hue Could you please avoid to use text `1`, `2`, `3`, .... ?
Using background color could be a better solution, like `background-color: rgb(calc(1 * var(--index)), calc(1 * var(--index)), calc(1 * var(--index)))`. (note you could replace the number `1` by another number.

packages/web-platform/web-elements/src/XList/XListWaterfall.ts
Sherry-hue: Don't need this now, I'll remove it

packages/web-platform/web-elements/src/XList/XListWaterfall.ts
Sherry-hue: Don't need this now, I'll remove it

Sherry-hue:
> @Sherry-hue Could you please avoid to use text `1`, `2`, `3`, .... ? Using background color could be a better solution, like `background-color: rgb(calc(1 * var(--index)), calc(1 * var(--index)), calc(1 * var(--index)))`. (note you could replace the number `1` by another number.

Looks nice, I will open a separate PR to fix this problem

PupilTong (approved):

Issue #295, opened 2025-03-21
[Feature]: Support `performance.printFileSize`


colinaaa:
### What problem does this feature solve?

Decide whether to print the file sizes after production build.

### What does the proposed API of configuration look like?

The same as Rsbuild: https://rsbuild.dev/config/performance/print-file-size

```js
export default {
  performance: {
    printFileSize: {
      total: false,
    },
  },
}
```

PR #336, opened 2025-03-25
feat(rspeedy): Support `performance.printFileSize`

babu-ch:
## Summary

close https://github.com/lynx-family/lynx-stack/issues/295

## Checklist


packages/rspeedy/core/src/config/performance/index.ts
```diff
@@ -6,6 +6,7 @@ import type {
   ChunkSplitBySize,
   ChunkSplitCustom,
 } from './chunk-split.js'
+import type { PrintFileSize } from './print-file-size.js'
```
colinaaa: I wonder if we could use directly use the type from Rsbuild? So that we do not need to maintain the types in Rspeedy.

E.g.:

```js
import { PerformanceConfig } from '@rsbuild/core'

export interface Performance {
  printFileSize?: PerformanceConfig['printFileSize'] | undefined
}
```

> We did not do that for `Performance.chunkSplit` since Rsbuild did not provide `PerformanceConfig` before :)

packages/rspeedy/core/src/config/performance/index.ts
```diff
@@ -67,4 +68,25 @@ export interface Performance {
...
+   * import { defineConfig } from '@lynx-js/rspeedy'
+   *
+   * export default defineConfig({
+   *   performance: {
+   *     printFileSize: false
+   *   },
+   * })
+   * ```
+   */
+  printFileSize?: boolean | PrintFileSize | undefined
```
colinaaa: Nitpick: Arrange in alphabetical order (e.g., place `printFileSize` before `removeConsole`).

packages/rspeedy/core/src/config/performance/index.ts
babu-ch: Oh, it worked. https://github.com/lynx-family/lynx-stack/pull/336/commits/c94825edce9c805659fafe3f7186289544502bf7
In this case, should I include a link to rsbuild in the description?

https://rsbuild.dev/config/performance/print-file-size

colinaaa: Certainly, that sounds wonderful.

babu-ch: added 👍 
https://github.com/lynx-family/lynx-stack/pull/336/commits/1750b7f1793c6c203ddcc5c962a7678a9e298e68

PR #337, opened 2025-03-25
feat: add disableMTSAndGesture option to shake MTS and gesture

upupming:
## Summary

Solves #19

`dist/main.lynx.bundle` bundle size compare:

|Case|Before|After using `disableMTSAndGesture: true` |
|-|-|-|
|react/example|97.0 kB|93.3 kB (-4%)|
|react/example, with `import '@lynx-js/react/experimental/lazy/import';`|122.7 kB|112.2 kB (-9%)|

## Checklist


packages/react/runtime/lazy/import.js
```diff
@@ -19,19 +19,102 @@ import {
 } from './target';
 
 Object.defineProperty(target, sExportsReact, {
-  value: ReactAPIs,
+  value: {
+    Children: ReactAPIs.Children,
```
colinaaa, line 23: Would this create extra member expression and slow down the execution of `main-thread.js`?

packages/react/runtime/lazy/import.js
```diff
@@ -19,19 +19,102 @@ import {
...
+    useState: ReactAPIs.useState,
+    useSyncExternalStore: ReactAPIs.useSyncExternalStore,
+    withInitDataInState: ReactAPIs.withInitDataInState,
+    // Shake worklet APIs if not used.
+    /* c8 ignore next */
+    ...(__DISABLE_MTS_AND_GESTURE__ && process.env['NODE_ENV'] !== 'test') ? {} : {
+      MainThreadRef: ReactAPIs.MainThreadRef,
+      runOnBackground: ReactAPIs.runOnBackground,
+      runOnMainThread: ReactAPIs.runOnMainThread,
+      useMainThreadRef: ReactAPIs.useMainThreadRef,
```
colinaaa, line 61: How can we ensure that all the APIs are added here?

packages/react/runtime/src/backgroundSnapshot.ts
```diff
@@ -254,9 +254,9 @@ export class BackgroundSnapshotInstance {
             if (!newSpreadValue) {
               continue;
             }
-            if ((newSpreadValue as any)._wkltId) {
+            if (!__DISABLE_MTS_AND_GESTURE__ && (newSpreadValue as any)._wkltId) {
```
colinaaa, line 257: Instead of using a macro, can we use something like a pluginSystem?

E.g.:

```js
// backgroundSnapshot.ts
for (let key in newSpread) {
  const newSpreadValue = newSpread[key]
  newSpread[key] = spreadPlugins.call(newSpreadValue)
}
```

And we only register MTS & gesture related plugins when using MTS/gesture related APIs

```js
export const runOnBackground = /*#__PURE__*/ (() => {
  spreadPlugins.tap('mtx', onPostWorkletCtx)

  return function runOnBackground() { /* impl */ }
})()
```

/cc @Yradex

colinaaa (commented):
1. I'm not sure if the opt-out approach is the best.
    1. Ideally, it would be better to achieve on-demand imports without requiring any configuration.
    2. If that's not possible, an opt-in mechanism like `enableMTS` seems more reasonable compared to `disableMTS`, as it ensures better default performance and smaller size, which feels more meaningful.

2. Additionally, if a config field must be added, why combine MTS and gesture into a single switch? Wouldn't splitting them into two separate toggles for independent control be a better solution?

PR #340, opened 2025-03-25
feat: support extractStr to reuse variables and reduce bundle size

upupming:
## Checklist


packages/webpack/react-webpack-plugin/src/ReactWebpackPlugin.ts
```diff
@@ -224,6 +247,99 @@ class ReactWebpackPlugin {
...
       );
+      if (options.extractStr) {
+        compilation.hooks.processAssets.tap(
+          {
+            name: 'compilation',
+            stage:
+              compiler.webpack.Compilation.PROCESS_ASSETS_STAGE_OPTIMIZE_INLINE,
+          },
+          (assets) => {
+            if (!options.extractStr) return;
```
gaoachao:
```suggestion

```
Line 250 ensures `options.extractStr` is neither `false` nor `undefined`

packages/webpack/react-webpack-plugin/src/ReactWebpackPlugin.ts
```diff
@@ -8,6 +8,8 @@ import { createRequire } from 'node:module';
 import type { Chunk, Compilation, Compiler } from '@rspack/core';
 import invariant from 'tiny-invariant';
 
+import { transformBundleResultSync } from '@lynx-js/react/transform';
```
hzy: @gaoachao Will using this export block us from migrating to the swc plugin?

packages/webpack/react-webpack-plugin/src/ReactWebpackPlugin.ts
upupming: Yes, here I use it as a type guard to avoid ts error, I may use non-null assertion here.

packages/webpack/react-webpack-plugin/test/cases/extract-str/basic/index.jsx
```diff
@@ -0,0 +1,36 @@
...
+  );
+  if (__JS__) {
+    expect(fn).toBeCalled();
+    expect(eval('_EXTRACT_STR').includes(['component', 'WillUnmount'].join('')))
+      .toBeTruthy();
+  } else {
+    expect(fn).not.toBeCalled();
+    const arr = eval('_EXTRACT_STR');
+    expect(arr.length).toBeGreaterThan(0);
+    expect(arr.includes(['component', 'WillUnmount'].join(''))).toBeTruthy();
```
hzy: Can we add a test: "Only strings live after minify and tree-shaking will be in `_EXTRACT_STR`"

.changeset/mighty-doodles-repeat.md
```diff
@@ -0,0 +1,6 @@
+---
+"@lynx-js/react-webpack-plugin": patch
+"@lynx-js/react-rsbuild-plugin": patch
+---
+
+feat: support extractStr to reuse variables and reduce bundle size
```
hzy, line 6:
```suggestion
feat: support extractStr to reuse string literal and reduce bundle size
```

And can we add a basic usage example here?

packages/webpack/react-webpack-plugin/src/ReactWebpackPlugin.ts
gaoachao: Yes, if we migrate to the SWC WASM plugin, there will be no `transformBundleResultSync` function exported.

upupming: We can keep the wasm or publish it as a separate package after we migrating to the swc plugin

packages/rspeedy/plugin-react/src/pluginReactLynx.ts
```diff
@@ -350,6 +358,7 @@ export function pluginReactLynx(
     engineVersion: '',
 
     experimental_isLazyBundle: false,
+    extractStr: false,
```
upupming, line 361: @hzy Shall we default to `true` at the present

packages/rspeedy/plugin-react/src/entry.ts
```diff
@@ -105,6 +107,7 @@ export function applyEntry(
       const backgroundEntry = entryName
 
       mainThreadChunks.push(mainThreadName)
+      backgroundChunks.push(backgroundName)
```
colinaaa, line 112: The `backgroundName` might include a hash such as `[contenthash:8]`. I'm uncertain if it will function correctly with `BannerPlugin.test`.

upupming: Thanks for pointing this out, now I have changed to use the following regex (which is used for `RuntimeWrapperWebpackPlugin` now):

https://github.com/lynx-family/lynx-stack/blob/24291512b61cb2cba1838d9ba46384e8c99e265a/packages/rspeedy/plugin-react/src/entry.ts#L202

packages/webpack/react-webpack-plugin/src/ReactWebpackPlugin.ts
```diff
@@ -157,8 +184,7 @@ class ReactWebpackPlugin {
       __PROFILE__: JSON.stringify(
         process.env['REACT_PROFILE'] ?? compiler.options.mode === 'development',
       ),
-      // TODO: config
-      __EXTRACT_STR__: JSON.stringify(false),
+      __EXTRACT_STR__: JSON.stringify(options.extractStr),
```
colinaaa: Should this be a true/false value?
```suggestion
      __EXTRACT_STR__: JSON.stringify(Boolean(options.extractStr)),
```

packages/webpack/react-webpack-plugin/src/ReactWebpackPlugin.ts
```diff
@@ -41,13 +43,24 @@ interface ReactWebpackPluginOptions {
...
    * Whether to enable lazy bundle.
    *
    * @alpha
    */
   experimental_isLazyBundle?: boolean;
+  /**
+   * Merge same string literals in JS and Lepus to reduce output bundle size.
+   * Set to `false` to disable.
+   *
+   * @defaultValue `{ strLength: 20 }`
```
colinaaa: This is not the same as the real default value.

packages/webpack/react-webpack-plugin/src/ReactWebpackPlugin.ts
```diff
@@ -224,6 +249,106 @@ class ReactWebpackPlugin {
             });
         },
       );
+      if (options.extractStr) {
+        compilation.hooks.processAssets.tap(
+          {
+            name: 'compilation',
```
colinaaa: nitpick:

```suggestion
            name: this.constructor.name,
```

packages/webpack/react-webpack-plugin/src/ReactWebpackPlugin.ts
```diff
@@ -224,6 +249,106 @@ class ReactWebpackPlugin {
...
+                }
+              },
+            );
+            jsAssets.forEach((key) => {
+              const entryIdx = options.backgroundChunks?.findIndex(
+                (chunkName) => (
+                  new RegExp(
+                    chunkName
+                      // escape
+                      .replace(/[-[\]{}()*+?.,\\^$|#\s]/g, '\\$&')
```
colinaaa: What are we escaping here? And why do we need to escape?

packages/webpack/react-webpack-plugin/src/ReactWebpackPlugin.ts
```diff
@@ -224,6 +249,106 @@ class ReactWebpackPlugin {
...
+                    new compiler.webpack.sources.SourceMapSource(
+                      result.code,
+                      key,
+                      result.map!,
+                      assets[key]!.source(),
+                      assets[key]!.map()!,
+                    ),
+                  );
+                  if (result.selectStrVec) {
+                    entryIdx2selectStrVec[entryIdx] = result.selectStrVec;
```
colinaaa: Should we only updateAsset when `result.selectStrVec` exists?

packages/webpack/react-webpack-plugin/src/ReactWebpackPlugin.ts
```diff
@@ -224,6 +249,106 @@ class ReactWebpackPlugin {
...
+              (key) => {
+                const entryIdx = options.mainThreadChunks?.findIndex(
+                  (chunkName) => key.includes(chunkName),
+                );
+                if (entryIdx !== undefined && entryIdx !== -1) {
+                  const result = transformBundleResultSync(
+                    assets[key]!.source().toString(),
+                    {
+                      filename: key,
+                      pluginName: 'transformBundleResult',
```
colinaaa: Can we use a different `pluginName` from the one used in the background? This way, if there are errors or warnings, we can differentiate them through the error messages.

packages/webpack/react-webpack-plugin/src/ReactWebpackPlugin.ts
```diff
@@ -224,6 +249,106 @@ class ReactWebpackPlugin {
...
+            const jsAssets = Object.keys(assets).filter(name =>
+              name.endsWith('.js')
+            );
+            jsAssets.forEach(
+              (key) => {
+                const entryIdx = options.mainThreadChunks?.findIndex(
+                  (chunkName) => key.includes(chunkName),
+                );
+                if (entryIdx !== undefined && entryIdx !== -1) {
+                  const result = transformBundleResultSync(
```
colinaaa: Should we also address `result.warning` and `result.error` if they occur?

packages/webpack/react-webpack-plugin/src/ReactWebpackPlugin.ts
```diff
@@ -224,6 +249,106 @@ class ReactWebpackPlugin {
             });
         },
       );
+      if (options.extractStr) {
```
colinaaa: nitpick: I think it would be great if you can write these code related with extractStr to a standalone file. E.g.: `ExtractStrWebpackPlugin.ts` so that it could be tested standalone

packages/webpack/react-webpack-plugin/src/ReactWebpackPlugin.ts
upupming: - we want to match `.rspeedy/main/background.8b9b08d9.js` with `.rspeedy/main/background.[contenthash:8].js`
- I simply construct a regex `.rspeedy/main/background..*.js`, but we need to escape the original string, for example escape `.rspeedy` to `\.rspeedy` and `.js` to `\.js`

packages/webpack/react-webpack-plugin/src/ReactWebpackPlugin.ts
hzy: if we use wasm plugin, we may need to dep a @swc/core to load it...

I prefer "publish it as a separate package", or we can just use the published one...

upupming:
See #391 instead

PR #342, opened 2025-03-25
feat(source-maps): Add `debugIds` option

krystofwoldrich:
## Summary

I've noticed that `rspeedy` is using SourceMapDevToolPlugin to generate source maps, but it doesn't expose the `debugIds` option. This option is helpful when debugging production errors as it links correct source map to the correct bundle. More at https://github.com/tc39/ecma426/blob/main/proposals/debug-id.md

I've tried the change locally with a new Lynx project and the debugIds comments are generated as expected above existing `sourcemappingurl`.

I'm curious if you would be open to add this option and what needs to be done to do so.

## Checklist


packages/rspeedy/core/src/config/output/index.ts
```diff
@@ -335,4 +335,14 @@ export interface Output {
...
+
+  /**
+   * The `debugIds` option controls whether to add debug IDs to the output files.
+   *
+   * @remarks
+   * When enabled, Rspeedy will add debug IDs to the output files to help with debugging.
+   * This is useful when you need to track down issues in production builds.
+   * See the {@link https://github.com/tc39/ecma426/blob/main/proposals/debug-id.md | TC39 sourcemap debug ID proposal} for more details.
+   */
+  debugIds?: boolean | undefined
```
colinaaa: We already have the `output.sourceMap.js` option, I think it would be better to directly use `-debugids` here just like Rspack/Webpack does.

```js
export default {
  output: {
    sourceMap: { js: 'source-map-debugids' },
  },
}
```

> Note that we would like to keep the configuration as same as Rsbuild. So I'm not sure if adding `output.debugIds` is necessary.

packages/rspeedy/core/src/plugins/sourcemap.plugin.ts
```diff
@@ -99,6 +109,7 @@ function applySourceMapPlugin(
       noSources,
       namespace: output?.devtoolNamespace,
       publicPath,
+      debugIds,
```
colinaaa, line 103: How about implementing it similarly to how it's done in Webpack (https://github.com/webpack/webpack/pull/18947)?

```js
const debugIds = devtool.includes('debugids')
```

packages/rspeedy/core/src/plugins/index.ts
```diff
@@ -64,7 +64,7 @@ export async function applyDefaultPlugins(
     ),
 
     import('./sourcemap.plugin.js').then(({ pluginSourcemap }) =>
-      pluginSourcemap()
+      pluginSourcemap(config.output)
```
colinaaa: Would you please add some tests in `packages/rspeedy/core/test/plugins/sourcemap.plugin.test.ts`?

colinaaa (commented):
This looks GREAT to me! I’ve added a few comments to better align it with Rspack.

krystofwoldrich:
Thank you for the feedback. It makes sense to me as well to align with webpack impl. 

I'll also add the tests.

packages/rspeedy/core/src/config/rsbuild/index.ts
```diff
@@ -50,7 +50,8 @@ export function toRsbuildConfig(
 
       polyfill: 'off',
 
-      sourceMap: config.output?.sourceMap,
+      // TODO: update the Rsbuild type to allow `sourceMap.js` to be `*-debugids` // cSpell:disable-line
```
krystofwoldrich: I'm not sure how to better handle the types, since Rsbuild is a project from a different reposity. I wanted to keep the change local to the lynx-stack.

colinaaa: I'm fine with this. We could remove this when Rspack/Rsbuild updates their types

colinaaa (dismissed):
LGTM! Thanks

Looks like the CI is failing. You need to:

1. Format the code with `pnpm dprint fmt`
2. Update API report with `pnpm turbo api-extractor -- --local`. See: https://github.com/lynx-family/lynx-stack/blob/main/CONTRIBUTING.md#update-api-references

krystofwoldrich:
@colinaaa Thank you for the help.

✅ formatted and updated the API.

colinaaa (approved):
Thanks!

packages/rspeedy/core/etc/rspeedy.api.md
```diff
@@ -284,7 +284,7 @@ export interface Source {
 
 // @public
 export interface SourceMap {
-    js?: Rspack.DevTool | undefined;
+    js?: Rspack.DevTool | undefined | `${Exclude<Rspack.DevTool, false | 'eval'>}-debugids`; // cSpell:disable-line
```
colinaaa: This file is auto-generated. Any change will be rejected by CI.

I think we should add `debugids` to `cspell.jsonc`.

krystofwoldrich: Ahh, thanks. I thought the CI wouldn't be happy.

krystofwoldrich: It should be fixed now.

colinaaa (approved):

Issue #140, opened 2025-03-09
[Feature]: Support for Hot Module Reloading on Lynx for Web


elliotnash:
### What problem does this feature solve?

Lynx on mobile devices currently supports HMR, which provides a fast feedback loop during development. However, Lynx for Web currently lacks this feature, making the development process slower and less efficient compared to mobile development. Adding HMR support to Lynx for Web would improve the developer experience and productivity when building web applications with Lynx.

### What does the proposed API of configuration look like?

N/A

elliotnash:
bump

sheepig:
您好，已收到您的信件！将尽快给您回复

PR #352, opened 2025-03-26
fix(rspeedy/core): chunk loading on Web platform

colinaaa:
## Summary

We can directly use the existing [`import-scripts`](https://rspack.dev/config/output#outputchunkloading) chunk loading runtime of Rspack/webpack on Web platform.

We also set `target: 'web'` to use the default [`chunkFormat`](https://rspack.dev/config/output#outputchunkformat) for Web.

> [!NOTE]
> I'm not sure if we should use `'web'` or `'webworker'` as the default value :) 

issue: #140

## Checklist


packages/rspeedy/core/src/plugins/target.plugin.ts
```diff
@@ -5,13 +5,22 @@
...
     name: 'lynx:rsbuild:target',
     setup(api) {
-      api.modifyBundlerChain((options) => {
-        options.target([getESVersionTarget()])
+      api.modifyBundlerChain((options, { environment }) => {
+        if (isWeb(environment)) {
+          options.target([
+            getESVersionTarget(),
+            // Add `target: 'web'` to make Rsbuild inject HMR related code.
+            'web',
```
colinaaa, line 19: We will keep using `'web'` instead of `'webworker'` since Rsbuild would currently only inject HMR entry on `'web'`.

See: https://github.com/web-infra-dev/rsbuild/blob/09e2b8eaa80965c3ff6c93fe72a0942e5b473e69/packages/core/src/server/compilationMiddleware.ts#L16

PupilTong (approved):

PR #371, opened 2025-03-27
fix: merge different chunk groups for same output filename

### LINKED ISSUE #370: [Bug]: layz bundle imported different times in main and background thread will cause incomplete bundle
The following usage will generate different async chunks groups (`./Foo.jsx` and `./Foo.jsx|./Foo.jsx`) of Foo.jsx:

```ts
const ans = await import('./Foo.jsx')
console.log('ans', ans)
if (__JS__) {
	const ans1 = await import('./Foo.jsx')
	console.log('ans1', ans1)
}
```

> This is because `Foo.jsx`  is imported 1 times in the main-thread, and 2 times in the background thread

Image

Image

Thus, we will only generate the first chunk group into the `output/.rspeedy/async/Foo.jsx/tasm.json`, and skip the second chunk group as it has the same filename output.

We will get an output tasm.json with no background.js:

Image

upupming:
## Summary

Fixes #370

## Checklist


packages/webpack/template-webpack-plugin/src/LynxTemplatePlugin.ts
```diff
@@ -681,23 +681,29 @@ class LynxTemplatePluginImpl {
...
-            .filter(chunkName => chunkName !== undefined);
-
-        const filename = Array.from(new Set(chunkNames)).join('_');
-
-        // If no filename is found, avoid generating async template
-        if (!filename) {
-          return;
-        }
+    // merge chunk groups with the same output `filename`
+    const filename2ChunkGroups: Record<string, ChunkGroup[]> = {};
```
colinaaa: By the screenshot you provided in the issue, it seems like an issue of the `getAsyncChunkGroups` method that we did not successfully `groupBy` the two background and main-thread `chunkGroup`.

So I think it would be reasonable to fix the problem there. Maybe you can group by `cg.name` instead of `origin.imports`.

colinaaa (commented):
Could you also add some tests to prevent future regression?

packages/webpack/react-webpack-plugin/test/cases/code-splitting/async-chunk-groups/rspack.config.js
```diff
@@ -0,0 +1,60 @@
...
+            expect(args.encodeData.lepusCode.root.name).toBeOneOf(
+              [ // main entry
+                'main:main-thread.js',
+                // foo.js lazy bundle
+                '.rspeedy/async/./foo.js-react:main-thread.js',
+              ],
+            );
+            expect(
+              Object.keys(args.encodeData.manifest).length,
+            ).toBe(1);
```
upupming, line 47: @colinaaa Here we assert the `root` and `manifest` must exist.

colinaaa (approved):
LGTM! Now the `groupBy` looks pretty nice

PR #372, opened 2025-03-27
fix(web): rsbuild will bundle 2 exactly same chunk for two same `new Worker` stmt

PupilTong:
the bundle size will be optimized about 28.2KB

### DIFF
--- a/.changeset/rotten-pigs-wash.md
+++ b/.changeset/rotten-pigs-wash.md
@@ -1,0 +1,7 @@
+---
+"@lynx-js/web-core": patch
+---
+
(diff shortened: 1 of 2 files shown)


.changeset/rotten-pigs-wash.md
```diff
@@ -0,0 +1,7 @@
+---
+"@lynx-js/web-core": patch
+---
+
+fix(web): rsbuild will bundle 2 exactly same chunk for two same `new Worker` stmt
+
+the bundle size will be optimized about 28.2KB
```
colinaaa, line 7: This sounds like this patch will add 28.2 kB to the bundle size :)

colinaaa (approved):

Issue #304, opened 2025-03-22
[Feature]: Support more CLI flags


colinaaa:
### What problem does this feature solve?

Enhance the flexibility of the Rspeedy CLI.

### What does the proposed API of configuration look like?

The same as Rsbuild: https://rsbuild.dev/guide/basic/cli#common-flags

### `--base`

Specify the base path of the server, see `server.base`

### `--mode`

Specify the build mode (`development`, `production` or `none`), see [`mode`](https://lynxjs.org/api/rspeedy/rspeedy.config.mode.html)

### `--environment`

Specify the name of environment to build, see [Build specified environment](https://rsbuild.dev/guide/advanced/environments#build-specified-environment)

### `--env-mode`

Following up to #230, specify the env mode to load the `.env.[mode]` file

colinaaa:
+`--no-env`

Ref:

- https://github.com/lynx-family/lynx-stack/pull/233#discussion_r2027156364
- https://github.com/web-infra-dev/rsbuild/pull/4947

PR #387, opened 2025-03-28
feat(rspeedy/core): support cli flag --base

gaoachao:
## Summary

As one of several PRs addressing issue #304

Support cli flag --base to specify the base path of the server.

## Checklist

### DIFF
--- a/.changeset/modern-keys-spend.md
+++ b/.changeset/modern-keys-spend.md
@@ -1,0 +1,5 @@
+---
+"@lynx-js/rspeedy": patch
+---
+
+Support cli flag `--base` to specify the base path of the server.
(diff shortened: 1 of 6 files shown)


packages/rspeedy/core/src/cli/inspect.ts
```diff
@@ -28,6 +29,11 @@ export async function inspect(
       configPath: inspectOptions.config,
     })
 
+    if (inspectOptions.base) {
```
colinaaa: I don't think we need `--base` for `rspeedy inspect` since there wouldn't be any server running.

Should we add it as a common option and use in `dev.ts` and `preview.ts`?

https://github.com/lynx-family/lynx-stack/blob/d7301017a383b8825cdc813a649ef26ce1c37641/packages/rspeedy/core/src/cli/commands.ts#L17-L23

colinaaa (changes requested):
You can also update the documentation located in the `website/` directory.

https://github.com/lynx-family/lynx-stack/blob/main/website/docs/en/guide/cli.md

The Lynx website will be updated during the release process.

colinaaa (approved):
LGTM!

PR #395, opened 2025-03-30
feat(rspeedy/core)!: build with Rslib

colinaaa:
## Summary

Build `@lynx-js/rspeedy` for:

1. Faster start-up time
2. Being able to upgrade dependencies like `typia` (we didn't upgrade before due to breaking changes in transitive peer dependencies)

> [!NOTE]
> We did not move any dependencies to devDependencies in this patch. We will bundle more dependencies in the future PRs. 

## Checklist

- [ ] Tests updated (or **not required**).

### DIFF
--- a/.changeset/every-turkeys-study.md
+++ b/.changeset/every-turkeys-study.md
@@ -1,0 +1,13 @@
+---
+"@lynx-js/rspeedy": minor
+---
+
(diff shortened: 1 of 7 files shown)


gaoachao (approved):

upupming:
@colinaaa It seems this relative path is breaking, let me fix in another PR
image

PR #402, opened 2025-03-31
feat(plugin-react): disable `module.generator.json.JSONParse` option

luhc228:
## Summary

Related [issue](https://github.com/webpack/webpack/issues/19319)

## Checklist

### DIFF
--- a/.changeset/violet-days-beam.md
+++ b/.changeset/violet-days-beam.md
@@ -1,0 +1,5 @@
+---
+"@lynx-js/react-rsbuild-plugin": patch
+---
+
+Disable `module.generator.json.JSONParse` option as it increases the bundle size of `main-thread.js`. For more detail, please see this [issue](https://github.com/webpack/webpack/issues/19319).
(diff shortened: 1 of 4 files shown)


packages/rspeedy/core/src/plugins/generator.plugin.ts
```diff
@@ -0,0 +1,20 @@
...
+export function pluginGenerator(): RsbuildPlugin {
+  return {
+    name: 'lynx:rsbuild:generator',
+    async setup(api) {
+      api.modifyBundlerChain(chain => {
+        // Avoid generating `JSON.parse()` for a JSON file which has more than 20 characters as it increases the bundle size of `main-thread.js`.
+        // For more detail, see https://github.com/webpack/webpack/issues/19319
+        chain.module
+          .generator.set('json', {
+            JSONParse: false,
```
colinaaa: I think we should only set `JSONParse: false` on the main-thread (adding `layer` in this rule). Since the `background.js` is not compiled to bytecode and it should be faster to use `JSON.parse`.

Maybe you should add this rule in `@lynx-js/react-rsbuild-plugin`.

luhc228: Done.

colinaaa (approved):
LGTM!

PR #414, opened 2025-03-31
feat(web): add dispatchNapiModules of onNapiModulesCall

Sherry-hue:
## Summary

feat: onNapiModulesCall function add new param: `dispatchNapiModules`, napiModulesMap val add new param: `handleDispatch`.

Now you can use them to actively communicate to napiModules (background thread) in onNapiModulesCall (ui thread).

## Checklist


packages/web-platform/web-core/src/uiThread/crossThreadHandlers/registerNapiModulesCallHandler.ts
```diff
@@ -13,6 +15,17 @@ export function registerNapiModulesCallHandler(
...
+    (
+      name: string,
+      data: Cloneable,
+      moduleName: string,
+    ) => {
+      return napiModulesCall(
+        name,
+        data,
+        moduleName,
+        (data) => rpc.invoke(dispatchNapiModuleEndpoint, [data]),
```
PupilTong: How about createCall and always use the same function

PupilTong (approved):

PR #428, opened 2025-04-01
fix: relative require path of WebSocketClient.js

upupming:
## Summary

Fix the path issue of #395

## Checklist


packages/rspeedy/core/src/plugins/dev.plugin.ts
```diff
@@ -128,7 +137,13 @@ export function pluginDev(
                   options?.client?.websocketTransport ?? require.resolve('@lynx-js/websocket'),
                   'default',
                 ],
-                __webpack_dev_server_client__: [require.resolve('../../client/hmr/WebSocketClient.js'), 'default'],
+                __webpack_dev_server_client__: [
+                  require.resolve(
+                    path.join(
```
colinaaa: Should we use `path.posix.join` to make it work on Windows?

packages/rspeedy/core/src/plugins/dev.plugin.ts
```diff
@@ -83,12 +85,19 @@ export function pluginDev(
...
+      const __dirname = path.dirname(fileURLToPath(import.meta.url))
 
       api.modifyBundlerChain((chain, { isProd }) => {
         if (isProd) {
           return
         }
         const rsbuildPath = require.resolve('@rsbuild/core')
+        const rspeedyDir = path.dirname(
+          require.resolve('@lynx-js/rspeedy/package.json', {
+            paths: [__dirname],
```
colinaaa: Why do we need this?

I think `require.resolve` starts from current directory by default

packages/rspeedy/core/src/plugins/dev.plugin.ts
upupming: It seems the path passed to `require.resolve` does not need to use `\\` for windows

colinaaa: Yes, so we have to use `path.posix.join` instead of `path.join`.

Otherwise, it would be `<rspeedy-dir>\client/hmr/WebSocketClient.js` on Windows

upupming: Oh yes, it seems that using `rspeedyDir` as `paths` options will be safer, because `<rspeedy-dir>` may contain windows `\` too. I have updated the code

colinaaa (approved):
LGTM!

PR #434, opened 2025-04-01
fix(react/runtime): ensure `ref` lifecycle events run after `firstScreen`

Yradex:
## Summary

This PR fixes a critical issue in the React runtime's lifecycle event handling, specifically in the background thread async render mode. The problem was that ref lifecycle events were being triggered before firstScreen events, which could lead to undefined refs when components try to access them.

The root cause was in the event ordering in the background thread. In the async render mode, we need to ensure that firstScreen events complete before ref events are processed, as components may depend on the firstScreen state being fully initialized.

The fix ensures proper event ordering by:
1. Modifying the lifecycle event delay mechanism
2. Updating the ref event handling to wait for firstScreen completion
3. Adding comprehensive test cases to verify the correct event sequence

This change improves the reliability of component initialization and ref handling in the background thread, preventing potential undefined ref errors that could occur during component mounting.

## Checklist

### DIFF
--- a/.changeset/petite-bobcats-travel.md
+++ b/.changeset/petite-bobcats-travel.md
@@ -1,0 +1,7 @@
+---
+"@lynx-js/react": patch
+---
+
+fix: ensure ref lifecycle events run after firstScreen in the background thread
+
+This patch fixes an issue where ref lifecycle events were running before firstScreen events in the background thread async render mode, which could cause refs to be undefined when components try to access them.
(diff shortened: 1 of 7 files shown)


.changeset/petite-bobcats-travel.md
```diff
@@ -0,0 +1,7 @@
+---
+"@lynx-js/react": patch
+---
+
+fix: ensure ref lifecycle events run after firstScreen in the background thread
+
+This patch fixes an issue where ref lifecycle events were running before firstScreen events in the background thread async render mode, which could cause refs to be undefined when components try to access them.
```
hzy, line 7: > an issue where ref lifecycle events were running before firstScreen events

Can I assume that this only happens when using ref inside a `<list/>`?

hzy (approved):

.changeset/petite-bobcats-travel.md
Yradex: ✅

PR #435, opened 2025-04-01
fix: add enableCSSInvalidation for encodeCSS of css HMR

upupming:
## Checklist


colinaaa (approved):
LGTM. But I think we need to find a way in the future to use the encode options of this compilation.

PR #438, opened 2025-04-01
fix(timing): The pipeline timing is not described correctly

partholon:
## Summary

lynx.performance.OnPipelineStart add param pipeline_options to convert more info like pipelineOrigin, dsl and stage.

When the pipeline is started by the framework, actively mark the start and end time of MtsRender.

Add FrameworkRenderingTiming typing

## Checklist


packages/react/runtime/src/lynx/performance.ts
```diff
@@ -22,8 +22,19 @@ enum PerformanceTimingKeys {
   patch_changes_end,
   hydrate_parse_snapshot_start,
   hydrate_parse_snapshot_end,
+  mtsRenderStart,
```
colinaaa, line 25: Can we use the same naming convention?
```suggestion
  mts_render_start,
```

packages/react/runtime/src/lynx/performance.ts
```diff
@@ -22,8 +22,19 @@ enum PerformanceTimingKeys {
...
+  mtsRenderEnd,
```
colinaaa, line 26:
```suggestion
  mts_render_end,
```

packages/react/runtime/src/lynx/performance.ts
```diff
@@ -22,8 +22,19 @@ enum PerformanceTimingKeys {
...
 }
 
+const PerformanceTimingFlags = {
+  reactLynxHydrate: 'react_lynx_hydrate',
+};
```
colinaaa: To make it more strict in type, use `as const`

```suggestion
} as const;
```

packages/react/runtime/src/lynx/performance.ts
```diff
@@ -22,8 +22,19 @@ enum PerformanceTimingKeys {
...
+
+const PipelineOrigins = {
+  reactLynxHydrate: 'reactLynxHydrate',
+  updateTriggeredByBts: 'updateTriggeredByBts',
+};
```
colinaaa: ditto
```suggestion
} as const;
```

packages/react/runtime/src/lynx/performance.ts
```diff
@@ -65,11 +76,18 @@ function markTimingLegacy(key: PerformanceTimingKeys, timingFlag_?: string): voi
   lynx.getNativeApp().markTiming?.(timingFlag!, PerformanceTimingKeys[key]);
 }
 
-function beginPipeline(needTimestamps: boolean, timingFlag?: string): void {
+function beginPipeline(needTimestamps: boolean, pipelineOrigin: string, timingFlag?: string): void {
```
colinaaa: Use strict type instead
```suggestion
function beginPipeline(needTimestamps: boolean, pipelineOrigin: PipelineOrigins[keyof PipelineOrigins], timingFlag?: string): void {
```

packages/react/runtime/src/lynx/performance.ts
```diff
@@ -65,11 +76,18 @@ function markTimingLegacy(key: PerformanceTimingKeys, timingFlag_?: string): voi
...
   globalPipelineOptions = lynx.performance?._generatePipelineOptions?.();
   if (globalPipelineOptions) {
     globalPipelineOptions.needTimestamps = needTimestamps;
-    lynx.performance?._onPipelineStart?.(globalPipelineOptions.pipelineID);
+    globalPipelineOptions.pipelineOrigin = pipelineOrigin;
+    globalPipelineOptions.dsl = 'reactLynx';
+    if (pipelineOrigin === PipelineOrigins.reactLynxHydrate) {
```
colinaaa: Use switch-case instead. So that we can make sure that we already handling all possible values of `PipelineOrigins`

colinaaa (changes requested):

packages/react/runtime/src/lynx/performance.ts
partholon: There is a topic here. The SDK Performance API uses camel case style, such as "[PipelineEntry](https://lynxjs.org/zh/api/lynx-api/performance-api/performance-entry/pipeline-entry.html)." Should we modify the timing keys in the framework to camel case as well, and let the SDK ensure compatibility with the old interfaces? (However, this will cause an issue where the new version of the framework combined with the old version of the SDK will lead to changes in the parameter style received by the client.)

partholon: Based on the discussion, we will make a breaking change to performance tracking: change all tracking point names to the camel case style expected by the Performance API.

colinaaa (dismissed):
LGTM

Yradex (dismissed):

hzy (approved):

colinaaa (approved):

PR #472, opened 2025-04-03
feat(web-elements): integrate web-security for input sanitization

BronzonTech-Cloud:
## Summary
This change integrates the web-security module with web-elements to provide automatic input sanitization, creating a robust defense against XSS attacks at the input level.

### Dependencies Added
- Enhanced the `package.json` of `web-elements` to include dependencies on:
  - `@lynx-js/web-constants`
  - `@lynx-js/web-core`
  - `@lynx-js/web-security`
- Introduced corresponding entries in `pnpm-lock.yaml` to ensure consistent dependency versions

### Technical Changes
1. **Added `sanitizeInput` Function Integration**
   - Imported the `sanitizeInput` function from `@lynx-js/web-security` module
   - Integrated the function into both `XInputEvents` and `XTextareaEvents` classes
   - Applied sanitization at the input source to prevent XSS attacks

2. **Event Handling Improvements**
   - Added event interception to prevent raw input events from propagating
   - Implemented the `blockHtmlEvent` method to stop original events
   - Created sanitized event dispatching to ensure only clean data is processed
   - Enhanced type safety through TypeScript types for events

3. **Sanitization Configuration**
   - Added support for configurable sanitization levels via attributes
   - Implemented default sanitization behavior that can be overridden
   - Created performance optimizations to minimize processing overhead

4. **Error Handling**
   - Added robust error handling to gracefully manage sanitization failures
   - Implemented logging for potentially dangerous input attempts

## Security Benefits
- **Automatic Protection**: All user input is automatically sanitized without requiring manual developer intervention
- **XSS Prevention**: Strips potentially dangerous content like script tags, event handlers, and JavaScript URIs
- **Defense in Depth**: Works in conjunction with other security measures like CSP for multi-layered protection
- **Consistent Security**: Ensures all input elements follow the same security standards across the application

## Usage Example
```typescript
// Example of how sanitized input is now handled automatically
const inputElement = document.createElement('x-input');
document.body.appendChild(inputElement);

// Behind the scenes, when a user types:
// <script>alert('xss')</script>
// The sanitized value will be:
// alert('xss')

// No additional code required - sanitization happens automatically
```

This implementation significantly improves the security posture of applications using web-elements by sanitizing user input at the source, before it can be processed by the application.

### DIFF
--- a/.changeset/all-melons-fall.md
+++ b/.changeset/all-melons-fall.md
@@ -1,0 +1,11 @@
+---
+"@lynx-js/web-mainthread-apis": patch
+"@lynx-js/web-worker-runtime": patch
+"@lynx-js/web-constants": patch
+"@lynx-js/web-webpack-plugin": patch
+"@lynx-js/web-core": patch
+---
+
+feat: support mts event handler (1/n)
+
+now the main-thread:bind handler could be invoked. The params of the handler will be implemented later.
--- a/packages/web-platform/web-constants/src/constants.ts
+++ b/packages/web-platform/web-constants/src/constants.ts
@@ -20,4 +20,6 @@ export const __lynx_timing_flag = '__lynx_timing_flag' as const;
 
 export const globalMuteableVars = [
   'registerDataProcessor',
+  'registerWorkletInternal',
+  'lynxWorkletImpl',
 ] as const;
--- a/packages/web-platform/web-core/src/utils/loadTemplate.ts
+++ b/packages/web-platform/web-core/src/utils/loadTemplate.ts
@@ -98,6 +98,7 @@ const mainThreadInjectVars = [
   '__SetCSSId',
   '__OnLifecycleEvent',
   '__FlushElementTree',
+  '__LoadLepusChunk',
 ];
 
 const backgroundInjectVars = [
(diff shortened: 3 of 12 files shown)


packages/web-platform/web-elements/src/XTextarea/XTextareaEvents.ts
```diff
@@ -5,122 +5,159 @@
...
-  }
+  #blockHtmlEvent = (e: InputEvent) => {
+    e.stopPropagation();
+    const textareaElement = this.#getTextareaElement() as HTMLTextAreaElement;
+    this.#prevBlockDomEnv.value = textareaElement.value;
+    this.#prevBlockDomEnv.selectionStart = textareaElement.selectionStart;
+    const inputEvent = new CustomEvent('input', {
+      ...commonComponentEventSetting,
+      detail: {
+        value: sanitizeInput(this.#prevBlockDomEnv.value), // Sanitize textarea input
```
PupilTong: Could you explain where is the vulnerability? Developer's Lynx code is running in a web worker. It cannot access the DOM therefore the `innerHTML` cannot be assigned by the lynx code.

PR #502, opened 2025-04-07
fix(rspeedy): duplicate chunks in chunkSplit with multiple entries

gaoachao:
## Summary

Fixed an issue in rspeedy where chunks weren't being properly shared between multiple entries during chunk splitting. This resulted in duplicate chunks being generated, increasing bundle size and reducing performance.

> [!CAUTION]
> **BREAKING CHANGE**:  `encodeData.manifest` now only contains the `/app-service.js` entry.

### TODO
Add support for synchronous chunk loading.

## Checklist

### DIFF
--- a/.changeset/light-donuts-sin.md
+++ b/.changeset/light-donuts-sin.md
@@ -1,5 +1,5 @@
 ---
-"@lynx-js/template-webpack-plugin": patch
+"@lynx-js/template-webpack-plugin": minor
 ---
 
 Fixed an issue in rspeedy where chunks weren't being properly shared between multiple entries during chunk splitting. This resulted in duplicate chunks being generated, increasing bundle size and reducing performance.
--- a/.changeset/light-donuts-sin.md
+++ b/.changeset/light-donuts-sin.md
@@ -4,4 +4,4 @@
 
 Fixed an issue in rspeedy where chunks weren't being properly shared between multiple entries during chunk splitting. This resulted in duplicate chunks being generated, increasing bundle size and reducing performance.
 
-**BREAKING CHANGE**: Exclude non-background files from the `encodeData.manifest`.
+**BREAKING CHANGE**: `encodeData.manifest` now only contains the `/app-service.js` entry.
--- a/.changeset/light-donuts-sin.md
+++ b/.changeset/light-donuts-sin.md
@@ -1,0 +1,7 @@
+---
+"@lynx-js/template-webpack-plugin": patch
+---
+
+Fixed an issue in rspeedy where chunks weren't being properly shared between multiple entries during chunk splitting. This resulted in duplicate chunks being generated, increasing bundle size and reducing performance.
+
+**BREAKING CHANGE**: Exclude non-background files from the `encodeData.manifest`.
--- a/packages/webpack/template-webpack-plugin/src/LynxEncodePlugin.ts
+++ b/packages/webpack/template-webpack-plugin/src/LynxEncodePlugin.ts
@@ -241,7 +241,7 @@ export class LynxEncodePluginImpl {
     if (this.#isBackground(name)) {
       return `/${name}`;
     }
-    return new URL(name, publicPath).toString();
+    return publicPath + name;
   }
 
   #isBackground(name: string): boolean {
--- a/packages/webpack/template-webpack-plugin/src/LynxEncodePlugin.ts
+++ b/packages/webpack/template-webpack-plugin/src/LynxEncodePlugin.ts
@@ -146,10 +146,16 @@ export class LynxEncodePluginImpl {
         const { encodeData } = args;
         const { manifest } = encodeData;
 
-        const publicPath =
-          typeof compilation?.outputOptions.publicPath === 'function'
-            ? compilation.outputOptions.publicPath({})
-            : compilation?.outputOptions.publicPath ?? '';
+        let publicPath = '';
+        if (typeof compilation?.outputOptions.publicPath === 'function') {
+          compilation.errors.push(
+            new compiler.webpack.WebpackError(
+              '`publicPath` as a function is not supported yet.',
+            ),
+          );
+        } else {
+          publicPath = compilation?.outputOptions.publicPath ?? '';
+        }
 
         if (!isDebug() && !isDev && !isRsdoctor()) {
           [
@@ -174,8 +180,6 @@ export class LynxEncodePluginImpl {
           //     lynx.requireModule('initial-chunk1')
           //     lynx.requireModule('initial-chunk2')
           //   `,
-          //   'initial-chunk1': `<content-of-initial-chunk>`,
-          //   'initial-chunk2': `<content-of-initial-chunk>`,
           // },
           // ```
           '/app-service.js': [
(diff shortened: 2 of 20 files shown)


packages/webpack/template-webpack-plugin/src/LynxEncodePlugin.ts
```diff
@@ -228,8 +235,15 @@ export class LynxEncodePluginImpl {
...
-    return `/${name}`;
+  #formatJSName(name: string, publicPath: string): string {
+    if (this.#isBackground(name)) {
+      return `/${name}`;
+    }
+    return publicPath + name;
+  }
+
+  #isBackground(name: string): boolean {
+    return /\.rspeedy\/[^/]+\/background(?:\.[a-f0-9]+)?\.js$/.test(name);
```
upupming: We should read `.rspeedy` from `distPath.intermediate`, see:

https://github.com/lynx-family/lynx-stack/blob/530da75ba11c0f1584e490da24ebe5fd1ed8f34a/packages/rspeedy/core/src/config/output/dist-path.ts#L22

gaoachao: https://github.com/lynx-family/lynx-stack/pull/502/commits/1ac35e9448b5be65e85426df9feaf79f83b4e539

.changeset/light-donuts-sin.md
```diff
@@ -0,0 +1,7 @@
+---
+"@lynx-js/template-webpack-plugin": patch
```
colinaaa: We should use `minor` instead of `patch` for breaking change
```suggestion
"@lynx-js/template-webpack-plugin": minor
```

packages/webpack/template-webpack-plugin/src/LynxEncodePlugin.ts
```diff
@@ -129,6 +129,11 @@ export class LynxEncodePluginImpl {
         const { encodeData } = args;
         const { manifest } = encodeData;
 
+        const publicPath =
+          typeof compilation?.outputOptions.publicPath === 'function'
+            ? compilation.outputOptions.publicPath({})
```
colinaaa: It doesn't make sense to pass an empty object to `publicPath`.

You may throw an error saying that we does not support using function as public path

We already have some TODO
https://github.com/lynx-family/lynx-stack/blob/94b1cb44ee136a93bb20ad9bd877dc4abfaf1fbd/packages/webpack/template-webpack-plugin/src/LynxTemplatePlugin.ts#L852

packages/webpack/template-webpack-plugin/src/LynxEncodePlugin.ts
```diff
@@ -228,8 +235,15 @@ export class LynxEncodePluginImpl {
...
-    return `/${name}`;
+  #formatJSName(name: string, publicPath: string): string {
+    if (this.#isBackground(name)) {
+      return `/${name}`;
+    }
+    return publicPath + name;
+  }
+
+  #isBackground(name: string): boolean {
+    return /\/background(?:\.[a-f0-9]+)?\.js$/.test(name);
```
colinaaa: It does not make sense to use `background.js` or `background.[hash].js` since:

1. The name can be customized by the framework.
2. `background.js` has no difference with other chunks(except it has webpack runtime). I don't think we need to make it special

packages/webpack/template-webpack-plugin/src/LynxEncodePlugin.ts
```diff
@@ -161,17 +166,19 @@ export class LynxEncodePluginImpl {
             Object.keys(manifest)
               .map((name) =>
                 `module.exports=lynx.requireModule('${
```
colinaaa, line 189: Using `lynx.requireModuleAsync` is preferable when dealing with multiple chunks, as it allows for parallel fetching of dependent chunks.

gaoachao: I will address this in a future PR.

packages/webpack/template-webpack-plugin/src/LynxEncodePlugin.ts
gaoachao: https://github.com/lynx-family/lynx-stack/pull/502/commits/e04f9e5a7d3d2e92b0d85240459723eb104637d9

packages/webpack/template-webpack-plugin/src/LynxEncodePlugin.ts
gaoachao: https://github.com/lynx-family/lynx-stack/pull/502/commits/e04f9e5a7d3d2e92b0d85240459723eb104637d9

.changeset/light-donuts-sin.md
```diff
@@ -0,0 +1,7 @@
+---
+"@lynx-js/template-webpack-plugin": minor
```
colinaaa, line 2: We would also need to update the `peerDependencies` of other plugins.

Otherwise, there would be an unexpected **major** bump. See: https://github.com/lynx-family/lynx-stack/pull/502#issuecomment-2784181504

packages/webpack/template-webpack-plugin/src/LynxEncodePlugin.ts
```diff
@@ -152,27 +163,19 @@ export class LynxEncodePluginImpl {
...
-                  this.#formatJSName(name)
+                  this.#formatJSName(name, publicPath)
                 }',globDynamicComponentEntry?globDynamicComponentEntry:'__Card__')`
               )
               .join(','),
             this.#appServiceFooter(),
           ].join(''),
-          ...(Object.fromEntries(
```
colinaaa, line 187: directly removing all the manifest would be a huge breaking change (and I don't think we are ready for that).

So maybe add an option like [`output.inlineScripts`](https://rsbuild.dev/config/output/inline-scripts)?

packages/webpack/template-webpack-plugin/src/LynxEncodePlugin.ts
```diff
@@ -152,27 +163,19 @@ export class LynxEncodePluginImpl {
...
           //   `,
-          //   'initial-chunk1': `<content-of-initial-chunk>`,
-          //   'initial-chunk2': `<content-of-initial-chunk>`,
           // },
           // ```
           '/app-service.js': [
             this.#appServiceBanner(),
             Object.keys(manifest)
               .map((name) =>
                 `module.exports=lynx.requireModule('${
```
colinaaa, line 189: Consider using `lynx.requireModuleAsync` when dealing with multiple chunks, as it allows scripts to load in parallel.

colinaaa, line 189: nitpick: can we avoid the multiple `module.exports`?

colinaaa: Maybe this could be done in future PRs. Using a new option like [`scriptLoading`](https://rsbuild.dev/config/html/script-loading)

colinaaa (changes requested):

gaoachao:
see #874

PR #533, opened 2025-04-14
fix(plugin-react): build background-only files with rslib

gaoachao:
## Summary

Introduced by #346

Include `background-only/error.ts` and `background-only/empty.ts` in `rslib.config.ts`.

## Checklist

### DIFF
--- a/.changeset/lucky-parrots-drum.md
+++ b/.changeset/lucky-parrots-drum.md
@@ -2,4 +2,4 @@
 "@lynx-js/react-rsbuild-plugin": patch
 ---
 
-Include `background-only/error.ts` and `background-only/empty.ts` in `rslib.config.ts` configuration.
+Include `background-only/error.ts` and `background-only/empty.ts` in `rslib.config.ts`.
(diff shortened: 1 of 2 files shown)


upupming (approved):

colinaaa (approved):

packages/rspeedy/plugin-react/rslib.config.ts
```diff
@@ -10,6 +10,8 @@ export default defineConfig({
       'loaders/ignore-css-loader': './src/loaders/ignore-css-loader.ts',
       'loaders/invalid-import-error-loader':
         './src/loaders/invalid-import-error-loader.ts',
+      'background-only/error': './src/background-only/error.ts',
```
colinaaa, line 13: Can we use the real [`background-only`](https://www.npmjs.com/background-only) npm package now?

.changeset/lucky-parrots-drum.md
```diff
@@ -0,0 +1,5 @@
+---
+"@lynx-js/react-rsbuild-plugin": patch
+---
+
+Include `background-only/error.ts` and `background-only/empty.ts` in `rslib.config.ts`.
```
colinaaa, line 5: Maybe it's better to use something like "fixing `background-only` not found" in CHANGELOG.

PR #541, opened 2025-04-14
fix(web): resize observer loop && x-list tests use background-color to distinguish

Sherry-hue:
## Summary

fix(web): x-list may cause error: 'Resizeobserver loop completed with undelivered notifications'.

detail: https://developer.mozilla.org/en-US/docs/Web/API/ResizeObserver#observation_errors

chore(web): x-list tests use background-color to distinguish, instead of number-text.

## Checklist


packages/web-platform/web-elements/src/XList/XListWaterfall.ts
```diff
@@ -179,9 +179,13 @@ export class XListWaterfall
   #resizeObserverInit = (spanCount: number, isScrollVertical: boolean) => {
     this.#resizeObserver?.disconnect();
     this.#resizeObserver = new ResizeObserver(() => {
-      this.#layoutListItem(
-        spanCount,
-        isScrollVertical,
+      // may cause: Resizeobserver loop completed with undelivered notifications
+      // https://developer.mozilla.org/en-US/docs/Web/API/ResizeObserver#observation_errors
+      requestAnimationFrame(() =>
```
PupilTong: Will this cause an infinite loop? Compare with temporary disabling  the resize observer, what's the pros and cons for using the requestAnimationFrame?

Also do you have a test to test this error?

Sherry-hue: 1. I tested that it does not cause a loop.

2. The problem is that ResizeObserver loop completed with undelivered notifications..
We can execute all resize observer events after the browser reflow, so i use raf.

If we disable the resize observer, we may not know when to unblock it.

3. I will add a test to show how many times the observer is executed

PupilTong (dismissed):

PupilTong (approved):

Issue #394, opened 2025-03-30
[Bug]: Property 'lynx-view' does not exist on type 'JSX.IntrinsicElements'


mcshaman:
### System Info

```
  System:
    OS: macOS 15.3.2
    CPU: (12) arm64 Apple M2 Pro
    Memory: 152.14 MB / 16.00 GB
    Shell: 5.9 - /bin/zsh
  Binaries:
    Node: 23.10.0 - /opt/homebrew/bin/node
    npm: 10.9.2 - /opt/homebrew/bin/npm
    pnpm: 10.4.1 - /opt/homebrew/bin/pnpm
    bun: 1.2.7 - /opt/homebrew/bin/bun
```

### Details

Typescript reporting type error for `linx-view` tag

```
❯ pnpm tsc
src/App.tsx:8:5 - error TS2339: Property 'lynx-view' does not exist on type 'JSX.IntrinsicElements'.

8     <lynx-view style={{ height: "100vh", width: "100vw" }} url="/main.web.bundle">
      ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~

src/App.tsx:9:5 - error TS2339: Property 'lynx-view' does not exist on type 'JSX.IntrinsicElements'.

9     </lynx-view>
      ~~~~~~~~~~~~


Found 2 errors in the same file, starting at: src/App.tsx:8
```

I would have expected that one of the lynx dependencies would provide type definitions for these tags.

### Reproduce link

https://github.com/lynx-family/lynx-examples/tree/main/examples/web-platform

### Reproduce Steps

1. pull repo
2. cd to `./examples/web-platform`
3. run `pnpm tsc`

PupilTong:
Hi @mcshaman,
Thanks for the issue.
This type error won't affect on the behaviour of lynx-view. 
We'll provide an example for the type configuration soon. 
Before that using a `//@ts-expect-error` could be a solution.

FreakeyPlays:
Hey @PupilTong, ignoring the error fixes the Error in the Editor and comping works, but it still throws errors in the Browser Console.
For me, I am getting these errors:

```console
TypeError: lynx.getJSModule is not a function
[...]

Issue #359, opened 2025-03-27
[Feature]: Use standalone `namespace JSX`


colinaaa:
### What problem does this feature solve?

We currently extend the `namespace JSX` within the `module 'react'`.

Refer to this link for details: https://github.com/lynx-family/lynx/blob/ca9b1a29c17f19b7d86d61d09bddb5123d002b64/js_libraries/types/types/common/element/element.d.ts#L40.

However, this approach may lead to conflicts with the types defined in `@types/react`. 

For instance, both `@lynx-js/types` and `@types/react` define `view` under `IntrinsicElements`.

Our goal is to establish a standalone `namespace JSX` that exclusively provides the correct Lynx types.

### What does the proposed API of configuration look like?

Requires [`jsxImportSource`](https://www.typescriptlang.org/tsconfig/#jsxImportSource) to be set to `@lynx-js/react`

```jsonc
{
  "compilerOptions": {
    "jsx": "react-jsx",
    "jsxImportSource": "@lynx-js/react",
  }
}
```

> [!NOTE]
> Note that this might be a breaking change

PR #545, opened 2025-04-15
feat(react): support `"jsx": "react-jsx"`

### LINKED ISSUE #544: [Bug]: monorepo project type error
### System Info

System:
OS: Windows 11 10.0.26100
CPU: (20) x64 12th Gen Intel(R) Core(TM) i7-12700
Memory: 3.31 GB / 15.75 GB
Binaries:
Node: 22.14.0 - D:\nvm\nvm-setup\node.js\node.EXE
npm: 10.9.2 - D:\nvm\nvm-setup\node.js\npm.CMD
pnpm: 10.6.5 - D:\nvm\nvm-setup\node.js\pnpm.CMD

### Details

I created an Integrate Lynx with Existing Apps (Web) using monorepo
The root package was created using `pnpm create rspeedy@latest` and the subpackages were created using `pnpm create rsbuild@latest`, but after installation VSCode reported an error because `ts.config.json` was configured with the wrong `compilerOptions`.
When `“jsx”: “preserve”`, VSCode thinks `<view>` is provided by `react`.

When I changed it to `“jsx”: “react-jsx”`, everything worked fine, and the `@lynx-js/types` type was used correctly.
Image

### Reproduce link

https://github.com/Mori-Yang/lynx-web-template

### Reproduce Steps

1. Run `pnpm i`
2. Find some error messages

Mori-Yang:
## Summary

fix #544 
## Checklist


packages/rspeedy/create-rspeedy/template-react-ts/tsconfig.json
```diff
@@ -1,6 +1,6 @@
 {
   "compilerOptions": {
-    "jsx": "preserve",
+    "jsx": "react-jsx",
```
colinaaa, line 3: I think using `"jsx": "react-jsx"` without `"jsxImportSource"` would also cause type error if using the same `@types/react` with Web projects.

colinaaa (commented):
Thank you for your contribution!

We are already having a issue (https://github.com/lynx-family/lynx-stack/issues/359) for using a standalone `JSX` namespace to prevent conflicts with `@types/react`.

Would you like to work on this? You may add a `namespace JSX` in `packages/react/runtime/jsx-runtime/index.d.ts` just like https://github.com/lynx-family/lynx-examples/blob/df7bb3db9fbd9254ab3c87025fe0eaca0dc2c86c/examples/with-solidjs/packages/solid/jsx-runtime/index.d.ts does.

colinaaa:
> BTW, I noticed that the web side doesn't provide a type for Web Component . I think this needs to be fixed as well, and this can be discussed further.

Yeah, we already received feature requests on this. See https://github.com/lynx-family/lynx-stack/issues/394.

We may want to provide an interface `IntrinsicElements` just like Web and Lynx does.

```js
export interface IntrinsicElements {
  'lynx-view': LynxViewProps
}
```

Mori-Yang:
Hi @colinaaa, check out the changes ;)

Let me know if there are any problems.

packages/react/runtime/jsx-dev-runtime/index.d.ts
```diff
@@ -3,3 +3,16 @@
...
 export { jsx, jsxs } from 'react/jsx-runtime';
+import { IntrinsicElements as _IntrinsicElements } from '@lynx-js/types';
+import { JSX as _JSX } from 'react';
+
+export namespace JSX {
+  interface IntrinsicElements extends _IntrinsicElements {
+  }
+
+  interface IntrinsicAttributes {
+    ref?: unknown | ((e: unknown) => void) | undefined;
```
colinaaa: The `unknown` does not look correct. You may keep the type same as before https://github.com/lynx-family/lynx-stack/blob/632bdddcf065262b98f06cf451a4d0d129e4a46a/packages/react/types/react.d.ts#L97

```suggestion
    ref?: Ref<NodesRef>;
```

colinaaa: Actually, I don't think we need `ref` here since we already have `ref` in the `StandardProps`, and it will be the base interface of `ViewProps` https://github.com/lynx-family/lynx/blob/8198dcbaaf8f0a3334eb730664fa15e6f138b8d9/js_libraries/types/types/common/element/view.d.ts#L7

```js
interface ViewProps extends StandardProps {}
```

packages/react/runtime/jsx-runtime/index.d.ts
```diff
@@ -2,3 +2,16 @@
...
 export { jsx, jsxs, Fragment } from 'react/jsx-runtime';
+import { IntrinsicElements as _IntrinsicElements } from '@lynx-js/types';
+import { JSX as _JSX } from 'react';
+
+export namespace JSX {
+  interface IntrinsicElements extends _IntrinsicElements {
+  }
+
+  interface IntrinsicAttributes {
+    ref?: unknown | ((e: unknown) => void) | undefined;
```
colinaaa: ditto here.

colinaaa (commented):
Basically LGTM.

Would you please also add some tests for the change? You may use the vitest for type testing. See https://vitest.dev/guide/testing-types.html for details.

> We will run `pnpm -r run test:type` on CI.

packages/react/runtime/jsx-dev-runtime/index.d.ts
Mori-Yang: Yes, I found the type of ref here:
https://github.com/lynx-family/lynx-stack/blob/47e5b22715a39b5fbd6c4c4cc734293a3ccd75b8/packages/react/types/react.d.ts#L97
I'll delete.

packages/react/runtime/jsx-dev-runtime/index.d.ts
```diff
@@ -3,3 +3,13 @@
 // LICENSE file in the root directory of this source tree.
 export { jsxDEV, Fragment } from 'react/jsx-dev-runtime';
 export { jsx, jsxs } from 'react/jsx-runtime';
+import { IntrinsicElements as _IntrinsicElements, type NodesRef } from '@lynx-js/types';
```
colinaaa:
```suggestion
import { IntrinsicElements as _IntrinsicElements } from '@lynx-js/types';
```

packages/react/runtime/jsx-dev-runtime/index.d.ts
```diff
@@ -3,3 +3,13 @@
...
+import { JSX as _JSX, type Ref } from 'react';
```
colinaaa:
```suggestion
import { JSX as _JSX } from 'react';
```

packages/react/runtime/jsx-runtime/index.d.ts
```diff
@@ -2,3 +2,13 @@
 // Licensed under the Apache License Version 2.0 that can be found in the
 // LICENSE file in the root directory of this source tree.
 export { jsx, jsxs, Fragment } from 'react/jsx-runtime';
+import { IntrinsicElements as _IntrinsicElements, type NodesRef } from '@lynx-js/types';
+import { JSX as _JSX, type Ref } from 'react';
```
colinaaa:
```suggestion
import { JSX as _JSX } from 'react';
```

packages/react/runtime/jsx-runtime/index.d.ts
```diff
@@ -2,3 +2,13 @@
 // Licensed under the Apache License Version 2.0 that can be found in the
 // LICENSE file in the root directory of this source tree.
 export { jsx, jsxs, Fragment } from 'react/jsx-runtime';
+import { IntrinsicElements as _IntrinsicElements, type NodesRef } from '@lynx-js/types';
```
colinaaa:
```suggestion
import { IntrinsicElements as _IntrinsicElements } from '@lynx-js/types';
```

packages/react/runtime/tsconfig.json
```diff
@@ -5,7 +5,8 @@
     "outDir": "./lib",
     "baseUrl": "./",
     "rootDir": "./src",
-    "jsx": "preserve",
+    "jsx": "react-jsx",
+    "jsxImportSource": "./",
```
Mori-Yang, line 9: I added these under the current package, but the test file still requires me to import 'React', so I'm importing 'React' in the test file to avoid ts errors.

colinaaa (commented):
I think maybe we should also export the `JSX` namespace from `@lynx-js/react`. Just like `@types/react` does.

This is helpful when annotating a function that returns JSX.

```ts
import type { JSX } from '@lynx-js/react'

export function App() {
  function renderFoo(): JSX.Element {
    return <text>Foo</text>
  }
  return renderFoo()
}
```

packages/react/runtime/__test__/typecheck/jsx-runtime.test-d.tsx
```diff
@@ -0,0 +1,21 @@
...
+  });
+
+  test('should validate the required props for raw-text', () => {
+    // @ts-expect-error: Missing required prop 'text'
+    const shouldError = <raw-text></raw-text>;
+
+    const rawTextELe = <raw-text text={'text'}></raw-text>;
+    assertType<JSX.Element>(rawTextELe);
+  });
+});
```
Mori-Yang, line 104: the test cases may not be comprehensive enough and I don't know what's missing.

Mori-Yang:
There are still a few uncertainties, which I have commented in CR

packages/react/runtime/tsconfig.json
colinaaa: This is because the `packages/react/runtime/tsconfig.json` does not include the test file 

https://github.com/lynx-family/lynx-stack/blob/07dc5e3d1b95c151c10902e0d4132bf7dbc8ca41/packages/react/runtime/tsconfig.json#L13-L20

You may create a new `tsconfig.json` under `packages/react/runtime/__test__`.

```jsonc
{
  "extends": "../../../../tsconfig.json",
  "compilerOptions": {
    "jsx": "react-jsx",
    "jsxImportSource": "@lynx-js/react",
    "noEmit": true,
  },
  "include": ["**/*.ts", "**/*.tsx"]
}
```

packages/react/runtime/__test__/typecheck/jsx-runtime.test-d.tsx
colinaaa: You may also add some tests to:

1. Unsupported tags (e.g.: `<div>`)
2. Ref (`<view ref={ref} />`, `<view ref={() => {}} />`
3. Event (`bind*`, `catch*`)
4. MainThread Ref (`<view main-thread:ref />`) & MainThread Event (`<view main-thread:bind* />`)

packages/react/runtime/__test__/typecheck/jsx-runtime.test-d.tsx
```diff
@@ -0,0 +1,31 @@
+// Copyright 2024 The Lynx Authors. All rights reserved.
+// Licensed under the Apache License Version 2.0 that can be found in the
+// LICENSE file in the root directory of this source tree.
+import { assertType, describe, test } from 'vitest';
+import { JSX } from '../../jsx-runtime/index';
+import React from 'react';
```
colinaaa:
```suggestion
```

Please add a `tsconfig.json` that extends the root `tsconfig.json` with `jsx` and `jsxImportSource`.

packages/react/runtime/__test__/typecheck/jsx-runtime.test-d.tsx
```diff
@@ -0,0 +1,31 @@
+// Copyright 2024 The Lynx Authors. All rights reserved.
+// Licensed under the Apache License Version 2.0 that can be found in the
+// LICENSE file in the root directory of this source tree.
+import { assertType, describe, test } from 'vitest';
+import { JSX } from '../../jsx-runtime/index';
```
colinaaa:
```suggestion
import type { JSX } from '../../jsx-runtime/index.js';
```

packages/react/runtime/__test__/typecheck/jsx-runtime.test-d.tsx
```diff
@@ -0,0 +1,100 @@
+// Copyright 2024 The Lynx Authors. All rights reserved.
+// Licensed under the Apache License Version 2.0 that can be found in the
+// LICENSE file in the root directory of this source tree.
+import type { JSX } from '../../jsx-runtime/index.js';
+import type { MainThread, NodesRef, TouchEvent } from '@lynx-js/types';
+import { assertType, describe, test } from 'vitest';
+import { useMainThreadRef } from '../../lib/lynx-api.js';
```
colinaaa: I think it's better to directly use `src/` in tests.

```suggestion
import { useMainThreadRef } from '../../src/lynx-api.js';
```

packages/react/runtime/__test__/typecheck/jsx-runtime.test-d.tsx
```diff
@@ -0,0 +1,100 @@
...
+import { useRef } from '../../lib/hooks/react.js';
```
colinaaa: ditto here
```suggestion
import { useRef } from '../../src/hooks/react.js';
```

packages/react/runtime/__test__/typecheck/jsx-runtime.test-d.tsx
```diff
@@ -0,0 +1,100 @@
...
+
+  test('should error on unsupported tags', () => {
+    // @ts-expect-error: Unsupported tag
+    const divElement = <div></div>;
+  });
+
+  test('should support event handlers', () => {
+    const viewWithBind = (
+      <view
+        bindtap={(e: TouchEvent) => {
```
colinaaa: Can we use `assertType<number>(e.detail.x)` to test about event type instead of `TouchEvent`?

packages/react/runtime/__test__/typecheck/jsx-runtime.test-d.tsx
```diff
@@ -0,0 +1,100 @@
...
+    );
+    assertType<JSX.Element>(viewWithBind);
+    assertType<JSX.Element>(viewWithCatch);
+  });
+
+  test('should support main-thread event handlers', () => {
+    const viewWithMainThreadEvent = (
+      <view
+        main-thread:bindtap={(e: MainThread.TouchEvent) => {
+          e.currentTarget;
```
colinaaa: ditto here. We should focus on testing if the `e` can be inferred as `MainThread.TouchEvent`.

colinaaa (commented):
LGTM overall. I need to fix the CI failure because it's currently incompatible with the main branch of a forked repository.

packages/react/runtime/__test__/typecheck/jsx-runtime.test-d.tsx
Mori-Yang: Okay, I'll change it now.

colinaaa:
> My first PR commit was from a branch of the fork repository, but all subsequent commits have been in the main branch of the fork repository, so I don't sure if it's me who's causing the CI to fail or not.

never mind. I fixed in https://github.com/lynx-family/lynx-stack/pull/584

packages/react/runtime/__test__/typecheck/jsx-runtime.test-d.tsx
```diff
@@ -0,0 +1,104 @@
+// Copyright 2024 The Lynx Authors. All rights reserved.
```
colinaaa, line 1: You would also need to exclude the coverage of this file.

Add `*.test-d.*` to

https://github.com/lynx-family/lynx-stack/blob/64f64ddc6f61bb9091fee57f5e8b5ba0efed77f1/packages/react/runtime/vitest.config.ts#L71

https://github.com/lynx-family/lynx-stack/blob/64f64ddc6f61bb9091fee57f5e8b5ba0efed77f1/vitest.config.ts#L11

packages/react/runtime/vitest.config.ts
```diff
@@ -91,6 +91,7 @@ export default defineConfig({
 
         '__test__/page.test.jsx',
         '**/*.d.ts',
+        '**/*.test-d.ts',
```
colinaaa: I think the added test is `test-d.tsx`?

Mori-Yang: sorry, my fault ;(

colinaaa (approved):
LGTM 👍

Mori-Yang:
BTW, I have two more confusions:
1. issue : https://github.com/lynx-family/lynx-stack/discussions/574
2. in [my monorepo test](https://github.com/Mori-Yang/lynx-web-template), every time I need to stop the terminal, repackage Lynx, and run it again. Is there a better solution? And what is the cause of this?

hzy (approved):

colinaaa (approved):
Thank you!

PR #562, opened 2025-04-16
fix: update react-rsbuild-plugin reference from @lynx-dev to @lynx-js

gaoachao:
## Summary

Update react-rsbuild-plugin reference from @lynx-dev to @lynx-js

## Checklist


colinaaa (approved):
👍

PR #577, opened 2025-04-17
ci: update workflow to add upstream remote for fetching branches

colinaaa:
## Summary

When fetch base branch, we should fetch from upstream instead of origin (which is the fork repo).

This would fix CI failure like https://github.com/lynx-family/lynx-stack/actions/runs/14495253377/job/40661745063

## Checklist


.github/workflows/workflow-build.yml
```diff
@@ -37,12 +37,16 @@ jobs:
         env:
           CLEAN_BASE_REF: ${{ steps.git-refs.outputs.base-ref }}
           CLEAN_HEAD_REF: ${{ steps.git-refs.outputs.head-ref }}
+          # We are not using
```
colinaaa:
```suggestion
```

PR #595, opened 2025-04-21
feat(web): x-list supports need-visible-item-info

Sherry-hue:
## Summary

feat: x-list supports `need-visible-item-info`, now you can get visible cells info in `scroll`、`scrolltoupper`、`scrolltolower` event.

## Checklist

### DIFF
--- a/packages/web-platform/web-elements/src/XList/XListEvents.ts
+++ b/packages/web-platform/web-elements/src/XList/XListEvents.ts
@@ -51,6 +51,10 @@ export class XListEvents
     '#lower-threshold-observer',
   );
   #getScrollDetail() {
+    const needVisibleItemInfo = this.#dom.getAttribute(
+      'need-visible-item-info',
+    ) !== null;
+
     const { scrollTop, scrollLeft, scrollHeight, scrollWidth } = this
       .#getListContainer();
     const detail = {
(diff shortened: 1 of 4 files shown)


packages/web-platform/web-elements/src/XList/XListEvents.ts
```diff
@@ -51,6 +51,12 @@ export class XListEvents
     '#lower-threshold-observer',
   );
   #getScrollDetail() {
+    const needVisibleItemInfo = this.#dom.getAttribute(
```
PupilTong, line 54: you don't need to consider the "false" value. we'll remove it unless it's a `notToFilterFalseAttributes` value.

packages/web-platform/web-elements/src/XList/XListEvents.ts
```diff
@@ -337,9 +346,7 @@ export class XListEvents
     this.#dom.dispatchEvent(
       new CustomEvent('lynxscroll', {
         ...commonComponentEventSetting,
-        detail: {
-          type: 'scroll',
-        },
+        detail: this.#getScrollDetail(),
```
PupilTong, line 347: Is there any way to avoid getScrollDetail for every scroll event?

Sherry-hue: I think it's impossible, getting the latest value every time you scroll seems inevitable?

PupilTong (approved):

PR #602, opened 2025-04-22
refactor: adopt background-only npm package

### LINKED ISSUE #600: [Feature]: Adopting `background-only` Package

Currently, we maintain our own built-in files for background-only functionality:
- https://github.com/lynx-family/lynx-stack/blob/main/packages/rspeedy/plugin-react/src/background-only/empty.ts
- https://github.com/lynx-family/lynx-stack/blob/main/packages/rspeedy/plugin-react/src/background-only/error.ts

We should replace our custom implementation with the established `background-only` package from npm:
https://www.npmjs.com/package/background-only

```json
"dependencies": {
    "background-only": "^0.0.1",
},
```

gaoachao:
## Summary

close #600 

Refactor: Replace built-in `background-only` implementation with npm package

Previously we maintained custom files:

- `empty.ts` for background thread
- `error.ts` for main thread validation

Now adopting the standard `background-only` npm package

## Checklist

### DIFF
--- a/packages/rspeedy/plugin-react-alias/etc/react-alias-rsbuild-plugin.api.md
+++ b/packages/rspeedy/plugin-react-alias/etc/react-alias-rsbuild-plugin.api.md
@@ -6,6 +6,11 @@
 
 import type { RsbuildPlugin } from '@rsbuild/core';
 
+// Warning: (ae-missing-release-tag) "createLazyResolver" is part of the package's API, but it is missing a release tag (@alpha, @beta, @public, or @internal)
+//
+// @public (undocumented)
+export function createLazyResolver(context: string, conditionNames: string[]): (request: string) => Promise<string>;
+
 // Warning: (ae-missing-release-tag) "Options" is part of the package's API, but it is missing a release tag (@alpha, @beta, @public, or @internal)
 //
 // @public (undocumented)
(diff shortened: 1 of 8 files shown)


packages/rspeedy/plugin-react-alias/etc/react-alias-rsbuild-plugin.api.md
```diff
@@ -6,6 +6,11 @@
 
 import type { RsbuildPlugin } from '@rsbuild/core';
 
+// Warning: (ae-missing-release-tag) "createLazyResolver" is part of the package's API, but it is missing a release tag (@alpha, @beta, @public, or @internal)
+//
+// @public (undocumented)
```
upupming, line 11: Should we add comment to this API?

packages/rspeedy/plugin-react/src/backgroundOnly.ts
```diff
@@ -27,7 +42,7 @@ export function applyBackgroundOnly(
       .alias
       .set(
         'background-only$',
-        path.resolve(__dirname, 'background-only', 'error.js'),
+        backgroundOnly.mainThread,
```
upupming, line 45: Can we just move this logic into `@lynx-js/react-alias-rsbuild-plugin`?

gaoachao: This logic is followed by the `invalid-import-error-loader` that handles `error.js`, which is why I placed it here.

colinaaa (approved):

PR #615, opened 2025-04-24
fix(testing-library): rollup dts files

upupming:
## Summary

Fix the type issue of testing library, we should bundle the dts file since we have bundled `@lynx-js/test-environment` package.

## Checklist

- [ ] Tests updated (or **not required**).


packages/react/testing-library/api-extractor.json
```diff
@@ -4,12 +4,21 @@
...
+  "mainEntryPointFilePath": "<projectFolder>/types/entry.d.ts",
   "compiler": {
     "overrideTsconfig": {
       "compilerOptions": {
         "skipLibCheck": true,
       },
     },
   },
+  "bundledPackages": [
+    "@testing-library/dom",
```
colinaaa: I don't think we need this. Since the only result of this is making the API report (`packages/react/testing-library/etc/react-lynx-testing-library.api.md`) a big mess (and also the API documentation, too).
```suggestion
```

packages/react/testing-library/api-extractor.json
```diff
@@ -4,12 +4,21 @@
...
+    "@lynx-js/test-environment",
+  ],
+  "dtsRollup": {
+    "enabled": true,
```
colinaaa: If we are using `dtsRollup` directly from `api-extractor`, I think we need to run `api-extractor` before/after `rslib build`.

Rslib also supports bundle types with api-extractor. Can we directly use that?

colinaaa (dismissed):

colinaaa (approved):

PR #647, opened 2025-04-28
chore: update css property map

PupilTong:
copied from packages/react/transform/src/css_property.rs

also removed the default value, which is useless


YellowFishWyfCPP:
There's actually offset-rotate in addition offset-path and offset-distance, it's default value is auto,

packages/web-platform/web-mainthread-apis/src/elementAPI/style/styleFunctions.ts
```diff
@@ -68,7 +68,13 @@ export function createStyleFunctions(
   ): void {
     let dashName: string | undefined;
     if (typeof key === 'number') {
-      dashName = queryCSSProperty(key).dashName;
+      const queryResult = queryCSSProperty(key);
+      dashName = queryResult.dashName;
+      if (queryResult.isX) {
+        console.error(
+          `[lynx-web] css property with index: ${key} is not supported. Please check your usage.`,
```
colinaaa: Can we log the real property name instead of the number index? I don't think it would help to have an error message like: "[lynx-web] css property with index: 143 is not supported. Please check your usage."

PupilTong:
> There's actually offset-rotate in addition offset-path and offset-distance, it's default value is auto,

The id should exactly match the value in reactlynx.

colinaaa (approved):

PR #652, opened 2025-04-28
fix(react/runtime): Accelerate the timing of effect and ref triggering

### LINKED ISSUE #405: [Bug]: `useEffect()` callbacks sometimes run later than expected
### System Info

  System:
    OS: macOS 14.0
    CPU: (10) arm64 Apple M1 Max
    Memory: 10.27 GB / 64.00 GB
    Shell: 5.9 - /bin/zsh
  Binaries:
    Node: 18.20.4 - ~/Library/Caches/fnm_multishells/88343_1743400732072/bin/node
    Yarn: 1.22.19 - /opt/homebrew/bin/yarn
    npm: 8.12.1 - /opt/homebrew/bin/npm
    pnpm: 7.26.3 - /opt/homebrew/bin/pnpm
  npmPackages:
    @lynx-js/qrcode-rsbuild-plugin: ^0.3.4 => 0.3.4 
    @lynx-js/react: ^0.106.1 => 0.106.1 
    @lynx-js/react-rsbuild-plugin: ^0.9.3 => 0.9.3 
    @lynx-js/rspeedy: ^0.8.6 => 0.8.6 
    @lynx-js/types: ^3.2.0 => 3.2.0 

### Details

The following code is expected to print

```
init: false
init: true
```

but it actually prints

```
init: true
init: true
```
.

```tsx
import { useEffect, useState } from '@lynx-js/react'

export function App() {
  const [initialized, setInitialized] = useState(false);

  useEffect(() => {
    console.log('init: ', initialized);
  }, [initialized]);

  if (!initialized) {
    Promise.resolve().then(() => {
      setInitialized(true);
    });
  }

  return <text>hello</text>;
}
```

### Reproduce link

_No response_

### Reproduce Steps

Use the code above.

Yradex:
## Summary

This PR fixes #405 by refactoring the ref and effect system to use Preact's ref implementation and accelerating the timing:

- Fixed effect hook variable capture issue where stale values might be captured in closure.
- Ensured `ref` and `useEffect()` side effects are now guaranteed to execute before event triggers.

**Breaking Changes**:

- The execution timing of `ref` and `useEffect()` side effects has been moved forward. These effects will now execute before hydration is complete, rather than waiting for the main thread update to complete.
- For components inside `<list />`, `ref` callbacks will now be triggered during background thread rendering, regardless of component visibility. If your code depends on component visibility timing, use `main-thread:ref` instead of regular `ref`.

## Checklist


.changeset/tired-drinks-give.md
```diff
@@ -0,0 +1,13 @@
+---
+"@lynx-js/react": minor
+---
+
+Fixes:
```
colinaaa: This might appear in the CHANGELOG with:

```
Fixes: (#652)
```

colinaaa (approved):
LGTM!

colinaaa (approved):

PR #653, opened 2025-04-28
fix: resolve page crash on development mode when enabling `experimental_isLazyBundle: true`

### LINKED ISSUE #504: [Bug]: dev mode separated lazy bundle cannot be loaded because of prefesh
Follow the guide: https://lynxjs.org/react/code-splitting.html#lazy-loading-standalone-project

If we use `rspeedy dev` to serve the lazy bundle, we will get an error `cannot convert to object`:

Image

Image

This error happens because the lazy module is wrapped by `prefesh` and the exports are not correct, we should disable prefresh for the lazy bundle.

Futher investigation is needed

upupming:
## Summary

Fixes #504 

In this PR, we disabled HMR for standalone lazy bundle to ensure it works normally when it is served using `rspeedy dev`.

When HMR is enabled, the bundle result will be like:

```js
/************************************************************************/
// module cache are used so entry inlining is disabled
// startup
// Load entry module and return exports
__webpack_require__("(react:background)/./src/index.tsx");
__webpack_require__("(react:background)/./node_modules/.pnpm/@rspack+core@1.3.6_@swc+helpers@0.5.17/node_modules/@rspack/core/hot/dev-server.js");
__webpack_require__("(react:background)/./node_modules/.pnpm/@lynx-js+webpack-dev-transport@0.1.2/node_modules/@lynx-js/webpack-dev-transport/lib/client/index.js?hostname=192.168.5.55&port=3000&pathname=/rsbuild-hmr&hot=true&live-reload=true&protocol=ws");
var __webpack_exports__ = __webpack_require__("(react:background)/./node_modules/.pnpm/@lynx-js+react@0.107.0_@lynx-js+types@3.2.1_@types+react@18.3.20/node_modules/@lynx-js/react/refresh/dist/index.js");
```

It wrongly exports `@lynx-js/react/refresh/dist/index.js` as the default export, however `(react:background)/./src/index.tsx` is expected.

## Checklist


packages/rspeedy/plugin-react/src/entry.ts
```diff
@@ -284,18 +295,26 @@ function getBackgroundFilename(
...
+  experimental_isLazyBundle: boolean,
+): string {
   if (typeof config.output?.filenameHash === 'string') {
     return config.output.filenameHash
       ? `.[${config.output.filenameHash}]`
       : EMPTY_HASH
   } else if (config.output?.filenameHash === false) {
     return EMPTY_HASH
-  } else if (isProd) {
+  } else if (isProd || experimental_isLazyBundle) {
```
colinaaa, line 323: Can we also add a config test to check if the filename contains `[contenthash]` (or `output.filenameHash`) when setting `experimental_isLazyBundle: true`?

Just like https://github.com/lynx-family/lynx-stack/blob/f4ac0f1bf034dedadf9722c83edc077ac20402db/packages/rspeedy/plugin-react/test/config.test.ts#L440

packages/rspeedy/plugin-react/src/entry.ts
```diff
@@ -131,7 +138,9 @@ export function applyEntry(
           import: imports,
           filename: backgroundName,
         })
-        .when(isDev && !isWeb, entry => {
+        // in standalone lazy bundle mode, we do not add
+        // other entries to avoid wrongly exporting from other entries
+        .when(isDev && !experimental_isLazyBundle && !isWeb, entry => {
```
colinaaa: Consider using `.prepend` instead of `.add` to resolve the issue.

.changeset/short-crabs-lead.md
```diff
@@ -0,0 +1,7 @@
+---
+"@lynx-js/runtime-wrapper-webpack-plugin": patch
+"@lynx-js/react-webpack-plugin": patch
+"@lynx-js/react-rsbuild-plugin": patch
+---
+
+fix: bundle result of `experimental_isLazyBundle: true` on development mode
```
colinaaa: Consider enhancing this by:

1. Enhancing the description, for example: "Resolve page crash when enabling `experimental_isLazyBundle: true` in development mode."
2. Dividing the changeset for clarity. We could provide a more detailed changeset for webpack plugins, such as: "Introduce `experimental_isLazyBundle` option."

colinaaa (dismissed):
Generally looks good to me.

hzy (approved):

colinaaa (approved):

Issue #659, opened 2025-04-29
[Bug]: OnPipelineStart throw error of param number


partholon:
### System Info

lynxsdk 2.16

### Details

`Error: OnPipelineStart arg count must == 1 ` when beginPipeline(...)

### Reproduce Steps

use lynxsdk below 3.0 and latest @lynx-js/*

PR #660, opened 2025-04-29
fix(timing): Fix Compatibility Issue in _onPipelineStart

partholon:
Resolved a compatibility problem caused by break changes in the _onPipelineStart method signature. For engine versions 3.1 and above, adjusted the method call to include PipelineOptions as an argument.

issue: #659

## Checklist

### DIFF
--- a/packages/react/runtime/src/lynx/performance.ts
+++ b/packages/react/runtime/src/lynx/performance.ts
@@ -6,6 +6,7 @@ import type { VNode } from 'preact';
 
 import { DIFF } from '../renderToOpcodes/constants.js';
 import { __globalSnapshotPatch } from '../lifecycle/patch/snapshotPatch.js';
+import { isSdkVersionGt } from '../utils.js';
 
 enum PerformanceTimingKeys {
   updateSetStateTrigger,
@@ -93,7 +94,11 @@ function beginPipeline(needTimestamps: boolean, pipelineOrigin: PipelineOrigin, 
         break;
     }
 
-    lynx.performance?._onPipelineStart?.(globalPipelineOptions.pipelineID, globalPipelineOptions);
+    if (isSdkVersionGt(3, 0)) {
+      lynx.performance?._onPipelineStart?.(globalPipelineOptions.pipelineID, globalPipelineOptions);
+    } else {
+      lynx.performance?._onPipelineStart?.(globalPipelineOptions.pipelineID);
+    }
     if (timingFlag) {
       lynx.performance?._bindPipelineIdWithTimingFlag?.(globalPipelineOptions.pipelineID, timingFlag);
     }
--- a/packages/react/runtime/types/types.d.ts
+++ b/packages/react/runtime/types/types.d.ts
@@ -260,7 +260,11 @@ declare module '@lynx-js/types/background' {
 
   interface Performance {
     _generatePipelineOptions?(): PipelineOptions;
-    _onPipelineStart?(pipelineID: string, options: PipelineOptions): void;
+    /**
+     * @param pipelineID - Returned by native when calling {@link _generatePipelineOptions}.
+     * @param options - Added since 3.1. See {@link PipelineOptions} for more details.
+     */
+    _onPipelineStart?(pipelineID: string, options?: PipelineOptions): void;
     _bindPipelineIdWithTimingFlag?(pipelineID: string, timingFlag: string): void;
     _markTiming?(pipelineID: string, key: string): void;
   }
--- a/.changeset/tall-spiders-punch.md
+++ b/.changeset/tall-spiders-punch.md
@@ -1,0 +1,5 @@
+---
+"@lynx-js/react": patch
+---
+
+Fix error `OnPipelineStart arg count must == 1` on app load.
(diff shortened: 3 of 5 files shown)


packages/react/runtime/src/lynx/performance.ts
```diff
@@ -93,7 +93,11 @@ function beginPipeline(needTimestamps: boolean, pipelineOrigin: PipelineOrigin,
         break;
     }
 
-    lynx.performance?._onPipelineStart?.(globalPipelineOptions.pipelineID, globalPipelineOptions);
+    if (SystemInfo.engineVersion >= '3.0') {
```
colinaaa: `SystemInfo.engineVersion` is `undefined` before 3.2. Use `SystemInfo.lynxSdkVersion` instead.
```suggestion
    if (SystemInfo.lynxSdkVersion >= '3.0') {
```

colinaaa: **DO NOT** use `>=` on string in JavaScript. There is a `isSdkVersionGt` function you may use.

https://github.com/lynx-family/lynx-stack/blob/f4ac0f1bf034dedadf9722c83edc077ac20402db/packages/react/runtime/src/utils.ts#L21

colinaaa (changes requested):

packages/react/runtime/types/types.d.ts
```diff
@@ -260,7 +260,15 @@ declare module '@lynx-js/types/background' {
...
     _generatePipelineOptions?(): PipelineOptions;
+    /**
+     * @since 3.1
+     */
     _onPipelineStart?(pipelineID: string, options: PipelineOptions): void;
+    /**
+     * @disabled since 3.1
+     * This version is disabled starting from version 3.1 due to additional required parameters.
+     */
+    _onPipelineStart?(pipelineID: string): void;
```
colinaaa: I don't think we need to use override here. Use optional parameter instead.

```suggestion
    _onPipelineStart?(pipelineID: string, options?: PipelineOptions): void;
```

packages/react/runtime/src/lynx/performance.ts
partholon: Done. changed to isSdkVersionGt(3, 1).

colinaaa (changes requested):
You may also add some tests for this

packages/react/runtime/src/lynx/performance.ts
partholon: changed to isSdkVersionGt(3, 1).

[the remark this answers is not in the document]
partholon: changed to isSdkVersionGt(3, 1).

packages/react/runtime/types/types.d.ts
partholon: Done. override function has been removed.

packages/react/runtime/types/types.d.ts
```diff
@@ -260,7 +260,11 @@ declare module '@lynx-js/types/background' {
 
   interface Performance {
     _generatePipelineOptions?(): PipelineOptions;
-    _onPipelineStart?(pipelineID: string, options: PipelineOptions): void;
+    /**
+     * @param pipelineID Returned by native when calling `_generatePipelineOptions`
+     * @param options since 3.1. PipelineOptions.
```
colinaaa: The format of this JSDoc is not correct.

```suggestion
     * @param options - Added since 3.1. See {@link PipelineOptions} for more details.
```

packages/react/runtime/types/types.d.ts
```diff
@@ -260,7 +260,11 @@ declare module '@lynx-js/types/background' {
 
   interface Performance {
     _generatePipelineOptions?(): PipelineOptions;
-    _onPipelineStart?(pipelineID: string, options: PipelineOptions): void;
+    /**
+     * @param pipelineID Returned by native when calling `_generatePipelineOptions`
```
colinaaa: ditto.
```suggestion
     * @param pipelineID - Returned by native when calling `_generatePipelineOptions`
```

colinaaa (changes requested):

packages/react/runtime/types/types.d.ts
partholon: Done

packages/react/runtime/types/types.d.ts
partholon: Done

partholon:
> You may also add some tests for this

add `timing api compatibility` test in `timing.test.jsx`

packages/react/runtime/src/lynx/performance.ts
```diff
@@ -93,7 +94,11 @@ function beginPipeline(needTimestamps: boolean, pipelineOrigin: PipelineOrigin,
         break;
     }
 
-    lynx.performance?._onPipelineStart?.(globalPipelineOptions.pipelineID, globalPipelineOptions);
+    if (isSdkVersionGt(3, 0)) {
+      lynx.performance?._onPipelineStart?.(globalPipelineOptions.pipelineID, globalPipelineOptions);
```
Yradex, line 98: Do we still need to check the existence for `lynx.performance._onPipelineStart()` when sdk version is above 3.1 ?

Yradex (approved):

colinaaa (approved):

PR #698, opened 2025-05-06
fix(feat): Improper code sanitization could lead path traversal

odaysec:
https://github.com/lynx-family/lynx-stack/blob/24ebce83f0b416aee3bdd7d4fe61a1b629aadeaa/packages/webpack/runtime-wrapper-webpack-plugin/src/RuntimeWrapperWebpackPlugin.ts#L242-L242

fix the issue need to sanitize the `targetSdkVersion` value before embedding it into the JavaScript code. This can be achieved by escaping potentially dangerous characters using a utility function like `escapeUnsafeChars`. This function will replace characters such as `<`, `>`, `/`, and others with their Unicode escape sequences, ensuring that the resulting string is safe for inclusion in JavaScript code.

The fix involves:
1. Defining an `escapeUnsafeChars` function to sanitize strings.
2. Applying this function to `JSON.stringify(targetSdkVersion)` before embedding it in the generated code.

Using string concatenation to construct JavaScript code can be error-prone, or in the worst case, enable code injection if an input is constructed by an attacker.

## POC
The below constructs a function that assigns the number 42 to the property `key` on an object `obj`. However, if `key` contains `</script>`, then the generated code will break out of a `</script>` if inserted into a `</script>` tag.

```js
function createObjectWrite() {
    const assignment = `obj[${JSON.stringify(key)}]=42`;
    return `(function(){${assignment}})` // NOT OK
}
```
The issue has been fixed by escaping potentially dangerous characters, as shown below.
```js
const charMap = {
    '<': '\\u003C',
    '>' : '\\u003E',
    '/': '\\u002F',
    '\\': '\\\\',
    '\b': '\\b',
    '\f': '\\f',
    '\n': '\\n',
    '\r': '\\r',
    '\t': '\\t',
    '\0': '\\0',
    '\u2028': '\\u2028',
    '\u2029': '\\u2029'
};

function escapeUnsafeChars(str) {
    return str.replace(/[<>\b\f\n\r\t\0\u2028\u2029]/g, x => charMap[x])
}

function createObjectWrite() {
    const assignment = `obj[${escapeUnsafeChars(JSON.stringify(key))}]=42`;
    return `(function(){${assignment}})` // OK
}
```
## References
[Code Injection](https://www.owasp.org/index.php/Code_Injection)
[CWE-94](https://cwe.mitre.org/data/definitions/94.html)
[CWE-79](https://cwe.mitre.org/data/definitions/79.html)
[CWE-116](https://cwe.mitre.org/data/definitions/116.html)

## Checklist

### DIFF
--- a/packages/webpack/runtime-wrapper-webpack-plugin/src/RuntimeWrapperWebpackPlugin.ts
+++ b/packages/webpack/runtime-wrapper-webpack-plugin/src/RuntimeWrapperWebpackPlugin.ts
@@ -62,6 +62,25 @@ const defaultInjectVars = [
   'window',
 ];
 
+const charMap = {
+  '<': '\\u003C',
+  '>': '\\u003E',
+  '/': '\\u002F',
+  '\\': '\\\\',
+  '\b': '\\b',
+  '\f': '\\f',
+  '\n': '\\n',
+  '\r': '\\r',
+  '\t': '\\t',
+  '\0': '\\0',
+  '\u2028': '\\u2028',
+  '\u2029': '\\u2029'
+};
+
+function escapeUnsafeChars(str: string): string {
+  return str.replace(/[<>\b\f\n\r\t\0\u2028\u2029]/g, x => charMap[x]);
+}
+
 /**
  * RuntimeWrapperWebpackPlugin adds runtime wrappers to JavaScript and allow to be loaded by Lynx.
  *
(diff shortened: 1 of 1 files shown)


colinaaa (commented):
Thank you for your PR! However, I believe this isn't necessary, as these options should be provided by the developer rather than the user. Additionally, they should follow the format `x.y` or `x.y.z`. I believe we don't need such an escape for now.

PR #711, opened 2025-05-07
feat(background-only): add source and integrate into monorepo

Huxpro:
Add back background-only package to monorepo.

Note that the package was already published to npm: https://www.npmjs.com/package/background-only

### DIFF
--- a/packages/background-only/README.md
+++ b/packages/background-only/README.md
@@ -1,4 +1,4 @@
-# @lynx-js/background-only
+# background-only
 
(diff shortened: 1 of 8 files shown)


Huxpro:
> ### ⚠️ No Changeset found

We are good as there is no need to republish the package

packages/background-only/README.md
```diff
@@ -0,0 +1,60 @@
+# @lynx-js/background-only
```
colinaaa: Do you mean `background-only`? Or should we change to `@lynx-js/background-only` in `package.json#name`?

```suggestion
# background-only
```

colinaaa (dismissed):

colinaaa (approved):

PR #721, opened 2025-05-08
chore(rspeedy/react): use `background-only` from workspace

colinaaa:
## Summary

Following up #711, use the `background-only` package from workspace.

## Checklist

### DIFF
--- a/pnpm-workspace.yaml
+++ b/pnpm-workspace.yaml
@@ -47,6 +47,7 @@ patchedDependencies:
 
 onlyBuiltDependencies:
   - dprint
+  - esbuild
 
 ignoredBuiltDependencies:
   - core-js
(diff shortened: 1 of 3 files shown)


pnpm-workspace.yaml
```diff
@@ -47,6 +47,7 @@ patchedDependencies:
 
 onlyBuiltDependencies:
   - dprint
+  - esbuild
```
Huxpro, line 50: how is this involved?

colinaaa: I received a warning while running `pnpm install`, so I updated the configuration to suppress it.

```
╭ Warning ─────────────────────────────────────────────────────────────────────╮
│                                                                              │
│   Ignored build scripts: esbuild.                                            │
│   Run "pnpm approve-builds" to pick which dependencies should be allowed     │
│   to run scripts.                                                            │
│                                                                              │
╰──────────────────────────────────────────────────────────────────────────────╯
```

gaoachao (approved):

Issue #726, opened 2025-05-08
[BE] Deprecate internal usages of old macros


Huxpro:
Many internal code paths still rely on `__LEPUS__` and `__JS__` for thread detection. We should migrate them over to `__MAIN_THREAD__` and `__BACKGROUND__` to improve code readability and maintainability for both human and LLM.

I'll do this in a relatively manual approach so I can also improve doc comments along the way, rather than just regex it.

PR #733, opened 2025-05-08
docs(react): README; add some comments; deprecate some old macros

Huxpro:
This diff does 3 things:
- add Reanimated, RSC and others into README
- try to help with https://github.com/lynx-family/lynx-stack/issues/726 whenever I see it
- add doc comments for files to clarify understandings for all

### DIFF
--- a/packages/react/runtime/src/snapshot/ref.ts
+++ b/packages/react/runtime/src/snapshot/ref.ts
@@ -137,12 +137,12 @@ function markRefToRemove(sign: string, ref: unknown): void {
 }
 
 export {
-  updateRef,
-  takeGlobalRefPatchMap,
-  updateBackgroundRefs,
-  unref,
-  transformRef,
   globalRefsToRemove,
   globalRefsToSet,
   markRefToRemove,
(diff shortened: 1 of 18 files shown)


packages/react/runtime/src/snapshot/ref.ts
```diff
@@ -137,12 +137,12 @@ function markRefToRemove(sign: string, ref: unknown): void {
...
 export {
-  updateRef,
-  takeGlobalRefPatchMap,
-  updateBackgroundRefs,
-  unref,
-  transformRef,
   globalRefsToRemove,
   globalRefsToSet,
   markRefToRemove,
+  takeGlobalRefPatchMap,
```
Huxpro, line 143: @colinaaa Does our formatter guarantee a stable and consistent output?

colinaaa: Yes, it does. It seems this was modified by the ESLint auto-fix (we're using `eslint-plugin-import`, which organizes the import/export order).

colinaaa (approved):

PR #770, opened 2025-05-12
fix(react/runtime): Accelerate the timing of effect and ref triggering

Yradex:
## Summary

This PR fixes https://github.com/lynx-family/lynx-stack/issues/405 by refactoring the ref and effect system to use
Preact's ref implementation and accelerating the timing:

- Fixed effect hook variable capture issue where stale values might be
captured in closure.
- Ensured `ref` and `useEffect()` side effects are now guaranteed to
execute before event triggers.

**Breaking Changes**:

- The execution timing of `ref` and `useEffect()` side effects has been
moved forward. These effects will now execute before hydration is
complete, rather than waiting for the main thread update to complete.
- For components inside `<list />`, `ref` callbacks will now be
triggered during background thread rendering, regardless of component
visibility. If your code depends on component visibility timing, use
`main-thread:ref` instead of regular `ref`.

## Checklist


colinaaa (dismissed):

packages/react/runtime/src/lifecycle/ref/delay.ts
```diff
@@ -0,0 +1,99 @@
...
+
+/**
+ * A proxy class designed for managing and executing reference-based tasks.
+ * It delays the execution of tasks until hydration is complete.
+ */
+class RefProxy {
+  private readonly refAttr: [snapshotInstanceId: number, expIndex: number];
+  private task: RefTask | undefined;
+
+  constructor(refAttr: [snapshotInstanceId: number, expIndex: number]) {
```
hzy, line 57:
```suggestion
  constructor(private refAttr: [snapshotInstanceId: number, expIndex: number]) {
```

packages/react/runtime/src/lifecycle/ref/delay.ts
```diff
@@ -0,0 +1,99 @@
...
+  shouldDelayUiOps.value = false;
+  delayedUiOps.length = 0;
+}
+
+/**
+ * A proxy class designed for managing and executing reference-based tasks.
+ * It delays the execution of tasks until hydration is complete.
+ */
+class RefProxy {
+  private readonly refAttr: [snapshotInstanceId: number, expIndex: number];
```
hzy, line 54:
```suggestion
```

packages/react/runtime/src/lifecycle/ref/delay.ts
colinaaa: I don't like the suggested change. Using `private` in constructor is indeed shorter, but it's not [erasbleSyntaxOnly](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-5-8.html#the---erasablesyntaxonly-option).

packages/react/runtime/src/hooks/react.ts
```diff
@@ -29,7 +29,7 @@ import type { DependencyList, EffectCallback } from 'react';
  * @deprecated `useLayoutEffect` in the background thread cannot offer the precise timing for reading layout information and synchronously re-render, which is different from React.
  */
 function useLayoutEffect(effect: EffectCallback, deps?: DependencyList): void {
-  return usePreactLayoutEffect(effect, deps);
+  return usePreactEffect(effect, deps);
```
hzy, line 32: Why not just use `usePreactLayoutEffect`?

packages/react/runtime/src/lifecycle/ref/delay.ts
hzy: Fine!

packages/react/transform/src/swc_plugin_snapshot/mod.rs
```diff
@@ -1589,6 +1597,21 @@ where
...
+          if target != TransformTarget::LEPUS
+            && (snapshot_values_has_ref || snapshot_values_has_spread)
+          {
+            snapshot_attrs.push(JSXAttrOrSpread::JSXAttr(JSXAttr {
+              span: DUMMY_SP,
+              name: JSXAttrName::Ident(IdentName::new("ref".into(), DUMMY_SP)),
+              value: Some(JSXAttrValue::JSXExprContainer(JSXExprContainer {
+                span: DUMMY_SP,
+                expr: JSXExpr::Expr(Box::new(quote!(
+                  r#"$runtime_id.applyRefs.bind([])"# as Expr,
```
hzy: this expr will be a part of jsx so every time a component renders, a new bound function will be created, and it is different with the previous,
1. the new bound function will be called (a.k.a ref), and the old one will be called with null (a.k.a unref)
2. in `applyDelayedRefs`, `refsToApply` will be consumed, and `applyRef` will be called, if the ref provided by the developer is a memoized function, it will still be called (ref), which means 
"double-ref".

Don't know if my analysis is correct, let me known if it is not :D

packages/react/runtime/src/internal.ts
```diff
@@ -26,7 +26,7 @@ export const __DynamicPartChildren_0: [DynamicPartType, number][] = [[DynamicPar
 
 export { updateSpread } from './snapshot/spread.js';
 export { updateEvent } from './snapshot/event.js';
-export { updateRef, transformRef } from './snapshot/ref.js';
+export { updateRef, transformRef, applyRefs } from './snapshot/ref.js';
```
hzy: nit: it is a function returns a ref callback, may be choose a more self-explanatory name?

packages/react/runtime/src/hooks/react.ts
Yradex: Using `usePreactLayoutEffect` results in a delay until `__flushElementTree` is finished, since it uses the same preact implementation as `componentDidMount`. I wonder if it's the intended behavior, since we've formerly made `useLayoutEffect` an alias to `useEffect`.

packages/react/transform/src/swc_plugin_snapshot/mod.rs
Yradex: That makes perfect sense.

I fixed this issue in the latest commit, however, continuing to use Preact's `ref` implementation introduced significant runtime overhead and added unnecessary code complexity. Therefore, I decided to revert back to our own implementation, which is more efficient for our specific needs.

colinaaa (approved):

PR #775, opened 2025-05-12
chore: enableSSR for all tests for web platform

PupilTong:


PupilTong:
Reactlynx doesn't support it with error `firstScreenSyncTiming must be jsReady`when SSR is enabled` . close.

PR #787, opened 2025-05-13
feat(webpack/runtime-wrapper): support `requestAnimationFrame`

colinaaa:
## Summary

Allow direct access to `requestAnimationFrame` and `cancelAnimationFrame` in background thread script.

close: #181 since `fetch` is now a parameter of the runtime wrapper.

## Checklist

### DIFF
--- a/packages/webpack/runtime-wrapper-webpack-plugin/src/RuntimeWrapperWebpackPlugin.ts
+++ b/packages/webpack/runtime-wrapper-webpack-plugin/src/RuntimeWrapperWebpackPlugin.ts
@@ -60,6 +60,32 @@ const defaultInjectVars = [
   'LynxJSBI',
   'lynx',
   'window',
+
+  // BOM API
+  'document',
+  'frames',
+  'self',
+  'location',
+  'navigator',
+  'localStorage',
+  'history',
+  'Caches',
+  'screen',
+  'alert',
+  'confirm',
+  'prompt',
+  'fetch',
+  'XMLHttpRequest',
+  '__WebSocket__', // We would provide `WebSocket` using `ProvidePlugin`
+  'webkit',
+  'Reporter',
+  'print',
+  '__Function__', // We should allow using `Function`
+  'global',
+
+  // Lynx API
+  'requestAnimationFrame',
+  'cancelAnimationFrame',
 ];
 
 /**
@@ -242,7 +268,9 @@ lynx.targetSdkVersion=lynx.targetSdkVersion||${
       JSON.stringify(targetSdkVersion)
     };
 ${overrideRuntimePromise ? `var Promise = lynx.Promise;` : ''}
-var fetch = lynx.fetch;
+fetch = fetch || lynx.fetch;
+requestAnimationFrame = requestAnimationFrame || lynx.requestAnimationFrame;
+cancelAnimationFrame = cancelAnimationFrame || lynx.cancelAnimationFrame;
 `
   );
 };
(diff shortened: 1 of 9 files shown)


packages/webpack/runtime-wrapper-webpack-plugin/src/RuntimeWrapperWebpackPlugin.ts
```diff
@@ -59,7 +59,33 @@ const defaultInjectVars = [
   'Behavior',
   'LynxJSBI',
   'lynx',
+
+  // BOM API
   'window',
```
colinaaa, line 64: The same order as https://github.com/lynx-family/lynx/blob/863965a80f332f5d3e21f6a9b201351b854d2577/js_libraries/lynx-core/src/app/app.ts#L634

lybvinci:
LGTM

packages/webpack/runtime-wrapper-webpack-plugin/src/RuntimeWrapperWebpackPlugin.ts
```diff
@@ -242,7 +268,9 @@ lynx.targetSdkVersion=lynx.targetSdkVersion||${
       JSON.stringify(targetSdkVersion)
     };
 ${overrideRuntimePromise ? `var Promise = lynx.Promise;` : ''}
-var fetch = lynx.fetch;
+fetch = fetch || lynx.fetch;
```
upupming, line 271: Why do we need this fallback here?

The `fetch` param seems already has been initialized with `lynx.fetch` by lynx-core:

https://github.com/lynx-family/lynx/blob/863965a80f332f5d3e21f6a9b201351b854d2577/js_libraries/lynx-core/src/app/app.ts#L647

packages/webpack/runtime-wrapper-webpack-plugin/test/cases/variables/request-animation-frame/index.js
```diff
@@ -0,0 +1,11 @@
+/*
+// Copyright 2024 The Lynx Authors. All rights reserved.
+// Licensed under the Apache License Version 2.0 that can be found in the
+// LICENSE file in the root directory of this source tree.
+*/
+expect(requestAnimationFrame).toStrictEqual(expect.any(Function));
+expect(cancelAnimationFrame).toStrictEqual(expect.any(Function));
+expect(requestAnimationFrame).not.toBe(lynx.requestAnimationFrame);
```
upupming, line 8: Same as above, in real Lynx environment, it seems `requestAnimationFrame` should have been initialized with `lynx.requestAnimationFrame`.

colinaaa: The `requestAnimationFrame` parameter was introduced in Lynx 3.x. To ensure compatibility with older versions of Lynx, you can use `requestAnimationFrame = requestAnimationFrame || lynx.requestAnimationFrame`, allowing you to use `requestAnimationFrame` directly.

packages/webpack/runtime-wrapper-webpack-plugin/src/RuntimeWrapperWebpackPlugin.ts
colinaaa: I'm not sure when the `fetch` param has been added in lynx-core. I would like to be as much compatible as possible :)

packages/webpack/runtime-wrapper-webpack-plugin/src/RuntimeWrapperWebpackPlugin.ts
colinaaa: 1. I didn't modify `window` :) It was added a long time ago. I assume you're referring to `document` and similar elements.
2. Prior to this update, `typeof document` returned `'undefined'`, and it remains `'undefined'` after the update.
3. Previously, accessing `document` would result in a runtime error. Now, accessing `document` no longer causes a runtime error but returns `undefined`. This might be considered a **BREAKING CHANGE**, though it is unlikely to impact most cases.

upupming (approved):

PR #789, opened 2025-05-13
feat: use lighthouse nodejs to replace lhci

Sherry-hue:
## Summary

use lighthouse nodejs to replace lhci

This is because some tests may crash, causing the entire CI to fail. Use lighthouse nodejs to retry.

## Checklist


PupilTong:
image

packages/web-platform/web-tests/package.json
```diff
@@ -14,7 +14,7 @@
     "build:cases": "rm -rf dist && node  ./scripts/generate-build-command.js",
     "coverage": "nyc report --cwd=$(realpath ../)",
     "coverage:ci": "nyc report --cwd=$(realpath ../) --reporter=lcov",
-    "lh": "pnpm dlx @lhci/cli autorun",
+    "lh": "node ./scripts/run-lighthouse.js && pnpm dlx @lhci/cli upload",
```
colinaaa, line 18: Could we turn this script into a standalone package that can be executed with `pnpm dlx`? This would allow us to avoid installing the `lighthouse` package in our workspace, saving significant time on CI.

.github/workflows/test.yml
```diff
@@ -68,6 +70,7 @@ jobs:
       is-web: true
       run: |
         export NODE_OPTIONS="--max-old-space-size=32768"
+        export CHROME_PATH=$HOME/google-chrome-stable/chrome
```
colinaaa: Why do we need another workflow instead of running `curl` here?

.github/workflows/test.yml
```diff
@@ -54,7 +56,7 @@ jobs:
       - name: Changeset Check
         run: pnpm changeset status --since=origin/main
   lighthouse:
-    needs: build
+    needs: build install-chrome
```
colinaaa: I don't think the syntax is correct.

```
The workflow is not valid. .github/workflows/test.yml (Line: 59, Col: 12): Job 'lighthouse' depends on unknown job 'build install-chrome'.
```

.github/workflows/test.yml
Sherry-hue: This is a bad usage, I will move it here, it is not ready yet

packages/web-platform/web-tests/scripts/run-lighthouse.js
```diff
@@ -0,0 +1,235 @@
...
+
+      while (retries < maxRetries) {
+        try {
+          const chrome = await launch({
+            chromeFlags: [
+              '--no-sandbox',
+              '--disable-gpu',
+              '--disable-dev-shm-usage',
+              '--browser-test',
+              '--headless',
```
PupilTong, line 76: did you try the --headless=new?

packages/web-platform/web-tests/scripts/run-lighthouse.js
```diff
@@ -0,0 +1,235 @@
...
+            ],
+          });
+          result = await lighthouse(
+            `http://localhost:${port}${cases[i].matchingUrlPattern}`,
+            {
+              output: 'json',
+              port: chrome.port,
+            },
+            {
+              extends: 'lighthouse:default',
```
PupilTong, line 86: remove packages/web-platform/web-tests/lighthouserc.js ?

PupilTong:
Have you tried the `upload`?

PupilTong (approved):

PR #658, opened 2025-04-29
feat(webpack): add StartupChunkDependenciesRuntimeModule to fix chunk splitting

gaoachao:
## Summary

add `StartupChunkDependenciesRuntimeModule` to fix chunk splitting

- Implement `StartupChunkDependenciesRuntimeModule` for handling chunk dependencies
- Fix `RuntimeGlobals.ensureChunkHandler` not found when using chunk splitting

Related Links:
- https://github.com/webpack/webpack/blob/main/lib/node/CommonJsChunkLoadingPlugin.js#L41
- https://github.com/webpack/webpack/blob/main/lib/runtime/StartupChunkDependenciesPlugin.js#L65

## Checklist

- [x] Tests updated (or not required).
- [x] Documentation updated (or not required).

PR #798, opened 2025-05-13
feat: reapply "feat(webpack): add StartupChunkDependenciesRuntimeModule to fix chunk splitting"

gaoachao:
## Summary

Rspack now has exposed `hasChunkEntryDependentChunks` on ChunkGraph.

Related links:
- https://github.com/lynx-family/lynx-stack/pull/658
- https://github.com/lynx-family/lynx-stack/pull/705
- https://github.com/web-infra-dev/rspack/releases/tag/v1.3.10
- https://github.com/web-infra-dev/rspack/pull/10286

## Checklist

### DIFF
--- a/packages/webpack/chunk-loading-webpack-plugin/src/StartupChunkDependenciesPlugin.ts
+++ b/packages/webpack/chunk-loading-webpack-plugin/src/StartupChunkDependenciesPlugin.ts
@@ -5,7 +5,24 @@ import type { Chunk, Compiler } from 'webpack';
 
 import { createStartupChunkDependenciesRuntimeModule } from './StartupChunkDependenciesRuntimeModule.js';
 
-import type { StartupChunkDependenciesPluginOptions } from './index.js';
+/**
+ * The options for StartupChunkDependenciesPlugin
+ */
+interface StartupChunkDependenciesPluginOptions {
+  /**
+   * Specifies the chunk loading method
+   * @defaultValue 'lynx'
+   * @remarks Currently only 'lynx' is supported
+   */
+  chunkLoading: string;
+
+  /**
+   * Whether to enable async chunk loading
+   * @defaultValue true
+   * @remarks Currently only async loading mode is supported
+   */
+  asyncChunkLoading: boolean;
+}
 
 const PLUGIN_NAME = 'StartupChunkDependenciesPlugin';
 
--- a/packages/webpack/chunk-loading-webpack-plugin/src/StartupChunkDependenciesPlugin.ts
+++ b/packages/webpack/chunk-loading-webpack-plugin/src/StartupChunkDependenciesPlugin.ts
@@ -1,0 +1,59 @@
+// Copyright 2024 The Lynx Authors. All rights reserved.
+// Licensed under the Apache License Version 2.0 that can be found in the
+// LICENSE file in the root directory of this source tree.
+import type { Chunk, Compiler } from 'webpack';
+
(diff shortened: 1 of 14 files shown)


packages/webpack/chunk-loading-webpack-plugin/src/StartupChunkDependenciesPlugin.ts
```diff
@@ -0,0 +1,79 @@
...
+            set.add(RuntimeGlobals.ensureChunk);
+            set.add(RuntimeGlobals.ensureChunkIncludeEntries);
+            compilation.addRuntimeModule(
+              chunk,
+              new StartupChunkDependenciesRuntimeModule(this.asyncChunkLoading),
+            );
+          }
+        },
+      );
+    });
```
colinaaa, line 96: It looks like we are missing the `compilation.hooks.runtimeRequirementInTree.for(RuntimeGlobals.startupEntrypoint)`

https://github.com/webpack/webpack/blob/f0e8744d5198fc777f5dadbefd1b09187ff4aa21/lib/runtime/StartupChunkDependenciesPlugin.js#L70

Is this expected?

gaoachao: This is expected. Since `RuntimeGlobals.startupEntrypoint` (i.e. `webpack_require.X`) hasn't been injected, this code wouldn't execute even if added. Therefore, I removed it.

The splitChunks can still work without `RuntimeGlobals.startupEntrypoint`.

gaoachao:
In this commit https://github.com/lynx-family/lynx-stack/pull/798/commits/e0e00e06930f21e8b7e0b7bbfea5767d0e1e1b17, I removed the test first because with the check of `chunkGraph.hasChunkEntryDependentChunks(chunk)`, `StartupChunkDependenciesRuntimeModule` won't be injected by default, so the test needs to be modified.

Since the test might involve splitChunk cases which would be a bit complex, I suggest we merge this PR first, and I will rewrite the tests in a subsequent PR. cc @colinaaa

packages/webpack/chunk-loading-webpack-plugin/src/StartupChunkDependenciesPlugin.ts
colinaaa: > Since `RuntimeGlobals.startupEntrypoint` (i.e. `webpack_require.X`) hasn't been injected, this code wouldn't execute even if added. Therefore, I removed it.

It might not be implemented immediately, but it could be effective in the future. I believe it would be better to align the implementation precisely with webpack.

It looks like this is required for chunks that does not have runtime but have entry modules. I think it may be added when `optimization.runtimeChunk` is enabled.

```js
compilation.hooks.additionalChunkRuntimeRequirements.tap(
  "CommonJsChunkFormatPlugin",
  (chunk, set, { chunkGraph }) => {
    if (chunk.hasRuntime()) return;
    if (chunkGraph.getNumberOfEntryModules(chunk) > 0) {
      set.add(RuntimeGlobals.require);
      set.add(RuntimeGlobals.startupEntrypoint);
      set.add(RuntimeGlobals.externalInstallChunk);
    }
  }
);
```

gaoachao: I've added `StartupEntrypointRuntimeModule`. Will add `CommonJsChunkFormatPlugin `(with `RuntimeGlobals.startupEntrypoint`) in the next PR.

colinaaa (approved):

PR #801, opened 2025-05-13
chore(deps): update swc_core to v23.2.0 and rust to v1.82.0

gaoachao:
## Summary

Rust 1.81.0 will crash due to the updated dependency `sourcemap` using `iter::repeat_n()`, so we need to upgrade to 1.82.0.

See:
- https://github.com/web-infra-dev/rspack/pull/10293
- https://github.com/rust-lang/rust/issues/104434

## Checklist


colinaaa (approved):

rust-toolchain
```diff
@@ -1 +1 @@
-1.81.0
+1.82.0
```
colinaaa: Can we upgrade to a higher version?

colinaaa (commented):
The test seems to be failing consistently.

colinaaa (approved):

PR #826, opened 2025-05-16
feat(react/runtime): add context tracking for worklet execution

Yradex:
## Summary

Allow the framework to get the currently running main thread function ctx.

## Checklist


packages/react/worklet-runtime/src/ctxTrace.ts
```diff
@@ -0,0 +1,22 @@
...
+import type { ClosureValueType, Worklet } from './bindings/types.js';
+
+interface CtxTrace {
+  ctx: Worklet;
+}
+
+export let currentCtx: CtxTrace | undefined;
+
+export function traceCtxCall(ctx: Worklet, _params: ClosureValueType[]): void {
+  console.log('traceCtxCall', ctx);
```
colinaaa:
```suggestion
```

colinaaa (changes requested):

PR #840, opened 2025-05-19
feat(react/runtime): support MTS functions to run at the first screen

Yradex:
## Summary

Support MTS functions to run before hydration finished. Notice that `runOnBackground` tasks are delayed until hydration finished.

This PR contains three parts:
- Support MTS functions to run at the first screen by renaming `_lepusWorkletHash` to `_wkltId`
- Add a MTS hydrate process, which merges the main thread value(higher priority) and the background initial value of the `MainThreadRef`s.
- Introduce a delay mechanism for `runOnBackground` tasks to ensure they are executed after hydration.

## Checklist

### DIFF
--- a/packages/react/runtime/src/gesture/processGesture.ts
+++ b/packages/react/runtime/src/gesture/processGesture.ts
@@ -10,7 +10,12 @@ function isSerializedGesture(gesture: GestureKind): boolean {
   return gesture.__isSerialized ?? false;
 }
 
-function getGestureInfo(gesture: BaseGesture, oldGesture: BaseGesture | undefined, isFirstScreen: boolean) {
+function getGestureInfo(
+  gesture: BaseGesture,
+  oldGesture: BaseGesture | undefined,
+  isFirstScreen: boolean,
+  dom: FiberElement,
+) {
   const config = {
     callbacks: [],
   } as GestureConfig;
@@ -25,7 +30,7 @@ function getGestureInfo(gesture: BaseGesture, oldGesture: BaseGesture | undefine
   ) {
     const callback = baseGesture.callbacks[key]!;
     const oldCallback = oldGesture?.callbacks[key];
-    onWorkletCtxUpdate(callback, oldCallback, isFirstScreen);
+    onWorkletCtxUpdate(callback, oldCallback, isFirstScreen, dom);
     config.callbacks.push({
       name: key,
       callback: callback,
@@ -72,7 +77,7 @@ export function processGesture(
     const baseGesture = gesture as BaseGesture;
     const oldBaseGesture = oldGesture as BaseGesture | undefined;
 
-    const { config, relationMap } = getGestureInfo(baseGesture, oldBaseGesture, isFirstScreen);
+    const { config, relationMap } = getGestureInfo(baseGesture, oldBaseGesture, isFirstScreen, dom);
     __SetGestureDetector(
       dom,
       baseGesture.id,
--- a/packages/react/runtime/src/snapshot/workletEvent.ts
+++ b/packages/react/runtime/src/snapshot/workletEvent.ts
@@ -23,7 +23,7 @@ function updateWorkletEvent(
   value._workletType = workletType;
 
   if (workletType === 'main-thread') {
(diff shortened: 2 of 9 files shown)


packages/react/worklet-runtime/src/workletRef.ts
```diff
@@ -7,20 +7,30 @@ import { profile } from './utils/profile.js';
...
 function initWorkletRef(): RefImpl {
   return (impl = {
     _workletRefMap: {},
+    /**
+     * Map of worklet refs that are created during first screen rendering.
+     * These refs are created with negative IDs and need to be hydrated
+     * when the app starts. The map is cleared after hydration is complete
+     * to free up memory.
+     */
+    _firstScreenWorkletRefMap: {},
```
f0rdream, line 30: In what case would we have non firstScreenWorkletRef ?

Supposing all the `useMainThreadRef`s would be run on render

packages/react/worklet-runtime/src/hydrate.ts
```diff
@@ -0,0 +1,93 @@
...
+ */
+function hydrateDelayRunOnBackgroundTasks(
+  fnObjs: Record<string, JsFnHandle>, // example: {"_jsFn1":{"_jsFnId":1}}
+  firstScreenFnObjs: Record<string, JsFnHandle>, // example: {"_jsFn1":{"_isFirstScreen":true,"_delayIndexes":[0]}}
+  execId: number,
+) {
+  for (const fnName in fnObjs) {
+    const fnObj = fnObjs[fnName]!;
+    const firstScreenFnObj: JsFnHandle | undefined = firstScreenFnObjs[fnName]!;
+    if (!firstScreenFnObj?._delayIndexes) {
```
f0rdream: Maybe **indices**

packages/react/runtime/src/gesture/processGesture.ts
```diff
@@ -63,13 +63,16 @@ export function processGesture(
   }
 
   if (gesture.type === GestureTypeInner.COMPOSED) {
-    for (const subGesture of (gesture as ComposedGesture).gestures) {
-      processGesture(dom, subGesture, { domSet: true });
+    for (const [index, subGesture] of (gesture as ComposedGesture).gestures.entries()) {
+      processGesture(dom, subGesture, (oldGesture as ComposedGesture)?.gestures[index], isFirstScreen, {
```
f0rdream, line 72: `processGesture` need more ajustment to meet new architecture. For example, the first pass has 5 gestures, and the second pass has 4 gestures, there will be out of index error

f0rdream: I need to think more about this

packages/react/worklet-runtime/src/workletRef.ts
Yradex: For example, some of the components won't render at the first screen at all. 
But I think I might be failed to understand your concern. Please comment if that's the case.

packages/react/runtime/src/gesture/processGesture.ts
Yradex: I don't think there would be an error in this case. But sure, there would be edge cases. 
However, I would be very grateful if you could help adding more test cases of gestures!

packages/react/worklet-runtime/src/hydrate.ts
```diff
@@ -0,0 +1,93 @@
...
+ * @param value The new value for the WorkletRef.
+ */
+function hydrateMainThreadRef(refId: WorkletRefId, value: WorkletRefImpl<unknown> | { current: unknown }) {
+  if ('_initValue' in value) {
+    // The ref has not been accessed yet.
+    return;
+  }
+  const ref = lynxWorkletImpl!._refImpl._workletRefMap[refId]!;
+  if (ref.current instanceof Element) {
+    // Monified by `main-thread:ref`
```
hzy: typo?
```suggestion
    // Modified by `main-thread:ref`
```

hzy (approved):

hzy (dismissed):

colinaaa (dismissed):

colinaaa (approved):

PR #842, opened 2025-05-19
WIP: add prefetchWorker of web-platform-rsbuild-plugin

Sherry-hue:
## Summary

add `prefetchWorker` configuration of web-platform-rsbuild-plugin.

When enable it, a prefetch link of the worker-async-chunk will be inserted into the generated HTML.

Before:

- Performance: image

- NetWork: image

After:

- Performance: image

- NetWork: image

## Checklist

### DIFF
--- a/packages/web-platform/web-explorer/index.html
+++ b/packages/web-platform/web-explorer/index.html
@@ -6,11 +6,9 @@
     <meta name="theme-color" content="#000000">
     <link rel="manifest" href="manifest.json">
     <link rel="icon" type="image/png" href="icons/icon-128x128.png">
-    <link rel="stylesheet" href="dist/static/css/index.css">
     <title>Lynx Explorer on Web Platform</title>
     <script type="module" src="https://unpkg.com/qr-scanner@1.4.2"></script>
-    <script type="module" src="preinstalled/vant-touch.js"></script>
-    <script type="module" src="./dist/static/js/index.js"></script>
+    <script type="module" src="../preinstalled/vant-touch.js"></script>
     <style>
     #install-button {
       position: fixed;
--- a/packages/web-platform/web-explorer/index.ts
+++ b/packages/web-platform/web-explorer/index.ts
@@ -9,7 +9,7 @@ let lynxView = document.getElementById('lynx-view') as LynxView;
 const backButton = document.getElementById('back-button') as HTMLDivElement;
 const nav = document.getElementById('nav') as HTMLDivElement;
 
-const homepage = 'preinstalled/homepage.json';
+const homepage = '../preinstalled/homepage.json';
 
 backButton.addEventListener('click', () => {
   setLynxViewUrl(homepage);
--- a/packages/web-platform/web-rsbuild-plugin/src/pluginWebPlatform.ts
+++ b/packages/web-platform/web-rsbuild-plugin/src/pluginWebPlatform.ts
@@ -4,6 +4,8 @@
 
 import type { RsbuildPlugin } from '@rsbuild/core';
 import path from 'path';
+import { merge } from 'webpack-merge';
+import { PrefetchWorkerPlugin } from './plugins/prefetch-worker.js';
 
 const __filename = new URL('', import.meta.url).pathname;
 const __dirname = path.dirname(__filename);
@@ -30,6 +32,12 @@ export interface PluginWebPlatformOptions {
    * When enabled, nativeModules will be packaged directly into the worker chunk instead of being transferred through Blob.
    */
   nativeModulesPath?: string;
+  /**
+   * Prefetch the worker resource.
+   *
+   * @default false
+   */
+  prefetchWorker?: boolean;
 }
 
 /**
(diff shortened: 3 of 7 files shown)


packages/web-platform/web-explorer/index.html
```diff
@@ -6,11 +6,9 @@
     <meta name="theme-color" content="#000000">
     <link rel="manifest" href="manifest.json">
     <link rel="icon" type="image/png" href="icons/icon-128x128.png">
-    <link rel="stylesheet" href="dist/static/css/index.css">
     <title>Lynx Explorer on Web Platform</title>
     <script type="module" src="https://unpkg.com/qr-scanner@1.4.2"></script>
-    <script type="module" src="preinstalled/vant-touch.js"></script>
-    <script type="module" src="./dist/static/js/index.js"></script>
+    <script type="module" src="../preinstalled/vant-touch.js"></script>
```
PupilTong, line 11: Why this url is changed?

packages/web-platform/web-explorer/index.ts
```diff
@@ -9,7 +9,7 @@ let lynxView = document.getElementById('lynx-view') as LynxView;
 const backButton = document.getElementById('back-button') as HTMLDivElement;
 const nav = document.getElementById('nav') as HTMLDivElement;
 
-const homepage = 'preinstalled/homepage.json';
+const homepage = '../preinstalled/homepage.json';
```
PupilTong, line 12: any explain about this?

PupilTong (approved):

packages/web-platform/web-rsbuild-plugin/src/pluginWebPlatform.ts
```diff
@@ -85,13 +93,36 @@ export function pluginWebPlatform(
...
           };
         }
+
+        if (options.prefetchWorker === true) {
+          if (!config.tools) {
+            config.tools = {};
+          }
+          const originToolsRspack = config.tools?.rspack;
+          if (typeof originToolsRspack === 'object') {
+            config.tools.rspack = (config, { HtmlPlugin }) => {
```
colinaaa, line 103: I don't think we need to deal with `tools.rspack` manually. You may use `api.modifyBundlerChain` or `api.modifyRsbuildConfig` to add your Webpack plugin.

colinaaa (changes requested):

packages/web-platform/web-rsbuild-plugin/src/plugins/prefetch-worker.ts
```diff
@@ -0,0 +1,129 @@
...
+  ) {
+    this.HtmlPlugin = options.HtmlPlugin;
+  }
+
+  apply(compiler: Compiler) {
+    if (
+      compiler.options.mode !== 'production'
+      || typeof compiler.options.output?.publicPath === 'function'
+    ) {
+      console.warn(
```
colinaaa, line 26: I would prefer throwing an error instead of console a warning and bailout.

packages/web-platform/web-rsbuild-plugin/src/plugins/prefetch-worker.ts
```diff
@@ -0,0 +1,129 @@
...
+        'PrefetchWorkerPlugin only works in production mode and when publicPath is a string',
+      );
+      return;
+    }
+
+    compiler.hooks.compilation.tap(
+      'PrefetchWorkerPlugin',
+      (compilation: Compilation) => {
+        if (!this.HtmlPlugin) {
+          console.warn('HtmlWebpackPlugin not found');
```
colinaaa, line 36: ditto here.

packages/web-platform/web-rsbuild-plugin/src/plugins/prefetch-worker.ts
```diff
@@ -0,0 +1,129 @@
...
+          return;
+        }
+
+        const htmlPluginHooks = (this.HtmlPlugin as any).getCompilationHooks(
```
colinaaa, line 40: Why do we need the `as any` here? It looks like we have correct type for `this.HtmlPlugin`

packages/web-platform/web-rsbuild-plugin/src/plugins/prefetch-worker.ts
```diff
@@ -0,0 +1,129 @@
...
+          compilation,
+        );
+        htmlPluginHooks.alterAssetTags.tap(
+          'PrefetchWorkerPlugin',
+          (data: any) => {
```
colinaaa, line 45: ditto. Why do we need `any`?

packages/web-platform/web-explorer/index.html
Sherry-hue: Now the html needs to use dist/index.html, and the path has changed.

Because this function depends on the html-rspack-config plugin to inject prefetch link, then I deleted `rspack.htmlPlugin`

packages/web-platform/web-rsbuild-plugin/src/plugins/prefetch-worker.ts
```diff
@@ -0,0 +1,129 @@
...
+              const outputRelativePath = compilation.getPath(
+                (compilation.options.output.chunkFilename as string)!.replace(
+                  /\[name\]/g,
+                  '[id]',
+                ),
+                {
+                  chunk: Array.from(compilation.chunks).find(i => i.id === id),
+                },
+              );
+              const finalPath = path.join(
```
colinaaa, line 57: I think this would not work on Windows. Should we use `path.posix` instead?
```suggestion
              const finalPath = path.posix.join(
```

Sherry-hue:
> BTW, I think this is quite the identical with the [`performance.prefetch`](https://rsbuild.dev/config/performance/prefetch) that Rsbuild provided. Can we use existing configuration instead of implement our own plugin?

We recommend that businesses use this method first, so this PR is closed first.

PR #850, opened 2025-05-19
chore: add more tests for fp-only-test

PupilTong:


.github/workflows/test.yml
```diff
@@ -232,6 +232,7 @@ jobs:
       - code-style-check
       - playwright-linux
       - playwright-linux-all-on-ui
+      - playwright-linux-fp-only
```
colinaaa, line 235: It looks like this job is not added to `test.yml`

colinaaa (approved):

colinaaa (approved):

Issue #848, opened 2025-05-19
[Bug]: Typescript errors in fresh `rspeedy` project when using `Suspense` or `Fragment` from `@lynx-js/react`


plesiecki:
### System Info

```
 System:
    OS: macOS 15.3.1
    CPU: (10) arm64 Apple M1 Pro
    Memory: 98.20 MB / 32.00 GB
    Shell: 5.9 - /bin/zsh
  Binaries:
    Node: 22.15.0 - ~/n/bin/node
    npm: 10.9.2 - ~/n/bin/npm
    pnpm: 10.6.5 - ~/n/bin/pnpm
    bun: 1.0.26 - ~/.bun/bin/bun
    Watchman: 2025.04.28.00 - /opt/homebrew/bin/watchman
  npmPackages:
    @lynx-js/qrcode-rsbuild-plugin: ^0.3.6 => 0.3.6 
    @lynx-js/react: ^0.108.1 => 0.108.1 
    @lynx-js/react-rsbuild-plugin: ^0.9.10 => 0.9.10 
    @lynx-js/rspeedy: ^0.9.5 => 0.9.5 
    @lynx-js/types: ^3.3.0 => 3.3.0 
```

### Details

Typescript reports errors in fresh `rspeedy` project created with `npm create rspeedy@latest` when using `Suspense` or `Fragment` from `@lynx-js/react`.

TS error:
```
'Fragment' cannot be used as a JSX component.
  Its return type 'ReactNode' is not a valid JSX element.
    Type 'undefined' is not assignable to type 'Element | null'.ts(2786)
```

I've noticed one workaround. Removing the `"jsxImportSource": "@lynx-js/react"` part from the `tsconfig.json` "fixes" TS errors.

### Reproduce Steps

1. Run `npm create rspeedy@latest`, select Typescript language and continue
2. `npm i`
3. Edit the `src/App.tsx` module as follows:
```diff
- import { useCallback, useEffect, useState } from '@lynx-js/react'
+ import { useCallback, useEffect, useState, Fragment } from '@lynx-js/react'
...
   return (
     <view>
+      <Fragment></Fragment>
       <view className='Background' />
```

PR #854, opened 2025-05-20
fix(react/types): export `JSX.ElementType`

colinaaa:
## Summary

Export types like `react/jsx-runtime` does. Including:

- `ElementType`: for `<Fragment>`, `<Suspense>`, `memo()`, `forwardRef()`, etc.
- `ElementClass`: for class component
- `ElementAttributesProperty`: for `props`
- `ElementChildrenAttribute`: for `children`
- `IntrinsicClassAttributes`: for `ref` on class component

Note that we did not change the `IntrinsicAttributes` since we don't have correct definition at `@lynx-js/types`.

The type test has been relocated from `packages/react/runtime` to `examples/react` to ensure accurate dependencies and references.

fix: #848

## Checklist

### DIFF
--- a/.changeset/chubby-icons-cross.md
+++ b/.changeset/chubby-icons-cross.md
@@ -1,0 +1,5 @@
+---
+"@lynx-js/react": patch
+---
+
+Fix type error when using `Suspense` with `"jsx": "react-jsx"`.
--- a/examples/react/package.json
+++ b/examples/react/package.json
@@ -5,7 +5,8 @@
   "type": "module",
   "scripts": {
     "build": "rspeedy build",
-    "dev": "rspeedy dev"
+    "dev": "rspeedy dev",
+    "test:type": "vitest --typecheck.only"
   },
   "dependencies": {
     "@lynx-js/react": "workspace:*"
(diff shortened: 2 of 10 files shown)


Yradex (approved):

plesiecki:
Why can't we extend `IntrinsicAttributes` with `React.JSX.IntrinsicAttributes`? This would resolve TS errors when using the `key` prop, like here for an example: https://github.com/lynx-family/lynx-examples/blob/main/examples/animation/src/transition_toggle/index.tsx#L46

colinaaa:
> Note that we did not change the `IntrinsicAttributes` since we don't have correct definition at `@lynx-js/types`.

As I mentioned in the PR description:

> Note that we did not change the IntrinsicAttributes since we don't have correct definition at @lynx-js/types.

plesiecki:
Ok, I think I know what you mean. This would require changing the definition in `@lynx-js/types` in the first place, wouldn't it?

plesiecki:
Would you like me to submit an issue regarding the `key` prop?

colinaaa:
> Ok, I think I know what you mean. This would require changing the definition in `@lynx-js/types` in the first place, wouldn't it?

Yeah, but I'm not sure what attributes we need. E.g.: the `key` is definitely not an attribute of Lynx Engine.

> Would you like me to submit an issue regarding the `key` prop?

Sure! That would be great.

PR #862, opened 2025-05-20
fix(web): x-list should observe property list-type change.

Sherry-hue:
## Summary

fix: x-list should observe property list-type change.

Before this commit, list-type only works when it was first assigned.

## Checklist


packages/web-platform/web-elements/src/XList/XListWaterfall.ts
```diff
@@ -182,7 +182,7 @@ export class XListWaterfall
     this.#resizeObserver = new ResizeObserver(() => {
       // may cause: Resizeobserver loop completed with undelivered notifications
       // https://developer.mozilla.org/en-US/docs/Web/API/ResizeObserver#observation_errors
-      boostedQueueMicrotask(() => {
+      requestAnimationFrame(() => {
```
PupilTong, line 185: if webkit crashes, how about we use requestAnimationFrame on webkit and use boostedQueueMicrotask for chromium and firefox?

Issue #871, opened 2025-05-21
TS errors when using the `key` prop on components


colinaaa:
Why can't we extend `IntrinsicAttributes` with `React.JSX.IntrinsicAttributes`? This would resolve TS errors when using the `key` prop, like here for an example: https://github.com/lynx-family/lynx-examples/blob/main/examples/animation/src/transition_toggle/index.tsx#L46

_Originally posted by @plesiecki in https://github.com/lynx-family/lynx-stack/issues/854#issuecomment-2893915164_

PR #872, opened 2025-05-21
fix(react/types): add `key` to `JSX.IntrinsicAttributes`

colinaaa:
## Summary

Add `key` to `JSX.IntrinsicAttributes`.

```jsx
<Foo key='foo' />
     ^^^ Type '{ key: string; }' is not assignable to type 'IntrinsicAttributes'.
           Property 'key' does not exist on type 'IntrinsicAttributes'.ts(2322)
```

> [!NOTE]
> We did not extends `React.JSX.IntrinsicAttributes` like [`@types/react`](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/981449c691b9be0fca569c48151fc57606f5ea7a/types/react/jsx-runtime.d.ts#L11) does.
> This is because the `React.JSX.IntrinsicAttributes` would be populated by `@lynx-js/types`, which contains incorrect properties.
 

fix: #871 

## Checklist

### DIFF
--- a/packages/react/types/react.d.ts
+++ b/packages/react/types/react.d.ts
@@ -95,7 +95,7 @@ declare module '@lynx-js/types' {
   interface StandardProps {
     children?: ReactNode;
     ref?: Ref<NodesRef>;
-    key?: Key;
+    key?: Key | null | undefined;
     'main-thread:ref'?: Ref<MainThread.Element>;
   }
 
--- a/.changeset/few-stars-write.md
+++ b/.changeset/few-stars-write.md
@@ -1,0 +1,5 @@
+---
+"@lynx-js/react": patch
+---
+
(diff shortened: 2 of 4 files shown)


packages/react/types/react.d.ts
```diff
@@ -95,7 +95,7 @@ declare module '@lynx-js/types' {
   interface StandardProps {
     children?: ReactNode;
     ref?: Ref<NodesRef>;
-    key?: Key;
+    key?: Key | null | undefined;
```
colinaaa, line 98: The `null | undefined` has been added to align with `React.Attributes`. See: https://github.com/DefinitelyTyped/DefinitelyTyped/blob/981449c691b9be0fca569c48151fc57606f5ea7a/types/react/index.d.ts#L248

ManiaciaChao:
LGTM

hzy (approved):

PR #874, opened 2025-05-21
feat: support output.inlineScripts

### LINKED ISSUE #870: [Feature]: Support `output.inlineScripts`

Follow up #502. Experimental switch. When set to `false`, `encodeData.manifest` will only contain the `/app-service.js` entry, which is to fix the duplicate chunks in chunkSplit with multiple entries.

```js
export default {
  output: {
    inlineScripts: false,
  },
};
```

gaoachao:
## Summary

#870  

follow up #502

Support `output.inlineScripts`, which controls whether to inline scripts files when LynxEncodePlugin generates the manifest file.

## Checklist

### DIFF
--- a/.changeset/afraid-meals-attend.md
+++ b/.changeset/afraid-meals-attend.md
@@ -1,0 +1,19 @@
+---
+"@lynx-js/template-webpack-plugin": patch
+"@lynx-js/react-rsbuild-plugin": patch
+"@lynx-js/rspeedy": patch
+---
+
+Support `output.inlineScripts`, which controls whether to inline scripts files when LynxEncodePlugin generates the manifest file.
+
+example:
+
+```js
+import { defineConfig } from '@lynx-js/rspeedy';
+
+export default defineConfig({
+  output: {
+    inlineScripts: false,
+  },
+});
+```
--- a/.changeset/afraid-meals-attend.md
+++ b/.changeset/afraid-meals-attend.md
@@ -4,7 +4,9 @@
 "@lynx-js/rspeedy": patch
 ---
 
-Support `output.inlineScripts`, which controls whether to inline scripts files when LynxEncodePlugin generates the manifest file.
+Support `output.inlineScripts`, which controls whether to inline scripts into Lynx bundle (`.lynx.bundle`).
+
+Only background thread scripts can remain non-inlined, whereas the main thread script is always inlined.
 
 example:
 
--- a/packages/rspeedy/core/src/config/output/index.ts
+++ b/packages/rspeedy/core/src/config/output/index.ts
@@ -312,7 +312,7 @@ export interface Output {
    *
    * @example
    *
-   * Disable inlining scripts.
+   * Disable inlining background thread scripts.
    * ```js
    * import { defineConfig } from '@lynx-js/rspeedy'
    *
--- a/packages/rspeedy/core/src/config/output/index.ts
+++ b/packages/rspeedy/core/src/config/output/index.ts
@@ -300,7 +300,7 @@ export interface Output {
   filenameHash?: boolean | string | undefined
 
   /**
-   * The {@link Output.inlineScripts} option controls whether to inline scripts files when LynxEncodePlugin generates the manifest file.
+   * The {@link Output.inlineScripts} option controls whether to inline scripts into Lynx bundle (`.lynx.bundle`).
    *
    * @remarks
    *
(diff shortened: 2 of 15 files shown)


.changeset/afraid-meals-attend.md
```diff
@@ -0,0 +1,19 @@
+---
+"@lynx-js/template-webpack-plugin": patch
+"@lynx-js/react-rsbuild-plugin": patch
+"@lynx-js/rspeedy": patch
+---
+
+Support `output.inlineScripts`, which controls whether to inline scripts files when LynxEncodePlugin generates the manifest file.
```
colinaaa:
```suggestion
Support `output.inlineScripts`, which controls whether to inline scripts into Lynx bundle (`.lynx.bundle`).

Only background thread scripts can remain non-inlined, whereas the main thread script is always inlined.
```

packages/rspeedy/core/src/config/output/index.ts
```diff
@@ -299,6 +299,30 @@ export interface Output {
    */
   filenameHash?: boolean | string | undefined
 
+  /**
+   * The {@link Output.inlineScripts} option controls whether to inline scripts files when LynxEncodePlugin generates the manifest file.
```
colinaaa: ditto.
```suggestion
   * The {@link Output.inlineScripts} option controls whether to inline scripts into Lynx bundle (`.lynx.bundle`).
```

packages/rspeedy/core/src/config/output/index.ts
```diff
@@ -299,6 +299,30 @@ export interface Output {
...
+   *
+   * @remarks
+   *
+   * If no value is provided, the default value would be `true`.
+   *
+   * This is different with {@link https://rsbuild.dev/config/output/inline-scripts | output.inlineScripts } since we normally want to inline scripts in Lynx manifest file.
+   *
```
colinaaa, line 312: We should also add the caveat about main thread scripts would currently always been inlined.

packages/rspeedy/core/src/config/output/index.ts
```diff
@@ -299,6 +299,30 @@ export interface Output {
...
+   * @example
+   *
+   * Disable inlining scripts.
```
colinaaa: Or maybe add the caveat here :)

packages/rspeedy/plugin-react/src/entry.ts
```diff
@@ -222,7 +226,7 @@ export function applyEntry(
         }])
         .end()
         .plugin(`${LynxEncodePlugin.name}`)
-        .use(LynxEncodePlugin, [{}])
+        .use(LynxEncodePlugin, [{ inlineScripts }])
```
colinaaa, line 230: Do we need this (disable inline background-thread/main-thread scripts into `.{lynx,web}.bundle`) for the `WebWebpackPlugin`? /cc @PupilTong

PupilTong: Currently It won't work on web.

packages/rspeedy/plugin-react/src/entry.ts
```diff
@@ -198,6 +198,10 @@ export function applyEntry(
     })
 
     if (isLynx) {
+      const inlineScripts = typeof config.output?.inlineScripts === 'boolean'
```
colinaaa: I think instead of `config.output`, we should use `   environment.config.output.inlineScripts`

Consider the following `lynx.config.js`:

```js
export default {
  environments: {
    lynx: { output: { inlineScripts: false } },
    web: { output: { inlineScripts: true } },
  },
}
```

You would get `true` when using `config.output.inlineScripts`, but `false` is expected.

colinaaa (commented):
Are we going to support an object for `output.inlineScripts` (maybe in next PRs)? Since Rsbuild supports that:

```ts
type InlineScriptsTest =
  | RegExp
  | ((params: { size: number; name: string }) => boolean);

type InlineScripts =
  | boolean
  | InlineScriptsTest
  | {
      enable?: boolean | 'auto';
      test: InlineScriptsTest;
    };
```

If we are going to do so, I since we should not close #870 in this PR, yet.

packages/rspeedy/plugin-react/src/entry.ts
gaoachao: https://github.com/lynx-family/lynx-stack/pull/874/commits/a56169a8954cfd3618cced590fb17d7fa317ffb6

colinaaa (approved):

PR #883, opened 2025-05-22
feat: extend inlineScripts type

### LINKED ISSUE #870: [Feature]: Support `output.inlineScripts`

gaoachao:
## Summary

follow up #874
close #870

Align `output.inlineScripts` with Rsbuild

```ts
type InlineChunkTestFunction = (params: {
  size: number;
  name: string;
}) => boolean;

type InlineChunkTest = RegExp | InlineChunkTestFunction;

type InlineChunkConfig =
  | boolean
  | InlineChunkTest
  | { enable?: boolean | 'auto'; test: InlineChunkTest };
```

Related links:
- https://rsbuild.dev/config/output/inline-scripts#outputinlinescripts

## Checklist


packages/webpack/template-webpack-plugin/src/LynxEncodePlugin.ts
```diff
@@ -144,27 +156,44 @@ export class LynxEncodePluginImpl {
...
       }, async (args) => {
-        const { encodeData } = args;
-        const { manifest } = encodeData;
+        const { encodeData, originManifest } = args;
+
+        const [inlinedManifest, externalManifest] = Object.entries(
+          originManifest,
+        )
+          .reduce(
+            ([inlined, external], [name, assert]) => {
```
colinaaa:
```suggestion
            ([inlined, external], [name, asset]) => {
```

packages/webpack/template-webpack-plugin/src/LynxTemplatePlugin.ts
```diff
@@ -812,6 +814,14 @@ class LynxTemplatePluginImpl {
 
     const { encodeData } = await hooks.beforeEncode.promise({
       encodeData: encodeRawData,
+      originManifest: Object.fromEntries(
```
colinaaa: I don't think we need this extra field. `hooks.beforeEncode` is a `AsyncSeriesWaterfallHook` that works like waterfall.

We should always use the `encodeData.manifest` and modify `encodeData.manifest` in taps.

You may get the `source.size()` using `compilation.getAsset()`.

packages/webpack/template-webpack-plugin/src/LynxEncodePlugin.ts
```diff
@@ -180,32 +209,35 @@ export class LynxEncodePluginImpl {
...
-          //   'initial-chunk2': `<content-of-initial-chunk>`,
+          //   'inlined-initial-chunk1': `<content>`,
+          //   'inlined-initial-chunk2': `<content>`,
           // },
           // ```
           '/app-service.js': [
             this.#appServiceBanner(),
-            Object.keys(manifest)
-              .map((name) =>
+            ...[externalManifest, inlinedManifest].flatMap(manifest =>
```
colinaaa, line 190: I don't think it's correct to use `lynx.requireModule` on external chunks.

1. It's slow. Network requests would be sent one-by-one and result in waterfall.
2. It' not necessary. The initial chunks would be requested by the entry chunk using webpack chunk-loading.

packages/webpack/template-webpack-plugin/src/LynxEncodePlugin.ts
```diff
@@ -250,6 +282,29 @@ export class LynxEncodePluginImpl {
...
+    if (inlineConfig instanceof RegExp) {
+      return inlineConfig.test(name);
+    }
+
+    if (typeof inlineConfig === 'function') {
+      return inlineConfig({ size, name });
+    }
+
+    if (typeof inlineConfig === 'object') {
+      // Currently, rspeedy does not support enable: 'auto'
```
colinaaa: I didn’t notice any code preventing the use of `enable: 'auto'`. I would prefer supporting it here, where `enable: 'auto'` implies `inlineScripts: true`.

packages/rspeedy/core/src/config/output/index.ts
```diff
@@ -318,12 +320,12 @@ export interface Output {
    *
    * export default defineConfig({
    *   output: {
-   *     inlineScripts: false,
+   *     inlineScripts: /[\\/]background\.\w+\.js$/,
```
colinaaa: It seems like this wouldn't match the description "Disable inlining background thread scripts." of this example

packages/rspeedy/plugin-react/src/entry.ts
```diff
@@ -198,10 +198,13 @@ export function applyEntry(
...
-      const inlineScripts =
-        typeof environment.config.output?.inlineScripts === 'boolean'
-          ? environment.config.output.inlineScripts
-          : true
+      let inlineScripts
+      if (experimental_isLazyBundle) {
+        // TODO: support inlineScripts in lazyBundle
+        inlineScripts = true
+      } else {
+        inlineScripts = environment.config.output?.inlineScripts ?? true
```
colinaaa, line 208: Should we also set `firstScreenSyncTiming` to `'jsReady'` when `inlineScripts !== true`?

.changeset/tired-times-glow.md
```diff
@@ -0,0 +1,5 @@
+---
+"@lynx-js/template-webpack-plugin": patch
+---
+
+fix: remove inlinedAssets from inlinedManifest
```
colinaaa: Do we still need this changeset?

packages/rspeedy/core/test/config/validate.test.ts
```diff
@@ -835,6 +835,24 @@ describe('Config Validation', () => {
...
+        {
+          inlineScripts: {
+            enable: 'auto',
+            test: /[\\/]background\.\w+\.js$/,
+          },
+        },
+        {
+          inlineScripts: {
+            enable: true,
+            test: /[\\/]background\.\w+\.js$/,
```
colinaaa, line 853: Can we also add a tests to check the `output.inlineScripts.test` is a required option?

.changeset/tangy-ends-act.md
```diff
@@ -0,0 +1,22 @@
+---
+"@lynx-js/rspeedy": patch
+"@lynx-js/react-rsbuild-plugin": patch
+"@lynx-js/react-webpack-plugin": patch
+"@lynx-js/template-webpack-plugin": patch
+---
+
+Align `output.inlineScripts` with Rsbuild
```
colinaaa: It would be better to something like:

"Enable fine-grained control for `output.inlineScripts`"

I don't think much of the user would be able to know or care about Rsbuild's `output.inlineScripts`

colinaaa: And also an example would be great to have.

colinaaa (approved):
The current implementation LGTM. But there are more to think about to make it better:

1. We should not only apply `inlineScripts` to `LynxEncodePlugin`, the `WebEncodePlugin` should also support this.
    - We may want a base class and implement some common logic (e.g.: inline script, deleting inline assets)
2. We may want to allow customize the retry and error handling of external chunks
    - In Rsbuild Web, we have https://github.com/rspack-contrib/rsbuild-plugin-assets-retry
3. We may want to allow injecting external & inlined chunks
    - In Rsbuild Web, we have `api.modifyHTMLTags` that can add `<script>` to the HTML.

PR #884, opened 2025-05-22
feat(web): pageConfig add new source

Sherry-hue:
## Summary

feat: output pageConfig add a new source: `encodeOptions.sourceContent.config`.

## Checklist

### DIFF
--- a/.changeset/puny-monkeys-rule.md
+++ b/.changeset/puny-monkeys-rule.md
@@ -1,0 +1,7 @@
+---
+"@lynx-js/template-webpack-plugin": patch
+---
+
+feat: Merge the absent configurations for `.web.bundle`.
+
+Before this change, the configuration of pageConfig in `.web.bundle` was `compilerOptions`. After this commit, pageConfig will be a combination of `compilerOptions` and `sourceContent.config`.
--- a/packages/webpack/template-webpack-plugin/src/WebEncodePlugin.ts
+++ b/packages/webpack/template-webpack-plugin/src/WebEncodePlugin.ts
@@ -73,7 +73,10 @@ export class WebEncodePlugin {
             },
             customSections: encodeData.customSections,
             cardType: encodeData.sourceContent.dsl.substring(0, 5),
-            pageConfig: encodeData.compilerOptions,
+            pageConfig: {
+              ...encodeData.compilerOptions,
+              ...encodeData.sourceContent.config,
+            },
           });
           return encodeOptions;
         });
@@ -87,7 +90,7 @@ export class WebEncodePlugin {
               styleInfo: encodeOptions['styleInfo'],
               manifest: encodeOptions.manifest,
               cardType: encodeOptions['cardType'],
-              pageConfig: encodeOptions.compilerOptions,
+              pageConfig: encodeOptions['pageConfig'],
               lepusCode: {
                 // flatten the lepusCode to a single object
                 ...encodeOptions.lepusCode.lepusChunk,
--- a/packages/webpack/template-webpack-plugin/test/cases/web/page-config/a.js
+++ b/packages/webpack/template-webpack-plugin/test/cases/web/page-config/a.js
@@ -1,0 +1,1 @@
+console.info('**aaa**');
--- a/packages/webpack/template-webpack-plugin/test/cases/web/page-config/b.js
+++ b/packages/webpack/template-webpack-plugin/test/cases/web/page-config/b.js
@@ -1,0 +1,13 @@
+/// <reference types="vitest/globals" />
+
+import * as path from 'node:path';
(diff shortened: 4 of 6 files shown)


colinaaa (approved):

PupilTong (dismissed):

PupilTong (dismissed):

packages/webpack/web-webpack-plugin/test/cases/main-thread/page-config/rspack.config.js
```diff
@@ -0,0 +1,64 @@
+import { LynxTemplatePlugin } from '@lynx-js/template-webpack-plugin';
```
colinaaa: We should also import `LynxTemplatePlugin` from `src/`.

packages/webpack/template-webpack-plugin/src/WebEncodePlugin.ts
```diff
@@ -87,7 +87,13 @@ export class WebEncodePlugin {
               styleInfo: encodeOptions['styleInfo'],
               manifest: encodeOptions.manifest,
               cardType: encodeOptions['cardType'],
-              pageConfig: encodeOptions.compilerOptions,
+              pageConfig: {
+                ...encodeOptions.compilerOptions,
+                ...((encodeOptions['sourceContent'] || {}) as Record<
```
colinaaa: Prefer using nullish coalescing operator (`??`) instead of a logical or (`||`), as it is a safer operator.eslint[@typescript-eslint/prefer-nullish-coalescing](https://typescript-eslint.io/rules/prefer-nullish-coalescing)

packages/webpack/template-webpack-plugin/src/WebEncodePlugin.ts
```diff
@@ -87,7 +87,13 @@ export class WebEncodePlugin {
               styleInfo: encodeOptions['styleInfo'],
               manifest: encodeOptions.manifest,
               cardType: encodeOptions['cardType'],
-              pageConfig: encodeOptions.compilerOptions,
+              pageConfig: {
```
colinaaa: We already have a `pageConfig` field on `encodeOptions`, which is equal to `encodeData.compilerOptions`:

https://github.com/lynx-family/lynx-stack/blob/1a65d8779cf13be892051653a3634c6ec26099c1/packages/webpack/template-webpack-plugin/src/WebEncodePlugin.ts#L76

I think you should merge the two configs there (in `beforeEncode`), instead of in `encode`.

packages/webpack/web-webpack-plugin/test/cases/main-thread/page-config/rspack.config.js
```diff
@@ -0,0 +1,64 @@
...
+            name: 'page-config-test',
+            stage:
+              compiler.webpack.Compilation.PROCESS_ASSETS_STAGE_PRE_PROCESS,
+          },
+          () => {
+            const hooks = LynxTemplatePlugin.getLynxTemplatePluginHooks(
+              // @ts-expect-error compilation is compatible with webpack
+              compilation,
+            );
+            hooks.beforeEncode.tap('page-config-test', (args) => {
```
colinaaa: This test doesn't seem to validate functionality effectively; it only verifies the implementation.  

Could you add an end-to-end (e2e) test to ensure that configurations can be properly read at runtime?

colinaaa (changes requested):

packages/webpack/web-webpack-plugin/test/cases/main-thread/page-config/rspack.config.js
Sherry-hue: I don't think this is needed in this PR, in this PR we are only concerned with the final outputs, as it only involves build time.

Of course, it makes sense to have another test at runtime, but it should not appear in this PR

.changeset/puny-monkeys-rule.md
```diff
@@ -0,0 +1,5 @@
+---
+"@lynx-js/template-webpack-plugin": patch
+---
+
+feat: output pageConfig add a new source: `encodeOptions.sourceContent.config`.
```
colinaaa:
```suggestion
Include the absent configurations in `.web.bundle`.
```

colinaaa (changes requested):
Ensure the changeset is clear and user-friendly:

1. Highlight that this is a Web-only change by incorporating `.web.bundle`.
2. Replace terms like `pageConfig` and `encodeOptions.sourceContent.config` with "configurations" for consistency and simplicity.

.changeset/puny-monkeys-rule.md
Sherry-hue: I have updated it, if there is anything inappropriate, please tell me

PR #885, opened 2025-05-22
fix: typos

gaoachao:
## Summary

fix typos

## Checklist


upupming (dismissed):

.typos.toml
```diff
@@ -0,0 +1,20 @@
+# https://github.com/crate-ci/typos
+# cargo install typos-cli
```
colinaaa, line 2: Could we integrate this into the CI pipeline to prevent future regressions?

You may use the `taiki-e/install-action`

cspell.jsonc
```diff
@@ -173,6 +173,7 @@
     "/patches",
```
colinaaa, line 175: Do we still need cspell now that we have crate-ci/typos?

In other words, are these typos not being detected because they are ignored in cspell.jsonc, or does cspell not recognize them as typos?

packages/web-platform/web-constants/src/types/UpdateDataOptions.ts
```diff
@@ -8,7 +8,7 @@ export const enum UpdateDataType {
   Unknown = 0,
 
   // update by `setState` or `setData`
-  UpdateExplictByUser = 1,
```
PupilTong, line 11: this is copied from lynx main repo， could you please double check it?

PupilTong (approved):

packages/web-platform/web-constants/src/types/UpdateDataOptions.ts
colinaaa: I think I wrote this typo in Lynx Engine :) 

It's fine to make change here since it's a `const enum`. And we would never use this value now.

cspell.jsonc
gaoachao: cspell did not catch these as typos.

In my opinion, we should replace cspell with typos since it's more powerful and has better performance. And I've integrated typos into our CI pipeline.However, I'm still working on implementing typos in lint-staged (or we can simply skip the typos check during the commit phase).

I tested the following code, but it failed to work.
```diff
const fix = {
  '*.{ts,tsx,js,jsx,cjs,mjs,css,sass,scss,less,html,yml,yaml,rs,toml,md,json}':
    [
      'eslint --cache --fix --no-warn-ignored',
      'biome lint --write --no-errors-on-unmatched --files-ignore-unknown=true',
      'dprint fmt --allow-no-files --',
    ],
  '*.rs': [
    'cargo fmt --',
  ],
  'package.json': 'sort-package-json',
  '*': [
-   'cspell lint --config cspell.jsonc --gitignore --no-must-find-files',
+  'typos',
  ],
};
```

colinaaa: > However, I'm still working on implementing typos in lint-staged (or we can simply skip the typos check during the commit phase)

I would prefer not to include `typos` in lint-staged, as it can be challenging to ensure all developers install it.

.github/workflows/rust.yml
```diff
@@ -63,3 +63,11 @@ jobs:
           cache-key: test-${{ runner.os }}
       - name: Format
         run: cargo fmt --check
+
+  typos:
```
colinaaa: The `rust.yml` workflow would be run **after** build finished, which I think is a little too late for checking typos.

Can we add this to `code-style-check` in `test.yml`?

packages/web-platform/web-core-server/src/createLynxView.ts
```diff
@@ -34,7 +34,7 @@ interface LynxViewConfig extends
   templateName?: string;
   hydrateUrl: string;
   injectStyles: string;
-  overrideElemenTemplates?: Record<
+  overrideElementTemplates?: Record<
```
colinaaa, line 37: It looks like this is a public API. Should we mark the `overrideElemenTemplates` as deprecated and use the fixed one? /cc @PupilTong

.github/workflows/rust.yml
gaoachao: https://github.com/lynx-family/lynx-stack/pull/885/commits/4aaba709512bf0e2371c56a6e3b580df7795861d

packages/web-platform/web-core-server/src/createLynxView.ts
PupilTong: server is not available at this time. I think it's fine.

.github/workflows/test.yml
```diff
@@ -55,6 +55,15 @@ jobs:
         run: pnpm turbo api-extractor
       - name: Changeset Check
         run: pnpm changeset status --since=origin/main
+
+  typos:
```
colinaaa, line 59: Nitpick: Sort the jobs in alphabetical order.

colinaaa, line 59: We should include "typos" in the "jobs.done.needs". I'm fine with adding it in the next PR after it has run for a while.

gaoachao: Got it.

PupilTong (approved):

PR #898, opened 2025-05-23
fix: the lepus arg of beforeEmit hook to contain root chunk

upupming:
## Checklist


colinaaa (approved):
LGTM! I believe this is a great opportunity to rename the `lepus` param in `beforeEmit` since we already have a `minor` change in today's release.

Maybe `mainThreadChunk`?

We can do that in the next PR

upupming:
@colinaaa Great, I will rename it in the next PR!

PR #936, opened 2025-05-27
refactor: replace enhanced-resolve with rust-based resolver

gaoachao:
## Summary

Replace [enhanced-resolve](https://github.com/webpack/enhanced-resolve) with rust-based resolver.

Related link:
- https://github.com/web-infra-dev/rspack-resolver
- https://github.com/unrs/unrs-resolver
- https://github.com/oxc-project/oxc-resolver

## Checklist


packages/rspeedy/plugin-react-alias/src/index.ts
```diff
@@ -180,21 +180,37 @@ export function pluginReactAlias(options: Options): RsbuildPlugin {
...
+) {
+  let lazyExports: Record<string, string>
+  let resolverLazy: (directory: string, request: string) => ResolveResult
 
   return async (
     request: string,
   ): Promise<string> => {
-    const { default: resolver } = await import('enhanced-resolve')
+    const { ResolverFactory } = await import('unrs-resolver')
+    const resolver = new ResolverFactory({ conditionNames })
```
colinaaa: It looks like we need to create a new `ResolverFactory` for each `resolve` request. Can we move this into the `if` below?

gaoachao: Good catch! See: https://github.com/lynx-family/lynx-stack/pull/936/commits/fd2a9e3dd6d73d4b92a496a486ba7ed433ef9ec1

PR #949, opened 2025-05-28
fix(rspeedy/core): remove the default `output.iife: false`

colinaaa:
## Summary

The `output.iife: false` setting has been removed in this patch to address the "SyntaxError: invalid redefinition of parameter name" error.

This issue occurs in production when a `const` or `let` variable is declared with the same name as a variable provided by the Lynx Engine (e.g., `fetch`, `requestAnimationFrame`).

Originally, this setting was introduced to mitigate performance issues in PrimJS caused by closures. However, since that problem was resolved years ago, it is now safe to remove this setting.

## Checklist


luhc228:
LGTM👍

PR #966, opened 2025-05-29
chore: use https to download from apt

PupilTong:


colinaaa (commented):
The `.github/workflows/workflow-bench.yml` is also failing because of `apt install`. Could you also replace the HTTPS link there?

PR #970, opened 2025-05-30
feat(react/runtime): support 'main-thread' as an alias for 'main thread'

Yradex:
## Summary

Support the 'main-thread' directive as an alias for 'main thread'.

This change would provide a better experience when using AI tools to generate code automatically.

## Checklist


colinaaa (approved):
👍

Issue #893, opened 2025-05-23
[Feature]: Improved `use-sync-external-store`


colinaaa:
### What problem does this feature solve?

The original [`use-sync-external-store`](https://www.npmjs.com/package/use-sync-external-store) package works with ReactLynx, but with the following limitation:

1. Only CommonJS (CJS) exports is supported.  
   - This results in the [`concatenateModules`](https://rspack.dev/config/optimization#optimizationconcatenatemodules) optimization bailout for `@lynx-js/react` (a dependency of `use-sync-external-store`), leading to increased bundle size and slower execution.
2. `default` import of `react` is used.
   - This results in the whole `@lynx-js/react` being bundled. The [`usedExports`](https://rspack.dev/config/optimization#optimizationusedexports) would take no effect.

Since the `use-sync-external-store` is used in some popular library like [`zustand`](https://github.com/pmndrs/zustand), we should provide our own version.

### What does the proposed API of configuration look like?

A new package `@lynx-js/use-sync-external-store` will be added, with the exact same APIs exported but in **PURE** ESM.

The framework (`@lynx-js/react-rsbuild-plugin`) will alias the `use-sync-external-store`. No code changes are required for end developers.

Huxpro:
@colinaaa so if preact already include it as source, is the forked package here intended to be used an internal replacement within the `@lynx-js/react` or to be used by a fork of those libraries that depends on `use-sync-external-store` e.g. zustand?

Huxpro:
@colinaaa if `@lynx-js/react` (a.k.a. `preact/compat`) already provides `useSyncExternalStore`, why does the Zustand fork (for ReactLynx, in this case) still need to depend on `use-sync-external-store` separately at all? Regardless whether it's CJS or ESM?

Huxpro:
@colinaaa could you update the README at https://github.com/lynx-family/lynx-stack/pull/899 to include the proper usages for this library?

It would be helpful to add more context that it's intended as a drop-in replacement for libraries that import `useSyncExternalStore` hooks directly from the `use-sync-external-store` package as a bundler-level alias, in order to preserve the optimizations mentioned above, that would otherwise be lost due to the original package being CJS and default importing `react`.

PR #980, opened 2025-06-03
feat(rspeedy/react): alias `use-sync-external-store`

colinaaa:
## Summary

Create an alias for `use-sync-external-store` pointing to `@lynx-js/use-sync-external-store`.

close: #893 

## Checklist

### DIFF
--- a/packages/rspeedy/plugin-react/src/backgroundOnly.ts
+++ b/packages/rspeedy/plugin-react/src/backgroundOnly.ts
@@ -2,6 +2,7 @@
 // Licensed under the Apache License Version 2.0 that can be found in the
 // LICENSE file in the root directory of this source tree.
 import path from 'node:path'
+import { fileURLToPath } from 'node:url'
 
 import type { RsbuildPluginAPI } from '@rsbuild/core'
 
@@ -16,6 +17,8 @@ export function applyBackgroundOnly(
   api: RsbuildPluginAPI,
 ): void {
   api.modifyBundlerChain(async chain => {
+    const __dirname = path.dirname(fileURLToPath(import.meta.url))
+
     const { resolve, resolveMainThread } = await import('./resolve.js')
 
     const [backgroundOnly, backgroundOnlyMainThread] = await Promise.all([


packages/rspeedy/plugin-react/src/useSyncExternalStore.ts
```diff
@@ -0,0 +1,28 @@
...
+    const useSyncExternalStoreEntries = [
+      'use-sync-external-store',
+      'use-sync-external-store/with-selector',
+      'use-sync-external-store/shim',
+      'use-sync-external-store/shim/with-selector',
+    ]
+
+    await Promise.all(
+      useSyncExternalStoreEntries.map(entry =>
+        resolve(`@lynx-js/${entry}`).then(value => {
```
colinaaa, line 19: I didn't make this in `@lynx-js/react-alias-rsbuild-plugin` since:

1. The react-alias-rsbuild-plugin would resolve from `@lynx-js/react` instead of the current plugin. We need to create a new resolver for it.
2. We don't need to reuse this in our internal projects.

.changeset/small-onions-hear.md
```diff
@@ -0,0 +1,7 @@
+---
+"@lynx-js/react-rsbuild-plugin": patch
+---
+
+Better [zustand](https://github.com/pmndrs/zustand) support by creating an alias for `use-sync-external-store`.
```
upupming, line 5: Can we add an example or testcase using zustand here?

upupming (approved):
LGTM!

.changeset/small-onions-hear.md
colinaaa: Maybe we can add one in lynx-family/lynx-examples. Then we could replace the example on https://lynxjs.org/react/state-management/zustand.html with `<Go />`. What do you think?

upupming: Sounds promising, we can add it after this version is published.

colinaaa: Actually, only `zustand@4` and `zustand/traditional` is still using `use-sync-external-store`.

As for a zustand@4 project, this patch could reduce 115.9 kB to 114.3 kB, with all modules concatenated.

I've created an example at https://github.com/lynx-family/lynx-examples/pull/82.

Issue #983, opened 2025-06-03
[Bug]: main-thread.js exception: ReferenceError: `__webpack_require__` is not defined


colinaaa:
### System Info

  System:
    OS: macOS 14.6.1
    CPU: (12) arm64 Apple M2 Max
    Memory: 3.09 GB / 96.00 GB
    Shell: 4.0.2 - /opt/homebrew/bin/fish
  Binaries:
    Node: 22.15.1 - ~/.local/state/fnm_multishells/67574_1748587337304/bin/node
    npm: 10.9.2 - ~/.local/state/fnm_multishells/67574_1748587337304/bin/npm
    pnpm: 9.15.9 - ~/.local/state/fnm_multishells/67574_1748587337304/bin/pnpm
    bun: 1.2.8 - ~/.bun/bin/bun
  Browsers:
    Chrome: 137.0.7151.55
    Safari: 17.6
  npmPackages:
    @lynx-js/qrcode-rsbuild-plugin: ^0.3.6 => 0.3.6 
    @lynx-js/react: ^0.109.0 => 0.109.1 
    @lynx-js/react-rsbuild-plugin: ^0.10.1 => 0.10.2 
    @lynx-js/rspeedy: ^0.9.7 => 0.9.8 
    @lynx-js/types: ^3.3.0 => 3.3.0 

### Details

When save the changes to JSX, an unexpected error occurred:

```
main-thread.js exception: ReferenceError: __webpack_require__ is not defined
```

It works fine when revert `@lynx-js/rspeedy` to v0.9.7.
It would also work when manually set `tools.rspack.output.iife: false`.

This should be related with #949.

### Reproduce link

https://github.com/lynx-family/lynx-stack/tree/main/examples/react

### Reproduce Steps

1. `pnpm install`
2. `pnpm run dev`
3. Change any JSX and save

PR #985, opened 2025-06-03
fix(react/runtime): should have `__webpack_require__` in `evaluate`

colinaaa:
## Summary

fix: https://github.com/lynx-family/lynx-stack/issues/983

## Checklist

### DIFF
--- a/packages/react/runtime/src/lifecycle/patch/snapshotPatchApply.ts
+++ b/packages/react/runtime/src/lifecycle/patch/snapshotPatchApply.ts
@@ -129,10 +129,20 @@ export function snapshotPatchApply(snapshotPatch: SnapshotPatch): void {
   }
 }
 
+declare global {
+  const __webpack_require__: (<T>(id: string) => T) | undefined;
+}
+
 /**
  * Evaluates a string as code with ReactLynx runtime injected.
  * Used for HMR (Hot Module Replacement) to update snapshot definitions.
  */
 function evaluate<T>(code: string): T {
+  // This is a workaround for webpack when `output.iife` is `true`.
+  // We should fully support main-thread HMR and remove this `evaluate` function entirely.
+  // See https://github.com/lynx-family/lynx-stack/issues/983 for more details.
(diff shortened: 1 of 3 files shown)


packages/react/runtime/src/lifecycle/patch/snapshotPatchApply.ts
```diff
@@ -129,10 +129,20 @@ export function snapshotPatchApply(snapshotPatch: SnapshotPatch): void {
...
  * Evaluates a string as code with ReactLynx runtime injected.
  * Used for HMR (Hot Module Replacement) to update snapshot definitions.
  */
 function evaluate<T>(code: string): T {
+  // This is a workaround for webpack when `output.iife` is `true`.
+  // We should fully support main-thread HMR and remove this `evaluate` function entirely.
+  // See https://github.com/lynx-family/lynx-stack/issues/983 for more details.
+  if (typeof __webpack_require__ !== 'undefined') {
+    return new Function('__webpack_require__', `return ${code}`)(__webpack_require__);
+  }
```
upupming: Adding `__webpack_require__` here in runtime might be harder to maintain for other bundling tools.

Since we need to access the local variable `__webpack_require__ `, can we just use `eval` here? Since it is only for dev, it may not affect production (I am not sure if eval is not a good choice)

colinaaa: We cannot use `eval()` directly. Since it would cause the ModuleConcatenation to bailout, which results in much larger bundle size change (86.4 kB -> 156.9 kB for examples/react).

```
"ModuleConcatenation bailout: Module uses eval()"
```

But we can use `if (__DEV__)` instead. I'll change it to `eval()`.

colinaaa:
BugBot run

colinaaa:
BugBot run

colinaaa:
BugBot run

colinaaa:
BugBot run

upupming (approved):

PR #1019, opened 2025-06-06
fix(react/runtime): make mts-runtime source map inline

Yradex:
## Summary

Support for locating errors in the source code directly on the device when exceptions occur when using MTS.

This requires Lynx engine v3.4 or later.

## Checklist


packages/react/worklet-runtime/scripts/build.js
```diff
@@ -20,7 +20,7 @@ const commonOptions = {
   write: true,
   bundle: true,
   format: 'iife',
-  sourcemap: true,
+  sourcemap: 'inline',
```
colinaaa: I think adding `sourcemap: 'inline'` to `commonOptions` would make `worklet-runtime.js` quite large in production. Can we instead just make it inline on development?

Yradex: Is it needed? I thought it would be encoded so it won't make any difference anyway

colinaaa: I do think it would be safer to not include the `//# sourceMappingURL` comment in a production build. We may not encode (e.g.: on Web Platform) in some cases, or we may use it in other environment directly.

Yradex: ok

colinaaa (approved):
👍

PR #1033, opened 2025-06-06
feat(css-hmr): use the compiler options from main template for css hmr

upupming:
## Summary

In #435, we have passed the `enableCSSInvalidation` compiler option for CSS hmr updates encode.

However, many other compiler options may cause CSS not to work or crash problems. We have found that missing `enableParallelElement` may cause a page crash in some circumstances. To fully solve this issue, we used the same encoding options from the main template to encode CSS hmr updates.

Since `CssExtractRspackPlugin` already depends on `LynxTemplatePlugin`, we tap the `encode` hook to make a copy of the real compiler options for the main template.

## Checklist

### DIFF
--- a/packages/rspeedy/plugin-react/test/lazy.test.ts
+++ b/packages/rspeedy/plugin-react/test/lazy.test.ts
@@ -90,10 +90,7 @@ describe('Lazy', () => {
           },
           output: {
             distPath: {
-              root: new URL(
-                './dist/standalone-lazy-bundle',
-                import.meta.url,
-              ).pathname,
+              root: './dist/standalone-lazy-bundle',
             },
           },
           plugins: [


packages/webpack/css-extract-webpack-plugin/src/CssExtractRspackPlugin.ts
```diff
@@ -169,6 +147,109 @@ class CssExtractRspackPluginImpl {
...
         || process.env['NODE_ENV'] === 'development'
       ) {
+        const hooks = LynxTemplatePlugin.getLynxTemplatePluginHooks(
+          // @ts-expect-error Rspack to Webpack Compilation
+          compilation,
+        );
+        // Save the encode options used for template of this compilation.
+        let baseEncodeOptions: EncodeOptions;
+        // Save the assets
+        let assets: Assets;
```
colinaaa: We don't need to store `assets` here. The `compilation.assets` can be used.

packages/webpack/css-extract-webpack-plugin/src/CssExtractRspackPlugin.ts
```diff
@@ -169,6 +147,109 @@ class CssExtractRspackPluginImpl {
...
+                );
+              }
+            }
+          }
+          this.hash = compilation.hash;
+          cssHmrJSONGenerated = true;
+        };
+
+        // @ts-expect-error only tap to get the encode options, not existing early
+        hooks.encode.tapPromise(this.name, async (args) => {
```
colinaaa: I don't think we should get the encode options here. You may use the `hooks.beforeEmit`, which has the final encode options.

packages/webpack/css-extract-webpack-plugin/src/CssExtractRspackPlugin.ts
```diff
@@ -169,6 +147,109 @@ class CssExtractRspackPluginImpl {
...
+              const importRuleNodes = value.filter(
+                (node) => node.type === 'ImportRule',
+              );
+
+              acc[key] = importRuleNodes.map(({ href }) => href);
+              return acc;
+            }, {});
+
+            try {
+              const encoded = await LynxEncodePlugin.encodeCSS(
```
colinaaa: We also don't need to use `LynxEncodePlugin.encodeCSS`, too. Since we already have `hooks.encode` and also `LynxTemplatePlugin.convertCSSChunksToMap`.

colinaaa (commented):
Here is a simplified version of CSS HMR using `hooks.beforeEmit` that works for me.

```js

        hooks.beforeEmit.tapPromise(this.name, async (args) => {
          for (const [filename, source] of Object.entries(compilation.assets)) {
            if (!filename.endsWith('.css')) {
              continue;
            }
            const content: string = source.source().toString('utf-8');
            const css = LynxTemplatePlugin.convertCSSChunksToMap(
              [content],
              options.cssPlugins,
              Boolean(
                args.finalEncodeOptions.compilerOptions['enableCSSSelector'],
              ),
            );
            const cssDeps = Object.entries(css.cssMap).reduce<
              Record<string, string[]>
            >((acc, [key, value]) => {
              const importRuleNodes = value.filter(
                (node) => node.type === 'ImportRule',
              );

              acc[key] = importRuleNodes.map(({ href }) => href);
              return acc;
            }, {});

            const { buffer } = await hooks.encode.promise({
              encodeOptions: {
                ...args.finalEncodeOptions,
                css,
                lepusCode: {
                  root: undefined,
                  lepusChunk: {},
                },
                manifest: {},
                customSections: {},
              },
              templateType: 'css-hmr',
            });
            const result = {
              content: buffer.toString('base64'),
              deps: cssDeps,
            };
            compilation.emitAsset(
              filename.replace(
                '.css',
                `${this.hash ? `.${this.hash}` : ''}.css.hot-update.json`,
              ),
              new compiler.webpack.sources.RawSource(
                JSON.stringify(result),
                true,
              ),
            );
          }

          this.hash = compilation.hash;

          return args;
        });
```

packages/webpack/css-extract-webpack-plugin/src/CssExtractRspackPlugin.ts
upupming: @colinaaa Oh I missed this hook, I choosed `encode` among `beforeEncode` and `encode`, and did not notice there is a `beforeEmit` hook, let me change to it, thanks!

packages/webpack/css-extract-webpack-plugin/src/CssExtractRspackPlugin.ts
```diff
@@ -10,10 +10,8 @@ import type {
   CssExtractRspackPluginOptions as ExternalCssExtractRspackPluginOptions,
 } from '@rspack/core';
 
-import {
-  LynxEncodePlugin,
-  LynxTemplatePlugin,
-} from '@lynx-js/template-webpack-plugin';
+import { removeFunctionWhiteSpace } from '@lynx-js/css-serializer/dist/plugins/removeFunctionWhiteSpace.js';
```
colinaaa: Use

```js
import { CSS } from '@lynx-js/template-webpack-plugin'
CSS.Plugins.removeFunctionWhiteSpace()
```

instead.

`@lynx-js/css-serializer` is a devDependency of `@lynx-js/css-extract-webpack-plugin`.

packages/webpack/css-extract-webpack-plugin/src/CssExtractRspackPlugin.ts
```diff
@@ -169,6 +145,69 @@ class CssExtractRspackPluginImpl {
...
+            >((acc, [key, value]) => {
+              const importRuleNodes = value.filter(
+                (node) => node.type === 'ImportRule',
+              );
+
+              acc[key] = importRuleNodes.map(({ href }) => href);
+              return acc;
+            }, {});
+
+            const { buffer } = await hooks.encode.promise({
```
colinaaa: Do we need to try-catch here?

packages/webpack/template-webpack-plugin/src/LynxTemplatePlugin.ts
```diff
@@ -93,10 +93,17 @@ export interface TemplateHooks {
...
-    intermediate: string;
-  }, { buffer: Buffer; debugInfo: string }>;
+  encode: AsyncSeriesBailHook<
+    {
+      encodeOptions: EncodeOptions;
+      intermediate: string;
+      templateType: 'main' | 'async';
+    } | {
+      encodeOptions: EncodeOptions;
+      templateType: 'css-hmr';
```
colinaaa: Do we still need the `templateType` parameter?

packages/webpack/template-webpack-plugin/etc/template-webpack-plugin.api.md
```diff
@@ -69,9 +83,6 @@ export class LynxEncodePlugin {
     static BEFORE_ENCODE_STAGE: number;
     static defaultOptions: Readonly<Required<LynxEncodePluginOptions>>;
     static ENCODE_STAGE: number;
-    static encodeCSS(cssChunks: string[], options: EncodeCSSOptions, plugins?: CSS.Plugin[], encode?: (options: any) => Promise<{
```
colinaaa, line 72: It looks like removing this `encodeCSS` method is a breaking change. We should make a minor bump.

packages/webpack/css-extract-webpack-plugin/src/CssExtractRspackPlugin.ts
```diff
@@ -23,26 +21,6 @@ import {
 interface CssExtractRspackPluginOptions
   extends ExternalCssExtractRspackPluginOptions
 {
-  /**
-   * {@inheritdoc @lynx-js/template-webpack-plugin#LynxTemplatePluginOptions.enableRemoveCSSScope}
-   */
-  enableRemoveCSSScope: boolean;
```
colinaaa, line 29: I would also say removing a required option is a breaking change. We should also make minor bump.

packages/webpack/template-webpack-plugin/etc/template-webpack-plugin.api.md
```diff
@@ -53,12 +53,26 @@ declare namespace CSSPlugins {
     }
 }
 
-// @public
-export interface EncodeCSSOptions {
-    enableCSSInvalidation: boolean;
-    enableCSSSelector: boolean;
-    enableRemoveCSSScope: boolean;
-    targetSdkVersion: string;
+// Warning: (ae-missing-release-tag) "EncodeOptions" is part of the package's API, but it is missing a release tag (@alpha, @beta, @public, or @internal)
```
luhc228: Should add more docs?

luhc228 (dismissed):

.changeset/fruity-lizards-thank.md
```diff
@@ -0,0 +1,5 @@
+---
+"@lynx-js/template-webpack-plugin": minor
```
colinaaa, line 2: You need to update the `peerDependencies` of `@lynx-js/react-webpack-plugin`, otherwise a "major" bump will be made.

packages/webpack/css-extract-webpack-plugin/package.json
```diff
@@ -59,7 +59,7 @@
     "webpack": "^5.99.9"
   },
   "peerDependencies": {
-    "@lynx-js/template-webpack-plugin": "^0.5.0 || ^0.6.0 || ^0.7.0"
+    "@lynx-js/template-webpack-plugin": "^0.5.0 || ^0.6.0 || ^0.7.0 || ^0.8.0"
```
colinaaa: I don't think this version is compatible with @lynx-js/template-webpack-plugin@0.5.0. Maybe set to `^0.8.0`?

PR #1040, opened 2025-06-09
fix(testing-library): __SetInlineStyles of raw styles should not normalize css

upupming:
## Summary

In JSDOM, `element.style.cssText = 'flex:1'` will normalize the style to `flex: 1 1 0%;`, this is not the  actual behavior of `__SetInlineStyles` in Lynx. We changed to use `element.setAttributeNS(null, 'style', 'flex:1');` to avoid issues.

Related change in `cssstyle`: https://github.com/jsdom/cssstyle/commit/79c88c407ee5da7f4fd9849725c9e9cc04b1a2c8#diff-72d4401db3e024bcef1fc57d7d43457be6f5b3f8f89a80b1580063362e280782

## Checklist

### DIFF
--- a/packages/testing-library/testing-environment/src/lynx/ElementPAPI.ts
+++ b/packages/testing-library/testing-environment/src/lynx/ElementPAPI.ts
@@ -284,15 +284,8 @@ export const initElementTree = () => {
       styles: string | Record<string, string>,
     ) {
       if (typeof styles === 'string') {
-        // Copied from https://github.com/jsdom/jsdom/blob/6197af431c46b95622eb4ce9d3e4df3010c66984/lib/jsdom/living/nodes/ElementCSSInlineStyle-impl.js#L8
-        // @ts-expect-error
-        if (!this._settingCssText) {
-          // @ts-expect-error
-          e._settingCssText = true;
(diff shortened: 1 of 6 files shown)


packages/testing-library/testing-environment/src/lynx/ElementPAPI.ts
```diff
@@ -284,7 +284,8 @@ export const initElementTree = () => {
       styles: string | Record<string, string>,
     ) {
       if (typeof styles === 'string') {
-        e.style.cssText = styles;
+        // Same as from https://github.com/jsdom/jsdom/blob/6197af431c46b95622eb4ce9d3e4df3010c66984/lib/jsdom/living/nodes/ElementCSSInlineStyle-impl.js#L8
```
colinaaa, line 287: Do we need a changeset for this?

colinaaa (approved):

packages/testing-library/testing-environment/src/lynx/ElementPAPI.ts
upupming: This change does not affect the user, but a changeset is still good, let me add it to the changeset.

PR #1060, opened 2025-06-10
fix(webpack/react): remove `:` from lazy bundle name

colinaaa:
## Checklist

### DIFF
--- a/packages/webpack/react-webpack-plugin/src/ReactWebpackPlugin.ts
+++ b/packages/webpack/react-webpack-plugin/src/ReactWebpackPlugin.ts
@@ -309,14 +309,14 @@ class ReactWebpackPlugin {
         return args;
       });
 
-      // The react-transform will add `-${LAYER}` to the webpackChunkName.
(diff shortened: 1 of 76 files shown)


packages/webpack/react-webpack-plugin/src/ReactWebpackPlugin.ts
```diff
@@ -309,14 +309,14 @@ class ReactWebpackPlugin {
         return args;
       });
 
-      // The react-transform will add `-${LAYER}` to the webpackChunkName.
+      // The react-transform will add `-react__${LAYER}` to the webpackChunkName.
```
upupming, line 312: Should we change the `LAYERS.BACKGROUND` variable to `react__background` directly in `packages/webpack/react-webpack-plugin/src/layer.ts`?

colinaaa: I didn't change it because it has already been used as a string literal in some `module.rule`. We should export the `LAYERS` before we change it (or we don't change it at all).

Issue #81, opened 2025-03-06
Review Windows support


PupilTong:
I just see MacOS instructions on the Quick Start: https://lynxjs.org/guide/start/quick-start.html

_Originally posted by @corysimmons in https://github.com/lynx-family/lynx/discussions/103_

coder-xiaomo:
Is current version support windows? 

I was creating a project on Windows 11, but it turned out this error: (see also https://github.com/lynx-family/lynx/issues/122)

```
E:\Project\lynx-app>yarn dev
yarn run v1.22.22
$ rspeedy dev
  Rspeedy v0.8.2 (Rsbuild v1.2.11, Rspack v1.2.3)

start   Building...
error   Compile error: 
Failed to compile, check the errors for troubleshooting.
× Error: Cannot find module 'E:\Project\lynx-app\node_modules\@lynx-js\tasm\build\win32\Release\lepus.node'
  │ Require stack:
  │ - E:\Project\lynx-app\node_modules\@lynx-js\tasm\index.js
  │     at Module._resolveFilename (node:internal/modules/cjs/loader:1225:15)
  │     at Module._load (node:internal/modules/cjs/loader:1051:27)
  │     at Module.require (node:internal/modules/cjs/loader:1311:19)
  │     at require (node:internal/modules/helpers:179:18)
  │     at encode_napi (E:\Project\lynx-app\node_modules\@lynx-js\tasm\index.js:48:17)
  │     at Object.fn (file:///E:/Project/lynx-app/node_modules/@lynx-js/template-webpack-plugin/lib/LynxEncodePlugin.js:145:71)

× Error: Cannot find module 'E:\Project\lynx-app\node_modules\@lynx-js\tasm\build\win32\Release\lepus.node'
  │ Require stack:
  │ - E:\Project\lynx-app\node_modules\@lynx-js\tasm\index.js
  │     at Module._resolveFilename (node:internal/modules/cjs/loader:1225:15)
  │     at Module._load (node:internal/modules/cjs/loader:1051:27)
  │     at Module.require (node:internal/modules/cjs/loader:1311:19)
  │     at require (node:internal/modules/helpers:179:18)
  │     at encode_napi (E:\Project\lynx-app\node_modules\@lynx-js\tasm\index.js:48:17)
  │     at Object.fn (file:///E:/Project/lynx-app/node_modules/@lynx-js/template-webpack-plugin/lib/LynxEncodePlugin.js:145:71)
```

[...]

PR #1061, opened 2025-06-10
test: replace `file://` URL

colinaaa:
## Summary

Since the error stack of Node.js is file URL (`file://`) instead of path (`C:\foo\bar.js` on Windows, `/foo/bar.js` on Posix), we should use `fileURLToPath` to make sure the snapshots are the same in different platform.

> [!NOTE]
> We are not implementing this in `path-serializer` because the `fileURLToPath` method from `node:url` exhibits varying behavior across platforms. I believe it’s not appropriate to include it there.

issue: #81

## Checklist

### DIFF
--- a/packages/webpack/test-tools/src/diagnostic.ts
+++ b/packages/webpack/test-tools/src/diagnostic.ts
@@ -2,6 +2,7 @@
 // Licensed under the Apache License Version 2.0 that can be found in the
 // LICENSE file in the root directory of this source tree.
 import path from 'node:path';
+import { fileURLToPath } from 'node:url';
 
 import {
   DiagnosticProcessor,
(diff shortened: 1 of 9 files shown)


packages/webpack/test-tools/src/diagnostic.ts
```diff
@@ -51,10 +52,18 @@ class RspeedyDiagnosticProcessor<Compiler extends ECompilerType>
 }
 
 const serializer = createSnapshotSerializer({
+  beforeSerialize(val) {
+    return val.replaceAll(/\(file:\/\/.*\)/g, (match) => {
```
SoonIter, line 56: how about below?

```ts
 beforeSerialize(val) {
    return val.replaceAll(pathToFileURL(process.cwd()), '<ROOT>')
 }
```

I can integrate it into the path-serializer

colinaaa: Nice catch! It would be great

PR #1065, opened 2025-06-11
feat(web): add _I18nResourceTranslation api in mts.

Sherry-hue:
## Summary

feat: add `_I18nResourceTranslation` api in mts && `init-i18n-resources` attr, `i18nResourceMissed` event of lynx-view.

`init-i18n-resource` is the complete set of i18nResources that need to be maintained on the container side. Note: You need to pass this value when lynx-view is initialized.

You can use `_I18nResourceTranslation` in MTS to get the corresponding i18nResource from `init-i18n-resources`. If it is undefined, the `i18nResourceMissed` event will be dispatched.

```js
// ui thread
lynxView.initI18nResources = [
  {
    options: {
      locale: 'en',
      channel: '1',
      fallback_url: '',
    },
    resource: {
      hello: 'hello',
      lynx: 'lynx web platform1',
    },
  },
];
lynxView.addEventListener('i18nResourceMissed', (e) => {
  console.log(e);
});

// mts
_I18nResourceTranslation({
  locale: 'en',
  channel: '1',
  fallback_url: '',
});
```

## Checklist


packages/web-platform/web-mainthread-apis/src/prepareMainThreadAPIs.ts
```diff
@@ -174,6 +178,16 @@ export function prepareMainThreadAPIs(
...
         createElement,
+        i18nResourceTranslation: (options: I18nResourceTranslationOptions) => {
+          const matchedInitI18nResources = initI18nResources.find(i =>
+            getCacheI18nResourcesKey(i.options)
+              === getCacheI18nResourcesKey(options)
+          );
+          if (matchedInitI18nResources) {
+            return matchedInitI18nResources.resource;
+          }
+          return i18nResourceTranslation(options);
```
PupilTong: Could you give it a better name? like `triggerI18nResourceLoading` ?

packages/web-platform/web-worker-runtime/src/mainThread/startMainThread.ts
```diff
@@ -24,6 +29,22 @@ export function startMainThreadWorker(
...
+    const cacheKey = getCacheI18nResourcesKey(options);
+
+    if (CacheI18nResources.has(cacheKey)) {
+      return CacheI18nResources.get(cacheKey);
+    }
+    backgroundThreadRpc.invoke(i18nResourceTranslationEndpoint, [options]).then(
+      res => {
+        if (res !== undefined) {
+          CacheI18nResources.set(cacheKey, res);
+          dispatchEvent(new Event('i18nResourceReady'));
```
PupilTong: ?

packages/web-platform/web-core/src/apis/LynxView.ts
```diff
@@ -292,6 +315,9 @@ export class LynxView extends HTMLElement {
         case 'init-data':
           this.#initData = JSON.parse(newValue);
           break;
+        case 'init-i18n-resources':
```
PupilTong: Seems it doesn't work.
How about we provide a property only?

PR #1071, opened 2025-06-12
fix(react): memory leak because of cycle reference

hzy:
Different from #1054, this fix avoids disconnecting the `SnapshotInstance` that is about to be reused, and avoids incompleteness when it is reused in the `<list/>`.

## Checklist

### DIFF
--- a/packages/react/runtime/src/list.ts
+++ b/packages/react/runtime/src/list.ts
@@ -3,7 +3,7 @@
 // LICENSE file in the root directory of this source tree.
 import { hydrate } from './hydrate.js';
 import { applyRefQueue } from './snapshot/workletRef.js';
-import type { SnapshotInstance } from './snapshot.js';
+import { type SnapshotInstance } from './snapshot.js';
 
 export interface ListUpdateInfo {
   flush(): void;
@@ -311,6 +311,9 @@ export function componentAtIndexFactory(
       recycleSignMap.delete(sign);
       hydrate(oldCtx, childCtx);
       oldCtx.unRenderElements();
+      if (!oldCtx.__id) {
+        oldCtx.tearDown();
+      }
       const root = childCtx.__element_root!;
       applyRefQueue();
       if (!enableBatchRender) {
--- a/packages/react/runtime/src/snapshot.ts
+++ b/packages/react/runtime/src/snapshot.ts
@@ -393,6 +393,14 @@ export class SnapshotInstance {
     return a;
   }
 
+  tearDown(): void {
+    traverseSnapshotInstance(this, v => {
+      v.__parent = null;
+      v.__previousSibling = null;
+      v.__nextSibling = null;
+    });
+  }
+
   // onCreate?: () => void;
   // onAttach?: () => void;
   // onDetach?: () => void;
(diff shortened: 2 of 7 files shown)


packages/react/runtime/src/list.ts
```diff
@@ -311,6 +311,16 @@ export function componentAtIndexFactory(
       recycleSignMap.delete(sign);
       hydrate(oldCtx, childCtx);
       oldCtx.unRenderElements();
+      if (!oldCtx.__id) {
+        traverseSnapshotInstance(oldCtx, (v) => {
+          // @ts-ignore
+          v.__parent = undefined;
```
upupming: Why it is `undefined` here but `null` in the below?

packages/react/runtime/src/snapshot.ts
```diff
@@ -569,6 +564,16 @@ export class SnapshotInstance {
...
       snapshotDestroyList(child);
     }
+
+    this.__removeChild(child);
+    traverseSnapshotInstance(child, v => {
+      v.__parent = null;
+      v.__previousSibling = null;
+      v.__nextSibling = null;
+      delete v.__elements;
+      delete v.__element_root;
```
upupming, line 586: Can we also add a test to ensure `__elements` are deleted correctly on first screen?

packages/react/runtime/src/list.ts
hzy: Sorry, this was a rookie mistake.

packages/react/runtime/src/snapshot.ts
hzy: WIP

hzy: I also want to add some tests about `FiberElement` being GCed, but not sure how... Maybe use `FinalizationRegistry`?

hzy: done

packages/react/testing-library/src/__tests__/list.test.jsx
```diff
@@ -226,4 +227,322 @@ describe('list', () => {
...
+                  <list-item item-key={`${v}`} key={v} full-span>
+                    <view>
+                      {showMask
+                        ? <text>{v}</text>
+                        : null}
+                      {showMask
+                        ? <text>{v}</text>
+                        : null}
+                    </view>
+                    {/* This will generate `__DynamicPartSlot` part for testing the hydration behavior of slot is as expceted */}
```
hzy:
```suggestion
                    {/* This will generate `__DynamicPartSlot` part for testing the hydration behavior of slot is as expected */}
```

packages/react/testing-library/src/__tests__/list.test.jsx
```diff
@@ -226,4 +227,322 @@ describe('list', () => {
...
+        3,
+      ]
+    `);
+    // Reuse the element 3
+    elementTree.leaveListItem(list, uid3);
+    elementTree.enterListItemAtIndex(list, 1);
+
+    expect(__CreateElement).toHaveBeenCalledTimes(0);
+    expect(__SetAttribute).toHaveBeenCalledTimes(4);
+    // The orginal FiberElement of element 3 is resued for
```
hzy:
```suggestion
    // The original FiberElement of element 3 is reused for
```

upupming (approved):

PR #1074, opened 2025-06-13
feat(css): remove backslash escapes when `enableCSSSelector` is disable

luhc228:
## Summary

the backslash escapes in selector should be removed when the `enableCSSSelector` option is disabled because the old CSS selector doesn't support. 

```css
.h-\[370px\] {
  height: 370px;
}
```

## Checklist


upupming (approved):

packages/tools/css-serializer/src/parse.ts
```diff
@@ -280,7 +281,11 @@ export function parse(content: string, options: {
           type: 'StyleRule',
           style: transformBlock(node.block, errors),
           selectorText: {
-            value: preludeText,
+            value: options.enableCSSSelector
+              ? preludeText
+              // When enableCSSSelector is disabled, remove backslash escapes to ensure CSS selector compatibility.
+              // For example, TailwindCSS generates selectors like `.h-\\[370px\\]` which require unescaping.
+              : preludeText.replace(/\\/g, ''),
```
colinaaa: I don't think it is correct to use `.replace()` for all backslash.

It's valid to use selector with `\` in legacy CSS Selector:

```css
.Backg\\round {
}
```

image

We should only escape the character that is meaningful in CSS. E.g.: `\[`, `\#`

colinaaa (changes requested):
Apologies for the delay; I need to pull this PR locally to test it thoroughly.

packages/tools/css-serializer/src/parse.ts
luhc228: OK! Thanks for your comment! @colinaaa

packages/tools/css-serializer/src/parse.ts
```diff
@@ -280,7 +281,14 @@ export function parse(content: string, options: {
...
           selectorText: {
-            value: preludeText,
+            value: options.enableCSSSelector
+              ? preludeText
+              // When enableCSSSelector is disabled, remove backslash escapes to ensure CSS selector compatibility.
+              // For example, TailwindCSS generates selectors like `.h-\\[370px\\]` which require unescaping.
+              : preludeText
+                .replaceAll('\\[', '[')
+                .replaceAll('\\]', ']')
+                .replaceAll('\\#', '#'),
```
colinaaa, line 291: Should we also escape `\.`?

E.g.:

```html
<div class="w-[0.6]" />
```

https://play.tailwindcss.com/T1v76iJNt8

PR #1076, opened 2025-06-13
refactor: adopt swc-plugin

gaoachao:
## Checklist


packages/rspeedy/plugin-react/src/loaders.ts
```diff
@@ -81,34 +114,87 @@ export function applyLoaders(
...
+    ...options,
+    jsc: {
+      ...options.jsc,
+      target: 'es2019',
+      experimental: {
+        ...options.jsc?.experimental,
+        plugins: [
+          ...(options.jsc?.experimental?.plugins ?? []),
+          [
+            require.resolve('@lynx-js/swc-plugin-react-lynx'),
```
colinaaa: We should not resolve `@lynx-js/swc-plugin-react-lynx` since it's a devDependency. Shall we use `@lynx-js/react/transform/swc-plugin-react-lynx` instead?

packages/rspeedy/plugin-react/src/loaders.ts
```diff
@@ -81,34 +114,87 @@ export function applyLoaders(
...
+  const swcLoaderOptions = {
+    ...options,
+    jsc: {
+      ...options.jsc,
+      experimental: {
+        ...options.jsc?.experimental,
+        plugins: [
+          ...(options.jsc?.experimental?.plugins ?? []),
+          [
+            require.resolve('@lynx-js/swc-plugin-react-lynx'),
```
colinaaa: ditto.

packages/webpack/react-webpack-plugin/src/loaders/options.ts
```diff
@@ -313,6 +234,39 @@ export function getBackgroundTransformOptions(
   };
 }
 
+export function getCompatOptions(
+  compat: CompatVisitorConfig | boolean,
+  layer: string,
+): CompatVisitorConfig | false {
+  if (typeof compat !== 'object') return false;
```
colinaaa, line 234: What if `compat === true`?

packages/rspeedy/plugin-react/src/loaders.ts
colinaaa: How about

```js
import { swcPluginReactLynx } from '@lynx-js/react/transform'
```

```js
export const swcPluginReactLynx = path.resolve(__dirname, './swc-plugin-reactlynx/swc_plugin_reactlynx.wasm')
```

packages/webpack/react-webpack-plugin/src/loaders/options.ts
gaoachao: Actually, `compat` can't be true. I should make the following change: 
 `compat: CompatVisitorConfig | boolean` ->  `compat: CompatVisitorConfig | undefined`,

packages/rspeedy/plugin-react/src/loaders.ts
gaoachao: Good catch!

packages/react/runtime/__test__/basic.test.jsx
```diff
@@ -10,6 +10,10 @@ import { SnapshotInstance, snapshotInstanceManager } from '../src/snapshot';
 
 const HOLE = null;
 
+const __SNAPSHOT__ = (snapshot) => {
```
upupming: Actually the input param is a `SnapshotInstance`/`BackgroundSnapshotInstance` here, which is created by `jsxDev`, and then `options.document.createElement`:

https://github.com/lynx-family/lynx-stack/blob/ab32e1cc162b47ee1c0d20e430d7f2bb9d63d6ac/packages/react/runtime/src/document.ts#L38

packages/react/runtime/vitest.config.ts
```diff
@@ -12,14 +12,23 @@ function transformReactLynxPlugin(): Plugin {
...
       ) as typeof import('@lynx-js/react-transform');
       const relativePath = path.basename(sourcePath);
-      const result = transformReactLynxSync(sourceText, {
+
+      const isTS = /\.[mc]?ts$/.exec(relativePath);
+      const syntax = isTS ? 'typescript' : 'ecmascript';
+      const syntaxConfig = {
+        syntax,
+        decorators: true,
+        tsx: !isTS,
```
upupming, line 27: This seems weird. Why do we enable `tsx` on non-TS files?

packages/react/runtime/__test__/basic.test.jsx
upupming: I think we can just add `__SNAPSHOT__` globally to `packages/react/runtime/__test__/utils/globals.js`, instead of adding it in each test file.

packages/react/testing-library/src/vitest.config.js
```diff
@@ -43,18 +43,27 @@ export const createVitestConfig = async (options) => {
           return '';
         }
 
-        const { transformReactLynxSync } = require(
-          '@lynx-js/react/transform',
+        const { transformReactLynx } = require(
+          '@lynx-js/react-transform',
```
upupming: Will `@lynx-js/react-transform` be published to npm? After this change testing-library might need to depends on `@lynx-js/react-transform`.

packages/rspeedy/plugin-react/src/loaders.ts
```diff
@@ -81,34 +115,87 @@ export function applyLoaders(
...
+          [
+            require.resolve('@lynx-js/react/transform/swc-plugin-reactlynx'),
+            mainThreadTransformOptions,
+          ],
+        ],
+      },
+    },
+  }
+
+  // When Passing to SWC, the swc-plugin-compat Must Be Placed Before swc-plugin-react-lyn
```
upupming:
```suggestion
  // When Passing to SWC, the swc-plugin-compat Must Be Placed Before swc-plugin-react-lynx
```

packages/react/transform/package.json
```diff
@@ -10,6 +10,7 @@
   "main": "./main.js",
   "types": "./index.d.ts",
   "scripts": {
+    "build": "rslib build",
```
colinaaa, line 13: Why do we need this `rslib build`?

Do we also need to change `main` and `types` after using Rslib?

colinaaa: We should also update `turbo.json` for this build.

packages/react/runtime/vitest.config.ts
colinaaa: This is for setting `tsx: false` for `.ts` files.

The SWC syntax config is split into two parts:

1. `syntax: 'ecmascript'` with `jsx`
2. `syntax: 'typescript'` with `tsx`

> https://swc.rs/docs/configuration/compilation#jscparser

Therefore, we should set

- `syntax: 'ecmascript'` with `jsx: true` for all `.js`, `.cjs`, `.mjs`, `.jsx` files.
- `syntax: 'typescript'` with `tsx: false` for all `.ts`, `.mts`, `.cts` files
- `syntax: 'typescript'` with `tsx: true` for all `.tsx` files

See https://github.com/web-infra-dev/rspack/issues/10636#issuecomment-2976977572 for more details.

packages/react/testing-library/src/vitest.config.js
```diff
@@ -43,18 +43,27 @@ export const createVitestConfig = async (options) => {
...
         // relativePath should be stable between different runs with different cwd
         const relativePath = normalizeSlashes(path.relative(
           __dirname,
           sourcePath,
         ));
         const basename = path.basename(sourcePath);
-        const result = transformReactLynxSync(sourceText, {
+
+        const isTS = /\.[mc]?ts$/.exec(relativePath);
+        const syntax = isTS ? 'typescript' : 'ecmascript';
```
colinaaa: ditto here, we need to set `syntax: 'typescript'` for `.tsx` files.

packages/react/transform/index.d.ts
```diff
@@ -25,7 +26,7 @@ export interface TransformReactLynxOptions {
    * @public
    * This is swc syntax config in JSON format
    */
-  syntaxConfig?: string;
+  syntaxConfig?: Record<string, any>;
```
colinaaa: Should we use `SwcLoaderParserConfig` from `@rspack/core` instead?

packages/webpack/react-webpack-plugin/src/loaders/options.ts
```diff
@@ -80,83 +70,32 @@ export interface ReactLoaderOptions {
...
-  } = this.getOptions();
+    defineDCE,
+  } = options;
 
-  const syntax = (/\.[mc]?tsx?$/.exec(this.resourcePath))
-    ? 'typescript'
-    : 'ecmascript';
-  // is '.ts' (one of '.js', '.jsx', '.ts', '.tsx')
-  const isTS = /\.[mc]?ts$/.exec(this.resourcePath);
+  const filename = 'test.js';
```
colinaaa: Instead of hard coded with `test.js`, we may get filename from SWC plugin context:

```rust
    let filename = if let Some(filename) =
        metadata.get_context(&TransformPluginMetadataContextKind::Filename)
    {
        FileName::Real(PathBuf::from(filename))
    } else {
        FileName::Anon
    };
```

https://github.com/swc-project/plugins/blob/bb0444292d46e5aee459ff5289bfaa127e2756ce/packages/relay/src/lib.rs#L34-L40

gaoachao: https://github.com/lynx-family/lynx-stack/pull/1206

It is split into another pr.

packages/react/testing-library/src/vitest.config.js
gaoachao: see: https://github.com/lynx-family/lynx-stack/pull/1208/files#diff-18ad6e397e4e4c086871a699d1d289ad9dc46f06f8be59814544b22215bcf4baR22

packages/react/runtime/vitest.config.ts
gaoachao: see: https://github.com/lynx-family/lynx-stack/pull/1208/files#diff-18ad6e397e4e4c086871a699d1d289ad9dc46f06f8be59814544b22215bcf4baR22

packages/react/testing-library/src/vitest.config.js
gaoachao: I will replace '@lynx-js/react-transform' with '@lynx-js/react/transform'

gaoachao: https://github.com/lynx-family/lynx-stack/pull/1209/commits/ea8aad44e545fe310b26533b03a873f9d39b83d2

packages/rspeedy/plugin-react/src/loaders.ts
gaoachao: https://github.com/lynx-family/lynx-stack/pull/1209/commits/ea8aad44e545fe310b26533b03a873f9d39b83d2

packages/react/runtime/__test__/basic.test.jsx
gaoachao: https://github.com/lynx-family/lynx-stack/pull/1209/commits/7a941cf8d12fa4c1658cfecaf1730e2c2272d947

packages/react/transform/index.d.ts
gaoachao: https://github.com/lynx-family/lynx-stack/pull/1209/commits/7a941cf8d12fa4c1658cfecaf1730e2c2272d947

packages/react/transform/package.json
gaoachao: I'm planning to convert `main.js` to TypeScript to achieve better type safety.

gaoachao: see: https://github.com/lynx-family/lynx-stack/pull/1209/commits/70c19c5bdd1918575f1c9734f0506eb28004dde1

PR #1088, opened 2025-06-16
feat(web): supports lynx.getI18nResource() and onI18nResourceReady ev…

Sherry-hue:
…ent in bts.

## Summary

feat: supports `lynx.getI18nResource()` and `onI18nResourceReady` event in bts.

- `lynx.getI18nResource()` can be used to get i18nResource in bts, it has two data sources:
  - the result of `_I18nResourceTranslation()`
  - lynx-view `updateI18nResources(data: InitI18nResources, options: I18nResourceTranslationOptions)`, it will be matched to the correct i8nResource as a result of `lynx.getI18nResource()`
- `onI18nResourceReady` event can be used to listen `_I18nResourceTranslation` and lynx-view `updateI18nResources` execution.

## Checklist

### DIFF
--- a/.changeset/heavy-sides-enter.md
+++ b/.changeset/heavy-sides-enter.md
@@ -1,0 +1,13 @@
+---
+"@lynx-js/web-mainthread-apis": patch
+"@lynx-js/web-worker-runtime": patch
+"@lynx-js/web-constants": patch
+"@lynx-js/web-core": patch
+---
+
+feat: supports `lynx.getI18nResource()` and `onI18nResourceReady` event in bts.
+
+- `lynx.getI18nResource()` can be used to get i18nResource in bts, it has two data sources:
+  - the result of `_I18nResourceTranslation()`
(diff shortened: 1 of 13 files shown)


packages/web-platform/web-worker-runtime/src/backgroundThread/background-apis/startBackgroundThread.ts
```diff
@@ -43,6 +48,7 @@ export function startBackgroundThread(
         nativeApp,
         mainThreadRpc,
         uiThreadRpc,
+        i18nResource,
```
PupilTong: could we read it from nativeApp? 
like the i18nResource object is belonging to the nativeApp instance. and Lynx instance read it from nativeApp?

Sherry-hue: yes, has done it

PupilTong (approved):

PR #1092, opened 2025-06-17
fix(web): when a list-item is deleted from list, the deleted list-ite…

Sherry-hue:
…m is still showed incorrectly

## Summary

fix: when a list-item is deleted from list, the deleted list-item is still showed incorrectly.

This is because the `enqueueComponent` method does not delete the node from the Element Tree. It is only to maintain all nodes on the RL side, and lynx web needs to delete the dom additionally.

## Checklist

## Summary by CodeRabbit

* **Bug Fixes**
  * Deleted list items no longer remain visible — UI now properly removes items from the DOM and display.

* **Tests**
  * Expanded update-list tests (insert/remove/mixed/randomized) and added a new visual test to verify list removal toggles and states.

* **Style**
  * Added stylesheet for the new list removal test page to ensure consistent layout.

* **Notes**
  * No public API signatures were changed.

### DIFF
--- a/.changeset/cuddly-roses-start.md
+++ b/.changeset/cuddly-roses-start.md
@@ -1,0 +1,9 @@
+---
+"@lynx-js/web-mainthread-apis": patch
+"@lynx-js/web-worker-rpc": patch
+"@lynx-js/web-constants": patch
+---
+
+fix: when a list-item is deleted from list, the deleted list-item is still showed incorrectly.
+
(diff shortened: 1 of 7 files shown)


hzy:
Good job! But the test feels a bit naive... A stress test like rendering a randomly shuffled array can help make sure your implementation correct.

```js
function shuffleArray(array) {
    for (let i = array.length - 1; i >= 1; i--) {
        const j = Math.floor(Math.random() * (i + 1));
        [array[i], array[j]] = [array[j], array[i]];
    }
}
```

Issue #1035, opened 2025-06-07
[Feature]: Simplify `@lynx-js/css-extract-webpack-plugin`


colinaaa:
### What problem does this feature solve?

Currently, we have a `@lynx-js/css-extract-webpack-plugin` that is a wrapper of `CssExtractRspackPlugin` (or `MiniCssExtractPlugin` on Webpack). It actually does nothing but modifying the content of `css-loader` result:

https://github.com/lynx-family/lynx-stack/blob/a86175445c47fd2948a6ffe8dcb89f118ced04d9/packages/webpack/css-extract-webpack-plugin/src/loader.ts#L228-L230

We would like to make it as simple as a normal loader that execute before `css-loader`.

This would avoid internal API changes between `CssExtractRspackPlugin.loader` and `CssExtractRspackPlugin` breaking our plugin.

It would be ideal to eliminate the CSS HMR logic and instead utilize the existing Rspack implementation.

### What does the proposed API of configuration look like?

A new `@lynx-js/scoped-css-loader` package will be created.

The loader sequence requires adjustment.

Before:

- [pitch] CssExtractRspackPlugin.loader (from `@lynx-js/css-extract-webpack-plugin`)
- css-loader
- (postcss-loader)
- (less/sass-loader)

After:

- [pitch] CssExtractRspackPlugin.loader (from `@rspack/core`)
- (`@lynx-js/scoped-css-loader`)
- css-loader
- (postcss-loader)
- (less/sass-loader)

PR #1096, opened 2025-06-17
fix(css-hmr): avoid duplicate css hmr files emitting

upupming:
## Summary

Fixes a bug of #1033 when using multiple entry:

image

This is because, in every `beforeEmit` of each entry, all CSS will be processed again.

## Checklist


luhc228 (approved):

packages/webpack/css-extract-webpack-plugin/src/CssExtractRspackPlugin.ts
```diff
@@ -155,6 +155,10 @@ class CssExtractRspackPluginImpl {
             if (!filename.endsWith('.css')) {
               continue;
             }
+            // the entry of `beforeEmit` hook should match the entry of current css
+            if (!args.cssChunks.some(c => c.name === filename)) {
```
colinaaa: Instead of filtering out on `args.cssChunks`, can we directly iterate on `args.cssChunks`?

```js
for (const {name: filename, source} of args.cssChunks) {
```

packages/webpack/template-webpack-plugin/src/LynxTemplatePlugin.ts
```diff
@@ -117,6 +117,7 @@ export interface TemplateHooks {
     template: Buffer;
     outputName: string;
     mainThreadAssets: Asset[];
+    cssChunks: Asset[];
```
colinaaa, line 120: I don't think we actually need this since there are enough information in `finalEncodeOptions`. It's fine to add this now, but I would like to make a refactor and avoid having these extra fields before #1098 and #1035.

colinaaa: E.g.:

```js
          const { buffer } = await hooks.encode.promise({
            encodeOptions: {
              ...args.finalEncodeOptions,
              lepusCode: {
                root: undefined,
                lepusChunk: {},
              },
              manifest: {},
              customSections: {},
            },
          });
```

upupming: In `finalEncodeOptions`, we cannot retrieve the CSS chunk filename; it only contains the `cssMap` and `cssSource` fields.

image

colinaaa: Yeah, I've found that. So I said "It's fine to add this now, but I would like to make a refactor"

PR #1107, opened 2025-06-18
fix(react/runtime): add snapshot id when throwing `ctx not found`

Yradex:
## Summary

Add snapshot id report when throwing `snapshotPatchApply failed: ctx not found` error.

We send the error to background thread to get the type of the snapshot.

before: 
```
[main-thread.js] Error: snapshotPatchApply failed: ctx not found'
```

after:
```
Error: snapshotPatchApply failed: ctx not found, snapshot type: '__Card__:__snapshot_a94a8_test_3'
```

## Checklist


packages/react/runtime/src/lifecycle/patch/error.ts
```diff
@@ -0,0 +1,46 @@
...
+
+export interface CtxNotFoundData {
+  id: number;
+}
+
+export function sendCtxNotFoundEventToBackground(id: number): void {
+  const error = new Error(errorMsg);
+  /* v8 ignore next 3 */
+  if (!lynx.getJSContext) {
+    throw error;
```
colinaaa: Nitpick: Should we create the Error here since the `error` isn't used elsewhere?

```suggestion
    throw new Error(errorMsg);
```

packages/react/runtime/src/lifecycle/patch/error.ts
```diff
@@ -0,0 +1,46 @@
...
+    .filter(({ 1: v }) => v === instance?.__snapshot_def)
+    .map(([k]) => k);
+
+  const snapshotType = snapshotDef[0] ?? 'null';
+  const error = new Error(`${errorMsg}, snapshot type: '${snapshotType}'`);
+  lynx.reportError(error);
+}
+
+export function addCtxNotFoundEventListener(): void {
+  lynx.getCoreContext?.().addEventListener(ctxNotFoundType, e => {
```
colinaaa: Should we also remove the listener somewhere?

packages/react/runtime/src/lifecycle/patch/error.ts
```diff
@@ -0,0 +1,46 @@
...
+    } as CtxNotFoundData,
+  });
+}
+
+export function reportCtxNotFound(data: CtxNotFoundData): void {
+  const id = data.id;
+  const instance = backgroundSnapshotInstanceManager.values.get(id);
+
+  const snapshotDef = [...snapshotManager.values.entries()]
+    .filter(({ 1: v }) => v === instance?.__snapshot_def)
```
colinaaa:
```suggestion
    .filter(([_, snapshot]) => snapshot === instance?.__snapshot_def)
```

colinaaa: Can we avoid iterating over `snapshotManager.values.entries()` when `instance === undefined`?

packages/react/runtime/src/lifecycle/patch/error.ts
```diff
@@ -0,0 +1,46 @@
...
+      id,
+    } as CtxNotFoundData,
+  });
+}
+
+export function reportCtxNotFound(data: CtxNotFoundData): void {
+  const id = data.id;
+  const instance = backgroundSnapshotInstanceManager.values.get(id);
+
+  const snapshotDef = [...snapshotManager.values.entries()]
```
colinaaa: Can we use for-of over `snapshotManager.values.entries()` to:

1. Avoid creating an array
2. Bailout when snapshot is found

colinaaa (approved):

PR #1111, opened 2025-06-19
fix(react/runtime): Accelerate the timing of didMount and didUpdate

colinaaa:
## Summary

Similar with https://github.com/lynx-family/lynx-stack/pull/770.

The `componentDidMount`, `componentDidUpdate`, `componentWillUnmount` and the callback of `setState` is not longer delayed until main-thread finish flushing.

close: m-6698445681

## Checklist


packages/react/runtime/__test__/lifecycle.test.jsx
```diff
@@ -200,7 +199,7 @@ describe('componentWillUnmount', () => {
       }
 
       componentDidMount() {
```
upupming, line 201: We should also add a test case of `componentDidUpdate` here to ensure it is "early" enough to get the correct state.

Before:

1st setState -> render ->  wait a long time (other 2nd setState may happen) -> componentDidUpdate get 2nd state

After: 

setState -> render -> componentDidUpdate  get the 1st state

colinaaa: I've added tests for `componentDidUpdate`:

- 0543e71 adds the tests that can pass on `main` branch
- 7d9ddf2 updates the tests (which shows `componentDidUpdate` can get correct state when there are multiple updates before main-thread callback)

PR #1117, opened 2025-06-20
fix(react/runtime): enable eslint for `src/lynx`

Yradex:


packages/react/runtime/src/lifecycle/event/delayLifecycleEvents.ts
```diff
@@ -1,9 +1,11 @@
 // Copyright 2025 The Lynx Authors. All rights reserved.
 // Licensed under the Apache License Version 2.0 that can be found in the
 // LICENSE file in the root directory of this source tree.
-const delayedLifecycleEvents: [type: string, data: any][] = [];
+import { LifecycleConstant } from '../../lifecycleConstant.js';
```
colinaaa:
```suggestion
import type { LifecycleConstant } from '../../lifecycleConstant.js';
```

PR #1122, opened 2025-06-20
fix(testing-library): fix flaky Windows tests

colinaaa:
## Summary

Fix the unhandled rejection on Windows ("lynx.getJSContext is not a function").

The `lynx.getJSContext` does exist on BTS, so does `lynx.getCoreContext` on MTS.

## Checklist


upupming (approved):
Thanks!

PR #1144, opened 2025-06-24
fix(react): memory leak because of cycle reference

hzy:
This PR tres to eliminate a memory leak by a cycle like: list(FiberElement) -> componentAtIndex -> ctx array -> full SnapshotInstance tree (by property like `__parent`, `__nextSibling` and `firstChild`) -> list(FiberElement, which closes the circle).

To minimize the impact on performance and behavior, only list-related logic was modified.

## Checklist


packages/react/runtime/src/list.ts
```diff
@@ -20,6 +21,22 @@ export function componentAtIndexFactory(
...
 ): [ComponentAtIndexCallback, ComponentAtIndexesCallback] {
+  // A hack workaround to ensure childCtx has no direct reference through `__parent` to list,
+  // to avoid memory leak.
+  ctx.forEach((childCtx) => {
+    if (Object.getOwnPropertyDescriptor(childCtx, '__parent')?.get) {
+      // do it only once
+    } else {
+      gParentWeakMap.set(childCtx, childCtx.parentNode!);
+      Object.defineProperty(childCtx, '__parent', {
+        get: () => gParentWeakMap.get(childCtx),
```
Yradex: It's returning `SnapshotInstance | undefined`, but the original `__parent` is returning `SnapshotInstance | null`

packages/react/runtime/src/list.ts
```diff
@@ -20,6 +21,22 @@ export function componentAtIndexFactory(
...
+        set: (value) => {
+          gParentWeakMap.set(childCtx, value);
```
Yradex, line 35: Unsafe argument of type `any` assigned to a parameter of type `SnapshotInstance`. eslint([@typescript-eslint/no-unsafe-argument](https://typescript-eslint.io/rules/no-unsafe-argument))

packages/react/runtime/src/list.ts
```diff
@@ -20,6 +21,22 @@ export function componentAtIndexFactory(
...
+  // A hack workaround to ensure childCtx has no direct reference through `__parent` to list,
+  // to avoid memory leak.
+  ctx.forEach((childCtx) => {
+    if (Object.getOwnPropertyDescriptor(childCtx, '__parent')?.get) {
+      // do it only once
+    } else {
+      gParentWeakMap.set(childCtx, childCtx.parentNode!);
+      Object.defineProperty(childCtx, '__parent', {
+        get: () => gParentWeakMap.get(childCtx),
+        set: (value) => {
```
Yradex: Does it accept `null` as a valid argument? It doesn't remove the previously stored entry when `null` is set; instead, it might be setting `null` into `gParentWeakMap`, which conflicts with its type `WeakMap<SnapshotInstance, SnapshotInstance>`.

Yradex:
I would prefer merging #1143 first to make sure ESLint is happy.

Yradex (approved):

Issue #1145, opened 2025-06-24
[Bug]: Duplicate root nodes are created after calling the lynxView.reload() method


Xieyongshen:
### System Info

System:
    OS: macOS 14.5
    CPU: (12) arm64 Apple M3 Pro
    Memory: 991.58 MB / 36.00 GB
    Shell: 5.9 - /bin/zsh
  Binaries:
    Node: 20.19.1 - ~/.nvm/versions/node/v20.19.1/bin/node
    npm: 10.8.2 - ~/.nvm/versions/node/v20.19.1/bin/npm
    pnpm: 10.12.1 - ~/.nvm/versions/node/v20.19.1/bin/pnpm
  npmPackages:
    @lynx-js/lynx-core: 0.1.2 => 0.1.2 
    @lynx-js/web-core: ^0.14.0 => 0.14.0 
    @lynx-js/web-elements: 0.7.4 => 0.7.4

### Details

Duplicate root nodes are created after calling the lynxView.reload() method with @lynx-js/web-core@0.14.0

### Reproduce Steps

1. run `pnpm install @lynx-js/web-core@0.14.0`
2. render the lynx page
3. call the method lynxView.reload()

PR #1147, opened 2025-06-25
fix(web): under the all-on-ui strategy, reload() will add two page el…

Sherry-hue:
…ements.

## Summary

#1145 

## Checklist

### DIFF
--- a/packages/web-platform/web-constants/src/constants.ts
+++ b/packages/web-platform/web-constants/src/constants.ts
@@ -16,6 +16,8 @@ export const lynxDatasetAttribute = 'l-dset' as const;
 
 export const lynxComponentConfigAttribute = 'l-comp-cfg' as const;
 
+export const lynxDisposedAttribute = 'l-disposed' as const;
+
 export const lynxDefaultDisplayLinearAttribute =
   'lynx-default-display-linear' as const;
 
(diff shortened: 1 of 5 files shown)


packages/web-platform/web-constants/src/constants.ts
```diff
@@ -16,6 +16,8 @@ export const lynxDatasetAttribute = 'l-dset' as const;
 
 export const lynxComponentConfigAttribute = 'l-comp-cfg' as const;
 
+export const lynxDisposeAttribute = 'l-dispose' as const;
```
PupilTong:
```suggestion
export const lynxDisposeAttribute = 'l-disposed' as const;
```

packages/web-platform/web-core/src/apis/LynxView.ts
```diff
@@ -369,6 +371,13 @@ export class LynxView extends HTMLElement {
   disconnectedCallback() {
     this.#instance?.dispose();
     this.#instance = undefined;
+    // under the all-on-ui strategy, when reload() triggers dsl flush, the previously removed pageElement will be used in __FlushElementTree.
```
PupilTong, line 374: use querySelector

PR #1154, opened 2025-06-25
refactor(rspeedy/core)!: deprecate `output.distPath.intermediate`

### LINKED ISSUE #1099: [BREAKING CHANGE]: Deprecated `output.distPath.intermediate`

The `output.distPath.intermediate` configuration currently exists but is not used. This could cause some confusion, as it might imply "enabling the emission of intermediate assets."

```diff
 export default defineConfig({
-  output: {
-    distPath: {
-      intermediate: '.rspeedy',
-    },
-  },
 })
```

gaoachao:
## Summary

close #1099

## Checklist

### DIFF
--- a/.changeset/fruity-showers-matter.md
+++ b/.changeset/fruity-showers-matter.md
@@ -4,4 +4,4 @@
 
 Deprecated `output.distPath.intermediate`
 
-This option is never readed and will be removed in the future version.
+This option is never read and will be removed in the future version.
--- a/packages/rspeedy/core/src/config/output/dist-path.ts
+++ b/packages/rspeedy/core/src/config/output/dist-path.ts
@@ -15,7 +15,7 @@ export interface DistPath extends DistPathConfig {
    *
    * @deprecated
    *
-   * This option is never readed and will be removed in the future version.
+   * This option is never read and will be removed in the future version.
    *
    * @remarks
    *


.changeset/eager-phones-shave.md
```diff
@@ -0,0 +1,5 @@
+---
+"@lynx-js/react-rsbuild-plugin": patch
+---
+
+Clean comments concerning `output.distPath.intermediate`.
```
colinaaa: I don't think this is needed to be shown in the CHANGELOG.

packages/rspeedy/core/src/plugins/inspect.plugin.ts
```diff
@@ -23,8 +22,6 @@ export function pluginInspect(config: Config): RsbuildPlugin {
           config,
           /** path */ path.join(
             api.context.distPath,
-            config.output?.distPath?.intermediate
```
colinaaa, line 26: We may want to keep this.

packages/rspeedy/core/src/plugins/output.plugin.ts
```diff
@@ -19,7 +18,7 @@ const defaultDistPathOptions = Object.freeze(
   {
     // We override the default value of Rsbuild(`static/css`) here.
     // Since all the CSS should be encoded into the template in Lynx.
-    css: DEFAULT_DIST_PATH_INTERMEDIATE,
```
colinaaa: ditto. We should also keep this.

packages/rspeedy/plugin-react/src/entry.ts
```diff
@@ -84,30 +82,15 @@ export function applyEntry(
...
-      const mainThreadName = path.posix.join(
-        isLynx
-          // TODO: config intermediate
-          ? DEFAULT_DIST_PATH_INTERMEDIATE
-          // For non-Lynx environment, the entry is not deleted.
-          // So we do not put it in the intermediate.
-          : '',
-        `${entryName}/main-thread.js`,
-      )
+      const mainThreadName = `${entryName}/main-thread.js`
```
colinaaa: I would like to keep the `.rspeedy` for JS chunks (at lease for the default value).

We may remove that when `chunkSplit.strategy !== 'all-in-one'`

packages/webpack/template-webpack-plugin/src/LynxTemplatePlugin.ts
```diff
@@ -386,7 +386,7 @@ export class LynxTemplatePlugin {
     .freeze<Required<LynxTemplatePluginOptions>>({
       filename: '[name].bundle',
       lazyBundleFilename: 'async/[name].[fullhash].bundle',
-      intermediate: '.rspeedy',
+      intermediate: '',
```
colinaaa: ditto here.

colinaaa (approved):

PR #1164, opened 2025-06-26
feat(react): support alog of component rendering

upupming:
## Summary

`console.alog` is added for component rendering hook:

image

## Checklist

## Summary by CodeRabbit

* **New Features**
  * Introduced an "alog" feature for logging component rendering in production, which can be enabled via environment variable or configuration.
  * Added support for a new console method, `console.alog`, with differentiation between main and background threads.
  * New global flag `__ALOG__` allows toggling of logging functionality.

* **Bug Fixes**
  * Improved test environment setup to ensure consistent logging and console behavior across threads.

* **Tests**
  * Added comprehensive tests to verify the new logging feature and its behavior during component rendering.

* **Chores**
  * Updated configuration to recognize "alog" as a valid word for typo checking.
  * Enhanced plugin and environment configuration for easier feature toggling.


PupilTong:
Whats the difference between log and alog?

packages/react/runtime/src/alog/render.ts
```diff
@@ -0,0 +1,26 @@
...
+
+import { DIFFED, DOM } from '../renderToOpcodes/constants.js';
+import type { SnapshotInstance } from '../snapshot.js';
+import { getDisplayName } from '../utils.js';
+
+export function initRenderAlog(): void {
+  const oldAfterDiff = options[DIFFED];
+  options[DIFFED] = function(vnode: VNode & { [DOM]: SnapshotInstance }) {
+    // Only log on component vnode
+    if (typeof vnode.type === 'function') {
```
upupming, line 15: @hzy 

Using `_render` is not not correct because on each diff, a new vnode will be created by [`constructNewChildrenArray`](https://github.com/preactjs/preact/blob/df481405b9454bbbc52a235e3a90bb4363852fb7/src/diff/children.js#L152) and the `__e` will always be undefined before the render.

I choose to use the `diffed` hook and a condition check to ensure it is a component (same check [as in the `diff` function](https://github.com/preactjs/preact/blob/df481405b9454bbbc52a235e3a90bb4363852fb7/src/diff/index.js#L81))

packages/webpack/react-webpack-plugin/src/ReactWebpackPlugin.ts
```diff
@@ -176,6 +176,8 @@ class ReactWebpackPlugin {
           ?? process.env['REACT_PROFILE']
           ?? compiler.options.mode === 'development',
       ),
+      // User can enable ALog by environment variable `REACT_ALOG=true`
+      __ALOG__: JSON.stringify(Boolean(process.env['REACT_ALOG'])),
```
hzy, line 180: Will this `DefinePlugin` override the one configured by `source.define`?

upupming: In fact, the opposite is true. The value defined in `source.define` will override the value here. This is the current behavior of `__PROFILE__`:

With:

```
source: {
    define: {
      __ALOG__: 'true',
      __PROFILE__: 'false'
    }
  }
```

using `DEBUG='rspeedy,rsbuild' pnpm dev` you can see:

```bash
warn    Build warnings: 
⚠ DefinePlugin:
  │ Conflicting values for '__ALOG__' ("true" !== "false")

⚠ DefinePlugin:
  │ Conflicting values for '__PROFILE__' ("false" !== "true")
```

And the final `.rsbuild/rspack.config.lynx.mjs`:

```js
new DefinePlugin(
      {
        __ALOG__: 'true',
        __PROFILE__: 'false'
      }
    ),
```

colinaaa (approved):

PR #1167, opened 2025-06-26
fix(web): class name contain special char not work

Sherry-hue:
## Summary

fix: when enableCSSSelector is false, the CSS style with className h-[40px] will not work.

This is because when the CSS selector contains special characters (such as `\`, `[`, `]`, `/`, `(`, `)`), the CSS style needs to be escaped with the character `\`, but the className does not need this.

The escape characters are filtered during building to ensure that that two parts are completely consistent.

## Checklist

### DIFF
--- a/.changeset/poor-carrots-build.md
+++ b/.changeset/poor-carrots-build.md
@@ -1,0 +1,9 @@
+---
+"@lynx-js/template-webpack-plugin": patch
+---
+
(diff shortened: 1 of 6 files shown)


colinaaa (changes requested):
Similar with https://github.com/lynx-family/lynx-stack/pull/1074#pullrequestreview-2935610500.

PR #1182, opened 2025-06-30
feat(testing-library): replace waitSchedule with act

upupming:
## Checklist


packages/react/testing-library/src/vitest.config.js
```diff
@@ -120,6 +154,7 @@ export const createVitestConfig = async (options) => {
       ),
       globals: true,
       setupFiles: [path.join(__dirname, 'vitest-global-setup')],
+      alias: [...runtimeOSSAlias, ...runtimeAlias, ...preactAlias],
```
colinaaa, line 157: Can we use [`resolve.dedupe`](https://vite.dev/config/shared-options.html#resolve-dedupe) instead?

colinaaa (approved):

packages/react/testing-library/src/vitest.config.js
colinaaa: I double-checked and confirmed that `resolve.dedupe` does not create an alias for `exports`.

colinaaa (approved):

PR #1201, opened 2025-07-02
fix(react): avoid manipulate `__pendingListUpdates` unless SnapshotInstance tree is changed

hzy:
We make `__pendingListUpdates.values` nullable, and set it to null to indicate manipulating it should be avoided, in the following scenarios:

1. setAttribute when hydrate
2. setAttribute when ensureElements

## Checklist

## Summary by CodeRabbit

* **Performance Improvements**
  * Enhanced efficiency of component rendering by reducing unnecessary updates when the component tree is unchanged.
  * Optimized internal update handling to improve execution speed.

* **Bug Fixes**
  * Prevented potential errors by adding safeguards when accessing internal update structures, ensuring more robust behavior.

* **Tests**
  * Updated test cases to reflect changes in internal update handling and to verify improved stability.

### DIFF
--- a/packages/react/runtime/src/hydrate.ts
+++ b/packages/react/runtime/src/hydrate.ts
@@ -221,7 +221,7 @@ export function hydrate(before: SnapshotInstance, after: SnapshotInstance, optio
     swap[before.__id] = after.__id;
   }
 
-  __pendingListUpdates.runWithNoUpdates(() => {
+  __pendingListUpdates.runWithoutUpdates(() => {
     after.__values?.forEach((value, index) => {
       const old = before.__values![index];
       if (value !== old) {
--- a/packages/react/runtime/src/hydrate.ts
+++ b/packages/react/runtime/src/hydrate.ts
@@ -221,12 +221,14 @@ export function hydrate(before: SnapshotInstance, after: SnapshotInstance, optio
     swap[before.__id] = after.__id;
   }
 
-  after.__values?.forEach((value, index) => {
-    const old = before.__values![index];
-    if (value !== old) {
-      after.__values![index] = old;
-      after.setAttribute(index, value);
-    }
+  __pendingListUpdates.runWithNoUpdates(() => {
+    after.__values?.forEach((value, index) => {
+      const old = before.__values![index];
+      if (value !== old) {
+        after.__values![index] = old;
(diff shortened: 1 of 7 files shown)


packages/react/runtime/src/pendingListUpdates.ts
```diff
@@ -5,14 +5,25 @@
...
-    });
-    this.clear();
+    if (this.values) {
+      Object.values(this.values).forEach(update => {
+        update.flush();
+      });
+      this.clear();
+    }
+  },
+  runWithNoUpdates(cb: () => void): void {
```
gaoachao:
```suggestion
  runWithoutUpdates(cb: () => void): void {
```

packages/react/runtime/src/hydrate.ts
```diff
@@ -221,12 +221,14 @@ export function hydrate(before: SnapshotInstance, after: SnapshotInstance, optio
     swap[before.__id] = after.__id;
   }
 
-  after.__values?.forEach((value, index) => {
-    const old = before.__values![index];
-    if (value !== old) {
-      after.__values![index] = old;
-      after.setAttribute(index, value);
-    }
+  __pendingListUpdates.runWithNoUpdates(() => {
```
gaoachao:
```suggestion
  __pendingListUpdates.runWithoutUpdates(() => {
```

packages/react/runtime/src/snapshot.ts
```diff
@@ -308,11 +308,13 @@ export class SnapshotInstance {
       }
     }
 
-    const values = this.__values;
-    if (values) {
-      this.__values = undefined;
-      this.setAttribute('values', values);
-    }
+    __pendingListUpdates.runWithNoUpdates(() => {
```
gaoachao:
```suggestion
      __pendingListUpdates.runWithoutUpdates(() => {
  ```

packages/react/runtime/src/hydrate.ts
Yradex: Given that `hydrate()` is utilized during `lynx.reload()` and `updatePage()`, is it appropriate to use `runWithNoUpdates()` within `hydrate()`?

Yradex (approved):

PR #1208, opened 2025-07-03
refactor: remove preserveJsx

gaoachao:
## Checklist

### DIFF
--- a/packages/react/transform/__test__/fixture.spec.js
+++ b/packages/react/transform/__test__/fixture.spec.js
@@ -300,7 +300,7 @@ Component, View
       expect(result.errors).toMatchInlineSnapshot(`[]`);
       expect(result.code).toMatchInlineSnapshot(`
         "/*#__PURE__*/ import { jsx as _jsx } from "@lynx-js/react/jsx-runtime";
-        import * as ReactLynx from "@lynx-js/react";
+        import * as ReactLynx from "@lynx-js/react-runtime";
         import * as ReactLynx1 from "@lynx-js/react/internal";
         const __snapshot_05fe4_4da58493_1 = /*#__PURE__*/ ReactLynx.createSnapshot("__snapshot_05fe4_4da58493_1", function() {
             const pageId = ReactLynx.__pageId;
--- a/packages/react/runtime/vitest.config.ts
+++ b/packages/react/runtime/vitest.config.ts
@@ -19,7 +19,8 @@ function transformReactLynxPlugin(): Plugin {
       const relativePath = path.basename(sourcePath);
 
       const isTS = /\.[mc]?ts$/.exec(relativePath);
-      const syntax = isTS ? 'typescript' : 'ecmascript';
+      const isTSX = /\.tsx$/.exec(relativePath);
+      const syntax = (isTS || isTSX) ? 'typescript' : 'ecmascript';
       const syntaxConfig = {
         syntax,
         decorators: true,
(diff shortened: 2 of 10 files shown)


packages/react/transform/__test__/fixture.spec.js
```diff
@@ -300,7 +300,7 @@ Component, View
       expect(result.errors).toMatchInlineSnapshot(`[]`);
       expect(result.code).toMatchInlineSnapshot(`
         "/*#__PURE__*/ import { jsx as _jsx } from "@lynx-js/react/jsx-runtime";
-        import * as ReactLynx from "@lynx-js/react";
+        import * as ReactLynx from "@lynx-js/react-runtime";
```
upupming, line 303: Is this snapshot change as expected? I see `compat` has been enabled.

gaoachao: I think it is expected because of https://github.com/lynx-family/lynx-stack/blob/refactor/remove-preserve-jsx/packages/react/transform/__test__/fixture.spec.js#L245.

gaoachao: The reason why the original inlineSnapshot or, in this case, `import * as ReactLynx from "@lynx-js/react"; ` behaves this way might be due to incorrect arguments passed to `transformReactLynx`, causing it to fall back to the default values.

gaoachao: In the main branch, we can modify https://github.com/lynx-family/lynx-stack/blob/main/packages/react/transform/__test__/fixture.spec.js#L323 as follows:
```javascript
      snapshot: {
        runtimePkg: '@lynx-js/react-runtime',
        filename: '',
        target: 'MIXED',
      },
```
After making this change, you will encounter the following error:
```
Error: Missing field `preserveJsx`
 ❯ napi_create_error dist/wasm.cjs:5126:19
 ❯ null.<anonymous> ../../../wasm:/wasm/00f0768e:1:506521
 ❯ null.<anonymous> ../../../wasm:/wasm/00f0768e:1:946946
 ❯ null.<anonymous> ../../../wasm:/wasm/00f0768e:1:944961
 ❯ dist/wasm.cjs:3733:48
 ❯ NodeEnv.callIntoModule dist/wasm.cjs:7715:15
 ❯ Module.<anonymous> dist/wasm.cjs:3732:30
 ❯ Module.transformReactLynx dist/wasm.cjs:3752:15
 ❯ __test__/fixture.spec.js:347:28
```
To resolve this, we can add `preserveJsx: false` to the snapshot configuration. After doing so, the following code will be generated as expected:
```javascript
import * as ReactLynx from "@lynx-js/react-runtime";
```

gaoachao: BTW: Prior to this commit, `preserveJsx` was always a required parameter:)

packages/react/transform/__test__/fixture.spec.js
```diff
@@ -300,7 +300,7 @@ Component, View
...
         import * as ReactLynx1 from "@lynx-js/react/internal";
         const __snapshot_05fe4_4da58493_1 = /*#__PURE__*/ ReactLynx.createSnapshot("__snapshot_05fe4_4da58493_1", function() {
```
upupming, line 305: It seems `createSnapshot` is not an export of `@lynx-js/react-runtime`:

https://github.com/lynx-family/lynx-stack/blob/main/packages/react/runtime/src/legacy-react-runtime/index.ts

PR #1222, opened 2025-07-04
feat(react/runtime): integrate with `preact/debug`

colinaaa:
## Summary

This patch add a `nodeType` property for `root` which make `@lynx-js/react` work with `preact/debug`.

Corresponding tests for supported `preact/debug` checks have been added.

> [!NOTE]
> The tests are divided into separate files because `preact/debug` stores the component stack in memory, which could lead to incorrect behavior if the initial render fails.

fix: m-6732088176

## Checklist

### DIFF
--- a/.changeset/grumpy-owls-argue.md
+++ b/.changeset/grumpy-owls-argue.md
@@ -1,0 +1,5 @@
+---
+"@lynx-js/react": patch
+---
+
+Make `preact/debug` work with `@lynx-js/react`.
--- a/packages/react/runtime/__test__/utils/setup.js
+++ b/packages/react/runtime/__test__/utils/setup.js
@@ -16,7 +16,12 @@ function inject() {
 
 inject();
 
-afterEach(() => {
+afterEach((context) => {
+  if (context.task.name.includes('preact/debug')) {
+    // Skip preact/debug tests since there would be error
+    return;
+  }
+
   // check profile call times equal end call times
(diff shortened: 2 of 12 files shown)


packages/react/runtime/__test__/utils/setup.js
```diff
@@ -16,7 +16,12 @@ function inject() {
 
 inject();
 
-afterEach(() => {
+afterEach((context) => {
+  if (context.task.name.includes('preact/debug')) {
+    // Skip preact/debug tests since there would be error
```
colinaaa:
```suggestion
    // Skip preact/debug tests since it would throw errors and abort the rendering process
```

.changeset/grumpy-owls-argue.md
```diff
@@ -0,0 +1,5 @@
+---
+"@lynx-js/react": patch
+---
+
+Make `preact/debug` work with `@lynx-js/react`.
```
hzy, line 5: Usage example in changelog will help.

colinaaa: The only usage at this point is `import preact/debug` in the entry (which we would alias to the correct position).

But I'm not sure if this is the best way to make it work.

We may provide a `@lynx-js/react/debug` export in the future PR, and we may update the changeset with an example there.

colinaaa: See: https://github.com/lynx-family/lynx-stack/pull/1250

PR #1226, opened 2025-07-05
chore(deps): pnpm approve-builds `unrs-resolver`

gaoachao:
## Summary

When running `git pull` and `pnpm i`, I encountered this warning. This PR addresses the issue by updating the configuration. Please refer to unrs-resolver@1.9.2's package.json for details.

image
image

## Checklist


pnpm-workspace.yaml
```diff
@@ -44,6 +44,7 @@ patchedDependencies:
 onlyBuiltDependencies:
   - dprint
   - esbuild
+  - unrs-resolver
 
 ignoredBuiltDependencies:
   - core-js
```
colinaaa: It looks like this `postinstall` is only for legacy `npm` version. See: https://github.com/unrs/unrs-resolver/issues/56

We should instead ignore the `postinstall` script.

```suggestion

ignoredBuiltDependencies:
  - core-js
  - unrs-resolver
```

colinaaa (changes requested):
Sorry for not taking a deeper look before.

PR #1230, opened 2025-07-07
feat: set JSONParse: false when extractStr is true

upupming:
## Summary

In https://github.com/web-infra-dev/rspack/pull/9666, we added `module.generator.json.JSONParse` and set `JSONParse: false` for the main thread only to let AOT compilation speed up performance.

However, different `JSONParse` configurations in the main thread and the background thread will prevent `extractStr` from working properly.

 In this PR, when `extractStr` is enabled, we should set `JSONParse` to `false` in both the main and background threads.

## Checklist

### DIFF
--- a/packages/rspeedy/plugin-react/src/generator.ts
+++ b/packages/rspeedy/plugin-react/src/generator.ts
@@ -5,7 +5,12 @@ import type { RsbuildPluginAPI } from '@rsbuild/core'
 
 import { LAYERS } from '@lynx-js/react-webpack-plugin'
 
-export function applyGenerator(api: RsbuildPluginAPI): void {
+import type { PluginReactLynxOptions } from './pluginReactLynx.js'
+
+export function applyGenerator(
+  api: RsbuildPluginAPI,
+  options: Required<PluginReactLynxOptions>,
+): void {
   api.modifyBundlerChain({
     order: 'pre',
     handler: chain => {
@@ -19,6 +24,19 @@ export function applyGenerator(api: RsbuildPluginAPI): void {
         .generator({
           JSONParse: false,
         })
+
+      // If `extractStr` is enabled, we also need to apply the same rule for the background layer.
+      // It will ensure that string literals are same in both main thread and background thread.
+      if (options.extractStr) {
+        chain.module
+          .rule(`json-parse:${LAYERS.BACKGROUND}`)
+          .issuerLayer(LAYERS.BACKGROUND)
+          .test(/\.json$/)
+          .type('json')
+          .generator({
+            JSONParse: false,
+          })
+      }
     },
   })
 }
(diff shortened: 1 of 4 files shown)


luhc228 (approved):

packages/rspeedy/plugin-react/src/generator.ts
```diff
@@ -19,6 +24,19 @@ export function applyGenerator(api: RsbuildPluginAPI): void {
         .generator({
           JSONParse: false,
         })
+
+      // If `extractStr` is enabled, we also need to apply the same rule for the background layer.
+      // It will ensure that string literals are same in both main thread and background thread.
+      if (options.extractStr) {
```
colinaaa: Does it save more bytes if we set `JSONParse: true` when `extractStr` is enabled?

```js
    if (!options.extractStr) {
      chain.module
        .rule(`json-parse:${LAYERS.MAIN_THREAD}`)
        .issuerLayer(LAYERS.MAIN_THREAD)
        .test(/\.json$/)
        .type('json')
        .generator({
          JSONParse: false,
        })
    }
```

packages/rspeedy/plugin-react/src/generator.ts
```diff
@@ -19,6 +24,19 @@ export function applyGenerator(api: RsbuildPluginAPI): void {
...
+        chain.module
+          .rule(`json-parse:${LAYERS.BACKGROUND}`)
+          .issuerLayer(LAYERS.BACKGROUND)
+          .test(/\.json$/)
+          .type('json')
+          .generator({
+            JSONParse: false,
```
colinaaa: If we are also doing this for the background thread, I would prefer add `JSONParse: false` to all layers:

```js
const rule = chain.module
        .rule(`json-parse:${LAYERS.MAIN_THREAD}`) // Maybe a better name
        .test(/\.json$/)
        .type('json')
        .generator({
          JSONParse: false,
        })

if (!options.extractStr) {
  rule.issueLayer(LAYERS.MAIN_THREAD)
}
```

packages/rspeedy/plugin-react/src/generator.ts
upupming: @colinaaa 

Here is the size report of the final bundle on a real project (with `extractStr.strLength=20`):

1. Main thread using `JSONParse: false` and background thread using `JSONParse: true` (current behavior): 1480.7 kB
2. Both threads using `JSONParse: true`: 1471.6 kB (The whole large JSON is stringified and reused)
3. Both threads using `JSONParse: false`: 1422.9 kB (JSON format is `{ key: 'value'}`, so only the `value` can be reused, `key` cannot be used because it is not a string)

Detailed info:

|Case|main-thread.js size|background.js size|bundle size|
|-|-|-|-|
|1|629.2 kB|842.9 kB|1480.7 kB|
|2|**612.8 kB**|**703.9 kB**|1471.6 kB|
|3|629.2 kB|785.2 kB|**1422.9 kB**|

Although case 2 has the smallest `main-thread.js` and `background.js` size, but the final bundle size is still larger than case 3. So I choose case 3 as the final result, what do you think?

colinaaa (approved):

PR #1232, opened 2025-07-07
fix(web): textarea placeholder inherits font-size by default

Liang5757:
## Summary

textarea placeholder inherits font-size by default

## Checklist

## Summary by CodeRabbit

* **Style**
  * Updated textarea placeholder to inherit font size by default, ensuring consistency with surrounding text.


packages/web-platform/web-elements/src/XTextarea/x-textarea.css
```diff
@@ -32,7 +32,6 @@ x-textarea::part(textarea), x-textarea::part(form) {
 }
 x-textarea::part(textarea) {
   --placeholder-color: grey;
-  --placeholder-font-size: 14px;
```
Sherry-hue, line 35: Consider the case where the --placeholder-font-size is not specified and the it is aligned with the text.

Sherry-hue:
Great ! Input also has the placeholder-font-size issue. Could  you help fix it also? It would be even better if you could add a few tests.

PR #1238, opened 2025-07-07
fix(react/runtime): improve error message of circular reference

colinaaa:
## Summary

Improve the error message when setting circular reference as attributes.

close: m-6732040310

## Checklist

## Summary by CodeRabbit

* **New Features**
  * Introduced enhanced debugging tools that provide component stack traces during rendering in development mode.
  * Improved error messages for circular references in deep equality checks, now including component stack information to aid debugging.
  * Added a new symbolic constant for runtime rendering.

* **Bug Fixes**
  * Added error handling for circular references in deep equality checks, preventing crashes and providing clearer error feedback.

* **Tests**
  * Added a new test case to verify correct error handling when circular references are encountered in component props.

* **Chores**
  * Updated test coverage settings to exclude debugging utilities from coverage reports.
  * Added an empty changeset file for tracking future changes.

### DIFF
--- a/packages/react/runtime/__test__/snapshot/spread.test.jsx
+++ b/packages/react/runtime/__test__/snapshot/spread.test.jsx
@@ -636,7 +636,7 @@ describe('spreadUpdate', () => {
     initGlobalSnapshotPatch();
     setSpread_(a);
 
-    expect(() => render(<Comp />, scratch)).toThrowErrorMatchingInlineSnapshot(`
+    expect(() => render(<Comp />, scratchBackground)).toThrowErrorMatchingInlineSnapshot(`
       [TypeError: Converting circular structure to JSON
           --> starting at object with constructor 'Object'
           --- property 'a' closes the circle


colinaaa:
bugbot run

Yradex (approved):

packages/react/runtime/src/utils.ts
```diff
@@ -1,14 +1,33 @@
...
-    return true;
+  try {
+    if (
+      typeof a == 'object' && a !== null && typeof b == 'object' && b !== null
+      && JSON.stringify(a) === JSON.stringify(b)
+    ) {
+      return true;
+    }
+  } catch (error) {
+    if (__DEV__ && (error as Error).message.includes('circular')) {
```
hzy: > https://github.com/FormidableLabs/react-fast-compare/blob/6f7d8afe02e4480c32f5af16f571367cccd47abc/index.js#L127

Should we check this like `react-fast-compare` does?

colinaaa: I've tested on both PrimJS, V8 and JSC and updated the error message

hzy (approved):

packages/react/runtime/src/debug/component-stack.ts
```diff
@@ -0,0 +1,174 @@
...
+ *   return (
+ *     <Foo><span /></Foo> // Foo's owner is Bar, span's owner is Bar
+ *   )
+ * }
+ * ```
+ *
+ * Note: A `vnode` may be hoisted to the root scope due to compiler
+ * optimization. In these cases the `_owner` will be different.
+ */
+let ownerStack: PatchedVNode[] = [];
```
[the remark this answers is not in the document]
colinaaa: This is fork directly from https://github.com/preactjs/preact/blob/main/debug/src/component-stack.js. I think the suggestion makes sense, but we will keep the preact version for now. And we may contribute to preact in the future.

PR #1250, opened 2025-07-08
feat(react): introduce `@lynx-js/react/debug`

colinaaa:
## Summary

Integrate with `preact/debug`.

A new exports `@lynx-js/react/debug` has been added.

```js
import '@lynx-js/react/debug';

import { App } from './App.jsx';

root.render(
  <App />,
);
```

close: m-6732088176

## Checklist

### DIFF
--- a/packages/react/runtime/debug/index.cjs
+++ b/packages/react/runtime/debug/index.cjs
@@ -1,0 +1,8 @@
+// Copyright 2025 The Lynx Authors. All rights reserved.
+// Licensed under the Apache License Version 2.0 that can be found in the
+// LICENSE file in the root directory of this source tree.
+
+if (process.env.NODE_ENV === 'development') {
+  // eslint-disable-next-line n/no-extraneous-require
+  require('preact/debug');
+}
--- a/.changeset/cruel-rocks-open.md
+++ b/.changeset/cruel-rocks-open.md
@@ -1,0 +1,5 @@
+---
+"create-rspeedy": patch
+---
+
+Add `import '@lynx-js/react/debug'` for all templates.
--- a/.changeset/smooth-bars-listen.md
+++ b/.changeset/smooth-bars-listen.md
@@ -1,0 +1,18 @@
+---
+"@lynx-js/react": patch
+---
+
+Introduce `@lynx-js/react/debug` which would include debugging warnings and error messages for common mistakes found.
+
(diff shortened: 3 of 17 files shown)


colinaaa:
bugbot run

packages/react/runtime/debug/index.cjs
```diff
@@ -0,0 +1,8 @@
+// Copyright 2025 The Lynx Authors. All rights reserved.
+// Licensed under the Apache License Version 2.0 that can be found in the
+// LICENSE file in the root directory of this source tree.
+
+if (process.env.NODE_ENV === 'development') {
+  // eslint-disable-next-line n/no-extraneous-require
+  require('preact/debug');
```
upupming, line 7: Will `require` resolve to `.js` instead of `.mjs` files? I wonder if there will be any reference to preact internals to have 2 different copy.

upupming (commented):
Do we need some test case to ensure `preact/debug` work as expected?

packages/react/runtime/debug/index.cjs
colinaaa: Currently, we have only two ways to integrate with `preact/debug`, both of them require using alias.

1. Use `require('preact/debug')` 👈 what I choosed in the PR

    - Alias `preact/debug` to `preact/debug/dist/index.mjs`
    - Removing in production by `if

    This would work with older version of `@lynx-js/react-rsbuild-plugin` since we already have alias to `preact/debug`.

2. Use `import 'preact/debug'`

    - Alias `preact/debug` to `preact/debug/dist/index.mjs`
    - Removing in production by aliasing `preact/debug` to `false`

    This would only works with the change of `@lynx-js/react-rsbuild-plugin`.

The ideal way to work would be like this:

1. Merge https://github.com/lynx-family/lynx-stack/pull/1238
2. Integrate with https://github.com/preactjs/preact/pull/4830

Then we would automatically have `preact/debug` imported and enabled in development, which would work without aliases.

plesiecki:
What would you say to injecting `@lynx-js/react/debug'` automatically only for dev build instead of introducing more boilerplate code?

plesiecki:
Sorry, by the boilerplate code I actually meant the source code here https://github.com/lynx-family/lynx-stack/pull/1250#issue-3212292997. The automatic injection would mean no import of the `@lynx-js/react/debug` module.

PR #1251, opened 2025-07-08
feat: `extractStr` is only supported in `all-in-one` chunkSplit strategy

upupming:
## Summary

In chunk split, `extractStr` will not work as expected. So we will report an error on both `extractStr` and chunk split is enabled.

For example, if a `lib-preact` is shared by entry A and entry B, they have different `__EXTRACT_STR__` array and index. if `lib-preact` follows the `__EXTRACT_STR__` of entry A, it will not work an entry B.

## Checklist

## Summary by CodeRabbit

* **Bug Fixes**
  * String extraction is now automatically disabled when the chunk splitting strategy is not "all-in-one".
  * A warning message is shown when string extraction is forcibly disabled due to an incompatible chunk splitting strategy.

* **Tests**
  * Added tests to verify string extraction behavior across different chunk splitting strategies.

### DIFF
--- a/packages/rspeedy/plugin-react/src/entry.ts
+++ b/packages/rspeedy/plugin-react/src/entry.ts
@@ -9,7 +9,6 @@ import type {
   RsbuildPluginAPI,
   Rspack,
 } from '@rsbuild/core'
-import { logger } from '@rsbuild/core'
 import type { UndefinedOnPartialDeep } from 'type-fest'
 
 import { LAYERS, ReactWebpackPlugin } from '@lynx-js/react-webpack-plugin'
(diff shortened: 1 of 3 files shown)


gaoachao (dismissed):

packages/rspeedy/plugin-react/src/entry.ts
```diff
@@ -240,6 +240,15 @@ export function applyEntry(
 
     const rsbuildConfig = api.getRsbuildConfig()
 
+    if (
+      extractStr
+      && rsbuildConfig.performance?.chunkSplit?.strategy !== 'all-in-one'
+    ) {
+      throw new Error(
```
colinaaa: Instead of throwing an error, can we disable extractStr when using splitChunks?

packages/rspeedy/plugin-react/src/entry.ts
```diff
@@ -9,6 +9,7 @@ import type {
   RsbuildPluginAPI,
   Rspack,
 } from '@rsbuild/core'
+import { logger } from '@rsbuild/core'
```
colinaaa: We should not import `@rsbuild/core` since it's not a peerDependency of plugin-react. Use `api.logger` (added in Rsbuild 1'r) or `rspeedyAPI.logger` instead.

colinaaa (changes requested):

PR #1254, opened 2025-07-08
refactor: move exposure system to web-core

PupilTong:
**THIS IS A BREAKING CHANGE**

@pupiltong : this helps us to start exposure detection after hydrate

Implement SSR for Web #52

For SSR and better performance, we moved the lynx's exposure system from web-element to web-core.

Before this commit, we create Intersection observers by creating HTMLElements.

After this commit, we will create such Intersection observers after dom stabled.

Also, the setInterval for exposure has been removed, now we use an on time lazy timer for such features.


packages/web-platform/web-core/package.json
```diff
@@ -36,6 +36,6 @@
   },
   "peerDependencies": {
     "@lynx-js/lynx-core": "0.1.2",
-    "@lynx-js/web-elements": ">=0.6.0"
+    "@lynx-js/web-elements": ">0.7.7"
```
Sherry-hue, line 39: `>= 0.8.0` is better?

PR #1270, opened 2025-07-10
ci: fix build issue

PupilTong:


packages/web-platform/web-style-transformer/package.json
```diff
@@ -11,10 +11,11 @@
   "sideEffects": false,
   "type": "module",
   "main": "dist/index.js",
-  "typings": "dist/index.d.ts",
+  "typings": "index.d.ts",
   "files": [
     "dist",
     "!dist/**/*.js.map",
+    "index.d.ts",
```
colinaaa, line 18: Maybe we could add `node -e "require('fs').copyFile('dist/index.d.ts', 'index.d.ts')"` in the build script

This would require updating the `turbo.json`, too.

PupilTong: Don't worries, this will be a pure rust module soon. Our goal is  rewritting the entire mts runtime.

PR #1279, opened 2025-07-14
fix: missing comma in app-service manifest on chunk split

upupming:
## Summary

Current when config regex `inlineScripts` config, we may get a wrong result such as:

```js
(function(){'use strict';function n({tt}){tt.define('/app-service.js',function(e,module,_,i,l,u,a,c,s,f,p,d,h,v,g,y,lynx){lynx.requireModuleAsync('http://1.1.1.1:8787/static/js/lib-preact-2.js')module.exports=lynx.requireModule('/.rspeedy/pages/product-list/background.e247bc2d.js',globDynamicComponentEntry?globDynamicComponentEntry:'__Card__')});return tt.require('/app-service.js');}return{init:n}})()

```

image

It misses a comma before the `module.exports`

image

## Checklist

### DIFF
--- a/packages/webpack/template-webpack-plugin/src/LynxEncodePlugin.ts
+++ b/packages/webpack/template-webpack-plugin/src/LynxEncodePlugin.ts
@@ -187,7 +187,7 @@ export class LynxEncodePluginImpl {
           // ```
           '/app-service.js': [
             this.#appServiceBanner(),
-            ...[externalManifest, inlinedManifest].flatMap(manifest =>
+            [externalManifest, inlinedManifest].flatMap(manifest =>
               Object.keys(manifest).map(name => {
                 if (manifest === externalManifest) {
                   return `lynx.requireModuleAsync('${
@@ -199,7 +199,7 @@ export class LynxEncodePluginImpl {
                   }',globDynamicComponentEntry?globDynamicComponentEntry:'__Card__')`;
                 }
               }).join(',')
-            ),
+            ).filter(arr => arr.length).join(','),
             this.#appServiceFooter(),
           ].join(''),
           ...Object.fromEntries(
--- a/packages/webpack/template-webpack-plugin/test/cases/inline-scripts/inline-partial/index.js
+++ b/packages/webpack/template-webpack-plugin/test/cases/inline-scripts/inline-partial/index.js
@@ -1,0 +1,29 @@
+/*
+// Copyright 2024 The Lynx Authors. All rights reserved.
+// Licensed under the Apache License Version 2.0 that can be found in the
+// LICENSE file in the root directory of this source tree.
+*/
+/// <reference types="vitest/globals" />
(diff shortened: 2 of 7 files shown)


packages/webpack/template-webpack-plugin/src/LynxEncodePlugin.ts
```diff
@@ -199,7 +199,7 @@ export class LynxEncodePluginImpl {
                   }',globDynamicComponentEntry?globDynamicComponentEntry:'__Card__')`;
                 }
               }).join(',')
-            ),
+            ).filter(arr => arr.length).join(','),
```
colinaaa: It looks like the `arr` is not actually an array since we are using `flatMap`.

packages/webpack/template-webpack-plugin/test/cases/inline-scripts/inline-partial/index.js
```diff
@@ -0,0 +1,29 @@
...
+/// <reference types="vitest/globals" />
+
+import { existsSync } from 'node:fs';
+import { readFile } from 'node:fs/promises';
+import { resolve } from 'node:path';
+
+import { foo } from './foo.js';
+import { bar } from './bar.js';
+
+it('should generate correct foo template', async () => {
```
colinaaa: This test does fail without the change. Since the `app-service.js` is not actually run.

packages/webpack/template-webpack-plugin/src/LynxEncodePlugin.ts
colinaaa: Instead of `.join`, can we split the `externalManifest` and `inlinedManifest` instead?


```js
            Object.keys(externalManifest).map(name =>
              `lynx.requireModuleAsync('${
                this.#formatJSName(name, publicPath)
              }')`
            ).join(','),
            ';module.exports=',
            Object.keys(inlinedManifest).map(name =>
              `lynx.requireModule('${
                this.#formatJSName(name, '/')
              }',globDynamicComponentEntry?globDynamicComponentEntry:'__Card__')`
            ).join(','),
            ';',
```

packages/webpack/template-webpack-plugin/test/cases/inline-scripts/inline-partial/index.js
upupming: I have added a `eval(manifest['/app-service.js'])` check to ensure the script is executed. This test now will fail on branch main (82405162) but success on current commit (84bac708)

packages/webpack/template-webpack-plugin/src/LynxEncodePlugin.ts
upupming: Thanks for advice, I have replaced with this!

colinaaa (approved):
👍

PR #1285, opened 2025-07-14
feat(react/runtime): render Suspense fallback on initial render

Yradex:
## Summary

Enable rendering of the `Suspense` fallback on initial render.

## Checklist


colinaaa:
bugbot run

packages/react/runtime/src/lynx/lazy-bundle.ts
```diff
@@ -157,19 +157,4 @@ function withSyncResolvers<T>() {
...
-      return Lazy(props);
-    } catch (e) {
-      // We should never throw at mainThread
-      return createElement(Fragment, {});
-    }
-  }
-
-  return _Lazy as T;
-}
+export const mainThreadLazy = backgroundLazy;
```
colinaaa: Can we instead avoid distinguishing `mainThreadLazy` and `backgroundLazy`? E.g.: always use `lazy` from `preact`

packages/react/runtime/src/renderToOpcodes/utils.ts
```diff
@@ -0,0 +1,12 @@
+export class Deferred<T> {
```
colinaaa: It looks like this is only used for test. Should we move it out of `src`?

packages/react/runtime/__test__/renderToOpcodes.test.jsx
```diff
@@ -4,21 +4,54 @@ import { Component, createContext, Fragment } from 'preact';
...
 import { setupPage, SnapshotInstance, snapshotInstanceManager } from '../src/snapshot';
 import { createElement, cloneElement } from '../lepus';
+import { Suspense } from 'preact/compat';
+import { Deferred } from '../src/renderToOpcodes/utils';
+
+const defaultChildren = <text>it works</text>;
+
+// `createSuspender()` is forked from `preact` unit test
+export function createSuspender() {
+  const deferred = new Deferred();
```
colinaaa: Since we are enforcing using node 22 and above, maybe we could use `Promise.withResolvers` instead.

packages/react/runtime/__test__/renderToOpcodes.test.jsx
```diff
@@ -380,6 +423,225 @@ describe('renderToOpcodes', () => {
...
+
+    const fallbackJsx1 = <text>loading1...</text>;
+    const fallbackJsx2 = <text>loading2...</text>;
+
+    const rendered = renderToString(
+      <Suspense fallback={fallbackJsx1}>
+        <view attr={Math.random()}>
+          <Suspender1>
+            <text className='foo'>bar</text>
+            <Suspender2 fallback={fallbackJsx2}>
```
colinaaa: Do you mean `Suspense` here?

packages/react/runtime/__test__/renderToOpcodes.test.jsx
```diff
@@ -380,6 +423,225 @@ describe('renderToOpcodes', () => {
...
+    await waitSchedule();
+
+    const fallbackJsx1 = <text>loading1...</text>;
+    const fallbackJsx2 = <text>loading2...</text>;
+    const resolvedJsx1 = <text className='foo'>bar</text>;
+    const resolvedJsx2 = <text className='foo'>bar</text>;
+
+    const rendered = renderToString(
+      <Suspense>
+        <Suspender1 fallback={fallbackJsx1}>
```
colinaaa: Do you mean setting fallback on `Suspense`? Or you want a `Suspense` without fallback?

Since it doesn't make any sense to have fallback on Suspender. 

```suggestion
      <Suspense fallback={fallbackJsx1}>
        <Suspender1>
```

packages/react/runtime/src/index.ts
```diff
@@ -33,18 +33,13 @@ import {
...
 export { PureComponent } from 'preact/compat';
 export * from './hooks/react.js';
 
-const lazy: typeof import('preact/compat').lazy = /*#__PURE__*/ (() => {
-  lynx.loadLazyBundle = loadLazyBundle;
-  return __LEPUS__
-    ? mainThreadLazy
-    : backgroundLazy;
-})();
+lynx.loadLazyBundle = loadLazyBundle;
```
colinaaa: We should keep the IIFE here to save some bytes when `lazy` is not used.

```suggestion
const lazy: typeof import('preact/compat').lazy = /*#__PURE__*/ (() => {
  lynx.loadLazyBundle = loadLazyBundle;
  return preactLazy;
});
```

This would save like 1 kB for the hello world project :)

colinaaa (changes requested):

colinaaa (approved):
👍

packages/react/runtime/src/renderToOpcodes/index.ts
```diff
@@ -241,11 +249,28 @@ function _renderToString(
...
+
+        if (component[DIRTY]) {
+          rendered = renderClassComponent(vnode, context);
+          component = vnode[COMPONENT];
+
+          opcodes.length = opcodesLength;
+          _renderToString(rendered, context, isSvgMode, selectValue, vnode, opcodes, opcodes.length);
+        }
+      } else {
+        throw e;
```
upupming, line 266: If we `throw e` from here, I think we still need to use `mainThreadLazy` to catch this error to avoid uncaught exception in main thread.

packages/react/runtime/src/renderToOpcodes/index.ts
```diff
@@ -241,11 +249,28 @@ function _renderToString(
...
+    } catch (e) {
+      if (e && typeof e === 'object' && e.then && component && /* _childDidSuspend */ component.__c) {
+        component.setState({ /* _suspended */ __a: true });
+
+        if (component[DIRTY]) {
+          rendered = renderClassComponent(vnode, context);
+          component = vnode[COMPONENT];
+
+          opcodes.length = opcodesLength;
+          _renderToString(rendered, context, isSvgMode, selectValue, vnode, opcodes, opcodes.length);
```
upupming, line 263: If there are multiple dynamic components such as `CompA` and `CompB`, it may throw 2 promises here, so will we call `_renderToString` 2 times here? Is there any performance downgrade?

Yradex: Not quite. Once the first promise is thrown, Suspense catches it and renders the fallback immediately. The second dynamic component won't even be rendered.

Yradex: If they are in separate Suspense boundaries, then yes. I think it's the expected behavior though.

PR #1305, opened 2025-07-17
feat: support rstest in testing library

upupming:
## Summary

In this PR, we added [`rstest`](https://github.com/web-infra-dev/rstest) support for ReactLynx Testing Library. All tests of RLTL are passed successfully.

User can use rstest with RLTL now:

```ts
import { defineConfig } from '@rstest/core';
import { pluginReactLynx } from '@lynx-js/react-rsbuild-plugin';
import { rstestTestingLibraryPlugin } from '@lynx-js/react/testing-library/plugins';

export default defineConfig({
  plugins: [
    rstestTestingLibraryPlugin(),
    pluginReactLynx(),
  ],
  testEnvironment: 'jsdom',
  setupFiles: [
    require.resolve('@lynx-js/react/testing-library/setupFiles/rstest'),
  ],
  globals: true,
});
```

The usage of `vitest` also changed from `mergeConfig` to a vite plugin:

```ts
import { defineConfig } from 'vitest/config';
import { vitestTestingLibraryPlugin } from '@lynx-js/react/testing-library/plugins';

export default defineConfig({
  plugins: [vitestTestingLibraryPlugin({
    runtimePkgName: '@lynx-js/react',
  })],
});

```

## Checklist


package.json
```diff
@@ -20,6 +20,7 @@
     "@microsoft/api-extractor": "catalog:",
     "@rslib/core": "^0.10.5",
     "@rspack/core": "catalog:rspack",
+    "@rstest/core": "^0.0.8",
```
gaoachao:
```suggestion
    "@rstest/core": "catalog:rstest",
```

To facilitate future migrations, it's advisable to centrally manage the version of `rstest` in the `pnpm-workspace.yaml`.

packages/rspeedy/create-rspeedy/template-react-rstest-rltl-js/package.json
```diff
@@ -0,0 +1,25 @@
...
+    "test": "rstest run"
+  },
+  "dependencies": {
+    "@lynx-js/react": "workspace:*"
+  },
+  "devDependencies": {
+    "@lynx-js/qrcode-rsbuild-plugin": "workspace:*",
+    "@lynx-js/react-rsbuild-plugin": "workspace:*",
+    "@lynx-js/rspeedy": "workspace:*",
+    "@rstest/core": "^0.0.8",
```
gaoachao: ditto

.changeset/chilly-pumas-mate.md
```diff
@@ -0,0 +1,10 @@
+---
+"@lynx-js/react": patch
+---
+
+Remove the `@lynx-js/react/testing-library/vitest-config` exports, user should replace it with the new `vitestTestingLibraryPlugin` in `vitest.config.js`:
```
colinaaa: This is a **BREAKING CHANGE**, we should use `minor` change and mark **BREAKING CHANGE** in changeset.

BTW, can we avoid making breaking change (i.e.: keep the `createVitestConfig`)

.changeset/red-lamps-arrive.md
```diff
@@ -0,0 +1,25 @@
...
+import { defineConfig } from '@rstest/core';
+import { pluginReactLynx } from '@lynx-js/react-rsbuild-plugin';
+import { rstestTestingLibraryPlugin } from '@lynx-js/react/testing-library/plugins';
+
+export default defineConfig({
+  plugins: [
+    rstestTestingLibraryPlugin(),
+    pluginReactLynx({
+      enableTestingLibrary: true,
+    }),
```
colinaaa: Instead of a new option, we may use the [`api.context.callerName`](https://rsbuild.rs/api/javascript-api/instance#contextcallername) to see if the plugin is applied by Rstest

```suggestion
    pluginReactLynx(),
```

packages/react/testing-library/vitest-polyfill.js
```diff
@@ -0,0 +1,6 @@
+// in order to make our test case work for
```
colinaaa: Use `.cjs` instead.

colinaaa (commented):
1. We may want to avoid breaking changes to the existing `vitest` users.
2. We don't need a new Rstest plugin, since the [api.context.callerName](https://rsbuild.rs/api/javascript-api/instance#contextcallername) would allow applying custom configuration in Rstest.
3. Rstest would have first-class support for existing Rspack, Rsbuild, Rslib projects (which would load the corresponding config instead of creating a separate `rstest.config.ts`). We may want the same thing for `lynx.config.js` (this would avoid writing multiple `resolve.alias`)

.changeset/chilly-pumas-mate.md
upupming: @colinaaa Good catch, the breaking change may affect users. I have reverted the breaking change. `mergeConfig` will still be supported, and we also export the Vite plugin for user customization.

packages/testing-library/examples/basic/package.json
```diff
@@ -6,7 +6,8 @@
   "scripts": {
     "build": "rspeedy build",
     "dev": "rspeedy dev",
-    "test": "vitest"
+    "test": "vitest",
+    "test:rs": "rstest run"
```
gaoachao: nitpick: 

```suggestion
    "rstest": "rstest"
```

or 
```suggestion
    "test:rstest": "rstest"
```


I personally think the `:rs` here doesn't intuitively indicate `rstest`, instead it gives me the impression of testing Rust. Also, the `run` should be unnecessary.

.changeset/red-lamps-arrive.md
```diff
@@ -0,0 +1,19 @@
+---
+"@lynx-js/react": patch
+---
+
+Support rstest for testing library, you can use rstest with RLTL now:
```
hzy, line 5:
```suggestion
Support rstest for testing library, you can use rstest with RLTL now:

Create a config file `rstest.config.ts` with the following content:
```

.changeset/red-lamps-arrive.md
```diff
@@ -0,0 +1,19 @@
...
+
+export default defineConfig({
+  ...lynxConfig as RstestConfig,
+  testEnvironment: 'jsdom',
+  setupFiles: [
+    require.resolve('@lynx-js/react/testing-library/setupFiles/rstest'),
+  ],
+  globals: true,
+});
+```
```
hzy, line 21: And, use `rstest` as usual:

```bash
$ rstest
```

For more usage detail, see https://rstest.rs/

hzy (approved):

packages/webpack/react-webpack-plugin/src/loaders/testing.ts
```diff
@@ -0,0 +1,119 @@
...
+import type { LoaderContext } from '@rspack/core';
+
+import { JSX_IMPORT_SOURCE, RUNTIME_PKG } from './options.js';
+import type { ReactLoaderOptions } from './options.js';
+
+function normalizeSlashes(file: string) {
+  return file.replaceAll(path.win32.sep, '/');
+}
+
+function backgroundLoader(
```
colinaaa: Should we rename this?
```suggestion
function testingLoader(
```

.github/workflows/test.yml
```diff
@@ -244,6 +267,7 @@ jobs:
       - test-type
       - test-typos
       - test-vitest
+      - test-rstest
```
colinaaa: Please sort in alphabetical order :)

.github/workflows/test.yml
```diff
@@ -229,6 +235,23 @@ jobs:
         --no-cache
         --logHeapUsage
         --silent
+  test-rstest:
```
colinaaa: ditto.

packages/react/package.json
```diff
@@ -101,9 +101,15 @@
       "default": "./testing-library/dist/pure.js"
     },
     "./testing-library/vitest-config": {
-      "types": "./testing-library/types/vitest-config.d.ts",
+      "types": "./testing-library/dist/vitest.config.d.ts",
       "default": "./testing-library/dist/vitest.config.js"
     },
+    "./testing-library/plugins": {
```
colinaaa: Why do we need this in this PR? It doesn't seem like to be related with Rstest.

And we are having too many `exports` now :)

packages/webpack/react-webpack-plugin/src/loaders/testing.ts
```diff
@@ -0,0 +1,119 @@
...
+  content: string,
+): void {
+  const require = createRequire(import.meta.url);
+  const { transformPath = '@lynx-js/react/transform' } = this.getOptions();
+  const { transformReactLynxSync } = require(
+    transformPath,
+  ) as typeof import('@lynx-js/react/transform');
+  const filename = normalizeSlashes(
+    path.relative(
+      require.resolve('@lynx-js/react/testing-library'),
```
colinaaa: Why do we need the relative path from `'@lynx-js/react/testing-library'` instead of `this.rootContext`?

packages/react/testing-library/rstest.config.ts
```diff
@@ -0,0 +1,52 @@
...
+          const newOptions = {
+            ...originalOptions,
+            jsc: {
+              ...originalOptions.jsc,
+              transform: {
+                ...((originalOptions.jsc?.transform)
+                  || {}),
+                // Keep our snapshot consistent with vitest
+                // class field will be undefined if not initialized
+                useDefineForClassFields: true,
```
colinaaa: If this is the only thing you want to change, you may use `tools.swc.jsc.transform.useDefineForClassFields`.

packages/rspeedy/create-rspeedy/src/index.ts
```diff
@@ -67,22 +67,30 @@ async function getTemplateName({ template }: Argv) {
...
-      required: false,
+  const tools = []
+  const testingTools = checkCancel<string>(
+    await select({
+      message: 'Select testing framework',
       options: [
         {
           value: 'vitest-rltl',
-          label: 'Add ReactLynx Testing Library for unit testing',
+          label: 'Add ReactLynx Testing Library with Vitest',
```
colinaaa:
```suggestion
          label: 'Vitest',
```

packages/rspeedy/create-rspeedy/src/index.ts
```diff
@@ -67,22 +67,30 @@ async function getTemplateName({ template }: Argv) {
...
+        },
+        {
+          value: 'rstest-rltl',
+          label: 'Add ReactLynx Testing Library with Rstest',
```
colinaaa:
```suggestion
          label: 'Rstest',
          hint: 'recommended',
```

You may remove the hint for now if you consider it not stable enough.

packages/react/testing-library/src/vitest.config.ts
```diff
@@ -0,0 +1,11 @@
+import { defineConfig, type ViteUserConfig } from 'vitest/config';
+import { vitestTestingLibraryPlugin } from './plugins/index.js';
+import type { TestingLibraryOptions } from './plugins/vitest.js';
+
+export async function createVitestConfig(options?: TestingLibraryOptions): Promise<ViteUserConfig> {
+  return defineConfig({
```
colinaaa: Can we avoid using `async` since there is no `await`.

```suggestion
export function createVitestConfig(options?: TestingLibraryOptions): ViteUserConfig {
  return defineConfig({
```

packages/react/package.json
```diff
@@ -101,9 +101,15 @@
...
+      "types": "./testing-library/dist/plugins/index.d.ts",
+      "default": "./testing-library/dist/plugins/index.js"
+    },
+    "./testing-library/setupFiles/vitest": "./testing-library/dist/setupFiles/vitest.js",
```
colinaaa: Can we avoid this `exports` by including the setup files in our vitest config (or plugin)?

packages/react/testing-library/vitest.config.ts
```diff
@@ -1,13 +1,13 @@
-import { defineConfig, mergeConfig } from 'vitest/config';
-import { createVitestConfig } from './dist/vitest.config';
+import { defineConfig } from 'vitest/config';
+import { vitestTestingLibraryPlugin } from './dist/plugins';
 
-const defaultConfig = await createVitestConfig({
-  runtimePkgName: '@lynx-js/react',
-});
-const config = defineConfig({
+export default defineConfig({
```
colinaaa: It looks like this is a **BREAKING CHANGE** here.

The new `vitest.config.ts` cannot work without changes:

```js
import { defineConfig, mergeConfig } from 'vitest/config'
import { createVitestConfig } from '@lynx-js/react/testing-library/vitest-config'

const defaultConfig = await createVitestConfig()
const config = defineConfig({
  test: {},
})

export default mergeConfig(defaultConfig, config)
```

packages/rspeedy/plugin-react/src/pluginReactLynx.ts
```diff
@@ -366,12 +367,23 @@ export function pluginReactLynx(
     name: 'lynx:react',
     pre: ['lynx:rsbuild:plugin-api'],
     async setup(api) {
+      const isRstest = api.context.callerName === 'rstest'
```
colinaaa: Can we add some test on `callerName === 'rstest'`?

packages/rspeedy/plugin-react/src/rstest.ts
```diff
@@ -0,0 +1,31 @@
...
+    return mergeRsbuildConfig({
+      output: {
+        // Allow using Node.js internal modules in testing framework directly
+        externals: /^node:/,
+      },
+    }, config)
+  })
+
+  // stub the `rspeedy.api`
+  api.expose(Symbol.for('rspeedy.api'), {
```
colinaaa: Rstest is not the only one needs this stub API. Either:

1. Remove most of the reference to 'rspeedy.api'. Since it can be replaced with `api.logger`
2. Move this stub API to another place (instead of calling it `rstest.ts`)

packages/webpack/react-webpack-plugin/src/ReactWebpackPlugin.ts
```diff
@@ -167,26 +174,28 @@ class ReactWebpackPlugin {
...
-      __EXTRACT_STR__: JSON.stringify(Boolean(options.extractStr)),
-      __FIRST_SCREEN_SYNC_TIMING__: JSON.stringify(
-        options.firstScreenSyncTiming,
-      ),
-      __ENABLE_SSR__: JSON.stringify(options.enableSSR),
-      __DISABLE_CREATE_SELECTOR_QUERY_INCOMPATIBLE_WARNING__: JSON.stringify(
-        options.disableCreateSelectorQueryIncompatibleWarning,
-      ),
-    }).apply(compiler);
+    if (options.enableDefine) {
```
colinaaa: Why do we need to disable these defined in Rstest?

It looks like all of them are used in `@lynx-js/react` and should be defined. And they are not something like `__MAIN_THREAD__` or `__BACKGROUND__` which should be determined during runtime.

packages/webpack/react-webpack-plugin/src/loaders/testing.ts
upupming: This is needed to keep the consistency of relative path generated by Rstest and Vitest. Same snapshotId will be generated in Rstest/Vitest such as `f24e5` in `__snapshot_f24e5_test_21`.

Same logic in Vitest: https://github.com/upupming/lynx-stack/blob/cf78e1a732c2fe66e9027641973d3e8e44d91d15/packages/react/testing-library/src/plugins/vitest.ts#L74

upupming: `this.rootContext` is not stable when running test case in `lynx-stack` and `lynx-stack/packages/testing-library`

upupming: Sorry, I found it can be stable both in Rstest and Vitest:

- In Rstest use `this.rootContext`
- In Vitest use `config.root` in [configResolved](https://vite.dev/guide/api-plugin.html#configresolved)

I have changed to use them.

packages/react/testing-library/vitest.config.ts
upupming: Actually, I supported both `vitestTestingLibraryPlugin` and `createVitestConfig`, so `createVitestConfig` is still supported.

Due to `vitestTestingLibraryPlugin` makes our `exports` too long and might be not related to this PR, I have removed the exports from this PR~

packages/webpack/react-webpack-plugin/src/ReactWebpackPlugin.ts
upupming: The testing framework will set `__MAIN_THREAD__` and `__BACKGROUND__ ` in the testing runtime to simulate a thread-switching behavior. Eg.

https://github.com/lynx-family/lynx-stack/blob/d29617030d83dd5d1ec2aaa8d79856df397d60b1/packages/react/runtime/__test__/utils/globals.js#L86-L87

This is because, in testing, we simulate a dual-thread behavior by using a single thread. Our code needs to be in a "MIXED" mode (and outputs **ONE single bundle**) that can run both the login that needs to be executed on the main thread and the background thread.

packages/rspeedy/plugin-react/src/rstest.ts
upupming: 1. [`config.output.filename.bundle`](https://github.com/lynx-family/lynx-stack/blob/f045d6365c7c0e2e2f44e41fcb24a0797c7efcb6/packages/rspeedy/plugin-react/src/entry.ts#L79) in `plugin-react/src/entry.ts` needs `api.config` to get the original Rspeedy config, but there is no good way to replace it with pure Rsbuild API.
2. I will try to use this method.

packages/rspeedy/plugin-react/src/pluginReactLynx.ts
upupming: Tests are added in `packages/rspeedy/plugin-react/test/config.test.ts` to make sure correct loaders are used.

packages/react/testing-library/vitest.config.ts
colinaaa: There is the `createVitestConfig` API, but the implementation and behavior is totally different.

We would have `setupFiles` set in `createVitestConfig` before, but after this change we need to manually added in `vitest.cofnig.ts`.

packages/rspeedy/plugin-react/src/rstest.ts
colinaaa: It seems like the [config.output.filename.bundle](https://github.com/lynx-family/lynx-stack/blob/f045d6365c7c0e2e2f44e41fcb24a0797c7efcb6/packages/rspeedy/plugin-react/src/entry.ts#L79) is only needed when using with Rspeedy.

Can we make an assertion on this instead of creating a stub one?

```js
import invariant from 'tiny-invariant'
import { ExposedAPI } from '@lynx-js/rspeedy'

function getRspeedyAPI(api: RsbuildPluginAPI): ExposedAPI {
  const rspeedyAPI = api.useExposed(Symbol.for('rspeedy.api'))
  invariant(rspeedyAPI, 'Should have rspeedy.api')
  return rspeedyAPI
}
```

packages/webpack/react-webpack-plugin/src/ReactWebpackPlugin.ts
colinaaa: I understand that we need `__MAIN_THREAD__` and `__BACKGROUND__` in the runtime. But it doesn't seem like they are set in the `DefinePlugin` here.

upupming: Oh yes, actually `__MAIN_THREAD__` and `__BACKGROUND__` are not defined here (but in jsx loaders). 

However, we may also want to test the behavior of different `__ALOG__` configurations (such as `packages/react/testing-library/src/__tests__/alog.test.jsx` requires `__ALOG__` to be true) or `__PROFILE__` configurations. If we define it here, the global injected defines will no longer take effect. What do you think about this?

Related logic:

- https://github.com/lynx-family/lynx-stack/blob/6baeb9f8db406b0740d59b7e94b064b36cb97f73/packages/testing-library/testing-environment/src/index.ts#L233
- https://github.com/lynx-family/lynx-stack/blob/6baeb9f8db406b0740d59b7e94b064b36cb97f73/packages/react/runtime/__test__/utils/globals.js#L82

colinaaa: Can we instead define `__ALOG__` in Rsbuild plugin?

packages/react/testing-library/vitest.config.ts
upupming: User does not need to specify `setupFiles` in `createVitestConfig`, because we have used `config` hook of vite plugin to specify `setupFiles` internally:

https://github.com/upupming/lynx-stack/blob/e884c07506dad55290708d9d3b0c12e1a426340e/packages/react/testing-library/src/plugins/vitest.ts#L134-L145

This should be an internal refactor and not affect the user. I am also open to move this change to another PR, what do you think?

packages/rspeedy/plugin-react/src/rstest.ts
upupming: Should been resolved in 9d8bd7cee1c27a9aa5e06e37de2af2d7a51fb53e

packages/webpack/react-webpack-plugin/src/ReactWebpackPlugin.ts
upupming: Should been resolved in 095bc2814afc9615162d6d623017d7505f9c63b8

packages/react/package.json
```diff
@@ -101,9 +101,10 @@
       "default": "./testing-library/dist/pure.js"
     },
     "./testing-library/vitest-config": {
-      "types": "./testing-library/types/vitest-config.d.ts",
+      "types": "./testing-library/dist/vitest.config.d.ts",
       "default": "./testing-library/dist/vitest.config.js"
     },
```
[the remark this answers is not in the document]
upupming: It will be built by `rslib build` in `packages/react/testing-library/package.json`.

packages/testing-library/testing-environment/src/index.ts
```diff
@@ -265,7 +252,10 @@ function injectMainThreadGlobals(target?: any, polyfills?: any) {
   target.requestAnimationFrame = setTimeout;
   target.cancelAnimationFrame = clearTimeout;
 
-  target.console = createPreconfiguredConsole();
+  target.console.profile = console.profile = () => {};
+  target.console.profileEnd = console.profileEnd = () => {};
+  // @ts-expect-error Lynx has console.alog
+  target.console.alog = console.alog = () => {};
```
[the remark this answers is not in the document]
upupming: It is needed to keep compatible with rstest because rstest will provide `console` in an iife.

packages/react/testing-library/src/plugins/vitest.ts
```diff
@@ -0,0 +1,177 @@
+// Copyright 2025 The Lynx Authors. All rights reserved.
+// Licensed under the Apache License Version 2.0 that can be found in the
+// LICENSE file in the root directory of this source tree.
+import type { ResolvedConfig, Vite } from 'vitest/node';
+import { VitestPackageInstaller } from 'vitest/node';
+import path from 'node:path';
```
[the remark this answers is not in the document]
upupming: We cannot access `vite` because it is not installed as devDep.

packages/react/testing-library/src/plugins/vitest.ts
```diff
@@ -0,0 +1,177 @@
...
+  const preactAlias = generateAlias('preact', preactDir, runtimeOSSDir);
+
+  let config: ResolvedConfig;
+
+  return {
+    name: 'transformReactLynxPlugin',
+    enforce: 'pre',
+    async buildStart() {
+      await ensurePackagesInstalled();
+    },
```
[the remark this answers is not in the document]
upupming: We do not want to change this behavior now, because it is copied from `packages/react/testing-library/src/vitest.config.js` before

packages/react/package.json
```diff
@@ -101,9 +101,10 @@
...
+    "./testing-library/setupFiles/rstest": "./testing-library/dist/setupFiles/rstest.js",
```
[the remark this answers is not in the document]
upupming: We won't expose them now as it will make our exports is too long. They are considered as internal only APIs.

PR #1310, opened 2025-07-18
fix(rspeedy/core): make `rspeedy dev --mode production` work

colinaaa:
## Summary

Previously, when using `rspeedy dev --mode production`, the publicPath would be `'/'` due to we only include the `dev.plugin.ts` when `mode === 'development'`.

This patch not only include the plugin when using `rspeedy dev --mode production`, also fix publicPath when using `rspeedy build --mode development`.

## Checklist

## Summary by CodeRabbit

* **Bug Fixes**
  * Ensure the asset public path is correctly applied when running the dev server with production mode, so assets load properly.

* **Tests**
  * Added tests verifying the dev server applies the asset public path in production-mode runs.
  * Updated test harness calls to explicitly specify action context for build-related tests.

* **Chores**
  * Added a package changeset recording the patch update.


colinaaa:
bugbot run

packages/rspeedy/core/src/cli/dev.ts
```diff
@@ -59,6 +59,12 @@ export async function dev(
 
     const rspeedy = await createRspeedy(createRspeedyOptions)
 
+    // When using `rspeedy dev --mode=production`, the `lynx:rsbuild:dev` plugin will not be loaded.
+    if (!rspeedy.isPluginExists('lynx:rsbuild:dev')) {
```
upupming: This seems to be a hack. Can we directly use:

```suggestion
    // When using `rspeedy dev --mode=production`
    if (rspeedyConfig.mode === 'production') {
```

Also I wonder if we can change it to use `command === 'dev'` here to solve this problem:

https://github.com/lynx-family/lynx-stack/blob/824051620ec2f92b00ee9d57ed566b7810709e3f/packages/rspeedy/core/src/plugins/index.ts#L84-L87

colinaaa: 1. Aside from `mode === 'production'`, we would also have `mode === 'none'` which should apply the dev plugins. wdyt about 
```suggestion
    // When using `rspeedy dev --mode=production`, the `lynx:rsbuild:dev` plugin will not be loaded.
    if (rspeedyConfig.mode !== 'development') {
```

2. There is no such `command` param in `createRspeedy` (which is the same as `createRsbuild`). There is `api.context.action === 'dev'`, but it is readonly and would be assigned **after** `rsbuild.startDevServer` called, which is too late for us.

upupming: 1. I think using `rspeedyConfig.mode === 'production' || rspeedyConfig.mode === 'none'` sounds good! It is more straightforward and does not rely on the implementation detail of `lynx:rsbuild:dev`.
2. Oh I get it, I also thought we can add a `command` to [`CreateRspeedyOptions`](https://github.com/lynx-family/lynx-stack/blob/19c7f916c6461bcda83f65be2337759a294f3208/packages/rspeedy/core/src/cli/dev.ts#L60), but I am not sure if it is a good way.

colinaaa: I do find a better way to do this.

Rsbuild will add a new `apply` which controls whether the plugin should be applied.

Using `apply` in `dev.plugin.ts` like this:

```js
{
  name: 'lynx:rsbuild:dev',
  apply(config, { action }) {
    return action === 'dev' || config.mode === 'development';
  },
}
```

could solve our problem.

See: https://github.com/web-infra-dev/rsbuild/pull/5653

colinaaa: I've updated the implementation to use `apply`. PTAL~

upupming (approved):

PR #1311, opened 2025-07-18
fix(react): avoid some unexpected `__SetAttribute` when attribute value is `undefined`

hzy:
## Checklist

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevented unexpected attribute operations when passing undefined as a prop in JSX elements.

* **New Features**
  * Added a test to ensure no unnecessary patch is emitted when hydrating elements with undefined props.

* **Refactor**
  * Updated the serialization format for snapshot values from arrays to objects with string keys to improve consistency.

* **Tests**
  * Adjusted existing test snapshots to reflect the new serialization format for values.
  * Added coverage for scenarios involving undefined and null props during hydration.

### DIFF
--- a/packages/react/runtime/src/snapshot.ts
+++ b/packages/react/runtime/src/snapshot.ts
@@ -247,7 +247,7 @@ export function traverseSnapshotInstance<I extends WithChildren>(
 export interface SerializedSnapshotInstance {
   id: number;
   type: string;
-  values?: any[] | undefined;
+  values?: Record<string, unknown> | undefined;
   extraProps?: Record<string, unknown> | undefined;
   children?: SerializedSnapshotInstance[] | undefined;
 }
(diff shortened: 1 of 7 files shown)


packages/react/runtime/src/snapshot.ts
```diff
@@ -625,7 +625,13 @@ export class SnapshotInstance {
     return {
       id: this.__id,
       type: this.type,
-      values: this.__values,
+      values: this.__values
+        // If `this.__values` is `[0, 1, 2]`, it can be transformed to `{ '0': 0, '1': 1, '2': 2 }`
+        // by `Object.assign({}, this.__values)`, so if `this.__values` is [undefined],
+        // it will be transformed to `{ '0': undefined }`, and transformed to `{}` by `JSON.stringify`.
+        // Therefore, `values[0] === undefined` is preserved during serialization.
+        ? Object.assign({}, this.__values) as unknown as Record<string, unknown>
```
colinaaa: @coderabbitai is there any better way to transform a `[0, 1, 2]` to `{ '0': 0, '1': 1, '2': 2 }`?

colinaaa (approved):

PR #1316, opened 2025-07-18
feat: introduce `@lynx-js/react/compat`

gaoachao:
## Summary

Introduce `@lynx-js/react/compat`.

```tsx
import { startTransition, useTransition } from '@lynx-js/react/compat';
```

## Checklist

### DIFF
--- a/packages/react/package.json
+++ b/packages/react/package.json
@@ -14,6 +14,11 @@
       "lazy": "./runtime/lazy/react.js",
       "default": "./runtime/lib/index.js"
     },
+    "./compat": {
+      "types": "./runtime/compat/index.d.ts",
+      "lazy": "./runtime/compat/index.js",
+      "default": "./runtime/compat/index.js"
+    },
     "./debug": {
       "types": "./runtime/debug/index.d.ts",
       "lazy": "./runtime/debug/index.js",
--- a/packages/react/runtime/compat/index.d.ts
+++ b/packages/react/runtime/compat/index.d.ts
@@ -1,0 +1,33 @@
+// Copyright 2025 The Lynx Authors. All rights reserved.
+// Licensed under the Apache License Version 2.0 that can be found in the
+// LICENSE file in the root directory of this source tree.
+export * from '@lynx-js/react';
+
+/**
+ * Starts a transition update, marking state updates in the callback as non-urgent
+ *
+ * Note: This is Preact's startTransition implementation, which differs fundamentally from React's:
+ * - React marks updates as low priority with interruption and time-slicing support
+ * - Preact simply executes the callback synchronously without true concurrent features
+ *
+ * @param cb - Callback function containing state updates, executed immediately and synchronously
+ *
+ * @public
+ */
+declare function startTransition(cb: () => void): void;
+/**
+ * Returns a tuple with a pending state and a startTransition function
+ *
+ * Note: This is Preact's useTransition implementation, which differs from React's:
+ * - React returns [isPending, startTransition] where isPending reflects actual transition state
+ * - Preact always returns [false, startTransition] since there's no real concurrent rendering
+ * - The isPending value is always false because Preact executes transitions synchronously
+ *
+ * @returns A tuple containing:
+ *   - isPending: Always false in Preact (unlike React where it indicates transition state)
+ *   - startTransition: Function to start a transition (executes callback synchronously)
+ *
+ * @public
+ */
+declare function useTransition(): [false, typeof startTransition];
+export { startTransition, useTransition };
--- a/packages/react/runtime/compat/index.js
+++ b/packages/react/runtime/compat/index.js
@@ -1,0 +1,15 @@
+// Copyright 2025 The Lynx Authors. All rights reserved.
+// Licensed under the Apache License Version 2.0 that can be found in the
+// LICENSE file in the root directory of this source tree.
+import { startTransition, useTransition } from 'preact/compat';
+
+import * as ReactLynx from '@lynx-js/react';
+
+export default {
+  ...ReactLynx,
+  startTransition,
+  useTransition,
+};
+
+export * from '@lynx-js/react';
+export { startTransition, useTransition };
--- a/packages/react/runtime/tsconfig.json
+++ b/packages/react/runtime/tsconfig.json
@@ -12,6 +12,7 @@
     "path": "../worklet-runtime/tsconfig.json",
   }],
   "include": [
+    "compat",
     "debug",
     "jsx-dev-runtime",
     "jsx-runtime",


packages/react/runtime/compat/index.js
```diff
@@ -0,0 +1,15 @@
...
+// LICENSE file in the root directory of this source tree.
+import { startTransition, useTransition } from 'preact/compat';
+
+import * as ReactLynx from '@lynx-js/react';
+
+export default {
+  ...ReactLynx,
+  startTransition,
+  useTransition,
+};
```
upupming: Does user those use TanStack Router need to config alias or import from `@lynx-js/react/compat` directly?

gaoachao: Nope. Just set alias: `react` → `@lynx-js/react/compat`

see: https://github.com/lynx-family/lynx-examples/pull/101/files#diff-edc78918892869228986650b3db4106a187f0bae72fcd6c4d686152c16e965d8R29

packages/react/package.json
```diff
@@ -14,6 +14,11 @@
       "lazy": "./runtime/lazy/react.js",
       "default": "./runtime/lib/index.js"
     },
+    "./compat": {
+      "types": "./runtime/compat/index.d.ts",
+      "lazy": "./runtime/compat/index.js",
```
colinaaa: Why are we using the real exports for `lazy`? This would make all the runtime being bundled into lazy bundle if `@lynx-js/react/compat` is used.

packages/react/package.json
```diff
@@ -14,6 +14,11 @@
...
+      "default": "./runtime/compat/index.js"
+    },
```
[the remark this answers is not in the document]
colinaaa: I do think this make sense. @gaoachao should we consider apply the suggestion?

packages/react/runtime/compat/index.d.ts
```diff
@@ -0,0 +1,33 @@
...
+ * - The isPending value is always false because Preact executes transitions synchronously
+ *
+ * @returns A tuple containing:
+ *   - isPending: Always false in Preact (unlike React where it indicates transition state)
+ *   - startTransition: Function to start a transition (executes callback synchronously)
+ *
+ * @public
+ */
+declare function useTransition(): [false, typeof startTransition];
+export { startTransition, useTransition };
```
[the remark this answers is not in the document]
colinaaa: @gaoachao I do think this is reasonable, too. We should provide types for `export default`.

> PS: `export * from '@lynx-js/react';` does not have `export default`

packages/react/runtime/compat/index.js
```diff
@@ -0,0 +1,15 @@
+// Copyright 2025 The Lynx Authors. All rights reserved.
+// Licensed under the Apache License Version 2.0 that can be found in the
+// LICENSE file in the root directory of this source tree.
+import { startTransition, useTransition } from 'preact/compat';
+
+import * as ReactLynx from '@lynx-js/react';
+
+export default {
+  ...ReactLynx,
```
colinaaa: Why are we spreading on the namespace import instead of the default import?

```suggestion
import ReactLynx from '@lynx-js/react';

export default {
  ...ReactLynx,
```

packages/react/runtime/__test__/lazy.test.js
```diff
@@ -22,6 +23,14 @@ describe('Lazy Exports', () => {
     );
   });
 
+  test('export APIs from "react/compat"', async () => {
+    expect(
+      new Set(Object.keys(ReactCompatExports)),
+    ).toStrictEqual(
+      new Set(['startTransition', 'useTransition']),
```
colinaaa: It doesn't seem correct to have only two exports. It should export exactly the same thing as `@lynx-js/react/compat` does.

packages/react/runtime/compat/index.d.ts
```diff
@@ -0,0 +1,40 @@
...
+ * @public
+ */
+declare function startTransition(cb: () => void): void;
+/**
+ * Returns a tuple with a pending state and a startTransition function
+ *
+ * Note: This is Preact's useTransition implementation, which differs from React's:
+ * - React returns [isPending, startTransition] where isPending reflects actual transition state
+ * - Preact always returns [false, startTransition] since there's no real concurrent rendering
+ * - The isPending value is always false because Preact executes transitions synchronously
```
colinaaa:
```suggestion
 * Note: This is Preact's `useTransition` implementation, which differs from React's:
 * - React returns `[isPending, startTransition]` where `isPending` reflects actual transition state
 * - Preact always returns `[false, startTransition]` since there's no real concurrent rendering
 * - The `isPending` value is always `false` because Preact executes transitions synchronously
```

packages/react/runtime/compat/index.d.ts
```diff
@@ -0,0 +1,40 @@
+// Copyright 2025 The Lynx Authors. All rights reserved.
+// Licensed under the Apache License Version 2.0 that can be found in the
+// LICENSE file in the root directory of this source tree.
+export * from '@lynx-js/react';
+
+/**
+ * Starts a transition update, marking state updates in the callback as non-urgent
+ *
+ * Note: This is Preact's startTransition implementation, which differs fundamentally from React's:
```
colinaaa:
```suggestion
 * Note: This is Preact's `startTransition` implementation, which differs fundamentally from React's:
```

packages/react/types/react.d.ts
```diff
@@ -208,3 +208,6 @@ declare module '@lynx-js/types' {
 }
 
 export type { JSX } from '../runtime/jsx-runtime/index.d.ts';
+
+declare const _default: typeof import('@lynx-js/react');
```
colinaaa: It seems like `@lynx-js/react` exactly points to `packages/react/types/react.d.ts`. It doesn't seem correct to use the types of itself.

packages/react/runtime/vitest.config.ts
```diff
@@ -60,6 +60,7 @@ export default defineConfig({
   ],
   resolve: {
     alias: {
+      '@lynx-js/react/compat': path.resolve(__dirname, './compat/index.js'),
```
colinaaa, line 63: We may also need an alias in `@lynx-js/react-alias-rsbuild-plugin`. Make sure it works with thd old version of `@lynx-js/react` which does not have `./compat` in `exports`

packages/react/runtime/compat/index.d.ts
colinaaa: I'm not sure if we should have "Preact" in our document or not. /cc @hzy @Huxpro wdyt
```suggestion
 * Note: This implementation differs fundamentally from React's:
```

packages/react/types/react.d.ts
gaoachao: ESLint will stage a protest if we remove this line. :)
image

gaoachao:
@colinaaa PTLA~

packages/react/types/react.d.ts
colinaaa: But this is not correct. We should not provide a wrong default export.

packages/rspeedy/plugin-react-alias/src/index.ts
```diff
@@ -61,13 +61,15 @@ export function pluginReactAlias(options: Options): RsbuildPlugin {
...
           reactLepusMainThread,
+          reactCompat,
         ] = await Promise.all([
           resolve('@lynx-js/react/jsx-runtime'),
           resolve('@lynx-js/react/lepus/jsx-runtime'),
           resolve('@lynx-js/react/jsx-dev-runtime'),
           resolve('@lynx-js/react/lepus/jsx-dev-runtime'),
           resolve('@lynx-js/react'),
           resolve('@lynx-js/react/lepus'),
+          resolve('@lynx-js/react/compat'),
```
colinaaa: This is not correct. This would failed to run on `@lynx-js/react` < 0.112.0

colinaaa: Either add a try-catch, or resolve on version condition.

colinaaa (changes requested):

packages/react/types/react.d.ts
gaoachao: OK. I use `eslint-disable`. see: https://github.com/lynx-family/lynx-stack/pull/1316/commits/f2560f6c00c2bc186ed21ee4567a901f7fc24b1a

packages/rspeedy/plugin-react-alias/src/index.ts
gaoachao: https://github.com/lynx-family/lynx-stack/pull/1316/commits/f2560f6c00c2bc186ed21ee4567a901f7fc24b1a

gaoachao:
@colinaaa Created compatibility shim for @lynx-js/react@<0.120.0. PTLA~

packages/rspeedy/plugin-react-alias/src/index.ts
```diff
@@ -1,10 +1,12 @@
 // Copyright 2024 The Lynx Authors. All rights reserved.
 // Licensed under the Apache License Version 2.0 that can be found in the
 // LICENSE file in the root directory of this source tree.
+import { readFileSync } from 'node:fs'
 import { createRequire } from 'node:module'
 import path from 'node:path'
 
 import type { RsbuildPlugin } from '@rsbuild/core'
+import semver from 'semver'
```
colinaaa: Avoid importing all the `semver` package. Use `semver/functions/gte.js` instead.

This would reduce the bundle size from 83.3 kB to 26.2 kB.

packages/rspeedy/plugin-react-alias/src/index.ts
```diff
@@ -40,6 +42,22 @@ export function pluginReactAlias(options: Options): RsbuildPlugin {
...
+      let version: string | undefined
+      try {
+        const pkg = JSON.parse(pkgContent) as { version?: string }
+        version = semver.coerce(pkg?.version ?? undefined)?.version
+      } catch {
+        throw new Error('Failed to parse @lynx-js/react version')
+      }
+
+      if (!version) {
+        throw new Error('version field not found in @lynx-js/react package')
```
colinaaa: It looks like this is unreachable.

packages/rspeedy/plugin-react-alias/src/index.ts
```diff
@@ -40,6 +42,22 @@ export function pluginReactAlias(options: Options): RsbuildPlugin {
...
         }),
       )
+
+      const pkgPath = path.join(reactLynxDir, 'package.json')
+      const pkgContent = readFileSync(pkgPath, 'utf-8')
+
+      let version: string | undefined
+      try {
+        const pkg = JSON.parse(pkgContent) as { version?: string }
+        version = semver.coerce(pkg?.version ?? undefined)?.version
```
colinaaa: I don't think we need `semver.coerce` explictly. The `gte` can receive the `string` as a param.

packages/rspeedy/plugin-react-alias/src/index.ts
```diff
@@ -40,6 +42,22 @@ export function pluginReactAlias(options: Options): RsbuildPlugin {
           paths: [rootPath ?? api.context.rootPath],
         }),
       )
+
+      const pkgPath = path.join(reactLynxDir, 'package.json')
+      const pkgContent = readFileSync(pkgPath, 'utf-8')
+
+      let version: string | undefined
+      try {
+        const pkg = JSON.parse(pkgContent) as { version?: string }
```
colinaaa: It would be simplier to use `require()` instead of `readFileSync` + `JSON.parse()`.

BTW, we already have the absolute path above using `require.resolve('@lynx-js/react/package.json)`. We may re-use that.

packages/rspeedy/plugin-react-alias/test/compat/legacy.test.ts
```diff
@@ -0,0 +1,47 @@
...
+        return JSON.stringify({ version: '0.111.0' })
+      } else {
+        return originalReadFileSync(filePath)
+      }
+    }),
+  }
+})
+
+describe('@lynx-js/react/compat - alias', () => {
+  test('alias with @lynx-js/react < 0.112.0', async () => {
```
colinaaa, line 34: Can we also add some tests for canary version?

E.g.:

- `@lynx-js/react@0.111.999-canary-20250728-1f7b2d07`
- `@lynx-js/react@0.112.0-canary-20250728-1f7b2d07`
- `@lynx-js/react@0.112.1-canary-20250730-1f7b2d07`

colinaaa, line 34: Can we add more test cases with canary version?

E.g.:

- `@lynx-js/react@0.111.999-canary-20250728-1f7b2d07`
- `@lynx-js/react@0.112.0-canary-20250728-1f7b2d07`
- `@lynx-js/react@0.112.999-canary-20250728-1f7b2d07`

packages/rspeedy/plugin-react-alias/src/index.ts
```diff
@@ -61,13 +79,17 @@ export function pluginReactAlias(options: Options): RsbuildPlugin {
...
           reactLepusMainThread,
+          reactCompat,
         ] = await Promise.all([
           resolve('@lynx-js/react/jsx-runtime'),
           resolve('@lynx-js/react/lepus/jsx-runtime'),
           resolve('@lynx-js/react/jsx-dev-runtime'),
           resolve('@lynx-js/react/lepus/jsx-dev-runtime'),
           resolve('@lynx-js/react'),
           resolve('@lynx-js/react/lepus'),
+          semver.gte(version, '0.112.0')
```
colinaaa: Considering the canary version, it would be better to use `gt(version, '0.111.9999')`

packages/rspeedy/plugin-react-alias/test/compat/legacy.test.ts
gaoachao: see: https://github.com/lynx-family/lynx-stack/pull/1316/commits/00bf2c40ed90ef28a2c09698f6057cb85faf1072

.changeset/all-planets-flow.md
```diff
@@ -0,0 +1,5 @@
+---
+"@lynx-js/rspeedy": patch
```
colinaaa: Why is this `@lynx-js/rspeedy`?

packages/react/runtime/compat/index.js
```diff
@@ -0,0 +1,16 @@
...
+import { startTransition, useTransition } from 'preact/compat';
+
+/* eslint-disable-next-line import/default */
+import ReactLynx from '@lynx-js/react';
+
+export default {
+  ...ReactLynx,
+  startTransition,
+  useTransition,
+};
```
colinaaa: The `...ReactLynx` may have side effect, which cause this module not tree shakeable. Maybe use `#__PURE__` with `Object.assign()` instead.


```suggestion
export default /*#__PURE__*/ Object.assign({}, ReactLynx, {
  startTransition,
  useTransition,
});
```

packages/rspeedy/plugin-react-alias/src/index.ts
```diff
@@ -35,11 +36,17 @@ export function pluginReactAlias(options: Options): RsbuildPlugin {
...
-      )
+      const reactLynxPkg = require.resolve('@lynx-js/react/package.json', {
+        paths: [rootPath ?? api.context.rootPath],
+      })
+      const reactLynxPkgContent = require(reactLynxPkg) as { version?: string }
+      const version = reactLynxPkgContent?.version
+
+      if (!version) {
+        throw new Error('version field not found in @lynx-js/react package')
+      }
```
colinaaa: Do we really have `@lynx-js/react/package.json` without `version`? I would prefer let it crash instead of try-catch everywhere.

packages/rspeedy/plugin-react-alias/test/compat/legacy.test.ts
```diff
@@ -0,0 +1,53 @@
...
+
+import { LAYERS } from '@lynx-js/react-webpack-plugin'
+
+vi.mock('node:module', async (importOriginal) => {
+  const original = await importOriginal()
+  const { createRequire: originalCreateRequire } =
+    original as typeof import('node:module')
+  return {
+    ...original as object,
+    createRequire: vi.fn().mockImplementation((path: string | URL) => {
```
colinaaa, line 17: Instead of mocking `createRequire` from `node:module`, can we set the `rootPath` option of `pluginReactAlias`? This could simplify the test.

gaoachao: `reactLynxDir` is used not only here but also in the subsequent `createLazyResolver`. If not resolved correctly, it will throw: `new Error(Failed to resolve ${request})`. I believe the best approach is to make it successfully resolve to the correct location in test files, while mocking only when reading `package.json.`

colinaaa (approved):

Huxpro:
Does this come with doc updates?

PR #1320, opened 2025-07-19
feat(tailwind-preset): improve transform transition compatibility

Dugyu:
## Summary

**New Features**

Improve transform transition compatibility with Lynx versions that do not support animating CSS variables. 

- Added solo transform utilities that avoid CSS variables: `solo-translate-x-*`, `solo-scale-*`, `solo-rotate-*` etc.

  Related to [Bug](https://github.com/lynx-family/lynx/commit/5741301d9b821d7ed5d71ba92af4528629952660). Transform animations using CSS variables are not supported in Lynx SDK versions before 3.4. This patch introduces standalone transform utilities as a workaround. These utilities are implemented without CSS variables using raw transform functions such as `translateX()`, `scale()` and `rotate()`. They are mutually exclusive and cannot be combined with normal transform utilities. 

- Enabled arbitrary values for `transform-[...]`: e.g. `transform-[translateX(20px)_rotate(10deg)]`, following Tailwind v4 behavior.

Fix `scale-*` utilities not supporting negative values. Now supports `-scale-x-100`, `-scale-y-50` as expected.

**Refactor**

Fix `autoBind` + `Bound` behavior in createPlugin to preserve generics and safely support destructuring methods from PluginAPI. Motivated by `transform` plugin DX needs, with minor adjustments across related plugins.

## Checklist

## Summary by CodeRabbit

* **New Features**
  * Added standalone transform utility classes for translate, scale, rotate, and skew (e.g., `solo-translate-x-*`, `solo-scale-*`, `solo-rotate-*`, `solo-skew-*`), improving compatibility and flexibility.
  * Enabled support for arbitrary transform values (e.g., `transform-[translateX(20px)_rotate(10deg)]`), aligning behavior with Tailwind CSS v4.

* **Bug Fixes**
  * Scale utilities now correctly support negative values (e.g., `-scale-x-100`, `-scale-y-50`).

* **Refactor**
  * Streamlined internal plugin and utility code for improved maintainability and type safety.
  * Refactored binding helper to improve function context safety and destructuring support.
  * Simplified type definitions for bound function utilities.

* **Tests**
  * Added comprehensive test suites for new and existing transform utilities to ensure correct behavior and input validation.


packages/third-party/tailwind-preset/src/__tests__/plugin.test.ts
```diff
@@ -15,11 +15,15 @@ describe('Lynx plugin coverage sanity check', () => {
       const addUtilities = vi.mocked(api.addUtilities);
       const addComponents = vi.mocked(api.addComponents);
       const addBase = vi.mocked(api.addBase);
+      const matchVariant = vi.mocked(api.matchVariant);
+      const matchComponent = vi.mocked(api.matchComponents);
```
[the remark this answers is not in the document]
Dugyu: Good catch!

packages/third-party/tailwind-preset/src/__tests__/plugins/soloTranslate.test.ts
```diff
@@ -0,0 +1,57 @@
...
+
+    expect(utils['solo-translate-x']?.(0)).toBeNull();
+
+    expect(utils['solo-translate-x']?.('1rem')).toEqual({
+      transform: `translateX(1rem)`,
+    });
+
+    expect(utils['solo-translate-y']?.('1rem')).toEqual({
+      transform: `translateY(1rem)`,
+    });
```
upupming, line 47: What will happen if user uses `solo-translate-x` and `solo-translate-y` simultaneously? In web, the latter style will override the previous so only `translateY` will take effects finally.

packages/third-party/tailwind-preset/src/__tests__/helper.test.ts
```diff
@@ -128,20 +128,43 @@ describe('helpers.ts', () => {
...
+  it('preserves callability for plain functions', () => {
+    expect(bound.echo('yo')).toBe('yo');
+  });
+
+  it('fails without autoBind for `this`-dependent methods', () => {
+    // eslint-disable-next-line @typescript-eslint/unbound-method
+    const { greet } = impl;
+    // Intentionally calling an unbound method to confirm that `this` is lost
+    expect(() => (greet as (this: unknown) => string)()).toThrow(TypeError);
+  });
```
upupming: I think it still will fail even it is bound with `autoBind` because `this` is `undefined`. So why is `autoBind` necessary?

packages/third-party/tailwind-preset/src/plugins/lynx/scale.ts
```diff
@@ -28,4 +28,4 @@ export const scale: Plugin = createUtilityPlugin('scale', [
       ]],
     ],
   ],
-]);
+], { supportsNegativeValues: true });
```
upupming, line 31: Should we extract this and make it as a `defaultOptions` and reuse it?

packages/third-party/tailwind-preset/src/plugins/lynx/soloRotate.ts
```diff
@@ -0,0 +1,24 @@
...
+    {
+      'solo-rotate': createFunctionCallUtility('transform', 'rotate'),
+      'solo-rotate-x': createFunctionCallUtility('transform', 'rotateX'),
+      'solo-rotate-y': createFunctionCallUtility('transform', 'rotateY'),
+      'solo-rotate-z': createFunctionCallUtility('transform', 'rotateZ'),
+    },
+    {
+      supportsNegativeValues: true,
+      values: theme('rotate'),
+    },
```
upupming, line 22: What is the purpose of this utility here, can we test the behavior of it?

upupming (changes requested):

packages/third-party/tailwind-preset/src/plugins/lynx/soloRotate.ts
Dugyu: `createFunctionCallUtility` helps create CSSRuleObject in the following format
```
{
    [property]: '${fn}(${value})',
}
```
This function will be used across several utility plugins in the following PRs, such as filter utilities: `blur-*`, `grayscale-*`

packages/third-party/tailwind-preset/src/plugins/lynx/scale.ts
Dugyu: No. Only a few length & percentage utilities use this option. We follow Tailwind's default behavior and does not make it default to `true`.

packages/third-party/tailwind-preset/src/__tests__/plugins/soloTranslate.test.ts
Dugyu: 1. VSCode Tailwind CSS IntelliSense will give a warning: 'solo-translate-y-*' applies the same CSS properties as 'solo-translate-x-*'.[cssConflict]
2. If not using conflict resolving libraries such as `tailwind-merge`, it is the user's responsibility to ensure no conflicting utility classes are applied on the same element. This follows the same behavior with other Tailwind utilities.
3. A common misconception is that the order of utility in the class attribute determines which one wins—that later utilities take precedence over earlier ones. In reality, every utility in the class list is applied, as selectors are pre-sorted by specificity at build time. (Doesn't matter whether it is on Web or Lynx). In Tailwind's property sorting list, `translateY()` will always appear after `translateX()`, so even if the user writes `className = {'solo-translate-y-10 solo-translate-x-10'}`, only `solo-translate-y-10` will take effect.

packages/third-party/tailwind-preset/src/__tests__/helper.test.ts
Dugyu: I should change `fails without autoBind for this-dependent methods` to `fails with autoBind for this-dependent methods`.
`autoBind` is not intended for `this-dependent` methods

Dugyu: `autoBind` is needed for the last scenario (together with `Bound` type)

```
const bound = autoBind(impl) as Bound<API>;

it('can be destructured without losing context', () => {
    const { echo } = bound;
    expect(echo('test')).toBe('test');
});
```

Dugyu: Added another commit to fix the confusing test logic.
```
 it('autoBind fails with `this`-dependent methods', () => {
    const { greet } = bound;
    // Intentionally calling an unbound method to confirm that `this` is lost
    expect(() => (greet as (this: unknown) => string)()).toThrow(TypeError);
  });

  it('`this`-independent methods can be destructured without losing context', () => {
    const { echo } = bound;
    expect(echo('test')).toBe('test');
  });
 ```
 
 Ensures `autoBind` returns `Bound<T>` directly:
 
 ```
 function autoBind<T extends object>(obj: T): Bound<T> {
  return Object.fromEntries(
    Object.entries(obj).map(([k, v]) =>
      isFunction(v) ? [k, v.bind(undefined)] : [k, v]
    ),
  ) as Bound<T>;
}
 ```

upupming (approved):

PR #1345, opened 2025-07-24
feat(tailwind-preset): add filter utilities and update related infrastructure

Dugyu:
## Summary
**Feature**

Add filter utilities: `blur-*`, `grayscale-*`.

- On Lynx, filter functions are mutually exclusive.
- Replaces Tailwind's variable-based filter system.

**Refactor**

Improve `createFunctionCallUtility` robustness.

- Accepts `unknown`, trims strings, adds `emptyFallback`, returns `null` on invalid input

**Docs**

Update plugin registration guide to reflect category-based registries.

## Summary by CodeRabbit

* **New Features**
  * Added new utility classes for blur and grayscale filters, allowing users to apply blur and grayscale effects using Tailwind-style classes.
  * Introduced a utility class for removing all filters.
  * Extended the theme configuration with predefined grayscale levels.

* **Documentation**
  * Improved and clarified documentation on how to add or remove utilities and plugins, including guidance on plugin registration and ordering.

* **Tests**
  * Added comprehensive tests for the new blur and grayscale utilities, ensuring correct behavior for various input scenarios.
  * Enhanced utility function tests for better input handling and fallback behavior.
  * Added tests to verify safe handling of missing or empty modifier maps.

## Checklist

### DIFF
--- a/packages/third-party/tailwind-preset/src/__tests__/plugins/grayscale.test.ts
+++ b/packages/third-party/tailwind-preset/src/__tests__/plugins/grayscale.test.ts
@@ -1,0 +1,55 @@
+// Copyright 2025 The Lynx Authors. All rights reserved.
+// Licensed under the Apache License Version 2.0 that can be found in the
+// LICENSE file in the root directory of this source tree.
+import { describe, expect, it, vi } from 'vitest';
+
+import { grayscale } from '../../plugins/lynx/grayscale.js';
+import { runPlugin } from '../utils/run-plugin.js';
+
+describe('grayscale plugin', () => {
+  it('covers all branches', () => {
+    const { api } = runPlugin(grayscale, {
+      theme: {
+        grayscale: {
+          none: '',
+          0: '0',
+          DEFAULT: '100%',
+        },
+      },
+    });
+
+    const matchUtilities = vi.mocked(api.matchUtilities);
+
+    type UtilityFn = (value: unknown) => Record<string, string> | null;
+
+    const utils = matchUtilities.mock.calls.reduce<Record<string, UtilityFn>>(
+      (acc, call) => {
+        const group = call[0] as Record<string, UtilityFn>;
+        for (const key in group) {
+          if (group[key]) {
+            acc[key] = group[key]!;
+          }
+        }
+        return acc;
+      },
+      {},
+    );
+
+    // Invalid input
+    expect(utils['grayscale']?.(null)).toBeNull();
+    expect(utils['grayscale']?.(undefined)).toBeNull();
+    expect(utils['grayscale']?.(false)).toBeNull();
+    expect(utils['grayscale']?.(1)).toBeNull();
+
+    // Special case: empty string / only whitespace
+    expect(utils['grayscale']?.('')).toEqual({ filter: 'none' });
+    expect(utils['grayscale']?.('   ')).toEqual({ filter: 'none' });
+
+    // Valid theme values
+    expect(utils['grayscale']?.('0')).toEqual({ filter: 'grayscale(0)' });
+    expect(utils['grayscale']?.('100%')).toEqual({ filter: 'grayscale(100%)' });
+
+    // Arbitrary value
+    expect(utils['grayscale']?.('60%')).toEqual({ filter: 'grayscale(60%)' });
+  });
+});
--- a/packages/third-party/tailwind-preset/src/plugins/lynx/soloTranslate.ts
+++ b/packages/third-party/tailwind-preset/src/plugins/lynx/soloTranslate.ts
@@ -17,7 +17,7 @@ export const soloTranslate: Plugin = createPlugin(
           'transform',
           'translateY',
         ),
-        'solo-translate-z': (value: string) => {
+        'solo-translate-z': (value: unknown) => {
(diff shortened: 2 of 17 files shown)


packages/third-party/tailwind-preset/src/theme.ts
```diff
@@ -121,4 +121,12 @@ export const lynxTheme: Partial<LynxThemeConfig> = {
...
     'in-out': 'cubic-bezier(0.4, 0, 0.2, 1)',
   },
+  grayscale: {
+    none: '',
+    0: '0',
+    25: '25%',
+    50: '50%',
+    75: '75%',
+    DEFAULT: '100%',
+  },
```
upupming: I see the values here is totallly same as the default value of tailwindcss, instead of maintain a full copy, can we extend the values from the default theme of tailwindcss and only define the difference?

packages/third-party/tailwind-preset/src/types/theme-types.ts
```diff
@@ -11,6 +11,7 @@ type LynxThemeOverrides = Pick<
   ThemeConfig,
   | 'aspectRatio'
   | 'boxShadow'
+  | 'grayscale'
```
upupming, line 14: It seems this `'grayscale'` type is not actually used.

packages/third-party/tailwind-preset/src/plugins/lynx/soloTranslate.ts
```diff
@@ -17,7 +17,7 @@ export const soloTranslate: Plugin = createPlugin(
           'transform',
           'translateY',
         ),
-        'solo-translate-z': (value: string) => {
+        'solo-translate-z': (value: unknown) => {
```
upupming, line 20: In tailwindcss, `value` will always be of type `string`

https://github.com/tailwindlabs/tailwindcss/blob/6314c8fb36cfd9011aa4552017a50c2124655452/packages/tailwindcss/src/compat/plugin-api.ts#L60

packages/third-party/tailwind-preset/src/__tests__/plugins/grayscale.test.ts
```diff
@@ -0,0 +1,55 @@
...
+    );
+
+    // Invalid input
+    expect(utils['grayscale']?.(null)).toBeNull();
+    expect(utils['grayscale']?.(undefined)).toBeNull();
+    expect(utils['grayscale']?.(false)).toBeNull();
+    expect(utils['grayscale']?.(1)).toBeNull();
+
+    // Special case: empty string / only whitespace
+    expect(utils['grayscale']?.('')).toEqual({ filter: 'none' });
```
upupming, line 45: Does this matches className `grayscale-`, I am not sure if supporting this is necessary because it is not supported by tailwindcss by default.

packages/third-party/tailwind-preset/src/__tests__/plugins/grayscale.test.ts
```diff
@@ -0,0 +1,55 @@
...
+    expect(utils['grayscale']?.('   ')).toEqual({ filter: 'none' });
```
upupming, line 46: Same for this case, I think user even cannot construct a className like this.

upupming (changes requested):

packages/third-party/tailwind-preset/src/plugins/lynx/soloTranslate.ts
Dugyu: In Tailwindcss v3, `value` is not always typed as `string`, should be `unknown` here:
  https://github.com/tailwindlabs/tailwindcss/blob/v3.4.17/types/config.d.ts#L280

packages/third-party/tailwind-preset/src/__tests__/plugins/grayscale.test.ts
Dugyu: User will write `grayscale-none`, `none` corresponds to empty string `''`

packages/third-party/tailwind-preset/src/__tests__/plugins/grayscale.test.ts
Dugyu: They may define a key-value pair `none: '    '` in theme config, this is used in Tailwind's `blur` plugin. Users are using `grayscale-none` in their class attribute.  We follow the same behavior:
```
blur: {
      0: '0',
      none: '',
      sm: '4px',
      DEFAULT: '8px',
      md: '12px',
      lg: '16px',
      xl: '24px',
      '2xl': '40px',
      '3xl': '64px',
},
```

packages/third-party/tailwind-preset/src/types/theme-types.ts
Dugyu: This is reserved for future PRs, `LynxThemeConfig` will be stricter (only allow theme keys supported plugins).
For now, just document what we have overriden.

packages/third-party/tailwind-preset/src/theme.ts
Dugyu: Extended in theme config.

upupming (approved):

PR #1370, opened 2025-07-29
feat: cache events until background thread ready

upupming:
## Summary by CodeRabbit

* **New Features**
  * Introduced a cache-events plugin that buffers Lynx native events during chunk splitting and replays them once the BTS chunk loads.
  * Added runtime globals to expose cache state and a setup list; plugin supports a setupList transformer for custom caching.
* **Refactor**
  * Removed forced “jsReady” override; firstScreenSyncTiming now uses configured value.
* **Documentation**
  * Added plugin README.
* **Tests**
  * Added cases for chunk-splitting, non-splitting, and setupList transformation.
* **Chores**
  * Added new package, dependencies, build/test configs, and project references.

## Checklist

### DIFF
--- a/.changeset/cold-boats-share.md
+++ b/.changeset/cold-boats-share.md
@@ -1,6 +1,5 @@
 ---
 "@lynx-js/webpack-runtime-globals": patch
-"@lynx-js/cache-events-webpack-plugin": patch
 ---
 
 Add `lynxCacheEventsSetupList` and `lynxCacheEvents` to RuntimeGlobals. It will be used to cache Lynx native events until the BTS chunk is fully loaded, and replay them when the BTS chunk is ready.
--- a/.changeset/chatty-monkeys-enter.md
+++ b/.changeset/chatty-monkeys-enter.md
@@ -1,0 +1,5 @@
+---
+"@lynx-js/cache-events-webpack-plugin": patch
+---
+
+Add new `LynxCacheEventsPlugin`, which will cache Lynx native events until the BTS chunk is fully loaded, and replay them when the BTS chunk is ready.
--- a/packages/webpack/cache-events-webpack-plugin/package.json
+++ b/packages/webpack/cache-events-webpack-plugin/package.json
@@ -1,7 +1,7 @@
 {
   "name": "@lynx-js/cache-events-webpack-plugin",
   "version": "0.0.0",
-  "description": "cache Lynx native events until the BTS chunk is fully loaded, and replay them when the BTS chunk is ready.",
+  "description": "Cache Lynx native events until the BTS chunk is fully loaded, and replay them when the BTS chunk is ready.",
   "keywords": [
     "webpack",
     "Lynx"


packages/rspeedy/plugin-react/src/entry.ts
```diff
@@ -236,7 +234,10 @@ export function applyEntry(
         }])
         .end()
         .plugin(`${LynxEncodePlugin.name}`)
-        .use(LynxEncodePlugin, [{ inlineScripts }])
+        .use(LynxEncodePlugin, [{
+          inlineScripts,
+          enableEventsCacheManifest: enableChunkSplitting,
+        }])
```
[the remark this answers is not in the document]
upupming: The runtime behavior of `events-cache.js` is manually tested, and runtime testing cases are hard to constuct.

.changeset/brave-melons-add.md
```diff
@@ -0,0 +1,9 @@
+---
+"@lynx-js/runtime-wrapper-webpack-plugin": patch
```
luhc228: How about minor?

packages/webpack/template-webpack-plugin/src/LynxEncodePlugin.ts
```diff
@@ -145,6 +157,39 @@ export class LynxEncodePluginImpl {
...
+          for (const entrypoint of compilation.entrypoints.values()) {
+            for (const file of entrypoint.getEntrypointChunk().files) {
+              if (file.endsWith('.js') && manifest[file]) {
+                backgroundScriptName = file;
+                break;
+              }
+            }
+            if (backgroundScriptName) break;
+          }
+
```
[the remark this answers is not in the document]
upupming: We are not choosing first js, but we can ensure the entry js exists in manifest must be the background.js we needed

.changeset/brave-melons-add.md
```diff
@@ -0,0 +1,7 @@
+---
+"@lynx-js/template-webpack-plugin": minor
+---
+
+**Breaking Changes**:
+
+Add `enableEventsCacheManifest` option to `LynxEncodePlugin`, which allows the plugin to generate a manifest file for cached events.
```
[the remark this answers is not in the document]
upupming: Our package version is `0.x.x`, so minor is for breaking change.

.changeset/two-moose-help.md
```diff
@@ -0,0 +1,7 @@
+---
+"@lynx-js/runtime-wrapper-webpack-plugin": minor
```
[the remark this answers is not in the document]
upupming: same as above

.changeset/brave-melons-add.md
```diff
@@ -0,0 +1,7 @@
+---
+"@lynx-js/template-webpack-plugin": minor
```
colinaaa: We need to update the `peerDependencies` of the webpack plugins.

https://github.com/lynx-family/lynx-stack/pull/1370#issuecomment-3132422249

> @lynx-js/css-extract-webpack-plugin	Major
> @lynx-js/react-webpack-plugin	Major

packages/webpack/runtime-wrapper-webpack-plugin/src/RuntimeWrapperWebpackPlugin.ts
```diff
@@ -151,7 +162,21 @@ class RuntimeWrapperWebpackPluginImpl {
...
+    compiler.hooks.environment.tap(
+      this.name,
+      () => {
+        if (
+          compiler.options.optimization.splitChunks
+          && compiler.options.output.iife
+        ) {
+          compiler.options.output.iife = false;
+          console.warn(
+            '`iife` is changed to `false` because it is only supported in `all-in-one` chunkSplit strategy, please set `performance.chunkSplit.strategy` to `all-in-one` to use `iife`.',
```
colinaaa: I don't think this make sense.

1. Mixing Rspeedy configuration (`performance.chunkSplit.strategy`) and Rspack configuration (`iife`) in the error message would be confusing.

2. I can't see why this is necessary. The IIFE is generated by the plugin now. Even if there are problems, you may use a webpack `RuntimeModule` instead of `BannerPlugin` to add the runtime.

packages/webpack/runtime-wrapper-webpack-plugin/src/RuntimeWrapperWebpackPlugin.ts
```diff
@@ -288,6 +321,26 @@ ${iife ? '' : '})();'}
...
+) => `
+${
+  iife ? '' : `
+
+// __webpack_exports__ will be a Promise when chunk splitting is enabled.
+// in this case, we should wait for the Promise to resolve before
+// sending any native events to BTS
+// We export the __webpack_exports__ here and the exports will be used
+// by a cache layer to wait for the Promise to resolve
+module.exports = ${runtimeGlobals.exports};
```
colinaaa: What if there is already an export? E.g.: having `export default` in the source code.

packages/webpack/runtime-wrapper-webpack-plugin/src/RuntimeWrapperWebpackPlugin.ts
colinaaa, line 1: The change to `RuntimeWrapperWebpackPlugin.ts` does not follow the idea of this plugin.

This plugin is designed to wrap the existing Webpack/Rspack chunks into chunks that can be loaded by `lynx.requireModule`.

- It should **always** apply to all chunks that is running on background thread.
- It should **never** change the behavior of the existing chunk.
- It should be able to use in both Rspeedy and other tools like Rslib to building a separate chunk that is loadable by Lynx.

And I don't think the newly added `backgroundChunks` option meet the current design.
Also, it looks unnecessary to me (see the following comments)

packages/webpack/template-webpack-plugin/static/events-cache.js
```diff
@@ -0,0 +1,157 @@
...
+// Licensed under the Apache License Version 2.0 that can be found in the
+// LICENSE file in the root directory of this source tree.
+(function() {
+  'use strict';
+  var g = (new Function('return this;'))();
+  function __init_card_bundle__(lynxCoreInject) {
+    g.__bundle__holder = undefined;
+    var tt = lynxCoreInject.tt;
+    tt.define('events-cache.js', function() {
+      const btsExports = tt.require('background.js');
```
colinaaa: We don't always have `background.js`. In production, we have `background.[contenthash:8].js` by default. And this could change in the future.

packages/webpack/template-webpack-plugin/static/events-cache.js
```diff
@@ -0,0 +1,157 @@
...
+          };
+          emitter.addListener(listenerKey, listener);
+          cleanupTasks.push(() => {
+            emitter.removeListener(listenerKey, listener);
+          });
+        });
+      }
+
+      btsExports
+        .then(() => {
```
colinaaa: Instead of using `.then` on the resolved promise, can we replay the cached event within the `setter`?

Here is the pseudo code:

```js
Object.defineProperty(tt, property, {
  get() {
    return mockedHandler;
  },
  set(value) {
    replayEvents();
    Object.defineProperty(tt, property, { value });
  },
})
```

packages/webpack/template-webpack-plugin/static/events-cache.js
```diff
@@ -0,0 +1,157 @@
...
+          'events-cache.js: background.js exports is not a promise, skip events cache',
+        );
+        return;
+      }
+
+      let btsResolved = false;
+
+      let cachedActions = [];
+
+      // ensure tt._appInstance is initialized to avoid TTApp this._appInstance.onFirstScreen() fail
```
colinaaa: Can we avoid these in the output of a ReactLynx project?

packages/webpack/template-webpack-plugin/src/LynxEncodePlugin.ts
```diff
@@ -145,6 +157,46 @@ export class LynxEncodePluginImpl {
...
+              compilation.warnings.push(
+                new compiler.webpack.WebpackError(
+                  `LynxEncodePlugin: ${enableEventsCacheManifestScriptName} is not injected because no background script found, manifest files: ${
+                    Object.keys(manifest).join(',')
+                  }`,
+                ),
+              );
+            }
+          } else {
+            inlinedManifest[enableEventsCacheManifestScriptName] =
```
colinaaa: Instead of modifying manifest, can we use webpack `RuntimeModule` instead?

It seems that the `events-cache.js` is not minified and cannot be customized or overrided.

packages/webpack/template-webpack-plugin/src/LynxEncodePlugin.ts
```diff
@@ -145,6 +157,46 @@ export class LynxEncodePluginImpl {
...
+              enableEventsCacheManifestScript.replace(
+                'background.js',
+                path.posix.basename(backgroundScriptName),
```
colinaaa: This could fail when there are comments that contain `background.js` :)

packages/webpack/template-webpack-plugin/static/events-cache.js
```diff
@@ -0,0 +1,157 @@
...
+    g.__bundle__holder = undefined;
+    var tt = lynxCoreInject.tt;
+    tt.define('events-cache.js', function() {
+      const btsExports = tt.require('background.js');
+      const maybePromise = typeof btsExports === 'object'
+        && btsExports !== null
+        && typeof btsExports.then === 'function';
+
+      if (!maybePromise) {
+        console.warn(
```
colinaaa: I don't think we have `console` on globalThis all the time.

luhc228 (dismissed):

packages/webpack/runtime-wrapper-webpack-plugin/src/RuntimeWrapperWebpackPlugin.ts
upupming: `export default` will not assign to `module.exports` currently, there are 2 cases now:

1. User directly export in their entry:

    ```
    "(react:background)/./src/index.tsx": (function (module, __webpack_exports__, __webpack_require__) {
    __webpack_require__.r(__webpack_exports__);
    __webpack_require__.d(__webpack_exports__, {
      "default": () => (__WEBPACK_DEFAULT_EXPORT__)
    });
    
    var __webpack_exports__ = __webpack_require__("(react:background)/./src/index.tsx");
    ```
    
    But `__webpack_exports__` is never used.

2. a lazy bundle `Foo.jsx` will produce output like this:

    ```
    exports.ids = ["./Foo.jsx-react__background"];
    exports.modules = {
    // ...
    
    "(react:background)/./src/Foo.tsx": (function (module, __webpack_exports__, __webpack_require__) {
    __webpack_require__.r(__webpack_exports__);
    __webpack_require__.d(__webpack_exports__, {
      "default": () => (Foo)
    });
    
    // ...
    }
    ```

packages/webpack/template-webpack-plugin/static/events-cache.js
upupming: I think replay in setter is not 100% safe. For example:

1. user may set `tt.updateCardData` in their code to intercept the one injected in [`injectTT`](https://github.com/lynx-family/lynx-stack/blob/19021abafe4e9cb78e5eadeeca39b92f269b02a5/packages/react/runtime/src/lynx/tt.ts#L27). If we do this we will replay to first `updateCardData` injected in `injectTT`, but not the one user defined.
2. There is also an special case for `lynx.performance`. User can add `lynx.performance.createObserver` in the end of their entry file. We have to wait the whole JS file loaded to replay the events to `lynx.performance.createObserver`'s listeners.

packages/webpack/template-webpack-plugin/src/LynxCacheEventsSetupListRuntimeModule.ts
```diff
@@ -0,0 +1,123 @@
...
+    extends webpack.RuntimeModule
+  {
+    constructor() {
+      super('Lynx cache events setup list', webpack.RuntimeModule.STAGE_NORMAL);
+    }
+
+    override generate(): string {
+      return `// lynx cache events setup list
+${LynxRuntimeGlobals.lynxCacheEvents} = {};
+${LynxRuntimeGlobals.lynxCacheEventsSetupList} = [
```
upupming: @colinaaa @luhc228 Now we maintaintain `LynxRuntimeGlobals.lynxCacheEventsSetupList`. It cantians 2 setup function by default, and different DSL can modify/push on it.

packages/webpack/template-webpack-plugin/src/LynxCacheEventsRuntimeModule.ts
```diff
@@ -0,0 +1,46 @@
...
+import { RuntimeGlobals as LynxRuntimeGlobals } from '@lynx-js/webpack-runtime-globals';
+
+type LynxCacheEventsRuntimeModule = new() => RuntimeModule;
+
+export function createLynxCacheEventsRuntimeModule(
+  webpack: typeof import('webpack'),
+): LynxCacheEventsRuntimeModule {
+  return class LynxCacheEventsRuntimeModule extends webpack.RuntimeModule {
+    constructor() {
+      super('Lynx cache events', webpack.RuntimeModule.STAGE_TRIGGER);
```
upupming: @colinaaa `webpack.RuntimeModule.STAGE_TRIGGER` ensures this runtime module is the last one in BTS:

image

packages/webpack/template-webpack-plugin/src/LynxCacheEventsSetupListRuntimeModule.ts
```diff
@@ -0,0 +1,123 @@
...
+  {
```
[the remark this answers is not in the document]
upupming: `loaded` and `cachedActions` is guaranteed by the `LynxCacheEventsRuntimeModule` runtime module. This runtime module is only responsible for the setup list initialization.

packages/webpack/template-webpack-plugin/src/LynxCacheEventsSetupListRuntimeModule.ts
```diff
@@ -0,0 +1,123 @@
...
+        while (cleanupTasks.length > 0) {
+          cleanupTasks.shift()();
+        }
+        // replay ${LynxRuntimeGlobals.lynxCacheEvents}.cachedActions
+        ${LynxRuntimeGlobals.lynxCacheEvents}.cachedActions.forEach(action => {
+          if (action.type === 'performanceEvent') {
+            const listenerKey = lynxPerformanceListenerKeys[action.data.type];
+            if (listenerKey && emitter) {
+              emitter.emit(listenerKey, action.data.args);
+            }
```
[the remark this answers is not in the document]
upupming: This API is designed to pass array as arg. (GlobalEventEmitter)

luhc228 (approved):

packages/webpack/webpack-runtime-globals/src/RuntimeGlobals.ts
```diff
@@ -22,4 +22,13 @@ export const RuntimeGlobals = {
    * A function to process the eval result of lazy bundle.
    */
   lynxProcessEvalResult: 'globalThis.processEvalResult',
+
+  /**
+   * A list of functions to setup the cache layer.
+   */
+  lynxCacheEventsSetupList: '__webpack_require__.__cache_events__.setupList',
```
colinaaa: Can we make this shorter, since the `__cache_events__` cannot be minified.

Also, maybe adding `lynx_` prefix to avoid conflict with future Webpack RuntimeGlobals.

packages/webpack/chunk-loading-webpack-plugin/src/StartupChunkDependenciesPlugin.ts
```diff
@@ -62,6 +64,20 @@ export class StartupChunkDependenciesPlugin {
           : globalChunkLoading;
         return chunkLoading === this.chunkLoading;
       };
+      const handler = (chunk: Chunk, runtimeRequirements: Set<string>) => {
```
colinaaa: I don't think it make sense to have this in the `StartupChunkDependenciesPlugin.ts`:

1. This is copied from webpack. We should keep it as it.
2. This should be placed within the same plugin that injecting the RuntimeModule. This could avoid making the two plugins coupled.

packages/webpack/template-webpack-plugin/src/LynxTemplatePlugin.ts
```diff
@@ -570,6 +576,45 @@ class LynxTemplatePluginImpl {
         );
       });
 
+      const LynxCacheEventsSetupListRuntimeModule =
```
colinaaa: Instead of in `LynxTemplatePlugin`, can we make a new standalone webpack plugin? Then we could test only the plugin itself.

You can apply the plugin inside the `LynxTemplatePlugin` or in react-rsbuild-plugin.

packages/webpack/template-webpack-plugin/test/cases/runtime-module/cache-events/index.js
```diff
@@ -0,0 +1,18 @@
+/// <reference types="vitest/globals" />
+
+import { add } from './lib-common.js';
+
+it('should have `__webpack_require__.__cache_events__`', () => {
+  expect(add(1, 2)).toBe(3);
+  expect(__webpack_require__['__cache_events__']).toBeTruthy();
```
colinaaa: Can we also add a test to see if `__webpack_require__['__cache_events__']` does not exist when no chunk split is set.

.changeset/rare-worms-vanish.md
```diff
@@ -0,0 +1,9 @@
+---
+"@lynx-js/react-rsbuild-plugin": minor
+---
+
+**Breaking Changes**:
```
colinaaa: Does this need to be a **BREAKING CHANGE**?

We didn't make it a **BREAKING CHANGE** before when changing to `'jsReady'` in https://github.com/lynx-family/lynx-stack/pull/883.

https://github.com/lynx-family/lynx-stack/pull/883/files#diff-e6cc62bac385f8ba6331777e7201c202a86d38d51b26934be11a3aed70907abdR212

And it turns out that it works fine until now. So I don't think it's necessary to be a breaking change when revert back.

.changeset/cold-boats-share.md
```diff
@@ -0,0 +1,7 @@
+---
+"@lynx-js/chunk-loading-webpack-plugin": patch
```
colinaaa: Please remember to remove this when reverting the change in chunk-loading-webpack-plugin.

.changeset/rare-worms-vanish.md
upupming: I was thinking that people using old version of `@lynx-js/template-webpack-plugin` and `@lynx-js/webpack-runtime-globals`.

If they update only `@lynx-js/react-rsbuild-plugin` they may break because they rely on 'jsReady' to work properly.

What do you think?

upupming: Let me change to patch first, I think user can still config 'jsReady' manually if the above case happens.

packages/webpack/cache-events-webpack-plugin/src/LynxCacheEventsRuntimeModule.ts
```diff
@@ -0,0 +1,46 @@
...
+
+    override generate(): string {
+      const { RuntimeGlobals } = this.compilation!.compiler.webpack;
+
+      return `// lynx cache events
+const cleanupList = []
+for (const { setup } of ${LynxRuntimeGlobals.lynxCacheEventsSetupList}) {
+  cleanupList.push(setup());
+}
+
```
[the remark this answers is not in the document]
upupming: We want the developer to ensure cleanup is correct type on compiler time. We want best performance and a simple runtime module.

colinaaa (approved):
👍

Issue #1371, opened 2025-07-29
[Bug]: Using `key` with multiple children cause wrong render result


colinaaa:
### System Info

  System:
    OS: macOS 26.0
    CPU: (12) arm64 Apple M2 Max
    Memory: 2.77 GB / 96.00 GB
    Shell: 5.9 - /bin/zsh
  npmPackages:
    @lynx-js/qrcode-rsbuild-plugin: workspace:* => 0.4.0 
    @lynx-js/react: workspace:* => 0.112.0 
    @lynx-js/react-rsbuild-plugin: workspace:* => 0.10.9 
    @lynx-js/rspeedy: workspace:* => 0.10.3 
    @lynx-js/types: 3.3.0 => 3.3.0 

### Details

When using a `key` with an expression, the rendering result could sometime incorrect.

This should be a regression of https://github.com/lynx-family/lynx-stack/pull/547.

In the error case, we would have multiple `__DynamicPartChildren` generated in the `slot` of snapshot, which is not supported in runtime.

```js
const __snapshot_835da_2d444_1 = /*#__PURE__*/ (__webpack_require__("(react:background)/../../packages/react/runtime/lib/internal.js")/* .createSnapshot */.createSnapshot)("__snapshot_835da_2d444_1", function() {
    const pageId = (__webpack_require__("(react:background)/../../packages/react/runtime/lib/internal.js")/* .__pageId */.__pageId);
    const el = __CreateView(pageId);
    const el1 = __CreateWrapperElement(pageId);
    __AppendElement(el, el1);
    const el2 = __CreateView(pageId);
    __AppendElement(el1, el2);
    const el3 = __CreateView(pageId);
    __AppendElement(el2, el3);
    const el4 = __CreateView(pageId);
    __AppendElement(el2, el4);
    return [
        el,
        el1,
        el2,
        el3,
        el4
    ];
}, [
    (snapshot, index, oldValue)=>(__webpack_require__("(react:background)/../../packages/react/runtime/lib/internal.js")/* .updateEvent */.updateEvent)(snapshot, index, oldValue, 0, "bindEvent", "tap", '')
], [
    [
        (__webpack_require__("(react:background)/../../packages/react/runtime/lib/internal.js")/* .__DynamicPartChildren */.__DynamicPartChildren),
        3
    ],
    [
        (__webpack_require__("(react:background)/../../packages/react/runtime/lib/internal.js")/* .__DynamicPartChildren */.__DynamicPartChildren),
[...]

PR #1380, opened 2025-07-30
fix: stricter jsx_has_dynamic_key

gaoachao:
As one of the PRs addressing #1371.

@coderabbitai summary

## Checklist


gaoachao:
@colinaaa I understand that the actual problem in #1371 lies in the runtime. #1380 merely enhances the detection logic to eliminate some avoidable cases. I will continue working on a proper fix for #1371.

Issue #1406, opened 2025-08-03
[Bug]: Error [ERR_MODULE_NOT_FOUND]: Cannot find module '<ROOT>/foo.js' imported from <ROOT>/lynx.config.ts


colinaaa:
### System Info

  System:
    OS: macOS 26.0
    CPU: (12) arm64 Apple M2 Max
    Memory: 1.02 GB / 96.00 GB
    Shell: 4.0.2 - /opt/homebrew/bin/fish
  Binaries:
    Node: 22.18.0 - ~/.local/state/fnm_multishells/1233_1753853167114/bin/node
    npm: 10.9.3 - ~/.local/state/fnm_multishells/1233_1753853167114/bin/npm
    pnpm: 9.15.9 - ~/.local/state/fnm_multishells/1233_1753853167114/bin/pnpm
    bun: 1.2.8 - ~/.bun/bin/bun
  npmPackages:
    @lynx-js/qrcode-rsbuild-plugin: ^0.4.0 => 0.4.0 
    @lynx-js/react: ^0.112.1 => 0.112.1 
    @lynx-js/react-rsbuild-plugin: ^0.10.10 => 0.10.10 
    @lynx-js/rspeedy: ^0.10.4 => 0.10.4 
    @lynx-js/types: 3.3.0 => 3.3.0 

### Details

This happens when enabling [builtin type stripping](https://nodejs.org/api/typescript.html#type-stripping) of Node.js (which is enabled by default in [v22.18.0](https://github.com/nodejs/node/releases/tag/v22.18.0)).

The `@lynx-js/rspeedy/register` we use has a `resolve` hook which resolve `.js` to `.ts` (the same as `resolveAlias` of Rspack). But it does not exist in Node.js.

### Reproduce Steps

```ts
// lynx.config.ts
import './foo.js'
```

```ts
// foo.ts
export {}
```

PR #1412, opened 2025-08-04
fix(rspeedy/core): deno compat

### LINKED ISSUE #1393: rspeedy doesn't resolve modules installed with `deno install`
TLDR: rspeedy doesn't resolve modules correctly when they are installed with `deno install`

I am very new to this codebase and definitely don't fully understand it but from what I do understand, (and this could be totally wrong) rspeedy essentially has its own "runtime" and, most importantly for right now, it's own module loader.

When running `deno install` with a package.json file, deno installs packages in the node-modules/.deno folder. What I assume is happening is that Rspeedy's module resolution is looking in the regular nod-modules folder but I could be wrong about that.

knotbin:
## Summary by CodeRabbit (edited by me)

* **Bug Fixes**
  * To fix compatibility with Deno which doesn't support module.register, environments that natively support TypeScript (including Deno) now 

closes #1393

## Checklist

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved compatibility with the Deno runtime, ensuring smooth configuration loading and TypeScript support detection.

* **Chores**
  * Added a changeset to document the compatibility update.


packages/rspeedy/core/register/index.js
```diff
@@ -13,6 +13,16 @@ import module from 'node:module'
  * @returns - The `unregister` function.
  */
 export function register(options) {
+  // Check for Deno environment first
```
colinaaa: We already have a `hasNativeTSSupport` function at:

https://github.com/lynx-family/lynx-stack/blob/15a46bc650cf95e92238fc6951df2ae7bad429c6/packages/rspeedy/core/src/config/loadConfig.ts#L155

which is based on `process.features.typescript`.

Although deno does not support `process.features.typescript` yet :)

Image

But it would be great to have the similar logic together.

packages/rspeedy/core/src/cli/start.ts
```diff
@@ -49,6 +49,12 @@ function tryGetPackageFolderFor(
 export function tryStartLocalRspeedy(
   root: string = process.cwd(),
 ): false | Promise<void> {
+  // Check for Deno environment and automatically use unmanaged mode
+  // @ts-expect-error - Deno global isn't be defined in Node.js
+  if (typeof Deno !== 'undefined' || process.versions?.deno) {
```
colinaaa: As for my understanding to Deno, it does support [installing dependencies to `node_modules`](https://docs.deno.com/runtime/fundamentals/node/#node_modules).

So I think it would not be necessary to bailout here for Deno. We would try-catch and use the unmanaged version when failed to find `@lynx-js/rspeedy` in `node_modules`.

packages/rspeedy/core/register/index.js
knotbin: I tried just adding the Deno logic to `hasNativeTSSupport` but it resulted in the same error when trying to run with Deno because the current logic assumes nativeTsSupport runtimes still support module.register.

I ended up using `hasNativeTSSupport` in a conditional for running `register`.

This only meant I had to change the `commonConfig.js` imports to `commonConfig.ts` in `web-tests` because these specific tests are run directly from the TypeScript source and not compiled into JS.

All other tests and libraries built perfectly fine with no changes and all tests pass.

I understand this change may be too much for just adding Deno support and I really did want to keep it isolated but that would have meant using the original change or making the conditional specific to only Deno. Please let me know if this meets your standards.

packages/rspeedy/core/src/cli/start.ts
knotbin: Yes you're right and I fixed this. I added this earlier when I thought the problem was that Deno installs with a different structure than npm and just symlinks the packages but that wasn't an issue once I added the other fix so I removed this.

packages/rspeedy/core/register/index.js
colinaaa: We are using `register` without condition to fix https://github.com/lynx-family/lynx-stack/issues/1406.

I do think it would be necessary to keep this way since there could be a lot of user that are doing this way (import with `.js` for a `.ts` file) since this is the default of TypeScript.

colinaaa: I would prefer something like this:


```js
  const unregister = shouldUseNativeImport(configPath)
    ? /** noop */ () => void 0
    : register({
      load: !hasNativeTSSupport(),
      resolve: true,
    })


function shouldUseNativeImport(configPath: string): boolean {
  return isJavaScriptPath(configPath) || isDeno()
}

function isDeno(): boolean {
  // @ts-expect-error - Deno global may not be defined in Node.js
  if (typeof Deno !== 'undefined' || process.versions?.deno) {
    return true
  }
  return false
}
```

wdyt

knotbin: That's a perfect solution! Sorry I tried a bunch of combinations of `isJavaScriptPath`, `hasNativeTSSupport` and the Deno logic but this one is clearly the most logical. Just updated it to reflect this and removed the web-test changes. Let me know what you think.

knotbin:
This is why the PR initially forced rspeedy into unmanaged mode. Rspeedy now works in Deno if you run `deno task build --unmanaged` which wasn't true before the PR.

Deno installs dependencies into the node-modules folder in a different pattern than node would but symlinks them to where node would expect them to be. I assume this symlinking has been specifically done to work with node and that's why running `npm run build` or `npm run dev` works after `deno install`. If you install dependencies with `npm install` and then run `deno task build` or `deno task dev` it works fine. 

It seems like the problem is specifically with resolving dependencies from that specific endpoint (`node_modules/@lynx-js/rspeedy/dist/cli/main.js` which if you command+click it, you'll see is actually symlinked to `node_modules/.deno/@lynx-js+rspeedy-canary@0.10.6-canary-20250806-cf183d9f/node_modules/@lynx-js/rspeedy-canary/dist/cli/main.js`) because I wasn't getting this error when I ran in the `examples/react` dir which I now realize was because the `@lynx-js/rspeedy` dependency was set to a workspace dependency.

I'll do some digging to make this work without `--unmanaged`. Rspack doesn't any errors like this so I'll look at the difference between the module resolution logic in the two (and if there's anything you can add about that, that would be a great help) and/or see if I can find why this might be happening on the Deno side.

knotbin:
@colinaaa I've traced the issue to the fact that [start.ts](https://github.com/lynx-family/lynx-stack/blob/main/packages/rspeedy/core/src/cli/start.ts) (called by [rspeedy.js](https://github.com/lynx-family/lynx-stack/blob/main/packages/rspeedy/core/bin/rspeedy.js)) seems to override the runtime's module resolution and implement a custom resolver using very jank filesystem logic to load the `main.ts`. It appears that loading `main.ts` this way without using `import` or `require` confuses Deno and causes it to load the file either without any context or with the context of start.ts.  `--unmanaged` fixes this because it just loads the `main.ts` file by importing it .

Trying to reimplement an import seems like a surefire way to introduce inconsistencies in resolution and bugs like the Deno one here. Package managers like PNPM, Deno, Bun, Yarn and even NPM now already handle resolving workspace dependencies gracefully. 

I might be missing something but the `--unmanaged` flag is something I've never seen before in a package. In all the other tools I've used, the functionality of the `--unmanaged` flag is built in as the default. There's probably some use case where it makes sense to have the "managed" mode but I've never seen it used in any other CLI including Rspack, so I think it makes much more sense (to me) to have unmanaged as the default.

I'm probably missing something, because the start.ts file is just part of the initial commit so it's hard for me to see why the functionality was added, so me know if there's something I'm not thinking of.

colinaaa:
Thank you @knotbin for your reply! Sorry for the delay :)

> I might be missing something but the `--unmanaged` flag is something I've never seen before in a package.

Both the idea and the [implementation](https://github.com/microsoft/rushstack/blob/main/apps/heft/src/startWithVersionSelector.ts) came from [`heft`](https://heft.rushstack.io/). There are tools like [`turbo`](https://turborepo.com/) also use the same mechanism.

It seems like the implementation is not good enough since it simulate the Node.js resolve.

PR #1438, opened 2025-08-05
fix(react): run all pending `renderComponent` before hydrate

hzy:
Run all pending `renderComponent` before hydrate, which ensures some immediate update can be applied in `hydrate`.

As background info, ReactLynx will use tree in background-thread as the source-of-truth, so this PR is helpful if main-thread renders more than background-thread's `root.render` by avoiding unwanted node removals.

## Summary by CodeRabbit

* **New Features**
  * Simplified background/non-main-thread rendering for more predictable startup.

* **Bug Fixes**
  * Deferred rethrow of certain lifecycle errors so first-screen work and hydration complete.
  * Ensures updates applied before hydration are preserved and hydration proceeds even if background processing errors.

* **Tests**
  * Added and streamlined tests covering pre-hydration updates, render flushing, and suspense/hydration flows.

## Checklist

### DIFF
--- a/packages/react/runtime/__test__/lifecycle/updateData.test.jsx
+++ b/packages/react/runtime/__test__/lifecycle/updateData.test.jsx
@@ -8,6 +8,7 @@ import { afterEach, beforeAll, beforeEach, describe, expect, it, vi } from 'vite
 import { replaceCommitHook } from '../../src/lifecycle/patch/commit';
 import { deinitGlobalSnapshotPatch } from '../../src/lifecycle/patch/snapshotPatch';
 import { InitDataConsumer, InitDataProvider, useInitData, withInitDataInState } from '../../src/lynx-api';
+import { useState } from '../../src/index';
 import { __root } from '../../src/root';
 import { globalEnvManager } from '../utils/envManager';
 import { elementTree, waitSchedule } from '../utils/nativeMethod';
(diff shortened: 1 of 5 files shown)


packages/react/runtime/__test__/lifecycle/updateData.test.jsx
```diff
@@ -864,3 +864,115 @@ describe('triggerDataUpdated when jsReady is enabled', () => {
...
+    // LifecycleConstant.jsReady
+    {
+      globalEnvManager.switchToMainThread();
+      rLynxJSReady();
+    }
+
+    // background updateCardData
+    {
+      globalEnvManager.switchToBackground();
+      lynxCoreInject.tt.updateCardData({ msg: 'update' });
```
upupming, line 941: Can we assert that `setState` is called when `updateCardData` is executed?

packages/react/runtime/__test__/lifecycle/updateData.test.jsx
```diff
@@ -864,3 +864,115 @@ describe('triggerDataUpdated when jsReady is enabled', () => {
...
+    }
+
+    // hydrate
+    {
+      globalEnvManager.switchToBackground();
+      // LifecycleConstant.firstScreen
+      lynxCoreInject.tt.OnLifecycleEvent(...globalThis.__OnLifecycleEvent.mock.calls[0]);
```
upupming, line 950: Can we assert that `process` is executed and state is updated? we can import `process` from preact and spy on it.

packages/react/runtime/__test__/lifecycle/updateData.test.jsx
```diff
@@ -864,3 +864,115 @@ describe('triggerDataUpdated when jsReady is enabled', () => {
...
+            />
+          </text>
+        </page>
+      `);
+    }
+
+    // background render
+    {
+      globalEnvManager.switchToBackground();
+      render(<Comp />, __root);
```
upupming, line 927: This render will use `update` because we share `lynx.__initData` across 2 threads in test framework.

packages/react/runtime/__test__/lynx/suspense.test.jsx
```diff
@@ -505,26 +410,6 @@ describe('suspense', () => {
       vi.runAllTimers();
     }
 
-    // render fallback
```
upupming, line 508: Why do we delete this?

packages/react/runtime/__test__/lynx/suspense.test.jsx
```diff
@@ -676,26 +562,6 @@ describe('suspense', () => {
...
-    {
-      globalEnvManager.switchToBackground();
```
upupming, line 681: Same as above

packages/react/runtime/src/lynx-api.ts
```diff
@@ -87,19 +87,8 @@ export const root: Root = {
       __root.__jsx = jsx;
     } else {
       __root.__jsx = jsx;
-      let preactProcess: (() => void) | undefined = undefined;
-      // eslint-disable-next-line @typescript-eslint/unbound-method
-      const oldDebounceRendering = options.debounceRendering;
-      options.debounceRendering = (cb) => {
-        preactProcess = cb;
-      };
```
upupming, line 96: Why do we remove this

upupming (changes requested):

packages/react/runtime/src/lynx-api.ts
hzy: This is redundant logic if we already call `process()` before hydrate.

packages/react/runtime/__test__/lynx/suspense.test.jsx
hzy: To make the tests pass.

In this group of tests, the `render(<Comp />, __root);` on background thread is calling directly to preact's `render`.

packages/react/runtime/__test__/lifecycle/updateData.test.jsx
hzy: Great catch!
I added a reset just now to ensure the `lynx.__initData` is the initial value in background thread.

packages/react/runtime/__test__/lifecycle/updateData.test.jsx
hzy: Done.

packages/react/runtime/__test__/lifecycle/updateData.test.jsx
hzy: I've added a test to assert it throwing an error.

PR #1452, opened 2025-08-06
refactor(react): always use `renderToString` when renderPage in test

hzy:
@coderabbitai summary

## Checklist


Yradex (approved):

packages/react/testing-library/src/__tests__/act.test.jsx
```diff
@@ -48,14 +48,8 @@ test('render calls useEffect immediately', async () => {
     // this is a different behavior from the testing library and real Lynx environment
     // in the testing library, `useEffect` will be called since we are using `render`
     // in the real Lynx environment, `useEffect` will not be called since we are not using `renderToString`
```
upupming:
```suggestion
```

packages/react/testing-library/src/__tests__/alog.test.jsx
```diff
@@ -38,16 +38,13 @@ describe('alog', () => {
     expect(lynxTestingEnv.mainThread.console.alog.mock.calls).toMatchInlineSnapshot(`
       [
         [
-          "[MainThread Component Render] name: ClassComponent, uniqID: __Card__:__snapshot_426db_test_2, __id: -6",
+          "[MainThread Component Render] name: ClassComponent, uniqID: undefined, __id: undefined",
```
upupming: Why it is changed to `undefined` here?

packages/react/runtime/src/renderToOpcodes/index.ts
```diff
@@ -304,7 +304,7 @@ function _renderToString(
     }
   }
 
-  if (typeof children === 'string') {
+  if (typeof children === 'string' || typeof children === 'number') {
```
upupming, line 310: Is this only for test, or will it affect the production environment too?

packages/react/testing-library/src/__tests__/alog.test.jsx
hzy: In the test environment, the vnode passed to `renderToString` is actually a real preact vnode, not a `SnapshotInstance` with a vnode shape, so the vnode passed to `options.diffed` is actually a real preact vnode, which means `type` (printed as `uniqID`) and `__id` is not present.

In the real environment, the vnode passed to `renderToString` comes from @lynx-js/react/lepus/jsx-runtime, so the result is printed normally.

packages/react/runtime/src/renderToOpcodes/index.ts
hzy: It will affect the production environment. This is to align with the behavior of preact on background-thread.

PupilTong (approved):

colinaaa (approved):

PR #1455, opened 2025-08-07
feat(react/runtime): support using a host element as direct child of Suspense

Yradex:
## Summary by CodeRabbit

* **New Features**
  * Added support for using a host element as a direct child of the Suspense component.

* **Bug Fixes**
  * Improved handling of Suspense fallback and content transitions, ensuring correct UI updates and cleanup in various scenarios.

* **Tests**
  * Introduced comprehensive tests for Suspense, covering rendering, hydration, parallel and nested boundaries, and error handling.
  * Updated test snapshots to reflect changes in element hierarchy with wrapper elements.

* **Refactor**
  * Enhanced internal logic for background snapshot management and subtree reconstruction to ensure robust element handling.

## Checklist

### DIFF
--- a/packages/react/runtime/src/internal.ts
+++ b/packages/react/runtime/src/internal.ts
@@ -2,7 +2,7 @@
 // Licensed under the Apache License Version 2.0 that can be found in the
 // LICENSE file in the root directory of this source tree.
 
-import { createElement, lazy } from 'preact/compat';
+import { Suspense, createElement, lazy } from 'preact/compat';
 import type { FC } from 'react';
 
 import './lynx.js';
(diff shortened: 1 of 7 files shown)


Yradex:
@upupming What do you think of the fix in 3f74f7d01083e9d8df6843b3aaf4b5396de78844?

packages/react/runtime/src/internal.ts
```diff
@@ -2,15 +2,14 @@
 // Licensed under the Apache License Version 2.0 that can be found in the
 // LICENSE file in the root directory of this source tree.
 
-import { createElement, lazy } from 'preact/compat';
+import { Suspense, createElement, lazy } from 'preact/compat';
```
upupming, line 5: If we use `Suspense` from `preact`, will the `Suspense` instance in component is polyfill different?

Yradex: Yes, but using Preact's Suspense here would suffice, and it wouldn't increase the package size either.

PR #1480, opened 2025-08-08
feat: add profile for list hydration

gaoachao:
@coderabbitai summary

## Checklist

### DIFF
--- a/packages/react/runtime/src/hydrate.ts
+++ b/packages/react/runtime/src/hydrate.ts
@@ -352,6 +352,9 @@ export function hydrate(before: SnapshotInstance, after: SnapshotInstance, optio
           updateAction,
         };
 
+        if (__PROFILE__) {
+          console.profile(`[hydrate] update-list-info: ${JSON.stringify(info)}`);
+        }
         const listElement = before.__elements![elementIndex]!;
         __SetAttribute(listElement, 'update-list-info', info);
         const [componentAtIndex, componentAtIndexes] = componentAtIndexFactory(afterChildNodes, hydrate);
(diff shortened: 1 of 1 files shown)


packages/react/runtime/src/hydrate.ts
```diff
@@ -366,6 +366,13 @@ export function hydrate(before: SnapshotInstance, after: SnapshotInstance, optio
           updateAction,
         };
 
+        if (__PROFILE__) {
+          console.profile(`[hydrate] update-list-info: ${JSON.stringify(info)}`);
```
hzy: Use `lynx.performance.profileStart`. See https://github.com/lynx-family/lynx-stack/pull/1533

packages/react/runtime/src/listUpdateInfo.ts
```diff
@@ -82,14 +83,30 @@ export class ListUpdateInfoRecording implements ListUpdateInfo {
     //   __SetAttribute(listElement, "update-list-info", pendingAttribute);
     //   __FlushElementTree(listElement);
     // });
-    __SetAttribute(listElement, 'update-list-info', this.__toAttribute());
+    const updateListInfo = this.__toAttribute();
+
+    if (__PROFILE__) {
+      const listID = __GetElementUniqueID(this.list.__elements[elementIndex]!);
```
upupming:
```suggestion
      const listID = __GetElementUniqueID(listElement);
```

packages/react/runtime/src/hydrate.ts
```diff
@@ -366,6 +367,15 @@ export function hydrate(before: SnapshotInstance, after: SnapshotInstance, optio
           updateAction,
         };
 
+        if (__PROFILE__) {
+          profileStart(`[hydrate] update-list-info`, {
```
hzy: Maybe consider a syntax more consistent with other profile event?

gaoachao: I have renamed it to `ReactLynx::listHydrate::updateListInfo`. PTLA~

PR #1482, opened 2025-08-11
fix: "emit different content to the same filename" error

### LINKED ISSUE #1481: [Bug]:  Conflict: Multiple assets emit different content to the same filename .rspeedy/some-lib/some-lib.699162460018c898.css.hot-update.json
### System Info

@lynx-js/css-extract-webpack-plugin: [0.6.1](https://www.npmjs.com/package/@lynx-js/css-extract-webpack-plugin/v/0.6.1)

### Details

With chunk split enabled in lynx.config.js and multiple entries:

```js
source: {
    entry: {
      main1: './src/index1.tsx',
      main2: './src/index2.tsx',
    },
  },
performance: {
    chunkSplit: {
      strategy: 'split-by-experience',
    },
  },
```

Both `src/index1.tsx` and `src/index2.tsx` use a shared lib `some-lib` (a component library so it will emit CSS assets).

### Reproduce link

_No response_

### Reproduce Steps

Using `rspeedy dev`, you will get the error like:

```bash
× Conflict: Multiple assets emit different content to the same filename .rspeedy/some-lib/some-lib.699162460018c898.css.hot-update.json
```

This "emit different content to the same filename" error happens because we encode `some-lib`'s css assets to hot update json files using different `compilerOptions.templateDebugUrl` config.

For example, the diff of two encode options of two css chunk may only be:

```diff
- "templateDebugUrl": "http://x.x.x.x:3000/.rspeedy/pages/main1/debug-info.json",
+ "templateDebugUrl": "http://x.x.x.x:3000/.rspeedy/pages/main2/debug-info.json",
```

upupming:
Fixes #1481

## Summary by CodeRabbit

- Bug Fixes
  - Fixed an issue where CSS extraction could emit different content to the same filename during chunk splitting and HMR, preventing inconsistent hot-update behavior.
- Tests
  - Added extensive chunk-splitting HMR test cases, fixtures, configs, and snapshots to validate CSS hot-update JSON generation across multiple entries.
- Chores
  - Added a patch-level changeset entry documenting the fix for the CSS extract plugin.

## Checklist


colinaaa (approved):
👍

PR #1498, opened 2025-08-12
fix(rspeedy/react): fix aliases are missing when using in environments

colinaaa:
## Summary by CodeRabbit

- New Features
  - Support using the React alias plugin multiple times across different environments; per-environment aliasing combines correctly with global plugins.

- Refactor
  - React plugin factory now returns multiple composable plugins (alias wrapper + main plugin) instead of a single plugin.

- Tests
  - Added tests for multi-environment configurations and plugin combinations; skipped an obsolete single-run alias test.

- Chores
  - Added a changeset entry declaring patch releases for related packages.

## Checklist


packages/rspeedy/plugin-react-alias/test/index.test.ts
```diff
@@ -257,4 +257,109 @@ describe('React - alias', () => {
...
+    test('alias with multiple environments', async () => {
+      vi.stubEnv('NODE_ENV', 'development')
+      const { pluginReactAlias } = await import('../src/index.js')
+
+      const rsbuild = await createRsbuild({
+        rsbuildConfig: {
+          environments: {
+            lynx: {
+              plugins: [
+                pluginReactAlias({
```
upupming, line 271: Shall we also add a test that uses `pluginReactLynx` here?

colinaaa: I've added more tests to `pluginReactLynx`, and one of them (using `plugins` with `environments['lynx'].plugins`) got stack overflow. It should be related with the way we modifying SWC/Rspack configuration. I would like to fix it in the future PRs.

PR #1504, opened 2025-08-12
feat: inlineScripts defaults to false, when chunkSplit strategy is not 'all-in-one'

### LINKED ISSUE #1221: [Feature]: inlineScripts defaults to false, when chunkSplit strategy is not 'all-in-one'.

`inlineScripts` defaults to false, when chunkSplit strategy is not 'all-in-one'.

N/A

gaoachao:
close #1221

@coderabbitai summary

## Checklist

### DIFF
--- a/packages/rspeedy/plugin-react/src/entry.ts
+++ b/packages/rspeedy/plugin-react/src/entry.ts
@@ -197,6 +197,12 @@ export function applyEntry(
         .end()
     })
 
+    const rsbuildConfig = api.getRsbuildConfig()
+    const userConfig = api.getRsbuildConfig('original')
+
+    const enableChunkSplitting =
+      rsbuildConfig.performance?.chunkSplit?.strategy !== 'all-in-one'
+
     let finalFirstScreenSyncTiming = firstScreenSyncTiming
 
     if (isLynx) {
@@ -205,7 +211,12 @@ export function applyEntry(
         // TODO: support inlineScripts in lazyBundle
         inlineScripts = true
       } else {
-        inlineScripts = environment.config.output?.inlineScripts ?? true
+        //  inlineScripts defaults to false, when chunkSplit strategy is not 'all-in-one'
+        if (enableChunkSplitting) {
+          inlineScripts = false
+        } else {
+          inlineScripts = environment.config.output?.inlineScripts ?? true
+        }
       }
 
       if (inlineScripts !== true) {
@@ -247,14 +258,8 @@ export function applyEntry(
         .end()
     }
 
-    const rsbuildConfig = api.getRsbuildConfig()
-    const userConfig = api.getRsbuildConfig('original')
-
     let extractStr = originalExtractStr
-    if (
-      rsbuildConfig.performance?.chunkSplit?.strategy !== 'all-in-one'
-      && originalExtractStr
-    ) {
+    if (enableChunkSplitting && originalExtractStr) {
       logger.warn(
         '`extractStr` is changed to `false` because it is only supported in `all-in-one` chunkSplit strategy, please set `performance.chunkSplit.strategy` to `all-in-one` to use `extractStr.`',
       )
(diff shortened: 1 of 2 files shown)


packages/rspeedy/plugin-react/src/entry.ts
```diff
@@ -205,7 +211,12 @@ export function applyEntry(
         // TODO: support inlineScripts in lazyBundle
         inlineScripts = true
       } else {
-        inlineScripts = environment.config.output?.inlineScripts ?? true
+        //  inlineScripts defaults to false, when chunkSplit strategy is not 'all-in-one'
+        if (enableChunkSplitting) {
+          inlineScripts = false
```
colinaaa: This is not **defaults** to `false`, but **forced** to `false`.

I do think we need to respect the userConfig on `inlineScripts`.

packages/rspeedy/plugin-react/test/config.test.ts
```diff
@@ -647,6 +647,34 @@ describe('Config', () => {
 
       expect(encodePlugin).toHaveProperty('options', { inlineScripts: true })
     })
+
+    test('output.inlineScripts defaults to `false`, when chunkSplit strategy is not `all-in-one`', async () => {
```
colinaaa, line 679: Please add test cases using `performance.chunkSplit.strategy: 'split-by-size'` with all the variants of `output.inlineScripts`

.changeset/gorgeous-adults-pretend.md
```diff
@@ -0,0 +1,5 @@
+---
+"@lynx-js/react-rsbuild-plugin": patch
```
colinaaa, line 3: Maybe we could make this change in `@lynx-js/rspeedy` (I'm not very sure about that). wdyt

gaoachao: LGTM

packages/rspeedy/plugin-react/src/entry.ts
gaoachao: Sorry, brain fart - didn't think that through:)

packages/rspeedy/plugin-react/src/entry.ts
```diff
@@ -205,7 +211,8 @@ export function applyEntry(
         // TODO: support inlineScripts in lazyBundle
         inlineScripts = true
       } else {
-        inlineScripts = environment.config.output?.inlineScripts ?? true
+        // environment.config.output?.inlineScripts is guaranteed to have a value
+        inlineScripts = environment.config.output?.inlineScripts
```
upupming, line 214: Do we only modify this default `true` here, It seems this file can keep unchanged as well.

gaoachao:
```js
inlineScripts = environment.config.output?.inlineScripts ?? !enableChunkSplitting
```
How about this?

packages/rspeedy/core/src/config/defaults.ts
```diff
@@ -26,8 +30,8 @@ export function applyDefaultRspeedyConfig(config: Config): Config {
       // from the `output.filename.bundle` field.
       filename: getFilename(config.output?.filename),
 
-      // inlineScripts should be enabled by default
-      inlineScripts: true,
+      // inlineScripts defaults to false when chunk splitting is enabled, true otherwise
+      inlineScripts: !enableChunkSplitting,
```
colinaaa, line 34: Can we add some tests for this?

E.g.:
- Default values
- Can be overriden by user config

packages/rspeedy/plugin-react/src/entry.ts
```diff
@@ -205,7 +211,8 @@ export function applyEntry(
         // TODO: support inlineScripts in lazyBundle
         inlineScripts = true
       } else {
-        inlineScripts = environment.config.output?.inlineScripts ?? true
+        inlineScripts = environment.config.output?.inlineScripts
```
colinaaa, line 214: We would also need a changeset for this change in react-rsbuild-plugin.

packages/rspeedy/core/src/config/defaults.ts
gaoachao: https://github.com/lynx-family/lynx-stack/pull/1504/commits/e5caf866042f1922cddfb78634ec4d2d272dbffa PTLA~

colinaaa (approved):

PR #1522, opened 2025-08-13
test: split vitest

colinaaa:
Test for the "Channel closed" error.

## Checklist


colinaaa:
The problem should be solved by https://github.com/lynx-family/lynx-stack/pull/1523.

PR #1531, opened 2025-08-14
fix(web): globalThis is never accessible in MTS

Sherry-hue:
@coderabbitai summary

## Checklist


packages/web-platform/web-constants/src/utils/generateTemplate.ts
```diff
@@ -106,6 +106,7 @@ const generateModuleContent = (
       `const ${nm} = lynx_runtime.${nm}?.bind(lynx_runtime);`
     ),
     ';var globDynamicComponentEntry = \'__Card__\';',
+    ';var navigator = undefined;var postMessage = undefined;',
```
PupilTong: How about create a new array like `disallowedGlobalVars` and configure it like `mainThreadInjectVars` ?

Sherry-hue: done~

packages/web-platform/web-constants/src/utils/generateTemplate.ts
```diff
@@ -106,6 +107,7 @@ const generateModuleContent = (
       `const ${nm} = lynx_runtime.${nm}?.bind(lynx_runtime);`
     ),
     ';var globDynamicComponentEntry = \'__Card__\';',
+    ...disallowedGlobalVars.map(nm => `var ${nm} = undefined;`),
```
PupilTong: try to produce
var a=b=c=undefined

packages/web-platform/web-mainthread-apis/src/createMainThreadGlobalThis.ts
```diff
@@ -857,6 +857,9 @@ export function createMainThreadGlobalThis(
   };
   mtsGlobalThis.globalThis = new Proxy(mtsGlobalThis, {
     get: (target, prop) => {
+      if (prop === 'navigator' || prop === 'postMessage') {
```
PupilTong: seems u should read the configuration&the configuration is supposed to be a set

Sherry-hue: I use the globalDisallowedVars array to maintain this so that generateTemaplte only needs to map once when injecting

PR #1533, opened 2025-08-14
refactor(react/runtime): use `lynx.performance.profileStart`

hzy:
@coderabbitai summary

## Checklist


packages/react/runtime/src/lynx/env.ts
```diff
@@ -51,7 +52,7 @@ export function setupLynxEnv(): void {
       let hasDefaultDataProcessorExecuted = false;
       globalThis.processData = (data, processorName) => {
         if (__PROFILE__) {
-          console.profile('processData');
+          profileStart('processData');
```
colinaaa, line 55: Do we need the `ReactLynx::` prefix here?

packages/react/runtime/src/debug/utils.ts
```diff
@@ -0,0 +1,23 @@
...
+/* eslint-disable @typescript-eslint/no-unnecessary-type-assertion */
+
+const noop = () => {};
+
+export const profileStart = /* @__PURE__ */ ((() => {
+  let p;
+  if (!(p = lynx.performance) || typeof p.profileStart !== 'function') {
+    return noop;
+  }
+  return p.profileStart.bind(p);
```
colinaaa, line 14: Can we add the `ReactLynx::` prefix here? This may save some bytes without gzip.

packages/react/runtime/src/lynx/env.ts
hzy: it is under globalThis so i think it should not？

packages/react/runtime/src/debug/utils.ts
hzy: Not all calls require this prefix, so I think it's better not to.

packages/react/runtime/src/backgroundSnapshot.ts
```diff
@@ -212,7 +213,7 @@ export class BackgroundSnapshotInstance {
 
   setAttribute(key: string | number, value: unknown): void {
     if (__PROFILE__) {
-      console.profile('setAttribute');
+      profileStart('ReactLynx::BSI::setAttribute');
```
upupming, line 216: User may not know  what `BSI` means, will we have doc on this?

PR #1562, opened 2025-08-19
fix: remove redundant internal-slot for pure text children

upupming:
@coderabbitai summary

## Checklist


packages/react/transform/src/swc_plugin_snapshot/jsx_helpers.rs
```diff
@@ -253,30 +253,35 @@ pub fn jsx_is_list_item(jsx: &JSXElement) -> bool {
...
+      !matches!(
+        child,
+        JSXElementChild::JSXExprContainer(JSXExprContainer {
+          expr: JSXExpr::JSXEmptyExpr(_),
+          ..
+        })
+      )
+    });
+
+    let only_has_empty_text = tmp.clone().all(|child| match child {
```
colinaaa: It looks like this `.clone()` can be avoided by using reference.


```suggestion
    let only_has_empty_text = &tmp.all(|child| match child {
```

colinaaa: Sorry, it looks like we are cloning an `iterator`. Can we rename `tmp` to something like `iter` to make the `.clone()` looks less expensive?

packages/react/transform/src/swc_plugin_snapshot/jsx_helpers.rs
```diff
@@ -253,30 +253,35 @@ pub fn jsx_is_list_item(jsx: &JSXElement) -> bool {
...
+      JSXElementChild::JSXText(text) => jsx_text_to_str(&text.value).is_empty(),
+      _ => false,
+    });
+
+    if !only_has_empty_text
+      && tmp.all(|child| match child {
         JSXElementChild::JSXText(text) => jsx_text_to_str(&text.value).is_empty(),
```
colinaaa: It looks like this is the same as line 279. We may filter out `JSXElementChild::JSXText` in this iteration.

Here is a simplified version created by GPT-5:

```rust
pub fn jsx_is_children_full_dynamic(n: &JSXElement) -> bool {
  if n.children.is_empty() {
    return false;
  }

  // A text node is considered empty if normalized string is empty (whitespace-only)
  let is_text_empty = |text: &JSXText| jsx_text_to_str(&text.value).is_empty();

  // Filter out:
  // - Empty expression containers (JSXEmptyExpr)
  // - Whitespace-only text
  // - Spread children (unsupported)
  // Keep:
  // - JSXElement
  // - JSXFragment
  // - Expr containers with a real Expr
  // - Non-empty text (will be treated as non-dynamic later)
  let mut filtered = n.children.iter().filter(|child| match child {
    JSXElementChild::JSXExprContainer(JSXExprContainer {
      expr: JSXExpr::JSXEmptyExpr(_),
      ..
    }) => false,
    JSXElementChild::JSXText(text) => !is_text_empty(text),
    JSXElementChild::JSXSpreadChild(_) => false,
    _ => true,
  });

  // Define “dynamic”:
  // - Custom JSX elements => depends on jsx_is_custom
  // - Expression containers with Expr(_) => true
  // - Fragments => false
  // - Non-empty text => false
  let is_dynamic = |child: &JSXElementChild| -> bool {
    match child {
      JSXElementChild::JSXElement(element) => jsx_is_custom(element),
      JSXElementChild::JSXExprContainer(JSXExprContainer {
        expr: JSXExpr::Expr(_),
        ..
      }) => true,
      JSXElementChild::JSXFragment(_) => false,
      JSXElementChild::JSXText(_) => false,
      // Any other cases we decided to drop earlier or are not dynamic
      _ => false,
    }
  };

  // No substantive children after filtering
  let Some(first) = filtered.next() else {
    return false;
  };

  let first_ok = is_dynamic(first);
  if !first_ok {
    return false;
  }

  filtered.all(is_dynamic)
}
```

packages/react/transform/src/swc_plugin_snapshot/mod.rs
```diff
@@ -1658,6 +1658,113 @@ mod tests {
...
+          unresolved_mark,
+          TransformMode::Test,
+        )),
+      )
+    },
+    full_static_children_self_close,
+    // Input codes
+    r#"
+    <view className="parent">
+			<view className="child"/>
```
colinaaa, line 1624: Can we add some tests about JSX comments?

```jsx
<view>
  {/** foo */}
</view>
```

colinaaa (dismissed):
👍

packages/react/transform/src/swc_plugin_snapshot/jsx_helpers.rs
```diff
@@ -253,47 +253,64 @@ pub fn jsx_is_list_item(jsx: &JSXElement) -> bool {
...
+      expr: JSXExpr::JSXEmptyExpr(_),
+      ..
+    }) => false,
+    JSXElementChild::JSXText(text) => !is_text_empty(text),
+    JSXElementChild::JSXSpreadChild(_) => {
+      unreachable!("Unexpected JSXSpreadChild in jsx_is_children_full_dynamic - not supported")
+    }
+    _ => true,
+  });
+
```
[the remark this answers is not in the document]
upupming: We do not want to modify the original logic about `unreachable!`, it is the old behavior, we keep it.

hzy (dismissed):

hzy (approved):

gaoachao (approved):

PR #1581, opened 2025-08-21
feat(react): support nested list

gaoachao:
## Summary by CodeRabbit

* **New Features**
  * Better nested-list support: insert/move/remove behavior and reuse after hydration improved.

* **Bug Fixes**
  * Reduced lingering pending-list state during reloads and delayed JS readiness, improving UI consistency.

* **Performance**
  * More targeted clearing and flushing of pending list updates to avoid unnecessary work.

* **Tests / Chores**
  * Added comprehensive nested-list tests and a patch changeset.

## Checklist

### DIFF
--- a/packages/react/runtime/__test__/utils/nativeMethod.ts
+++ b/packages/react/runtime/__test__/utils/nativeMethod.ts
@@ -139,6 +139,10 @@ export const elementTree = new (class {
     e.props[key] = value;
   }
 
+  __GetAttributes(e: Element): Record<string, any> {
+    return e.props;
+  }
+
   __AddEvent(e: Element, eventType: string, eventName: string, event: string | Record<string, any>) {
     if (typeof event === 'undefined') {
       if (e.props.event) {
--- a/packages/react/runtime/src/listUpdateInfo.ts
+++ b/packages/react/runtime/src/listUpdateInfo.ts
@@ -7,7 +7,8 @@ import { componentAtIndexFactory, enqueueComponentFactory } from './list.js';
 import type { SnapshotInstance } from './snapshot.js';
 
 export interface ListUpdateInfo {
-  flush(): void;
+  flush(): number | undefined;
+  getAttachedListId(): number | undefined;
   onInsertBefore(
     newNode: SnapshotInstance,
     existingNode?: SnapshotInstance,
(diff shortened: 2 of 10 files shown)


packages/react/runtime/src/listUpdateInfo.ts
```diff
@@ -86,6 +90,15 @@ export class ListUpdateInfoRecording implements ListUpdateInfo {
       enqueueComponentFactory(),
       componentAtIndexes,
     );
+    return this.list.__id;
+  }
+
+  getAttachedListId(): undefined | number {
+    /* v8 ignore next 3 */
```
hzy: Can we add tests to cover those lines?

packages/react/runtime/src/snapshot.ts
```diff
@@ -331,6 +331,7 @@ export class SnapshotInstance {
 
     if (isListHolder) {
       // never recurse into list's children
+      __pendingListUpdates.flushWithId(this.__id);
```
hzy, line 338: Could you add some commets or a test to explain why this is needed?

packages/react/runtime/__test__/utils/nativeMethod.ts
```diff
@@ -139,6 +139,10 @@ export const elementTree = new (class {
     e.props[key] = value;
   }
 
+  __GetAttributes(e: Element): Record<string, any> {
```
hzy, line 142: Is this API used in this PR?

gaoachao: This was left behind during the code migration - we used to need the check for `custom-list-name` being `'list-container'`. I think it's harmless though, since it's only used in tests. WDYT?

packages/react/runtime/src/snapshot.ts
gaoachao: see: https://github.com/lynx-family/lynx-stack/pull/1581/commits/7c5459b6055902e3359f812a369c652d7eebccf9

packages/react/runtime/src/listUpdateInfo.ts
gaoachao: I've added a test to cover these lines. see: https://github.com/lynx-family/lynx-stack/pull/1581/commits/6f10ca3210519e4a79efa186fbffc27d0f9a0030

PR #1598, opened 2025-08-25
feat: use item key as beforeMap key in diffArrayLepus

gaoachao:
@coderabbitai summary

## Checklist

### DIFF
--- a/packages/react/runtime/src/backgroundSnapshot.ts
+++ b/packages/react/runtime/src/backgroundSnapshot.ts
@@ -470,6 +470,8 @@ export function hydrate(
             (a, b) => {
               helper(a, b);
             },
+            // Should be `false` in hydrate as SerializedSnapshotInstance has no item-key
+            false,
           );
           diffArrayAction(
             beforeChildNodes,
--- a/packages/react/runtime/src/hydrate.ts
+++ b/packages/react/runtime/src/hydrate.ts
@@ -5,9 +5,12 @@
 import { componentAtIndexFactory, enqueueComponentFactory, gRecycleMap, gSignMap } from './list.js';
 import { __pendingListUpdates } from './pendingListUpdates.js';
 import { DynamicPartType } from './snapshot/dynamicPartType.js';
+import type { PlatformInfo } from './snapshot/platformInfo.js';
 import { unref } from './snapshot/ref.js';
 import type { SnapshotInstance } from './snapshot.js';
 import { isEmptyObject } from './utils.js';
+
+const UNREACHABLE_ITEM_KEY_NOT_FOUND = 'UNREACHABLE_ITEM_KEY_NOT_FOUND';
 
 export interface DiffResult<K> {
   $$diff: true;
@@ -21,6 +24,7 @@ export interface DiffResult<K> {
 
 export interface Typed {
   type: string;
+  __listItemPlatformInfo?: PlatformInfo;
 }
 
 export function isEmptyDiffResult<K>(diffResult: DiffResult<K>): boolean {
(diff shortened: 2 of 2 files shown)


packages/react/runtime/src/backgroundSnapshot.ts
```diff
@@ -471,6 +471,8 @@ export function hydrate(
             (a, b) => {
               helper(a, b);
             },
+            // Should be `false` in hydrate as SerializedSnapshotInstance has no item-key
+            false,
```
upupming, line 475: This seems too hacky. The parameter is named as `isListChildren` and we passed `false` here. But it truely can be a `DynamicPartType.ListChildren`.

packages/react/runtime/src/hydrate.ts
```diff
@@ -5,10 +5,13 @@
 import { componentAtIndexFactory, enqueueComponentFactory, gRecycleMap, gSignMap } from './list.js';
 import { __pendingListUpdates } from './pendingListUpdates.js';
 import { DynamicPartType } from './snapshot/dynamicPartType.js';
+import type { PlatformInfo } from './snapshot/platformInfo.js';
 import { unref } from './snapshot/ref.js';
 import type { SnapshotInstance } from './snapshot.js';
 import { isEmptyObject } from './utils.js';
 
+const UNREACHABLE_ITEM_KEY_NOT_FOUND = 'UNREACHABLE_ITEM_KEY_NOT_FOUND';
```
upupming, line 13: Why we use this instead of throw an error directly?

packages/react/runtime/src/hydrate.ts
```diff
@@ -46,12 +51,18 @@ export function diffArrayLepus<A extends Typed, B extends Typed>(
...
+    (beforeMap[key] ??= new Set()).add([node, i]);
   }
 
   for (let i = 0; i < after.length; i++) {
     const afterNode = after[i]!;
-    const beforeNodes = beforeMap[afterNode.type];
+    const key = isListChildren
+      ? afterNode.__listItemPlatformInfo?.['item-key'] ?? UNREACHABLE_ITEM_KEY_NOT_FOUND
+      : afterNode.type;
+    const beforeNodes = beforeMap[key];
```
upupming, line 65: What if the key is same but type is different, will it be reused wrongly?

packages/react/runtime/src/backgroundSnapshot.ts
gaoachao: How about change `isListChildren` to `isListHasItemKey`?

packages/react/runtime/src/hydrate.ts
gaoachao: Because the list SDK will throw an error when the `item-key` is not provided.

packages/react/runtime/src/hydrate.ts
gaoachao: We have a logic check `isSameType`.
Refer: https://github.com/lynx-family/lynx-stack/blob/main/packages/react/runtime/src/hydrate.ts#L62

.changeset/clear-numbers-call.md
```diff
@@ -0,0 +1,23 @@
+---
+"@lynx-js/react": patch
+---
+
+During hydration, replace update with insert + remove for same-type `<list-item />` with different `item-key` so the Lynx SDK detects changes.
```
hzy:
```suggestion
During hydration, replace update with insert + remove for same-type `<list-item />` with different `item-key` so the Lynx Engine detects changes.
```

PR #1641, opened 2025-09-01
fix: should only recycle off-screen list-item in recursive hydration

gaoachao:
## Summary by CodeRabbit

- Bug Fixes
  - Refined hydration so only off-screen list items are considered for recycling, preventing visible items from being reused and improving reload stability.
- Tests
  - Updated tests to assert the tightened recycling behavior during hydration.
- Chores
  - Added a changeset for a patch release of @lynx-js/react.

## Checklist

### DIFF
--- a/packages/react/runtime/__test__/list.test.jsx
+++ b/packages/react/runtime/__test__/list.test.jsx
@@ -1945,6 +1945,11 @@ describe('list reload', () => {
     expect(signMap.get(__GetElementUniqueID(d1.__element_root))).toBe(d2_); // note: d1 is reused by d2_
     expect(signMap.get(__GetElementUniqueID(d2_.__element_root))).toBe(d2_);
     expect(signMap.get(__GetElementUniqueID(d3_.__element_root))).toBe(d3_);
+    // d1 was enqueued as well, so it should be in the recycling pool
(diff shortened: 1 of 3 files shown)


packages/react/runtime/__test__/list.test.jsx
```diff
@@ -1940,7 +1945,6 @@ describe('list reload', () => {
     expect(signMap.get(__GetElementUniqueID(d1.__element_root))).toBe(d2_); // note: d1 is reused by d2_
     expect(signMap.get(__GetElementUniqueID(d2_.__element_root))).toBe(d2_);
     expect(signMap.get(__GetElementUniqueID(d3_.__element_root))).toBe(d3_);
-    expect(recycleSignMap.get(__GetElementUniqueID(d3_.__element_root))).toBe(d3_);
```
hzy, line 1943: Since `componentAtIndex` and `enqueueComponent` are called above (https://github.com/lynx-family/lynx-stack/pull/1641/files#diff-f904e2ac8b183a8d7d7cd945d1b38ba8b1f1ed23e9204e178cabb99065a10fa2R1856-R1862), it is recommended to do some (strict) assert for `recycleSignMap` here.

PR #1650, opened 2025-09-03
ci(benchmark/react): add benchmark for various attribute update

hzy:
- **fix(react): run all pending `renderComponent` before hydrate (#1438)**
- **ci(benchmark/react): enable list in benchmark**
- **ci(benchmark/react): add benchmark for various attribute update**

## Summary by CodeRabbit

* New Features
  * Added a new React benchmark case that exercises a variety of attribute updates with runtime timing, enabling more granular performance insights.
  * Introduced dedicated scripts to run and profile this benchmark case for quicker benchmarking and trace capture.

* Chores
  * Updated benchmark configuration to register the new case in the build and run pipeline.
  * Exposed utilities for determining execution context to support benchmark behavior across environments.

## Checklist

### DIFF
--- a/benchmark/react/cases/004-various-update/index.tsx
+++ b/benchmark/react/cases/004-various-update/index.tsx
@@ -1,0 +1,204 @@
+// Copyright 2025 The Lynx Authors. All rights reserved.
+// Licensed under the Apache License Version 2.0 that can be found in the
+// LICENSE file in the root directory of this source tree.
+
+import { root, useEffect, useState } from '@lynx-js/react';
+import { SnapshotInstance } from '@lynx-js/react/internal';
+import type { CSSProperties, MainThread, NodesRef } from '@lynx-js/types';
+
+import { hook, isMainThread } from '../../src/hook.js';
+
+if (__MAIN_THREAD__) {
+  hook(
+    SnapshotInstance.prototype,
+    'setAttribute',
+    function(this: SnapshotInstance, old, key, value) {
+      if (!this.__elements) {
+        // skip when there is no underlying FiberElement
+        return old!.call(this, key, value);
+      }
+
+      let name = '';
+      switch (key) {
+        case 0:
+          name = 'Attr';
+          break;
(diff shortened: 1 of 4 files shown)


benchmark/react/cases/004-various-update/index.tsx
```diff
@@ -0,0 +1,204 @@
...
+    SnapshotInstance.prototype,
+    'setAttribute',
+    function(this: SnapshotInstance, old, key, value) {
+      if (!this.__elements) {
+        // skip when there is no underlying FiberElement
+        return old!.call(this, key, value);
+      }
+
+      let name = '';
+      switch (key) {
```
upupming, line 22: Is it guaranteed that there is only one single `SnapshotInstance`?  I see we only switch on `key` but not `this`.

benchmark/react/cases/004-various-update/index.tsx
```diff
@@ -0,0 +1,204 @@
...
+      return ret;
+    },
+  );
+}
+
+function F() {
+  const [stopBenchmark, setStopBenchmark] = useState(false);
+  const [values, setValues] = useState(
+    [
+      'some-exposure-id' as string,
```
upupming, line 91: I am not sure if this string is too short to get a stable benchmark result. If we found false positive we can generate fixed larger string and object here.

hzy: We do not to pressure test to benchmark. The benchmark is underlying using valgrind, which instrument machine code and count its number of instructions executed, number of memory reads / writes, number of cache misses and etc, to CALCULATE the time of benchmark. See https://codspeed.io/docs/instruments/cpu/overview.

benchmark/react/cases/004-various-update/index.tsx
hzy: Yes, the patch code and the benchmark case are tightly coupled and need to be modified together.

PR #1651, opened 2025-09-03
feat(core): set `output.iife` to false to avoid memory leak

luhc228:
## Summary by CodeRabbit

- Bug Fixes
  - Improved build stability for Lynx targets by disabling a bundling wrapper, reducing main-thread memory issues.

- Chores
  - Added changesets for patch releases across related packages.
  - Updated React plugin peer requirements to include @lynx-js/rspeedy ^0.11.1 for compatibility.

- Tests
  - Removed obsolete test cases related to bundling wrapper behavior.

## Checklist


packages/rspeedy/plugin-react/package.json
```diff
@@ -64,7 +64,8 @@
     "typia-rspack-plugin": "2.2.1"
   },
   "peerDependencies": {
-    "@lynx-js/react": "^0.103.0 || ^0.104.0 || ^0.105.0 || ^0.106.0 || ^0.107.0 || ^0.108.0 || ^0.109.0 || ^0.110.0 || ^0.111.0 || ^0.112.0"
+    "@lynx-js/react": "^0.103.0 || ^0.104.0 || ^0.105.0 || ^0.106.0 || ^0.107.0 || ^0.108.0 || ^0.109.0 || ^0.110.0 || ^0.111.0 || ^0.112.0",
+    "@lynx-js/rspeedy": "^0.11.1"
```
colinaaa, line 68: Adding a new `peerDependencies` is a breaking change. We would like to avoid this.

packages/rspeedy/plugin-react/src/pluginReactLynx.ts
```diff
@@ -405,14 +405,6 @@ export function pluginReactLynx(
             })
           }
 
-          // This is used to avoid the IIFE in main-thread.js, which would cause memory leak.
```
colinaaa, line 408: This is also a **BREAKING CHANGE**. For those who are using `@lynx-js/rspeedy` < 0.11.2 with latest `@lynx-js/react-rsbuild-plugin`. We would like to keep this for now.

packages/rspeedy/core/src/plugins/chunkLoading.plugin.ts
```diff
@@ -35,6 +35,7 @@ export function pluginChunkLoading(): RsbuildPlugin {
           .output
             .chunkLoading('lynx')
             .chunkFormat('commonjs')
+            .iife(false)
```
colinaaa, line 38: Can we also add some tests about if the default value of `output.iife` can be overwritten (e.g.: by `tools.rspack.output.iife`)

colinaaa (changes requested):

PR #1654, opened 2025-09-04
feat(web): add simple support for `rpx` unit

AtarisMio:
@coderabbitai summary

## Checklist


PupilTong:
String matching is not the best solution. The tokenizer has token type outputs. The transformer only needs to check the dimensions.

packages/web-platform/web-core/src/apis/LynxView.ts
```diff
@@ -501,6 +501,31 @@ export class LynxView extends HTMLElement {
               );
             }
           }
+          {
```
PupilTong: This is incorrect. remove this.
better solution: in shadowroot `--rpx: inherit`

packages/web-platform/web-mainthread-apis/src/createMainThreadGlobalThis.ts
```diff
@@ -798,5 +798,27 @@ export function createMainThreadGlobalThis(
     enumerable: true,
   });
 
+  const style =
```
PupilTong: remove

packages/web-platform/web-style-transformer/src/transformer/transform.rs
```diff
@@ -52,17 +81,17 @@ pub fn query_transform_rules<'a>(
...
+        ("background-clip".to_string(), "text".to_string()),
+        ("--lynx-text-bg-color".to_string(), value.clone()),
       ]);
     } else {
       result.extend([
-        ("--lynx-text-bg-color", "initial"),
-        ("-webkit-background-clip", "initial"),
-        ("background-clip", "initial"),
-        ("color", value),
+        ("--lynx-text-bg-color".to_string(), "initial".to_string()),
```
PupilTong: why?

packages/web-platform/web-tests/tests/react-rpx-style-transformer.spec.ts
```diff
@@ -0,0 +1,333 @@
+// Copyright 2023 The Lynx Authors. All rights reserved.
```
PupilTong: do not create a new one, reuse react.spec.ts

packages/web-platform/web-tests/tests/web-style-transformer.spec.ts
```diff
@@ -0,0 +1,499 @@
+// Copyright 2024 The Lynx Authors. All rights reserved.
```
PupilTong: ditto

packages/web-platform/web-core/src/apis/LynxView.ts
AtarisMio: it should not be a default `--rpx` value for each lynx-view?

packages/web-platform/web-style-transformer/src/transformer/transform.rs
AtarisMio: Because the `transformed_source` generated by the `parser` I introduced is a `String`. I changed the type of `CSSPair`

```
// type CSSPair<'a> = (&'a str, &'a str);
type CSSPair = (String, String);
```

There might be a better way to handle it?

PR #1655, opened 2025-09-04
ci(benchmark/react): background setAttribute `BatchedValues`

hzy:
## Summary by CodeRabbit

- New Features
  - Added background benchmarking hooks and instrumentation for batched attribute updates; runtime now records executed benchmarks when triggered.

- Bug Fixes
  - Instrumentation is gated so it only runs when the benchmarking environment is present, avoiding unintended overhead.

- Chores
  - Simplified thread-detection logic and expanded runtime surface to include background snapshot handling.

## Checklist

### DIFF
--- a/benchmark/react/cases/004-various-update/index.tsx
+++ b/benchmark/react/cases/004-various-update/index.tsx
@@ -3,12 +3,15 @@
 // LICENSE file in the root directory of this source tree.
 
 import { root, useEffect, useState } from '@lynx-js/react';
-import { SnapshotInstance } from '@lynx-js/react/internal';
+import {
+  BackgroundSnapshotInstance,
+  SnapshotInstance,
+} from '@lynx-js/react/internal';
 import type { CSSProperties, MainThread, NodesRef } from '@lynx-js/types';
 
 import { hook, isMainThread } from '../../src/hook.js';
 
-if (__MAIN_THREAD__) {
+if (typeof Codspeed !== 'undefined' && __MAIN_THREAD__) {
   hook(
     SnapshotInstance.prototype,
     'setAttribute',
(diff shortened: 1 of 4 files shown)


benchmark/react/cases/004-various-update/index.tsx
```diff
@@ -48,18 +51,15 @@ if (__MAIN_THREAD__) {
           name = 'Ref';
           break;
         case 9:
-          name = 'TimingFlag';
-          break;
-        case 10:
           name = 'MT_Ref';
           break;
-        case 11:
+        case 10:
```
upupming: Is there any way to avoid hard coding this index?

benchmark/react/cases/004-various-update/index.tsx
```diff
@@ -84,6 +84,31 @@ if (__MAIN_THREAD__) {
...
+      if (
+        key === 'values' && values[values.length - 1] === 'stop-benchmark-true'
+      ) {
+        // we only care about the update that stops the benchmark
+
+        Codspeed.startBenchmark();
+        const ret = old!.call(this, key, value);
+        Codspeed.stopBenchmark();
+        Codspeed.setExecutedBenchmark(
+          `${__REPO_FILEPATH__}::${__webpack_chunkname__}-setAttribute__BatchedValues`,
```
upupming, line 105: Why we only report `setAttribute__BatchedValues` in background thread, will we add other cases like main thread?

Do we add `BTS::` to avoid naming conflict with main thread?

upupming (dismissed):

benchmark/react/cases/004-various-update/index.tsx
hzy: In background thread, `setAttribute` will only be called with batched values. because we generate snapshot JSX with `values` props.

We do not need to add `BTS::` because `__webpack_chunkname__` helps us distinguish benchmarks

upupming (approved):

benchmark/react/cases/004-various-update/index.tsx
hzy: We didn't generate any info about each value in `values` so sadly no for now :)

PR #1668, opened 2025-09-08
fix(web): svg use img tag to render

Sherry-hue:
@coderabbitai summary

## Checklist


packages/web-platform/web-elements/src/XSvg/x-svg.css
```diff
@@ -8,3 +8,27 @@ x-svg {
   background: no-repeat center;
   background-size: 100% 100%;
 }
+
+x-svg::part(img) {
```
PupilTong, line 11: consider about migrating this with x-image?

packages/web-platform/web-elements/src/XSvg/XSvg.ts
```diff
@@ -7,60 +7,70 @@ import {
...
-      const src = 'data:image/svg+xml;base64,'
-        + btoa(unescape(encodeURIComponent(content)));
-      this.#loadEventInvoker.src = src;
-      return `url("${src}")`;
-    },
-  );
+  #handleContent(content: string | null) {
+    if (!content) return;
+    const src = 'data:image/svg+xml;base64,'
+      + btoa(unescape(encodeURIComponent(content)));
```
colinaaa: Instead of raw data-uri, we should use modern `Blob` API.

Here are some quote from the legacy `<svg>` implementation:
 
> Not using data-uri(`data:image/svg+xml;utf8,${props.content}`)
> since it has follow limitations:
>   1. < and > must be encoded to %3C and %3E.
>   2. Double quotes must be converted to single quotes.
>   3. Colors must use a non-hex format because # will not work inside data-uri.
> See: https://codepen.io/zvuc/pen/BWNLJL
> Instead, we use modern Blob API to create SVG URL that have the same support

```suggestion
    const blob = new Blob([content], {
      type: 'image/svg+xml;charset=UTF-8',
    });
    const src = URL.createObjectURL(blob);
```

packages/web-platform/web-elements/src/XSvg/XSvg.ts
```diff
@@ -7,60 +7,72 @@ import {
...
-      this.#loadEventInvoker.src = src;
-      return `url("${src}")`;
-    },
-  );
+  #handleContent(content: string | null) {
+    if (!content) return;
+    const blob = new Blob([content], {
+      type: 'image/svg+xml;charset=UTF-8',
+    });
+    const src = URL.createObjectURL(blob);
```
PupilTong, line 46: This url needs to be cleaned manually

packages/web-platform/web-elements/src/XSvg/XSvg.ts
Sherry-hue: Thanks! now it will be used as blob

packages/web-platform/web-elements/src/XSvg/XSvg.ts
Sherry-hue: done

Issue #1518, opened 2025-08-13
[Feature]: Support using `env`, `command` in `defineConfig`


colinaaa:
### What problem does this feature solve?

Make it easier to distinguish build env in `lynx.config.js`.

### What does the proposed API of configuration look like?

```ts
import type { Config } from '@lynx-js/rspeedy'

function defineConfig(config: ({ env, command }: { env: string, command: 'build' | 'dev' }) => Config | Promise<Config>): Config
```

The same as [Rsbuild](https://rsbuild.rs/guide/configuration/rsbuild#export-function)

PR #1669, opened 2025-09-08
feat(rspeedy/core): support config paramenters

colinaaa:
## Summary by CodeRabbit

- New Features
  - Config files can now export a function that receives { command, env } (sync or async), enabling conditional settings based on CLI command and environment.
  - Sensible defaults applied when not provided: command from CLI (or “build”) and env from NODE_ENV (or “production”).
  - Allows toggling behaviors like minification and build caching based on context.

- Documentation
  - Changelog updated to reflect support for command and env in configuration functions.
  - API docs adjusted to hide internal helper types from navigation.

close: #1518

## Checklist


packages/rspeedy/core/src/config/defineConfig.ts
```diff
@@ -4,14 +4,48 @@
...
+   *
+   * Possible values:
+   *
+   * - `'build'`
+   *
+   * - `'dev'`
+   *
+   * - `'preview'`
+   */
+  command: string
```
upupming: Why don't we use union type just like `env`:

```suggestion
  command: 'build' | 'dev' | 'preview' | (string & Record<never, never>)
```

packages/rspeedy/core/src/config/loadConfig.ts
```diff
@@ -136,20 +137,22 @@ export async function loadConfig(
...
-    const content = validate(
-      'default' in exports ? exports.default : exports,
-      configPath,
-    )
+    const configExport = 'default' in exports ? exports.default : exports
+
+    const rawContent: Config = typeof configExport === 'function'
+      ? await configExport({
+        command: process.argv[2] ?? 'build',
+        env: process.env['NODE_ENV'] ?? 'production',
```
upupming, line 149: What if user runs `rspeedy dev --mode production`, shall we pass `production` to user?

colinaaa: This is the same as `rsbuild dev --mode production` (with `env: 'development'`)

image

packages/rspeedy/core/src/config/defineConfig.ts
colinaaa: I've updated the types.

PR #1677, opened 2025-09-08
feat(web-platform): migrate environment.web output format and update tests for runtime wrapper plugin

PupilTong:
## Summary by CodeRabbit

* **New Features**
  * Unified web runtime/output format with iOS/Android for consistent behavior across platforms.
  * Runtime wrapper now enabled on web builds for improved compatibility.
  * Note: Requires upgrading your web platform version.

* **Tests**
  * Updated web tests to validate presence of the runtime wrapper.

* **Chores**
  * Bumped package version to reflect changes.
  * Updated encoded web bundle format to include version 2.

## Checklist

### DIFF
--- a/.changeset/plenty-actors-cheer.md
+++ b/.changeset/plenty-actors-cheer.md
@@ -1,0 +1,15 @@
+---
+"@lynx-js/react-rsbuild-plugin": minor
+---
+
+feat(web-platform): migrate environment.web output format and update tests for runtime wrapper plugin
+
+**Breaking Change, you will need to upgrade your web platform version**
+
+Before this commit, the environment.web output has its own chunk format.
+
+After this commit, we migrate those into the format running on iOS/Android.
+
+The new format has such special features:
+
+- some global variables, is exported by the IIFE wrapper
--- a/packages/rspeedy/plugin-react/src/entry.ts
+++ b/packages/rspeedy/plugin-react/src/entry.ts
@@ -196,6 +196,29 @@ export function applyEntry(
     const enableChunkSplitting =
       rsbuildConfig.performance?.chunkSplit?.strategy !== 'all-in-one'
 
+    chain
+      .plugin(PLUGIN_NAME_RUNTIME_WRAPPER)
+      .use(RuntimeWrapperWebpackPlugin, [{
+        injectVars(vars) {
+          const UNUSED_VARS = new Set([
+            'Card',
+            'Component',
+            'ReactLynx',
+            'Behavior',
+          ])
+          return vars.map(name => {
+            if (UNUSED_VARS.has(name)) {
+              return `__${name}`
+            }
+            return name
+          })
+        },
+        targetSdkVersion,
+        // Inject runtime wrapper for all `.js` but not `main-thread.js` and `main-thread.[hash].js`.
+        test: /^(?!.*main-thread(?:\.[A-Fa-f0-9]*)?\.js$).*\.js$/,
+        experimental_isLazyBundle,
+      }])
+      .end()
     if (isLynx) {
       let inlineScripts
       if (experimental_isLazyBundle) {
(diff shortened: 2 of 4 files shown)


Sherry-hue (dismissed):

packages/rspeedy/plugin-react/src/entry.ts
```diff
@@ -200,6 +200,29 @@ export function applyEntry(
     const enableChunkSplitting =
       rsbuildConfig.performance?.chunkSplit?.strategy !== 'all-in-one'
 
+    chain
```
colinaaa, line 199: This is also a **BREAKING CHANGE**. The new version of plugin would not work with old Rspeedy (with `output.chunkLoading: 'import-scripts'`).

Either:

1. Mark this as a minor change for the plugin.
2. Decide whether to use `RuntimeWrapperWebpackPlugin` based on `output.chunkLoading`

packages/webpack/template-webpack-plugin/src/WebEncodePlugin.ts
```diff
@@ -73,6 +73,7 @@ export class WebEncodePlugin {
             },
             customSections: encodeData.customSections,
             cardType: encodeData.sourceContent.dsl.substring(0, 5),
+            appType: encodeData.sourceContent.appType,
```
colinaaa: We need a changeset for this. Otherwise, the template plugin would not be published.

packages/rspeedy/core/src/plugins/chunkLoading.plugin.ts
```diff
@@ -7,36 +7,23 @@ import type { RsbuildPlugin } from '@rsbuild/core'
...
+        // dprint-ignore
+        chain
+        .plugin('lynx:chunk-loading')
+          .use(ChunkLoadingWebpackPlugin)
+        .end()
+        .plugin('lynx:cache-events')
+          .use(LynxCacheEventsPlugin)
+        .end()
+        .output
+          .chunkLoading('lynx')
```
colinaaa: I think it would be better to not change the value of `chunkLoading` in `rspeedy/core`. This could avoid making a breaking change.

.changeset/plenty-actors-cheer.md
```diff
@@ -0,0 +1,19 @@
+---
+"@lynx-js/react-rsbuild-plugin": minor
+"@lynx-js/rspeedy": minor
+---
+
+feat(web-platform): use same chunk loading and chunk format with iOS/Android
```
colinaaa: We should do better on this changeset (CHANGELOG).

1. If we are making **BREAKING CHANGE** for two packages, try to split the changeset so that we could know what has changed for this package and how to migrate.

2. Make sure this is for end-user, not maintainer. I don't think most of the user would care about "chunk format" or "IIFE function wrapper".

.changeset/plenty-actors-cheer.md
```diff
@@ -0,0 +1,19 @@
...
+
+**Breaking Change, you will need to upgrade your web platform version**
```
colinaaa, line 7: This is not the only BREAKING CHANGE. The react-rsbuild-plugin would also require `@lynx-js/rspeedy` v0.12.0 (the new minor version).

colinaaa (changes requested):
I would prefer the following way (try our best to avoid breaking changes):

1. In react-rsbuild-plugin, add RuntimeWrapperWebpackPlugin based on `output.chunkLoading`:

```js
if (output.chunkLoading === 'lynx') {
  appendPlugin(RuntimeWrapperWebpackPlugin)
}
```

2. In Rspeedy, add ChunkLoadingWebpackPlugin in web environments.

In this way, the user could opt-in to use the new chunk loading by

1. Upgrade Rspeedy and react-rsbuild-plugin
2. Set `tools.rspack.output.chunkLoading` to `'lynx'`

And we avoid making **BREAKING CHANGES** to the existing projects.

We may change the default value later when we found this stable.

PR #1680, opened 2025-09-08
ci(benchmark): use `uv` instead of `python` directly

colinaaa:
@coderabbitai summary

## Checklist


packages/lynx/benchx_cli/scripts/build.mjs
```diff
@@ -104,13 +104,17 @@ git cherry-pick -n ${PICK_COMMIT}
...
+source .venv/bin/activate
+uv pip install pyyaml
 source tools/envsetup.sh
 ../habitat/venv/bin/hab sync .
 `.pipe(process.stdout);
 
 // build from source
 await $`
 cd lynx
+source .venv/bin/activate
```
colinaaa: I think this is unnecessary. But I don't want to try again :)

hzy (approved):

PR #1692, opened 2025-09-09
fix: should update signMap before __FlushElementTree

gaoachao:
## Summary by CodeRabbit

- Bug Fixes
  - Updated list rendering to update internal mappings earlier, preventing missing or corrupted items in deferred or reused list-item scenarios.

- Tests
  - Added tests that verify correct ordering between list mapping updates and element-tree flush for both standard and reuse flows.

- Chores
  - Patch release for @lynx-js/react; no public API changes.

## Checklist

### DIFF
--- a/.changeset/healthy-maps-clean.md
+++ b/.changeset/healthy-maps-clean.md
@@ -1,0 +1,7 @@
+---
+"@lynx-js/react": patch
+---
+
+Fix "TypeError: cannot read property 'o' of undefined" in deferred list-item scenarios.
+
+Deferred `componentAtIndex` causes nodes that quickly appear/disappear to be enqueued without `__elements`. Update `signMap` before `__FlushElementTre` to resolve the issue.
--- a/.changeset/healthy-maps-clean.md
+++ b/.changeset/healthy-maps-clean.md
@@ -4,4 +4,4 @@
 
 Fix "TypeError: cannot read property '0' of undefined" in deferred list-item scenarios.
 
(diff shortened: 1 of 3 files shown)


.changeset/healthy-maps-clean.md
```diff
@@ -0,0 +1,7 @@
+---
+"@lynx-js/react": patch
+---
+
+Fix "TypeError: cannot read property 'o' of undefined" in deferred list-item scenarios.
```
hzy:
```suggestion
Fix "TypeError: cannot read property '0' of undefined" in deferred list-item scenarios.
```

packages/react/runtime/__test__/list.test.jsx
```diff
@@ -3032,6 +3032,113 @@ describe('list componentAtIndexes', () => {
...
+    const __originalFlushElementTree = globalThis.__FlushElementTree;
+    globalThis.__FlushElementTree = (_, options = {}) => {
+      mockFlushElementTree(options);
+    };
+
+    const listID = __GetElementUniqueID(listRef);
+    const signMap = gSignMap[listID];
+    const recycleMap = gRecycleMap[listID];
+
+    const __originalSet = signMap.set;
```
hzy: Maybe use `spyOn` instead?

gaoachao: see: https://github.com/lynx-family/lynx-stack/pull/1692/commits/3a98e653b8b7fc8246daf666585f3e8e59dd4d45

PR #1705, opened 2025-09-09
fix: disable parallel element by default

upupming:
@coderabbitai summary

## Checklist


PupilTong (dismissed):

.changeset/clever-maps-wait.md
```diff
@@ -0,0 +1,6 @@
+---
+"@lynx-js/template-webpack-plugin": minor
```
colinaaa, line 2: We should update the `peerDependencies` to avoid `major` change like https://github.com/lynx-family/lynx-stack/pull/1705#issuecomment-3270811095

.changeset/clever-maps-wait.md
```diff
@@ -0,0 +1,6 @@
...
+"@lynx-js/react-rsbuild-plugin": minor
+---
+
+**BREAKING CHANGE:** Remove the `enableParallelElement` and `pipelineSchedulerConfig` options. Since thread element resolution is still in experimental stage and may have stability risks, it will be disabled by default after this change.
```
colinaaa: Let's make it two lines for better readability

```suggestion
**BREAKING CHANGE:** Remove the `enableParallelElement` and `pipelineSchedulerConfig` options.

Since the thread element resolution is still in experimental stage and may have stability risks, it will be disabled by default after this change.
```

colinaaa (approved):

PR #1739, opened 2025-09-12
fix(web): temporarily remove support for chunk split

PupilTong:
## Summary by CodeRabbit

* **Revert**
  * Temporarily removed chunk splitting support to prevent runtime issues in split chunks while a robust fix is prepared.
* **Bug Fixes**
  * App loading now fails fast with a clear error when a requested resource is missing from the manifest, avoiding silent failures and aiding diagnosis.
* **Tests**
  * Skipped split-chunk related tests while the feature is disabled to keep test suite stable.

## Checklist

### DIFF
--- a/.changeset/many-actors-count.md
+++ b/.changeset/many-actors-count.md
@@ -1,0 +1,7 @@
+---
+"@lynx-js/web-core": minor
+---
+
+break(web): temporary remove support for chunk split
+
+Since the global variables cannot be accessed in the splited chunk, we temporary remove supporting for chunk spliting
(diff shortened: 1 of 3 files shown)


.changeset/many-actors-count.md
```diff
@@ -0,0 +1,7 @@
+---
+"@lynx-js/web-core": minor
+---
+
+break(web): temporary remove support for chunk split
```
colinaaa, line 5: Can we add a simple migration guide here?

```js
import { defineConfig } from '@lynx-js/rspeedy'

export default defineConfig({
  performance: {
    chunkSplit: {
      strategy: 'all-in-one',
    },
  },
})
```

colinaaa (dismissed):

PR #1458, opened 2025-08-07
refactor: integrate swc wasm plugin

gaoachao:
@coderabbitai summary

## Checklist

- [ ] Tests updated (or not required).
- [ ] Documentation updated (or not required).
- [ ] Changeset added, and when a BREAKING CHANGE occurs, it needs to be clearly marked (or not required).

PR #1744, opened 2025-09-12
refactor: add `swc-plugin-reactlynx-compat`

gaoachao:
A cherry-pick of #1458.
Follow up #1740, so we should merge it first.

## Summary by CodeRabbit

* **New Features**
  * Added a configurable React Lynx compatibility transform: JSX rewriting, component wrapping, event-prop normalization, runtime selection, dark-mode and related config options.

* **Tests**
  * Expanded snapshot coverage for wrapping, embedded components, runtime changes, event props, view renames, constructor simplification, recursion, and scope/aliasing.

* **Chores**
  * Broadened lint/format ignore patterns, added build/task config for the transform, updated workspace inclusion and build flags for WASM artifact.

## Checklist

### DIFF
--- a/packages/react/.dprint.jsonc
+++ b/packages/react/.dprint.jsonc
@@ -11,7 +11,6 @@
     "transform/index.d.ts",
     "transform/index.js",
 
-
     // REMOVE ME
     "transform/tests/__swc_snapshots__/**",
     // Snapshot
--- a/packages/react/transform/swc-plugin-reactlynx-compat/build.js
+++ b/packages/react/transform/swc-plugin-reactlynx-compat/build.js
@@ -10,11 +10,12 @@ import { fileURLToPath } from 'node:url';
 const __filename = fileURLToPath(import.meta.url);
 const __dirname = path.dirname(__filename);
 
+const RUSTFLAGS = `${
+  process.env.RUSTFLAGS ? process.env.RUSTFLAGS + ' ' : ''
+}-C link-arg=--export-table -C link-arg=-s`;
+
 execSync('cargo build-wasi --release', {
-  env: {
-    ...process.env,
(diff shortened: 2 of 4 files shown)


packages/react/transform/swc-plugin-reactlynx-compat/src/lib.rs
```diff
@@ -1 +1,1870 @@
+use convert_case::{Case, Casing};
```
colinaaa, line 1: Instead of copying the whole file, can we re-use the existing mod or crate? E.g.:

- crates/swc_plugin_* <-- pure Rust crates that contains swc plugins
- packages/react/transform <-- napi-rs binding
- packages/react/transform/swc-plugin-* <-- WASM plugin binding

gaoachao: While this is theoretically achievable, it would require some refactoring costs:

1. The current `swc_plugin_*` mods are not completely pure - they contain `#[napi(object)]`
2. In the WASM plugin, we would need `serde` for serialization and deserialization
3. If napi-rs bindings are to be completely removed in the future, this refactoring would be redundant

However, the current approach of directly copying source code has some drawbacks:
1. Need to maintain consistency between two codebases
2. Loss of commit history, making git blame unavailable

colinaaa: > The current swc_plugin_* mods are not completely pure - they contain #[napi(object)]

Yes it is. So what you need to do is to extract the swc plugin into a pure Rust crate (which could be reused by napi and WASM plugin). And handling the napi binding in the napi crate.

> If napi-rs bindings are to be completely removed in the future, this refactoring would be redundant

We may still making changes before finishing the refactor. And we could make pure Rust benchmark & tests after this refactor.

PR #1757, opened 2025-09-16
fix(react/runtime): prevent `reloadVersion` from cancelling the MTRef initial value patch

Yradex:
@coderabbitai summary

## Checklist


hzy (approved):
LGTM!

PR #1768, opened 2025-09-17
feat(react/runtime): support `SelectorQuery` `animation` APIs

Yradex:
## Summary by CodeRabbit

- New Features
  - Support for SelectorQuery animation APIs (animate, play/pause/cancel).
- Bug Fixes
  - More reliable flushing/execution of delayed UI operations to avoid stuck tasks.
- Refactor
  - Streamlined ref method forwarding for consistent, less-boilerplate chaining behavior.
- Tests
  - Added tests for delayed UI ops and improved snapshot serializers for proxy-like objects.
- Chores
  - Added release metadata for a patch update.

## Checklist

### DIFF
--- a/packages/react/runtime/__test__/snapshot/ref.test.jsx
+++ b/packages/react/runtime/__test__/snapshot/ref.test.jsx
@@ -6,6 +6,8 @@
 import { render } from 'preact';
 import { afterEach, beforeAll, beforeEach, describe, expect, it, vi } from 'vitest';
 
+import { runDelayedUiOps, shouldDelayUiOps } from '../../src/lifecycle/ref/delay';
+
 import { Component, createRef, useState } from '../../src/index';
 import { clearCommitTaskId, replaceCommitHook } from '../../src/lifecycle/patch/commit';
 import { injectUpdateMainThread } from '../../src/lifecycle/patch/updateMainThread';
@@ -1907,5 +1909,17 @@ describe('ui operations', () => {
         ]
       `);
     }
+  });
+});
+
+describe('runDelayedUiOps helper', () => {
+  it('should reset shouldDelayUiOps when no tasks queued', () => {
+    // flush any queued tasks from previous tests
+    shouldDelayUiOps.value = false;
+    runDelayedUiOps();
+
+    shouldDelayUiOps.value = true;
+    runDelayedUiOps();
+    expect(shouldDelayUiOps.value).toBe(false);
   });
 });
--- a/packages/react/runtime/__test__/snapshot/ref.test.jsx
+++ b/packages/react/runtime/__test__/snapshot/ref.test.jsx
@@ -6,7 +6,7 @@
 import { render } from 'preact';
 import { afterEach, beforeAll, beforeEach, describe, expect, it, vi } from 'vitest';
 
-import { runDelayedUiOps, shouldDelayUiOps } from '../../src/lifecycle/ref/delay';
+import { RefProxy, runDelayedUiOps, shouldDelayUiOps } from '../../src/lifecycle/ref/delay';
 
 import { Component, createRef, useState } from '../../src/index';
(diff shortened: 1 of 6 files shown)


packages/react/runtime/src/lifecycle/ref/delay.ts
```diff
@@ -56,34 +64,41 @@ class RefProxy {
 
   constructor(refAttr: [snapshotInstanceId: number, expIndex: number]) {
     this.refAttr = refAttr;
+    this.task = undefined;
+
+    return new Proxy(this, {
```
upupming, line 69: Can we add a codspeed benchmark to ref (cc @hzy ), I am not sure Proxy will slow down performance.

packages/react/runtime/src/lifecycle/ref/delay.ts
```diff
@@ -39,11 +45,13 @@ function runOrDelay(task: () => void): void {
  * Executes all delayed UI operations.
  */
 function runDelayedUiOps(): void {
-  for (const task of delayedUiOps) {
+  const tasks = delayedUiOps.slice();
```
upupming, line 48: `slice` may take some cost, why do we use slice here?

packages/react/runtime/src/lifecycle/ref/delay.ts
```diff
@@ -56,34 +64,41 @@ class RefProxy {
...
+        if (
+          typeof prop === 'symbol'
+          || prop === 'then'
+          || prop in target
+          || typeof prop !== 'string'
+        ) {
+          return Reflect.get(target, prop, receiver);
+        }
+
+        const forward = <K extends ForwardableNodesRefMethod>(method: K) => {
```
upupming, line 80: Shall we add some type testing to ensure field like `playAnimation` is existed here?

packages/react/runtime/src/lifecycle/ref/delay.ts
Yradex: @coderabbitai should we use slice here?

packages/react/runtime/__test__/snapshot/ref.test.jsx
```diff
@@ -1909,3 +1911,15 @@ describe('ui operations', () => {
     }
   });
 });
+
+describe('runDelayedUiOps helper', () => {
```
hzy, line 1982: The PR titles itself as supporting animation api but in the test there is no assert about the api can be called on the `Proxy`ed SelectorQuery, could u pls add some?

packages/react/runtime/src/lifecycle/ref/delay.ts
Yradex: Didn't add tests for `playAnimation` because I figured the existing ones for stuff like invoke and fields were enough. Since they all just get proxied by `RefProxy`, testing one should cover it. Can add more if you'd like, though!

Yradex: @coderabbitai Can you help to generate unit tests for calling the following animation apis on `RefProxy`?

```js
export interface KeyframeEffectV2 {
  readonly keyframes: Array<Record<string, any>>;
  readonly options: Record<string, any>;
}

export interface AnimationV2 {
  readonly effect: KeyframeEffectV2;
  readonly id: string;
}

animate(animations: AnimationV2[] | AnimationV2): SelectorQuery;

playAnimation(ids: string[] | string): SelectorQuery;

pauseAnimation(ids: string[] | string): SelectorQuery;

cancelAnimation(ids: string[] | string): SelectorQuery;
```

packages/react/runtime/__test__/snapshot/ref.test.jsx
Yradex: See https://github.com/lynx-family/lynx-stack/pull/1768#discussion_r2354932477

hzy (approved):

PR #1774, opened 2025-09-18
feat(tailwind-preset): enable lynxUIPlugins by default

Dugyu:
## Summary

Enable `lynxUIPlugins` by default in the Tailwind preset, so that UI plugins (such as `uiVariants`) are always available without extra configuration.  
This makes component-state styling easier out of the box.

## Motivation

- Lynx core currently lacks pseudo-selectors and data-attribute selectors.  
- `uiVariants` and similar UI plugins bridge this gap, providing a recommended way to express component states.  
- Default-on ensures developers don't need to opt in to essential functionality.

## Details

- Changed default of `createLynxPreset({ lynxUIPlugins })` from `false` → `true`.  
- Per-plugin configuration and opt-out remain supported.
- Added tests covering default enablement, explicit opt-out, empty-config behavior, and defensive/error logging scenarios.
- README and plugin docs updated to explain default-enabled behavior, how to disable globally or per-plugin, configuration examples, advanced customization, and version notes.  
- Added a changeset for a minor release and simplified enablement UX.
  
## Future Compatibility

If Lynx later adds attribute selectors or pseudo-selectors at the core level, this decision remains compatible.  
UI plugins will still work as extensions, but users may choose to disable them explicitly.

## Checklist

### DIFF
--- a/packages/third-party/tailwind-preset/src/core.ts
+++ b/packages/third-party/tailwind-preset/src/core.ts
@@ -193,21 +193,6 @@ export function toEnabledSet(
   return set;
 }
 
-export function toEnabledLynxUIPluginSet(
-  opt: LynxUIPluginsOption = true,
-): Set<LynxUIPluginName> {
-  if (opt === true) return new Set(ORDERED_LYNX_UI_PLUGIN_NAMES);
-  if (opt === false) return new Set();
-  if (Array.isArray(opt)) return new Set(opt);
-
-  const set = new Set(ORDERED_LYNX_UI_PLUGIN_NAMES);
-  for (const [k, on] of Object.entries(opt)) {
-    if (on === false) set.delete(k as LynxUIPluginName);
-    else if (on === true) set.add(k as LynxUIPluginName);
-  }
-  return set;
-}
-
 export function resolveUIPluginEntries(
   raw: LynxUIPluginsOption,
 ): {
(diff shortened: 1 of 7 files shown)


packages/third-party/tailwind-preset/src/lynx.ts
```diff
@@ -20,13 +19,50 @@ import type {
...
+ * // enable debug mode
+ * createLynxPreset({ debug: true });
+ *
+ * @since 0.1.0
  */
-
 function createLynxPreset({
   lynxPlugins = true,
-  lynxUIPlugins = false,
+  lynxUIPlugins = true,
```
Dugyu, line 65: Core change of this PR.

packages/third-party/tailwind-preset/src/core.ts
```diff
@@ -193,21 +193,6 @@ export function toEnabledSet(
   return set;
 }
 
-export function toEnabledLynxUIPluginSet(
```
Dugyu, line 196: Make sure `lynUIPluginOptions` is resolved only in `resolveUIPluginEntries`, keep single source of truth.

packages/third-party/tailwind-preset/src/lynx.ts
```diff
@@ -53,10 +88,24 @@ function createLynxPreset({
 
   // Lynx UI Plugins
   for (const [name, options] of resolveUIPluginEntries(lynxUIPlugins)) {
-    if (uiSetEnabled.has(name)) {
-      const fn = LYNX_UI_PLUGIN_MAP[name];
+    const fn = LYNX_UI_PLUGIN_MAP[name];
+
+    // Invariant: map must contain every ordered name
+    if (!fn) {
```
Dugyu, line 94: Defensive branches.

gaoachao (approved):

Issue #1691, opened 2025-09-09
[Feature]: Support `server.compress`


colinaaa:
### What problem does this feature solve?

Configure whether to enable [gzip compression](https://developer.mozilla.org/en-US/docs/Glossary/gzip_compression) for static assets served by the dev server or preview server.

### What does the proposed API of configuration look like?

The same as [Rsbuild](https://rsbuild.rs/config/server/compress):

```js
export default {
  server: {
    compress: false,
  },
};
```

PR #1799, opened 2025-09-20
feat(rspeedy): Support `server.compress`

babu-ch:
https://github.com/lynx-family/lynx-stack/issues/1691

@coderabbitai summary

## Checklist

### DIFF
--- a/.changeset/evil-pigs-doubt.md
+++ b/.changeset/evil-pigs-doubt.md
@@ -1,0 +1,5 @@
+---
+"@lynx-js/rspeedy": patch
+---
+
+Support server.compress
--- a/.changeset/evil-pigs-doubt.md
+++ b/.changeset/evil-pigs-doubt.md
@@ -2,4 +2,4 @@
 "@lynx-js/rspeedy": patch
 ---
 
-Support server.compress
+Support `server.compress`
--- a/packages/rspeedy/core/src/config/rsbuild/index.ts
+++ b/packages/rspeedy/core/src/config/rsbuild/index.ts
@@ -88,6 +88,8 @@ export function toRsbuildConfig(
     server: {
       base: config.server?.base,
 
+      compress: config.server?.compress ?? true,
+
       headers: config.server?.headers,
 
       host: config.server?.host,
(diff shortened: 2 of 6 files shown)


.changeset/evil-pigs-doubt.md
```diff
@@ -0,0 +1,5 @@
+---
+"@lynx-js/rspeedy": patch
+---
+
+Support server.compress
```
colinaaa:
```suggestion
Support `server.compress`
```

packages/rspeedy/core/src/config/rsbuild/index.ts
```diff
@@ -88,6 +88,8 @@ export function toRsbuildConfig(
     server: {
       base: config.server?.base,
 
+      compress: config.server?.compress ?? true,
```
colinaaa: Maybe we could remove the `true` and respect the default value of Rsbuild.
```suggestion
      compress: config.server?.compress,
```

packages/rspeedy/core/test/config/rsbuild.test.ts
```diff
@@ -618,6 +618,15 @@ describe('Config - toRsBuildConfig', () => {
...
   })
 
+  describe('Server', () => {
+    test('transform empty server.compress', () => {
+      const rsbuildConfig = toRsbuildConfig({
+        server: {},
+      })
+      expect(rsbuildConfig.server?.compress).toBe(true)
+    })
+  })
```
colinaaa: Can we also add some validation tests? Especially for using object format.

packages/rspeedy/core/src/config/rsbuild/index.ts
babu-ch: That was certainly the case!
https://github.com/lynx-family/lynx-stack/pull/1799/commits/6c96039ff70745a49e7660103d6975996547aea6

packages/rspeedy/core/test/config/rsbuild.test.ts
babu-ch: Is this test okay?
https://github.com/lynx-family/lynx-stack/pull/1799/commits/308d0cc5d7480e6119f5f1099aa30007afc5ad67

colinaaa: Actually I mean adding tests in `validate.test.ts` :) But it's fine to keep these tests.

PR #1813, opened 2025-09-23
feat: delay first componentDidMount after hydrate

upupming:
@coderabbitai summary

In #1111, we accelerate the timing of didMount. But it may make `hydrate`/`firstScreen` in BTS happens more later than before, thus event handlers are later to be executed. We want to make hydrate before first componentDidMount again.

## Checklist

### DIFF
--- a/.changeset/strong-wasps-smell.md
+++ b/.changeset/strong-wasps-smell.md
@@ -1,0 +1,5 @@
+---
+"@lynx-js/react": patch
+---
+
+Delay first componentDidMount after hydrate
--- a/packages/react/runtime/src/lifecycle/patch/commit.ts
+++ b/packages/react/runtime/src/lifecycle/patch/commit.ts
@@ -20,12 +20,14 @@
  */
 
 import { options } from 'preact';
+import type { VNode } from 'preact';
+import type { Component } from 'preact/compat';
 
 import type { RunWorkletCtxData } from '@lynx-js/react/worklet-runtime/bindings';
 
 import { LifecycleConstant } from '../../lifecycleConstant.js';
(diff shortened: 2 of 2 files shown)


.changeset/strong-wasps-smell.md
```diff
@@ -0,0 +1,5 @@
+---
+"@lynx-js/react": patch
+---
+
+Delay first componentDidMount after hydrate
```
HuJean, line 5: Delay by default? Will it break the original behavior?

upupming: It will BREAK the original behavior in #1111. But it is only a experiment now, let us decide if we should accept this change after performance result. (If yes, we can also add config for this behavior)

packages/react/runtime/src/lifecycle/patch/commit.ts
```diff
@@ -118,13 +120,50 @@ function replaceCommitHook(): void {
...
+      let renderCallbacks:
+        | Array<
+          {
+            component: Component<any, {}>;
+            [RENDER_CALLBACKS]: ((this: Component<any, {}>) => void)[];
+            [VNODE]: VNode<any> | null | undefined;
+          }
+        >
+        | undefined;
+      if (isFirst) {
```
gaoachao, line 134: Should we use an experimental flag like `experimental_delayInitialMount`?

.changeset/strong-wasps-smell.md
HuJean: Different cases may expect different behaviors. Maybe we can add config by default to customize different behaviors.

Issue #1806, opened 2025-09-22
[Feature] Improve Web Platform development experiences


PupilTong:
- [ ] HMR https://github.com/lynx-family/lynx-stack/issues/140
    - [x] Javascript
    - [ ] CSS
- [ ] Ship web platform shell project in rspeedy dev mode
    - [x] implement a dev middleware for rsbuild dev server to provide the  shell project 
    - [x] enable the middleware on dev
    - [x] preview service shipped with rspeedy 
    - [ ] lazy encode on dev mode
    - [ ] rspeedy uses http headers to infer responses with web/native template
- [ ] add new MCP server for Lynx development with Web Platform

PupilTong:
Note for MCP:
I'm not sure if the puppeteer is essential. Not sure if I can create a MCP inside the browser.

adamsimonini:
This would be a fantstic feature to ship. Cheering for you!

PR #1816, opened 2025-09-23
feat: enhance QR code plugin with web environment support and middleware integration

PupilTong:
#1806 

## Summary by CodeRabbit

- New Features
  - Web platform preview added alongside Lynx; QR display and console shortcuts now support both Lynx and Web URLs. Dev URL generation improved to compute web base URLs automatically.

- Documentation
  - Added guidance and example config for enabling Web preview (environments.web / environments.lynx).

- Tests
  - Updated tests to cover multi-platform previews, entry/env permutations, and dual-URL onPrint behavior.

- Chores
  - Added middleware dependency, TypeScript project references, and a changeset entry.

## Checklist

### DIFF
--- a/packages/rspeedy/plugin-qrcode/src/generateDevUrls.ts
+++ b/packages/rspeedy/plugin-qrcode/src/generateDevUrls.ts
@@ -7,12 +7,10 @@ import type { ExposedAPI } from '@lynx-js/rspeedy'
 
 import type { CustomizedSchemaFn } from './index.js'
 
-export default function generateDevUrls(
+function generateFileNameBase(
   api: RsbuildPluginAPI,
-  entry: string,
-  schemaFn: CustomizedSchemaFn,
   port: number,
-): Record<string, string> {
+): { name: string, assetPublicPath: string } {
   const { dev: { assetPrefix } } = api.getNormalizedConfig()
   const { config } = api.useExposed<ExposedAPI>(
     Symbol.for('rspeedy.api'),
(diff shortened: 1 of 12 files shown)


packages/rspeedy/plugin-qrcode/src/generateDevUrls.ts
```diff
@@ -35,15 +35,36 @@ export default function generateDevUrls(
...
     new URL(
       name.replace('[name]', entry).replace('[platform]', 'lynx'),
-      // <port> is supported in `dev.assetPrefix`, we should replace it with the real port
-      assetPrefix.replaceAll('<port>', String(port)),
+      base,
     ).toString(),
   )
 
-  return typeof customSchema === 'string'
+  const outputPathname = new URL(
```
colinaaa: We should respect the `schemaFn` option, even for Web.

packages/rspeedy/plugin-qrcode/src/index.ts
```diff
@@ -138,23 +145,29 @@ export function pluginQRCode(
...
+        const lynxEntries = Object.keys(environments['lynx']?.entry ?? {})
+        const webEntries = Object.keys(environments['web']?.entry ?? {})
+        const allEntries = [
+          ...new Set([
+            ...lynxEntries,
+            ...webEntries,
+          ]),
+        ]
+        if (allEntries.length === 0) {
+          // Not lynx or web environment, skip print
```
colinaaa, line 154: nitpick: the comment does not match the condition.

packages/rspeedy/plugin-qrcode/src/shortcuts.ts
```diff
@@ -32,22 +36,29 @@ export async function registerConsoleShortcuts(
...
     options.api,
     currentEntry,
     options.schema,
     options.port,
   )
 
-  const value: string | symbol = Object.values(devUrls)[0]!
-  await options.onPrint?.(value)
-  showQRCode(value)
+  const hasLynxEntry = options.entries.lynx.includes(currentEntry)
```
colinaaa: Actually, I don't think it's a good idea to have environments mixed-up with entries. It's too complex.

Can we only display the QRCode of Lynx in the terminal, with the `http://<host>:<port>/web` printed as the entrypoint of the whole Web Platform, on which page display links to each entry.

colinaaa (commented):
Like I said in https://github.com/lynx-family/lynx-stack/pull/1816/files#r2384891713. This change would make the QRCode too complex. And also would make it **REQUIRED** for Lynx development.

The QRCode plugin is only made for quickly display the URLs of `.lynx.bundle` that can be scanned with LynxExplorer. Since we already have a web-dev-middleware, I think we could use it for Web Platform development (and even Lynx development).

Here are some common use cases:

### 1. Use Rspeedy without qrcode-plugin

On the terminal:

We use `server.printUrls` to print the URL to the development page (without the `?casename` query)

```
  ➜  Local:    http://localhost:3000/lynx
  ➜  Local:    http://localhost:3000/web
  ➜  Network:  http://<host>:3000/lynx
  ➜  Network:  http://<host>:3000/web
```

#### On the web page (`http://<host>:<port>/web`):

We display URLs for all the entries.
Or we could even create a `<lynx-view>` for each entry.

#### On the lynx page (`http://<host>:<port>/lynx`):

We display QRCode and URLs for all the entires.

### 2. Use Rspeedy with qrcode-plugin

there is nothing special, only the QRCode for the `.lynx.bundle` are displayed on the terminal:

```
●  Scan with Lynx
│
◆ <qrcode>
│
◆  http://10.67.172.118:3000/main2.lynx.bundle?fullscreen=true

  ➜  Local:    http://localhost:3000/lynx
  ➜  Local:    http://localhost:3000/web
  ➜  Network:  http://<host>:3000/lynx
  ➜  Network:  http://<host>:3000/web

◆  Usage
│   r  Switch entries 
│   a  Switch schema 
│   h  Help 
│   q  Quit
```

What do you think

PR #1835, opened 2025-09-25
feat: add event.stopPropagation

f0rdream:
Add `event.stopPropagation` and `event.stopImmediatePropagation` in MTS, to help with event propagation control

```typescript
function App() {
  function handleInnerTap(event: MainThread.TouchEvent) {
    'main thread';
    event.stopPropagation();
    // Or stop immediate propagation with
    // event.stopImmediatePropagation();
  }

  // OuterTap will not be triggered
  return (
    <view main-thread:bindtap={handleOuterTap}>
      <view main-thread:bindtap={handleInnerTap}>
        <text>Hello, world</text>
      </view>
    </view>
  );
}
```

## Summary by CodeRabbit

- **New Features**
  - Event objects now support stopPropagation and stopImmediatePropagation and expose combined propagation result.

- **Bug Fixes**
  - Removed an obsolete runtime warning related to stopPropagation handling.

- **Documentation**
  - Added usage example and changelog guidance; note to update related SDK in lazy-loading setups.

- **Tests**
  - Added tests validating propagation methods, event-result wrapping, and SDK gating.

## Checklist

### DIFF
--- a/.changeset/clear-crabs-swim.md
+++ b/.changeset/clear-crabs-swim.md
@@ -1,0 +1,27 @@
+---
+'@lynx-js/react': patch
+---
+
+Add `event.stopPropagation` and `event.stopImmediatePropagation` in MTS, to help with event propagation control
+
+```tsx
+function App() {
+  function handleInnerTap(event: MainThread.TouchEvent) {
+    'main thread';
+    event.stopPropagation();
+    // Or stop immediate propagation with
+    // event.stopImmediatePropagation();
+  }
+
+  // OuterTap will not be triggered
+  return (
+    <view main-thread:bindtap={handleOuterTap}>
+      <view main-thread:bindtap={handleInnerTap}>
+        <text>Hello, world</text>
+      </view>
+    </view>
+  );
+}
+```
+
+Note, if this feature is used in [Lazy Loading Standalone Project](https://lynxjs.org/react/code-splitting.html#lazy-loading-standalone-project), both the Producer and the Consumer should update to latest version of `@lynx-js/react` to make sure the feature is available.
--- a/packages/react/transform/crates/swc_plugin_compat/lib.rs
+++ b/packages/react/transform/crates/swc_plugin_compat/lib.rs
@@ -1043,22 +1043,8 @@ where
   }
 
   fn visit_mut_call_expr(&mut self, n: &mut CallExpr) {
-    // check if it is e.stopPropagation()
     if let Callee::Expr(e) = &mut n.callee {
       if let Expr::Member(m) = &**e {
-        // check if it is e.stopPropagation
-        if let MemberProp::Ident(id) = &m.prop {
-          if id.sym == "stopPropagation" {
-            HANDLER.with(|handler| {
-              handler
-              .struct_span_warn(
-                n.span,
-                "BROKEN: e.stopPropagation() takes no effect and MUST be migrated in ReactLynx 3.0",
-              )
-              .emit()
-            });
-          }
-        }
         if let Expr::This(_) = &*m.obj {
           if let MemberProp::Ident(id) = &m.prop {
             match id.sym.to_string().as_str() {
--- a/packages/react/worklet-runtime/src/eventPropagation.ts
+++ b/packages/react/worklet-runtime/src/eventPropagation.ts
@@ -1,0 +1,56 @@
+// Copyright 2024 The Lynx Authors. All rights reserved.
+// Licensed under the Apache License Version 2.0 that can be found in the
+// LICENSE file in the root directory of this source tree.
+
+import type { ClosureValueType, EventCtx, RunWorkletOptions } from './bindings/types.js';
+import { RunWorkletSource } from './bindings/types.js';
+
+// EventResult enum values
+export const EventResult = {
+  kDefault: 0x0,
+  kStopPropagationMask: 0x1,
+  kStopImmediatePropagationMask: 0x2,
+} as const;
+
+type EventLike = Record<string, ClosureValueType>;
+
+export function isEventObject(
+  ctx: ClosureValueType[],
+  options?: RunWorkletOptions,
+): ctx is [EventLike, ...ClosureValueType[]] {
+  if (!Array.isArray(ctx) || typeof ctx[0] !== 'object' || ctx[0] === null) {
+    return false;
+  }
+  if (options && options.source === RunWorkletSource.EVENT) {
+    return true;
+  }
+  return false;
+}
+
+/**
+ * Add event methods to an event object if needed
+ * @param ctx The event object to enhance
+ * @param options The run worklet options
+ * @returns A tuple of boolean and the event return result
+ */
+export function addEventMethodsIfNeeded(ctx: ClosureValueType[], options?: RunWorkletOptions): [boolean, EventCtx] {
+  if (!isEventObject(ctx, options)) {
+    return [false, {}];
+  }
+  const eventCtx: EventCtx = {};
+  const eventObj = ctx[0];
+
+  // Add stopPropagation method
+  eventObj['stopPropagation'] = function() {
+    eventCtx._eventReturnResult = (eventCtx._eventReturnResult ?? EventResult.kDefault)
+      | EventResult.kStopPropagationMask;
+  };
+
+  // Add stopImmediatePropagation method
+  eventObj['stopImmediatePropagation'] = function() {
+    eventCtx._eventReturnResult = (eventCtx._eventReturnResult ?? EventResult.kDefault)
+      | EventResult.kStopImmediatePropagationMask;
+  };
+
+  return [true, eventCtx];
+}
--- a/packages/react/worklet-runtime/src/workletRuntime.ts
+++ b/packages/react/worklet-runtime/src/workletRuntime.ts
@@ -2,10 +2,11 @@
 // Licensed under the Apache License Version 2.0 that can be found in the
 // LICENSE file in the root directory of this source tree.
 import { Element } from './api/element.js';
-import type { ClosureValueType, Worklet, WorkletRefImpl } from './bindings/types.js';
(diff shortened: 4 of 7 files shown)


.changeset/clear-crabs-swim.md
```diff
@@ -0,0 +1,25 @@
+---
+'@lynx-js/react': patch
+---
+
+Add `event.stopPropagation` and `event.stopImmediatePropagation` in MTS, to help with event propagation control
+
+```typescript
```
Yradex:
```suggestion
```tsx
```

packages/react/worklet-runtime/src/eventPropagation.ts
```diff
@@ -0,0 +1,52 @@
...
+  kStopImmediatePropagationMask: 0x2,
+} as const;
+
+type EventLike = Record<string, ClosureValueType>;
+
+export function isEventObject(ctx: ClosureValueType[]): ctx is [EventLike, ...ClosureValueType[]] {
+  if (!Array.isArray(ctx) || typeof ctx[0] !== 'object') {
+    return false;
+  }
+  const eventObj = ctx[0] as Record<string, ClosureValueType>;
```
Yradex:
```suggestion
  const eventObj = ctx[0] as Record<string, ClosureValueType> | null;
```

packages/react/worklet-runtime/src/eventPropagation.ts
```diff
@@ -0,0 +1,52 @@
...
+  if (eventObj && 'target' in eventObj && 'currentTarget' in eventObj) {
+    return true;
+  }
+  return false;
+}
+
+/**
+ * Adds event propagation methods to an event object
+ * @param eventObj The event object to enhance
```
Yradex: Doesn't match actual parameters.

packages/react/worklet-runtime/src/workletRuntime.ts
Yradex: I would prefer adding a new parameter to track its source. E.g.

```ts
function runWorklet(ctx: Worklet, params: ClosureValueType[], options?: {source: 'event'}): unknown {
```

packages/react/transform/crates/swc_plugin_compat/lib.rs
```diff
@@ -1043,22 +1043,8 @@ where
...
-            HANDLER.with(|handler| {
-              handler
-              .struct_span_warn(
-                n.span,
-                "BROKEN: e.stopPropagation() takes no effect and MUST be migrated in ReactLynx 3.0",
-              )
-              .emit()
-            });
-          }
-        }
```
gaoachao, line 1061: I think this is necessary for background. We need to determine whether the build target is `JS` or `LEPUS`. You can refer to the implementation here: https://github.com/lynx-family/lynx-stack/blob/main/packages/react/transform/crates/swc_plugin_compat/lib.rs#L1102

.changeset/clear-crabs-swim.md
```diff
@@ -0,0 +1,25 @@
...
+  function handleInnerTap(event: MainThread.TouchEvent) {
+    'main thread';
+    event.stopPropagation();
+    // Or stop immediate propagation with
+    // event.stopImmediatePropagation();
+  }
+
+  // OuterTap will not be triggered
+  return (
+    <view bindtap={handleOuterTap}>
```
hzy:
```suggestion
    <view main-thread:bindtap={handleOuterTap}>
```

.changeset/clear-crabs-swim.md
```diff
@@ -0,0 +1,25 @@
...
+      <view bindtap={handleInnerTap}>
```
hzy:
```suggestion
      <view main-thread:bindtap={handleInnerTap}>
```

.changeset/clear-crabs-swim.md
hzy: Suggestion also applies on git commit message.

packages/react/worklet-runtime/src/workletRuntime.ts
```diff
@@ -70,7 +71,25 @@ function runWorkletImpl(ctx: Worklet, params: ClosureValueType[]): unknown {
...
+  }
+  const result = profile('runWorklet', () => worklet(...params_));
+
+  const eventReturnResult = eventCtx._eventReturnResult;
+
+  // Check if any event object has _eventResult set and include it in return
+  if (isEventObject(params_)) {
+    return {
+      returnValue: result,
+      __EventReturnResult: eventReturnResult,
```
hzy: Can this name still be modified? I think the "__" prefix is ​​unnecessary.

f0rdream: Yes, this can be modified. If modifined, what about using eventReturnResult? cc @Yradex

Yradex: Maybe you could just ignore the return value and only return `eventReturnResult`. It's probably faster and you don't need to give it a name.

Yradex: I just noticed `eventReturnResult` is just a number. I guess it's OK to keep it as what it is now then.

packages/react/worklet-runtime/src/workletRuntime.ts
```diff
@@ -70,7 +72,25 @@ function runWorkletImpl(ctx: Worklet, params: ClosureValueType[]): unknown {
...
-  return profile('runWorklet', () => worklet(...params_));
+
+  const eventCtx: EventCtx = {};
+
+  if (isEventObject(params_, options)) {
+    addEventPropagationMethods(params_, eventCtx);
+  }
+  const result = profile('runWorklet', () => worklet(...params_));
+
+  const eventReturnResult = eventCtx._eventReturnResult;
```
Yradex: Could this assignment be moved inside `if (isEventObject(params_, options)) {}` ?

packages/react/worklet-runtime/src/workletRuntime.ts
```diff
@@ -70,7 +72,25 @@ function runWorkletImpl(ctx: Worklet, params: ClosureValueType[]): unknown {
...
+
+  // Check if any event object has _eventResult set and include it in return
+  if (isEventObject(params_, options)) {
```
Yradex: Is it possible to avoid calling `(isEventObject(params_, options)` twice?
It would be even better to give it a more precise name.

packages/react/worklet-runtime/src/eventPropagation.ts
```diff
@@ -0,0 +1,56 @@
...
+ * @param eventCtx The event context to store the event return result
+ */
+export function addEventPropagationMethods(ctx: [EventLike, ...ClosureValueType[]], eventCtx: EventCtx): void {
+  const eventObj = ctx[0];
+
+  // Add stopPropagation method
+  eventObj['stopPropagation'] = function() {
+    if (!isSdkVersionGt(3, 4)) {
+      throw new Error('stopPropagation requires Lynx sdk version 3.5');
+    }
```
Yradex: I think `stopPropagation()` can now only be called when `options.source` is `Event`. Do we still need these checks?

packages/react/transform/crates/swc_plugin_compat/lib.rs
f0rdream: This compile-time warning is used for RL2 to RL3 migiration. But there's no `stopPropagation` in BTS of RL2, so no need for BTS version of check

packages/react/worklet-runtime/src/eventPropagation.ts
f0rdream: No we don't need that now

packages/react/worklet-runtime/src/workletRuntime.ts
```diff
@@ -48,8 +49,9 @@ function registerWorklet(_type: string, id: string, worklet: (...args: unknown[]
  * Native event touch handler will call this function.
  * @param ctx worklet object.
  * @param params worklet params.
+ * @param options run worklet options.
  */
-function runWorklet(ctx: Worklet, params: ClosureValueType[]): unknown {
+function runWorklet(ctx: Worklet, params: ClosureValueType[], options?: RunWorkletOptions): unknown {
   if (!validateWorklet(ctx)) {
```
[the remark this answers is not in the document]
f0rdream: This is not needed, cause this pass is preserved for comatible, where MTS not supporting Initial Frame Rendering.

Now, this pass will not be touched.

Yradex (approved):

gaoachao (approved):

packages/react/worklet-runtime/src/eventPropagation.ts
```diff
@@ -0,0 +1,56 @@
...
+ */
+export function addEventMethodsIfNeeded(ctx: ClosureValueType[], options?: RunWorkletOptions): [boolean, EventCtx] {
+  if (!isEventObject(ctx, options)) {
+    return [false, {}];
+  }
+  const eventCtx: EventCtx = {};
+  const eventObj = ctx[0];
+
+  // Add stopPropagation method
+  eventObj['stopPropagation'] = function() {
```
upupming, line 44: Will we also support `stopPropagation` for BTS?

Maybe we should also update our docs, currently we advice user uses `catchtap` instead of `stopPropagation`: https://lynxjs.org/react/introduction.html#different-event-naming.

f0rdream: Its not possible to support `stopPropgation` for BTS, due to the aysnc natrue of BTS, we can not simply **stop** an event.

But yeah, the docs needs to be updated. Considering this feature is only avaible on Lynx 3.5, I am working on adding more examples and also update the React part of events.

https://lynxjs.org/next/api/lynx-api/event/event.html  Here's already an update of docs for event

hzy (approved):

PR #1836, opened 2025-09-25
feat(rspeedy/cli): add `--root` option for project directory

### LINKED ISSUE #1817: [Feature]: Support `--root` CLI flag

Set the project root directory (absolute path or relative to [cwd](https://nodejs.org/api/process.html#processcwd))

The same as [Rsbuild](https://rsbuild.rs/guide/basic/cli#common-flags):

```
rspeedy --root ./project-foo
rspeedy -r ./project-foo
rspeedy --root /path/to/project-bar
```

rznn7:
## Summary by CodeRabbit

- **New Features**
  - Added a --root (-r) option to build, dev, inspect, and preview commands.
  - Resolves provided paths (relative or absolute) to an absolute project root.
  - Defaults to the current working directory when --root is not specified.

- **Tests**
  - Added CLI tests covering root handling, argument validation, path resolution, and interaction with other options (e.g., --config).

- **Documentation**
  - Updated CLI docs to document the new --root option.

- **Chores**
  - Added a changeset entry for the patch release.

Fixes #1817

## Checklist

### DIFF
--- a/packages/rspeedy/core/src/cli/commands.ts
+++ b/packages/rspeedy/core/src/cli/commands.ts
@@ -2,6 +2,8 @@
 // Licensed under the Apache License Version 2.0 that can be found in the
 // LICENSE file in the root directory of this source tree.
 
+import path from 'node:path'
+
 import type { RsbuildMode } from '@rsbuild/core'
 import type { Command } from 'commander'
 
@@ -16,6 +18,7 @@ export interface CommonOptions {
   envMode?: string
   noEnv?: boolean
   mode?: RsbuildMode
+  root?: string
 }
 
 function applyCommonOptions(command: Command) {
@@ -36,10 +39,18 @@ function applyCommonOptions(command: Command) {
       '-m --mode <mode>',
       'specify the build mode, can be `development`, `production` or `none`',
     )
+    .option(
+      '-r --root <root>',
+      'set the project root directory (absolute path or relative to cwd)',
+    )
+}
+
+function resolveRoot(cwd: string, root?: string): string {
+  if (!root) return cwd
+  return path.isAbsolute(root) ? root : path.resolve(cwd, root)
 }
 
 export function apply(program: Command): Command {
-  // TODO(cli): support custom cwd
   const cwd = process.cwd()
 
   program
@@ -65,12 +76,11 @@ export function apply(program: Command): Command {
       '--watch',
       'Enable watch mode to automatically rebuild on file changes',
     )
-    .action(
-      (buildOptions: BuildOptions) =>
-        import('./build.js').then(({ build }) =>
-          build.call(buildCommand, cwd, buildOptions)
-        ),
-    )
+    .action(async (buildOptions: BuildOptions) => {
+      const actualRoot = resolveRoot(cwd, buildOptions.root)
+      const { build } = await import('./build.js')
+      return await build.call(buildCommand, actualRoot, buildOptions)
+    })
 
   const devCommand = program.command('dev')
   devCommand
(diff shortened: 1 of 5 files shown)


packages/rspeedy/core/test/cli/root.test.ts
```diff
@@ -0,0 +1,88 @@
+// Copyright 2024 The Lynx Authors. All rights reserved.
```
colinaaa: nitpick:

```suggestion
// Copyright 2025 The Lynx Authors. All rights reserved.
```

packages/rspeedy/core/src/cli/commands.ts
```diff
@@ -66,10 +78,11 @@ export function apply(program: Command): Command {
       'Enable watch mode to automatically rebuild on file changes',
     )
     .action(
-      (buildOptions: BuildOptions) =>
-        import('./build.js').then(({ build }) =>
-          build.call(buildCommand, cwd, buildOptions)
-        ),
+      async (buildOptions: BuildOptions) => {
+        const actualRoot = resolveRoot(cwd, buildOptions.root)
```
colinaaa: Since the `root` has been added to `CommonOptions`, I think it should not only applied to the `build` command, but also the `dev`, `inspect` and `preview` command.

packages/rspeedy/core/test/cli/root.test.ts
```diff
@@ -0,0 +1,88 @@
...
+      'build',
+    ])
+
+    expect(build).toHaveBeenCalledWith(
+      process.cwd(),
+      expect.objectContaining({}),
+    )
+  })
+
+  test('passes relative root path to build command', async () => {
```
colinaaa: Can we also add a test that uses absolute value?

packages/rspeedy/core/test/cli/root.test.ts
```diff
@@ -0,0 +1,88 @@
...
+    const { apply } = await import('../../src/cli/commands.js')
+    const { build } = await import('../../src/cli/build.js')
+
+    program.exitOverride()
+    await apply(program).parseAsync([
+      'node',
+      'rspeedy',
+      'build',
+      '--root',
+      './my-project',
```
colinaaa: This may not work when running test on Windows :)

rznn7:
Added support for each command and some tests.
Are the tests working on Windows now? I tried a fix but I am not sure.

Waiting for your review!

packages/rspeedy/core/test/cli/root.test.ts
rznn7: Not sure it is fixed by my changes. Need feedback please

packages/rspeedy/core/src/config/loadConfig.ts
```diff
@@ -63,7 +63,7 @@ export const resolveConfigPath = (
  */
 export interface LoadConfigOptions {
   configPath?: string | undefined
-  cwd?: string | undefined
+  root?: string | undefined
 }
```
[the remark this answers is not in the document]
colinaaa: @rznn7 I think this make sense. The `LoadConfigOptions.cwd` is a public API. And we don't want to make a **BREAKING CHANGE** on it.

So let's keep the `cwd` instead of renaming to `root`.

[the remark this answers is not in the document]
rznn7: Sure, I understand. I thought it would be clearer but didn't think it would break the public API. So we're not deprecating cwd in favor of a new 'root' either?
I find 'cwd' a bit confusing, as it doesn't precisely represent the current working directory.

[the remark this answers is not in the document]
colinaaa: > So we're not deprecating cwd in favor of a new 'root' either?

Yeah, the API is inherited from [Rsbuild](https://rsbuild.rs/api/javascript-api/core#options). So it would be better to keep the `cwd`.

colinaaa (approved):
LGTM! Thank you. I'll update the PR with the changeset and documentation.

rznn7:
Thanks for your reviews!

PR #458, opened 2025-04-02
feat(rspeedy/core): add `dev.hmr` and `dev.liveReload`

colinaaa:
## Summary

fix: #298

## Checklist

- [x] Tests updated (or not required).
- [x] Documentation updated (or not required).

PR #1882, opened 2025-10-11
feat(rspeedy): apply `dev.hmr` and `dev.liveReload` config to Rsbuild config

luhc228:
Related PR: https://github.com/lynx-family/lynx-stack/pull/458

If we set `dev.hmr` or `dev.liveReload` to `false` in `lynx.config.js`:

```js
import { defineConfig } from '@lynx-js/rspeedy'

export default defineConfig({
  dev: {
    hmr: false,
    liveReload: false,
  }
})
```
Now we will get different value of `dev.hmr` and `dev.liveReload` in rsbuild plugin:

```js
function myRsbuildPlugin(api) {
  // first way
  // this config is user config
  const { config } = api.useExposed(Symbol.for('rspeedy.api'))!
  console.log(config.dev.hmr, config.dev.liveReload) // false false

  // second way
  // from environment config
  api.modifyBundlerChain((chain, { environment }) => {
    const { config } = environment
    console.log(config.dev.hmr, config.dev.liveReload) // true true
  })
}
```

We can also set these two config in `environment` config and above can't read the real value by the first way above:

```js
import { defineConfig } from '@lynx-js/rspeedy'

export default defineConfig({
  environment: {
    lynx: {
      hmr: false,
      liveReload: false,
    }
  }
})
```

I think Rspeedy should apply the two configs to rsbuild config and other plugins should read the config from `environment.config` first.

## Summary by CodeRabbit

* **New Features**
  * Development builds now default to Hot Module Replacement (HMR) and Live Reload enabled when not explicitly configured; these flags honor environment-specific overrides.
* **Tests**
  * Expanded coverage to verify dev.hmr and dev.liveReload behavior, including environment-specific override cases.
* **Chores**
  * Added a changeset to publish a patch release reflecting the default dev.hmr and dev.liveReload settings.

## Checklist

### DIFF
--- a/packages/rspeedy/core/src/config/rsbuild/index.ts
+++ b/packages/rspeedy/core/src/config/rsbuild/index.ts
@@ -17,15 +17,14 @@ export function toRsbuildConfig(
 ): UndefinedOnPartialDeep<RsbuildConfig> {
   return {
     dev: {
+      hmr: config.dev?.hmr ?? true,
       lazyCompilation: false,
+      liveReload: config.dev?.liveReload ?? true,
       watchFiles: config.dev?.watchFiles,
       // We expect to use different default writeToDisk with Rsbuild
       writeToDisk: config.dev?.writeToDisk ?? true,
 
       progressBar: config.dev?.progressBar ?? true,
-
-      hmr: config.dev?.hmr ?? true,
-      liveReload: config.dev?.liveReload ?? true,
     },
     environments: config.environments ?? { lynx: {} },
     mode: config.mode,


packages/rspeedy/core/src/config/rsbuild/index.ts
```diff
@@ -23,6 +23,9 @@ export function toRsbuildConfig(
       writeToDisk: config.dev?.writeToDisk ?? true,
 
       progressBar: config.dev?.progressBar ?? true,
+
+      hmr: config.dev?.hmr ?? true,
+      liveReload: config.dev?.liveReload ?? true,
```
colinaaa: nitpick: we should sort in alphabetical order

PR #1885, opened 2025-10-13
fix(rspeedy/react): only inject hot clients on needed

colinaaa:
## Summary by CodeRabbit

* **Improvements**
  * Dev servers now respect dev.hmr and dev.liveReload settings when deciding whether to load hot clients.
  * When both are disabled, hot clients are not loaded; when only HMR is disabled, CSS HMR is off but live reload remains active.

* **Tests**
  * Added tests verifying hot client behavior when HMR and/or live reload are disabled.

* **Chores**
  * Prepared a patch version bump with a changeset entry.

## Checklist

### DIFF
--- a/packages/rspeedy/plugin-react/src/entry.ts
+++ b/packages/rspeedy/plugin-react/src/entry.ts
@@ -63,6 +63,8 @@ export function applyEntry(
     const isLynx = environment.name === 'lynx'
     const isWeb = environment.name === 'web'
 
+    const { hmr, liveReload } = environment.config.dev ?? {}
+
     chain.entryPoints.clear()
 
     const mainThreadChunks: string[] = []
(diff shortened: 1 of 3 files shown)


packages/rspeedy/plugin-react/src/entry.ts
```diff
@@ -137,25 +139,28 @@ export function applyEntry(
...
+              })
+              .prepend({
+                layer: LAYERS.BACKGROUND,
+                import: '@lynx-js/webpack-dev-transport/client',
+              })
+              // This is aliased in `./refresh.ts`
+              .prepend({
+                layer: LAYERS.BACKGROUND,
+                import: '@lynx-js/react/refresh',
+              })
```
upupming, line 161: I think`@lynx-js/react/refresh` is only needed on HMR, am I right?

luhc228, line 161: - When enable `hmr`, prepend `@lynx-js/react/refresh` and `@rspack/core/hot/dev-server`
- When enable `liveReload` or `hmr`, prepend `@lynx-js/webpack-dev-transport/client`

PR #1887, opened 2025-10-14
fix(web): layoutchange result top, left was 0 incorrectly

Sherry-hue:
## Summary by CodeRabbit

- Bug Fixes
  - Corrected layoutchange event data: detail.top and detail.left now report accurate values based on the element’s bounding rectangle, improving position reliability.

- Tests
  - Added an end-to-end test and example that validate layoutchange behavior and UI updates when expected positions are reported.

- Chores
  - Added a changeset entry for a patch release of the web elements package.

## Checklist

### DIFF
--- a/packages/web-platform/web-elements/src/common/CommonEventsAndMethods.ts
+++ b/packages/web-platform/web-elements/src/common/CommonEventsAndMethods.ts
@@ -26,10 +26,10 @@ export class CommonEventsAndMethods {
       if (!this.#resizeObserver) {
         this.#resizeObserver = new ResizeObserver(([entry]) => {
           if (entry) {
-            // The layoutchange event is the border box of the element
-            const { width, height, left, right, top, bottom } =
-              entry.contentRect;
             const id = this.#dom.id;
+            // Reuse getBoundingClientRect to get left and top, because ResizeObserver-contentRect is calculated based on padding box
+            const { top, bottom, left, right, width, height } = entry.target
+              .getBoundingClientRect();
             this.#dom.dispatchEvent(
               new CustomEvent('layoutchange', {
                 detail: {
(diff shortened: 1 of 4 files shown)


packages/web-platform/web-elements/src/common/CommonEventsAndMethods.ts
```diff
@@ -27,9 +27,10 @@ export class CommonEventsAndMethods {
         this.#resizeObserver = new ResizeObserver(([entry]) => {
           if (entry) {
             // The layoutchange event is the border box of the element
-            const { width, height, left, right, top, bottom } =
-              entry.contentRect;
+            const { width, height, right, bottom } = entry.contentRect;
             const id = this.#dom.id;
+            // Reuse getBoundingClientRect to get left and top, because ResizeObserver-contentRect is calculated based on padding box
+            const { top, left } = entry.target.getBoundingClientRect();
```
PupilTong: Why don't we use weight/h/r/bottom value from the bounding client rect?

packages/web-platform/web-tests/tests/react/api-bindlauoutchange/index.jsx
```diff
@@ -0,0 +1,37 @@
+// Copyright 2023 The Lynx Authors. All rights reserved.
+// Licensed under the Apache License Version 2.0 that can be found in the
+// LICENSE file in the root directory of this source tree.
+import { root, useState } from '@lynx-js/react';
+function App() {
+  const [color, setColor] = useState('pink');
+  const handleLayoutChange = (e) => {
+    if (e.detail.top === 50 && e.detail.left === 50) {
```
PupilTong: e.detail.top >=50 would be better?

packages/web-platform/web-elements/src/common/CommonEventsAndMethods.ts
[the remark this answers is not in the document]
PupilTong: the top and left are paddingtop and paddingleft

PR #1899, opened 2025-10-28
feat: delay the createSnapshot operation

upupming:
**BREAKING CHANGE**: Delay the `createSnapshot` operation to `Snapshot` constructor to speed up IFR.

This change refactors how snapshots are created and registered:

- Removed the `entryUniqID` function
- Snapshots are now lazily created via `snapshotCreatorMap` instead of eagerly at bundle load time
- Snapshot IDs are generated at compile time and only prefixed with `${globDynamicComponentEntry}:` for standalone lazy bundles

**⚠️ Lazy Bundle Compatibility:**

- **Backward compatibility (new runtime → old lazy bundles)**: ✅ **Supported**. Old lazy bundles will work with the new runtime.

- **Forward compatibility (old runtime → new lazy bundles)**: ❌ **NOT Supported**. Lower version consumers **will not be able to load lazy bundles produced by this version** due to the changed snapshot creation mechanism.

**Migration guidance**:
If you are using lazy bundles, ensure all consumers are upgraded to this version or later **before** deploying lazy bundles built with this version. For monorepo setups, coordinate the upgrade across all consuming applications.

## Summary by CodeRabbit

* **Breaking Changes**
  * Snapshot creation is delayed and lazy: older runtimes cannot load lazy bundles produced by this release; coordinate upgrades. Removed a previously public helper (entryUniqID).

* **New Features**
  * Public snapshot registry (snapshotCreatorMap) enables on-demand snapshot creation to reduce eager initialization and speed startup/DEV.

* **Tests**
  * Updated test fixtures and inline snapshots to match lazy registration and simplified snapshot type names.

* **Chores**
  * Minor package version bump for @lynx-js/react.

## Checklist


hzy (changes requested):
Pls fix the "4 failing checks"

upupming:
@hzy Sorry for that, tests are passed now!

packages/react/runtime/__test__/lynx/timing.test.jsx
```diff
@@ -98,7 +98,7 @@ describe('setState timing api', () => {
 
     expect(mtCallbacks[0][1]).toMatchInlineSnapshot(`
       {
-        "data": "{"patchList":[{"id":3,"snapshotPatch":[0,"__Card__:__snapshot_a94a8_test_2",3,0,null,4,3,4,0,1,1,3,4,null,1,-2,3,null]}],"flushOptions":{"__lynx_timing_flag":"__lynx_timing_actual_fmp"}}",
+        "data": "{"patchList":[{"id":3,"snapshotPatch":[100,"__snapshot_a94a8_test_2","function() {\\n  const pageId = __vite_ssr_import_1__.__pageId;\\n  const el = __CreateText(pageId);\\n  return [\\n    el\\n  ];\\n}",[],[[3,0]],null,"__Card__",0,"__snapshot_a94a8_test_2",3,0,null,4,3,4,0,1,1,3,4,null,1,-2,3,null]}],"flushOptions":{"__lynx_timing_flag":"__lynx_timing_actual_fmp"}}",
```
upupming: This `SnapshotOperation.DEV_ONLY_AddSnapshot` is redundant, I will try to remove it.

packages/react/transform/crates/swc_plugin_snapshot/tests/__swc_snapshots__/lib.rs/should_static_extract_dynamic_inline_style.js
```diff
@@ -1,62 +1,66 @@
...
+        const el = __CreateView(pageId);
+        return [
+            el
+        ];
+    }, [
+        function(ctx) {
+            if (ctx.__elements) {
+                __SetInlineStyles(ctx.__elements[0], ctx.__values[0]);
+            }
         }
```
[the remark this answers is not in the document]
upupming: All js files inside `packages/react/transform/crates/swc_plugin_snapshot/tests/__swc_snapshots__`  are auto generated test snapshot files, they will not be manually updated.

packages/react/runtime/__test__/snapshotPatch.test.jsx
```diff
@@ -714,62 +722,73 @@ describe('DEV_ONLY_addSnapshot', () => {
...
-          el2,
-        ];
-      },
-      /* v8 ignore stop */
-      null,
-      null,
-    );
-
+    const uniqID1 = 'basic-0';
+    // We have to use `snapshotCreatorMap[uniqID1] =` so that is can be created after `initGlobalSnapshotPatch`
```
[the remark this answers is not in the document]
upupming: generate a comittable change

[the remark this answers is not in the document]
upupming: generate a Committable suggestion using diff

packages/react/runtime/__test__/snapshotPatch.test.jsx
```diff
@@ -714,62 +722,73 @@ describe('DEV_ONLY_addSnapshot', () => {
...
+            el,
+            el1,
+            el2,
+          ];
+        },
+        /* v8 ignore stop */
+        null,
+        null,
+      );
+    };
```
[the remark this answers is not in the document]
upupming: We keep this for the compatibility with loading old version lazy bundles, they will still use `const snapshot_xxx = createSnapshot(...)` directly.

packages/react/runtime/src/snapshot.ts
```diff
@@ -193,10 +215,6 @@ export const backgroundSnapshotInstanceManager: {
   },
 };
 
-export function entryUniqID(uniqID: string, entryName?: string): string {
-  return entryName ? `${entryName}:${uniqID}` : uniqID;
-}
-
 export function createSnapshot(
```
HuJean, line 222: Is this implementation compatible within Lazy Bundle with lower versions?

upupming: Yes, it is compatible within Lazy Bundle with lower versions. However it has a limitation: **lower version lazy bundle will lost the prefix (such as `https://xxx/lynx.bundle`) of their entryName**:

- before: they have a snapshot type such as `https://xxx/lynx.bundle:__snapshot_5ab44_710c8_1`
- after: they have a snapshot type such as `__snapshot_5ab44_710c8_1` directly

I expect this change will not break anything since snapshot id uses `filename_hash`, `content_hash` and `snapshot_counter` to generate unique id.

```rs
let snapshot_uid = format!(
      "__snapshot_{}_{}_{}",
      self.filename_hash, self.content_hash, self.snapshot_counter
    );
```

This change is to avoid time consuming string concatenatation blocks main-thread.js execution on loaded (inside the `createSnapshot` call).

Now we generate the unique snapshot type at compile time, and only add `${globDynamicComponentEntry}:` prefix when it is a standalone lazy bundle,see: 

https://github.com/lynx-family/lynx-stack/blob/3c50cdaac0f6ef4f29498a18d1171eb5ea2acfe8/packages/react/transform/crates/swc_plugin_snapshot/tests/__swc_snapshots__/lib.rs/should_extract_css_id_dynamic_component_without_css_id.js#L2

upupming: A risk is that **lower version consumers will not be able to load the lazy bundle produced by this version**. I think we should mention this explicitly in our changelog, @coderabbitai Could you please suggest a change to the changeset file?

.changeset/full-brooms-care.md
```diff
@@ -0,0 +1,5 @@
+---
+"@lynx-js/react": minor
+---
+
+**BREAKING CHANGE**: Delay the `createSnapshot` operation to `Snapshot` constructor to speed up IFR.
```
upupming, line 5:
```suggestion
---
"@lynx-js/react": minor
---

**BREAKING CHANGE**: Delay the `createSnapshot` operation to `Snapshot` constructor to speed up IFR.

This change refactors how snapshots are created and registered:
- Removed the `entryUniqID` function
- Snapshots are now lazily created via `snapshotCreatorMap` instead of eagerly at bundle load time
- Snapshot IDs are generated at compile time and only prefixed with `${globDynamicComponentEntry}:` for standalone lazy bundles

**⚠️ Lazy Bundle Compatibility:**
- **Backward compatibility (new runtime → old lazy bundles)**: ✅ **Supported**. Old lazy bundles will work with the new runtime, but will lose their entryName prefix (e.g., `https://xxx/lynx.bundle:__snapshot_xxx` becomes `__snapshot_xxx`). This should not break functionality since snapshot IDs use `filename_hash`, `content_hash`, and `snapshot_counter` for uniqueness.
  
- **Forward compatibility (old runtime → new lazy bundles)**: ❌ **NOT Supported**. Lower version consumers **will not be able to load lazy bundles produced by this version** due to the changed snapshot creation mechanism.

**Migration guidance**: 
If you are using lazy bundles, ensure all consumers are upgraded to this version or later **before** deploying lazy bundles built with this version. For monorepo setups, coordinate the upgrade across all consuming applications.

```

.changeset/full-brooms-care.md
```diff
@@ -0,0 +1,20 @@
...
+
+- Removed the `entryUniqID` function
+- Snapshots are now lazily created via `snapshotCreatorMap` instead of eagerly at bundle load time
+- Snapshot IDs are generated at compile time and only prefixed with `${globDynamicComponentEntry}:` for standalone lazy bundles
+
+**⚠️ Lazy Bundle Compatibility:**
+
+- **Backward compatibility (new runtime → old lazy bundles)**: ✅ **Supported**. Old lazy bundles will work with the new runtime, but will lose their entryName prefix (e.g., `https://xxx/lynx.bundle:__snapshot_xxx` becomes `__snapshot_xxx`). This should not break functionality since snapshot IDs use `filename_hash`, `content_hash`, and `snapshot_counter` for uniqueness.
+
+- **Forward compatibility (old runtime → new lazy bundles)**: ❌ **NOT Supported**. Lower version consumers **will not be able to load lazy bundles produced by this version** due to the changed snapshot creation mechanism.
```
HuJean, line 17: What exception will this case cause? Can give an example?

packages/react/transform/__test__/css/__snapshots__/transform-imports-css-scoped-modules.js
```diff
@@ -4,17 +4,18 @@ import './foo.css';
...
-    return [
-        el
-    ];
-}, [
-    function(ctx) {
-        if (ctx.__elements) __SetClasses(ctx.__elements[0], ctx.__values[0] || '');
-    }
-], null, 1185352, globDynamicComponentEntry, null);
+const __snapshot_2d408_test_1 = "__snapshot_2d408_test_1";
+ReactLynx.snapshotCreatorMap[__snapshot_2d408_test_1] = (__snapshot_2d408_test_1)=>ReactLynx.createSnapshot(__snapshot_2d408_test_1, function() {
```
Yradex, line 8: Why do we need this argument `__snapshot_2d408_test_1` of the arrow function instead of using `__snapshot_2d408_test_1` directly?

upupming: To make HMR (`DEV_ONLY_AddSnapshot`) happy.

In this way, the function will not depend on the outer scope variable `__snapshot_2d408_test_1`, and it can be safely serialized in BTS and deserialzed in MTS and then execute.

.changeset/full-brooms-care.md
upupming: See 1a82f927e4399089b7236769995264962f60f757

Now it will give verbose info on how to deal with it:

image

.changeset/full-brooms-care.md
```diff
@@ -0,0 +1,20 @@
...
+
+This change refactors how snapshots are created and registered:
+
+- Removed the `entryUniqID` function
+- Snapshots are now lazily created via `snapshotCreatorMap` instead of eagerly at bundle load time
+- Snapshot IDs are generated at compile time and only prefixed with `${globDynamicComponentEntry}:` for standalone lazy bundles
+
+**⚠️ Lazy Bundle Compatibility:**
+
+- **Backward compatibility (new runtime → old lazy bundles)**: ✅ **Supported**. Old lazy bundles will work with the new runtime, but will lose their entryName prefix (e.g., `https://xxx/lynx.bundle:__snapshot_xxx` becomes `__snapshot_xxx`). This should not break functionality since snapshot IDs use `filename_hash`, `content_hash`, and `snapshot_counter` for uniqueness.
```
hzy: Consider different lazy bundles bundles a same JSX file, the uniqueness provided by hashes is insufficient.

e.g.:

```txt
(src/index.tsx, ..., a.common.css) -> lazyA.lynx.bundle
(src/index.tsx, ..., b.common.css) -> lazyB.lynx.bundle
```

`__SetCSSID` should be called with different params between lazy bundles for all snapshots, and the key diff is `entryName`.

hzy (changes requested):
insufficient uniqueness

.changeset/full-brooms-care.md
upupming: In this case, in lazyA.lynx.bundle, src/index.tsx will import a.common.css and src/index.tsx will import b.common.css, so their `content_hash` will be different, am I right?

upupming: If they both import the same `a.common.css` file, but the css content differs in lazyA and lazyB, I think it will cause problems.

upupming: Solved in https://github.com/lynx-family/lynx-stack/pull/1899/commits/c2e080f4890db869a2bd4d8125175d5ad91f4792

packages/react/runtime/__test__/snapshotPatch.test.jsx
```diff
@@ -1352,10 +1514,59 @@ describe('missing snapshot', () => {
...
+      /** entryName */
+      'https://example.com/main.lynx.bundle',
+      null,
+      /** isLazySnapshotSupported */
+      false,
+    );
+    expect(snapshotManager.values.size).toBe(oriSize + 1);
+    const si = new SnapshotInstance(uniqieId);
+    expect(si.type).toBe('https://example.com/main.lynx.bundle:snapshot-1');
   });
```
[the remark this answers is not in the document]
upupming: @coderabbitai generate a applicable suggestion

packages/react/runtime/__test__/snapshotPatch.test.jsx
```diff
@@ -1352,10 +1514,59 @@ describe('missing snapshot', () => {
...
+    };
+    expect(snapshotManager.values.size).toBe(oriSize);
+    const si = new SnapshotInstance('snapshot-0');
+    expect(snapshotManager.values.size).toBe(oriSize + 1);
+    expect(si.type).toBe('snapshot-0');
+  });
+  it('legacy sync createSnapshot should work', () => {
+    let oriSize = snapshotManager.values.size;
+    expect(snapshotManager.values.size).toBe(oriSize);
+    const uniqieId = globalThis.createSnapshot(
```
upupming:
```suggestion
    const uniqueId = globalThis.createSnapshot(
```

packages/react/runtime/__test__/snapshotPatch.test.jsx
```diff
@@ -1352,10 +1514,59 @@ describe('missing snapshot', () => {
...
+      null,
+      undefined,
+      /** entryName */
+      'https://example.com/main.lynx.bundle',
+      null,
+      /** isLazySnapshotSupported */
+      false,
+    );
+    expect(snapshotManager.values.size).toBe(oriSize + 1);
+    const si = new SnapshotInstance(uniqieId);
```
upupming:
```suggestion
    const si = new SnapshotInstance(uniqueId);
```

hzy (approved):

PupilTong (approved):

gaoachao (approved):

PR #1900, opened 2025-10-28
fix: Use item-key to determine whether list-item can be reused in hyd…

DwwWxx:
…rate

## Release Notes

* **Bug Fixes**
  * Enhanced list item hydration to consider both type and item-key attributes when determining if existing nodes can be reused, ensuring more accurate node matching and proper reuse behavior during the hydration process.

## Checklist

### DIFF
--- a/packages/react/runtime/src/hydrate.ts
+++ b/packages/react/runtime/src/hydrate.ts
@@ -303,7 +303,8 @@ export function hydrate(before: SnapshotInstance, after: SnapshotInstance, optio
         const diffResult = diffArrayLepus(
           beforeChildNodes,
           afterChildNodes,
-          (a, b) => a.type === b.type,
+          (a, b) =>
+            a.type === b.type && a.__listItemPlatformInfo?.['item-key'] === b.__listItemPlatformInfo?.['item-key'],
           (a, b, _oldIndex, newIndex) => {
             if (
               JSON.stringify(a.__listItemPlatformInfo)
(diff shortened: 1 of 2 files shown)


packages/react/runtime/src/hydrate.ts
```diff
@@ -303,7 +303,8 @@ export function hydrate(before: SnapshotInstance, after: SnapshotInstance, optio
         const diffResult = diffArrayLepus(
           beforeChildNodes,
           afterChildNodes,
-          (a, b) => a.type === b.type,
+          (a, b) =>
+            a.type === b.type && a.__listItemPlatformInfo?.['item-key'] === b.__listItemPlatformInfo?.['item-key'],
```
upupming, line 307: @DwwWxx What case of error does this fix, can we add some unit test cases to avoid future regression? Thanks!

DwwWxx: When a vertical list nested some horizontal lists and we scroll the vertical list, one horizontal list A with it's children item-key [A1, A2, A3...] may be reused by the other horizontal list B with it's children's item-key [B1, B2, B3....], and this diffArrayLepus() method will be invoked. If not use item-key to determine whether list-item can be reused, the diff result will not contain remove:[0, 1, 2...] and insert:[0, 1, 2...] which is not expected by list.

DwwWxx: By the way, this issue can also be solved by https://github.com/lynx-family/lynx-stack/pull/1902, so no need to merge this pr.

PR #1919, opened 2025-11-05
feat: reduce `__SetInlineStyles` call when styles in spread not changed

upupming:
## Summary by CodeRabbit

* **Performance**
  * Reduced redundant attribute and inline-style updates during re-renders when spread props or styles remain unchanged, improving UI update efficiency and reducing unnecessary DOM work.

* **Tests**
  * Added tests ensuring spread props and inline styles remain stable across re-renders and no redundant attribute/style updates are produced.

## Checklist

### DIFF
--- a/packages/react/runtime/src/snapshot/spread.ts
+++ b/packages/react/runtime/src/snapshot/spread.ts
@@ -91,11 +91,13 @@ function updateSpread(
   let match: RegExpMatchArray | null = null;
   for (const key in newValue) {
     const v = newValue[key];
-    if (!isDirectOrDeepEqual(v, oldValue[key])) {
+    if (v !== oldValue[key]) {
       if (key === 'className') {
         __SetClasses(snapshot.__elements[elementIndex]!, v as string);
       } else if (key === 'style') {
-        __SetInlineStyles(snapshot.__elements[elementIndex]!, v as string);
+        if (!isDirectOrDeepEqual(v, oldValue[key])) {
+          __SetInlineStyles(snapshot.__elements[elementIndex]!, v);
+        }
       } else if (key === 'id') {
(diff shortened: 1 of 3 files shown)


packages/react/runtime/src/snapshot/spread.ts
```diff
@@ -91,7 +91,7 @@ function updateSpread(
   let match: RegExpMatchArray | null = null;
   for (const key in newValue) {
     const v = newValue[key];
-    if (v !== oldValue[key]) {
+    if (!isDirectOrDeepEqual(v, oldValue[key])) {
```
HuJean: {a:1, b:2} is equal to {b:2, a:1} ?

upupming: Sure, I believe they should be treated as same attribute.

When not using spread, `{a:1, b:2}` and  `{b:2, a:1}` are already treated as same value now in BTS and skip generating a patch, see:

https://github.com/lynx-family/lynx-stack/blob/83952a0151ce3347e0f93e2d6f1c64481d0ad130/packages/react/runtime/src/backgroundSnapshot.ts#L350

Yradex: I wonder if it's possible to perform this check only for styles. It might affect performance(and saves nothing) for most of the other keys.

HuJean: isDirectOrDeepEqual is base on JSON.stringify. We redefine the behavior of stringify?
image

upupming: No, we do not redefine `JSON.stringify`. It is my mistake, {a:1, b:2} and {b:2, a:1} are compared using `JSON.stringify` so it will cause a setAttribute/setInlineStyles.

Yradex (approved):

PR #1924, opened 2025-11-06
feat(webpack-plugin): init externals-loading-webpack-plugin

luhc228:
## Summary by CodeRabbit

* **New Features**
  * Adds an Externals Loading Webpack plugin for runtime externals loading (sync/async), deduplication, and layer-aware background vs main-thread handling; public plugin API and types included.

* **Examples**
  * Adds a React externals example demonstrating bundling and runtime loading strategies.

* **Documentation**
  * Adds plugin README, changelog, API report and example docs.

* **Packaging**
  * Introduces a self-contained package with package metadata and build/test scripts.

* **Tests**
  * New end-to-end test suite, fixtures and helpers covering async/sync loading, layer routing, and duplicate-filtering.

## Checklist

### DIFF
--- a/.changeset/chilly-ravens-smile.md
+++ b/.changeset/chilly-ravens-smile.md
@@ -2,7 +2,7 @@
 "@lynx-js/externals-loading-webpack-plugin": patch
 ---
 
-Introduce `ExternalsLoadingWebpackPlugin`.
+Introduce `@lynx-js/externals-loading-webpack-plugin`.
 
 ```js
 // webpack.config.js


packages/webpack/externals-loading-webpack-plugin/src/index.ts
```diff
@@ -0,0 +1,359 @@
...
+// Licensed under the Apache License Version 2.0 that can be found in the
+// LICENSE file in the root directory of this source tree.
+
+/**
+ * @packageDocumentation
+ *
+ * A webpack plugin to load externals in Lynx. Use `lynx.fetchBundle()` and `lynx.loadScript()` API to load and parse the externals.
+ *
+ * @remark
+ * It requires Lynx >= 3.4.
```
upupming: Will we have a build time check to throw an error if targetSDKVersion < 3.4?

packages/webpack/externals-loading-webpack-plugin/src/index.ts
```diff
@@ -0,0 +1,359 @@
...
+   *
+   * Load `lodash` library in background layer and `main-thread` layer.
+   *
+   * ```js
+   * module.exports = {
+   *  plugins: [
+   *    new ExternalsLoadingPlugin({
+   *      externals: {
+   *        lodash: {
+   *          libraryName: "Lodash",
```
upupming: I think we can delete this config since it is only used by us internally. We can generate a unique id for each external module.

packages/webpack/externals-loading-webpack-plugin/src/index.ts
```diff
@@ -0,0 +1,359 @@
...
+       *
+       * @defaultValue `undefined`
+       */
+      mainThread?: LayerOptions;
+      /**
+       * The wait time in milliseconds.
+       *
+       * @defaultValue `100`
+       */
+      timeout?: number;
```
upupming, line 200: What will happen if the timeout is reached? I think we should add a test case that timeout is reached but load failed.

luhc228: > What will happen if the timeout is reached? I think we should add a test case that timeout is reached but load failed.

It will throw an error directly by Lynx engine.

packages/webpack/externals-loading-webpack-plugin/src/index.ts
luhc228: We can do this check in the externals-rsbuild-plugin.

packages/webpack/externals-loading-webpack-plugin/src/index.ts
luhc228: Now `libraryName` is an optional option. The default [external library name](https://webpack.js.org/configuration/externals/#string) is same as pkg name. We can set the `libraryName` option when there are multi modules in one bundle to avoid duplicate loading.

luhc228:
@upupming @colinaaa PTAL~

packages/webpack/externals-loading-webpack-plugin/test/cases/externals-loading/async/rspack.config.js
```diff
@@ -0,0 +1,27 @@
...
+          libraryName: 'Foo',
+          url: 'foo',
+          async: true,
+          background: {
+            sectionPath: 'background',
+          },
+          mainThread: {
+            sectionPath: 'mainThread',
+          },
+        },
```
upupming, line 23: For a package like`@lynx-js/react`, we may want to bundle several sub-packages into a single bundle. So I prefer to use some config like this:

```js
externals: {
  ReactLynx: [
    { exports: 'React', package: '@lynx-js/react' },
    { exports: 'ReactInternal', package: '@lynx-js/react/internal' },
    { exports: 'Preact', package: 'preact' },
  ]
}
```

--- 

This is how I bundle this:

```js
export * as React from "@lynx-js/react";
export * as ReactInternal from "@lynx-js/react/internal";
export * as ReactLazyImport from "@lynx-js/react/experimental/lazy/import";
export * as ReactLegacyRuntime from "@lynx-js/react/legacy-react-runtime";
export * as ReactComponents from "@lynx-js/react/runtime-components";
export * as ReactWorkletRuntime from "@lynx-js/react/worklet-runtime/bindings";
export * as ReactDebug from "@lynx-js/react/debug";
export * as Preact from "preact";

```

```js
import { defineExternalBundleRslibConfig, LAYERS } from '@lynx-js/lynx-bundle-rslib-config'
import { ReactWebpackPlugin } from '@lynx-js/react-webpack-plugin'

export default defineExternalBundleRslibConfig({
  id: 'react',
  tools: {
    rspack: {
      module: {
        rules: [
          {
            layer: LAYERS.BACKGROUND,
            loader: ReactWebpackPlugin.loaders.BACKGROUND
          },
          {
            layer: LAYERS.MAIN_THREAD,
            loader: ReactWebpackPlugin.loaders.MAIN_THREAD
          },
        
        ]
      }
    }
  },
  source: {
    entry: {
      'ReactLynx': './entry.js',
    }
  },
  output: {
    minify: false
  }
})

```

luhc228: BTW, I am not sure whether the `@lynx-js/react/debug` is need in prod or not.

upupming: @luhc228 In production it will be an empty module, but when in development it is needed, I think we can let user decide to build a dev/prod bundle of ReactLynx.

upupming (dismissed):

upupming (approved):

PR #1948, opened 2025-11-20
fix: `<list-item />` with dynamic `key` & spread attributes

### LINKED ISSUE #1920: [Bug]: The `<list-item>` produces incorrect diff results during re-render, causing `update-list-info` to be empty.
### System Info

N/A

### Details

```tsx
<list-item
     item-key={itemKey}
     key={key}
     {...(itemEstimatedHeightPx ? { 'estimated-height-px': itemEstimatedHeightPx } : {})}
/>
```

### Reproduce link

_No response_

### Reproduce Steps

N/A

gaoachao:
close #1920

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed a runtime error that could occur when using dynamic `key` values together with spread attributes on `<list-item>`.
  * Improved list-item spread handling so list-item related platform attributes are applied more reliably during updates.

* **Chores**
  * Updated the package changeset to release a patch version for the fix.

## Checklist


packages/react/runtime/src/snapshot/snapshot/spread.ts
```diff
@@ -75,6 +76,9 @@ function updateSpread(
       } as SnapshotInstance;
       updateListItemPlatformInfo(fakeSnapshot, index, oldPlatformInfo, elementIndex);
     }
+  } else if (isListItem) {
+    const platformInfo = pick(newValue, platformInfoAttributes);
```
HuJean, line 110: How to handle other attributes

gaoachao: Addressed in 521f67a5b. The list-item branch only seeds `__listItemPlatformInfo` before the snapshot is attached to a list holder; non-platform spread attributes continue through the normal update loop below. I also made `isListItem` default to `false` and updated runtime-created/clone snapshot call sites to pass `false` explicitly.

HuJean (approved):

PR #1951, opened 2025-11-24
fix(rspeedy/react)!: remove `experiments.layers`

colinaaa:
## Summary by CodeRabbit

* **Breaking Changes**
  * This release requires @lynx-js/rspeedy version 0.12.0 or higher
  * Rsbuild experiments configuration is no longer automatically managed by the React plugin

Resolve the deprecated warning of Rspack 1.6:

```
(node:48125) DeprecationWarning: `experiments.layers` config has been deprecated and will be removed in Rspack v2.0. Feature layers will be always enabled. Please remove this option from your Rspack configuration.
```

## Checklist

### DIFF
--- a/.changeset/sharp-plants-repair.md
+++ b/.changeset/sharp-plants-repair.md
@@ -1,0 +1,5 @@
+---
+"@lynx-js/react-rsbuild-plugin": minor
+---
+
+**BREAKING CHANGE**: Require `@lynx-js/rspeedy` 0.12.0.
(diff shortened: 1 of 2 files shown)


.changeset/sharp-plants-repair.md
```diff
@@ -0,0 +1,5 @@
+---
+"@lynx-js/react-rsbuild-plugin": minor
+---
+
+**BREAKING CHANGE**: Require `@lynx-js/rspeedy` 0.12.0.
```
HuJean, line 5: What happens with lower versions?

colinaaa: Rspack 1.6 [stabilized `experiments.layers`](https://rspack.rs/blog/announcing-1-6#stabilized-layers-feature). So the `experiments.layers` is not needed anymore in the newer viersion.

But it is needed in the older version for layers to work.

If using the new plugin with the old Rspeedy, there would be an error saying that `experiments.layers` is not enabled.

Issue #1954, opened 2025-11-25
[Bug]: lynx.loadLazyBundle is not a function


HuJean:
### System Info

None

### Details

Dynamic import script and not use lazy. This would lead to an error because lynx.loadLazyBundle is not defined.

### Reproduce Steps

```
export function App() {
  const handleTap = ()=>{
    import("./test").then((res)=>{
      console.log('dynamic import', res)
    })
  }
  return (
    <view bindtap={handleTap}>...</view>
  )
}
```

PR #1956, opened 2025-11-25
feat(react/transform): auto define lynx.loadLazyBundle

HuJean:
Auto define lynx.loadLazyBundle when using `import(/* relative path */)`.

close https://github.com/lynx-family/lynx-stack/issues/1954

## Summary by CodeRabbit

* **New Features**
  * Auto-defines lynx.loadLazyBundle when using relative dynamic imports to simplify lazy-bundle usage.

* **Examples & Documentation**
  * Added a React lazy-bundle example project with config, sample app, and utility modules demonstrating dynamic imports.

* **Tests**
  * Added and updated tests and snapshots covering dynamic import transformation and lazy-bundle loading behavior.

## Checklist


packages/react/transform/__test__/fixture.spec.js
```diff
@@ -938,12 +939,14 @@ describe('dynamic import', () => {
     );
 
     expect(result.code).toMatchInlineSnapshot(`
-      "import "@lynx-js/react/experimental/lazy/import";
+      "import "@lynx-js/react/experimental/lazy/lynx";
+      import "@lynx-js/react/experimental/lazy/import";
```
upupming: I think this test case should be split and combined correctly:

- Only `lazy/lynx` is generated
- Only `lazy/import` is generated
- Both `lazy/lynx` and `lazy/import` are generated

packages/react/package.json
```diff
@@ -28,6 +28,10 @@
       "types": "./runtime/lazy/import.d.ts",
       "default": "./runtime/lazy/import.js"
     },
+    "./experimental/lazy/lynx": {
+      "types": "./runtime/lazy/lynx.d.ts",
+      "default": "./runtime/lazy/lynx.js"
+    },
```
upupming: It is only used internally; can we avoid exporting them? (Since we have too many exports now)

packages/react/transform/crates/swc_plugin_dynamic_import/lib.rs
```diff
@@ -196,6 +217,7 @@ where
           text: format!("webpackChunkName: \"{}-{}\"", str_lit, self.opts.layer).into(),
         },
       );
+      self.has_dynamic_import = true;
```
upupming: `has_lazy_bundle` may be better because we also have dynamic import in the else branch

```suggestion
      self.has_lazy_bundle = true;
```

packages/react/runtime/lazy/lynx.js
```diff
@@ -0,0 +1,2 @@
+import { loadLazyBundle } from '@lynx-js/react/internal';
+lynx.loadLazyBundle = loadLazyBundle;
```
upupming: I am thinking about removing `loadLazyBundle` from `lynx` and using it on demand. If so, we do not need to maintain an extra export like so.

CC @Yradex, what do you think?

Yradex: lgtm

HuJean: now this is used in webpack runtime, can export for it?

packages/react/transform/__test__/fixture.spec.js
HuJean: there are tests in rust implement

packages/react/package.json
HuJean: it should be a separated export because it has side effect.

packages/react/runtime/vitest.config.ts
```diff
@@ -36,6 +36,7 @@ function transformReactLynxPlugin(): Plugin {
           filename: 'test',
           target: 'MIXED',
         },
+        dynamicImport: false,
```
upupming, line 39: Why do we need `false` here?

HuJean: vite cannot handle data url like `import "data:text/javascript;..."`

PR #1968, opened 2025-11-28
feat(react): shared runtime imports

Yradex:
## Summary by CodeRabbit

* **New Features**
  * Support for declaring cross-thread shared modules via import attributes (e.g. import with runtime: "shared") so main-thread code can reference imported identifiers directly.

* **Documentation**
  * Added docs describing usage and limitations (only directly imported identifiers are treated as shared; shared modules don't auto-promote functions; accessing main-thread-only APIs may error).

* **Tests**
  * New tests and snapshots covering named/default/namespace/shared imports and invalid/non-string runtime values.

## Checklist

### DIFF
--- a/packages/react/transform/crates/swc_plugin_worklet/lib.rs
+++ b/packages/react/transform/crates/swc_plugin_worklet/lib.rs
@@ -13,7 +13,7 @@ use serde::Deserialize;
 use std::collections::HashSet;
 use std::vec;
 use swc_core::common::util::take::Take;
-use swc_core::common::{Spanned, DUMMY_SP};
+use swc_core::common::{errors::HANDLER, Span, Spanned, DUMMY_SP};
 use swc_core::ecma::ast::*;
 use swc_core::ecma::utils::prepend_stmts;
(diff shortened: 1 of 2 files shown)


packages/react/transform/crates/swc_plugin_worklet/lib.rs
```diff
@@ -678,6 +733,210 @@ function X(event) {
...
+          custom_global_ident_names: None,
+          runtime_pkg: "@lynx-js/react".into(),
+        }
+      )),
+      hygiene()
+    ),
+    should_skip_shared_identifiers_lepus,
+    r#"
+import { sharedRuntime } from './utils.js' with {
+    runtime: "shared"
```
HuJean, line 949: runtime: "main-thread"
runtime: "background"

Is there such a scenario in the future?

check  `runtime: "other"` and throw invalid info?

Yradex: No such plan. Checks added.

gaoachao (approved):

PR #1972, opened 2025-11-28
fix(web): capture event listener should be called before bind event l…

Sherry-hue:
…istener

## Release Notes

* **Bug Fixes**
  * Fixed event listener capture and bind functionality to trigger correctly during different event phases (capturing vs. bubbling).

## Checklist


packages/web-platform/web-mainthread-apis/ts/createMainThreadGlobalThis.ts
```diff
@@ -269,9 +268,13 @@ export function createMainThreadGlobalThis(
         /**
          * remove handler
          */
-        element.removeEventListener(eventName, currentRegisteredHandler, {
-          capture: isCapture,
-        });
+        element.removeEventListener(
+          eventName,
+          (e) => currentRegisteredHandler(e, isCapture),
```
PupilTong: I think this won't work.

Sherry-hue: fixed

Issue #1977, opened 2025-12-03
[Bug]: set hmr to false would cause an exception


HuJean:
### System Info

Node

### Details

```
dev: {
    hmr: false
}
```

The exception is 'loadCard failed Error: [HMR] Hot Module Replacement is disabled.'

### Reproduce link

Node

### Reproduce Steps

Node

PR #1980, opened 2025-12-03
fix: avoid injecting hot update runtime

HuJean:
Closed https://github.com/lynx-family/lynx-stack/issues/1977

## Summary by CodeRabbit

* **New Features**
  * React plugin now respects dev.hmr and dev.liveReload flags and avoids injecting hot-update runtimes when those are disabled.

* **Bug Fix / Behavior**
  * Adjusted ordering of dev-time runtime imports to ensure correct layering of transport and refresh runtimes.

* **Tests**
  * Added comprehensive tests covering hot-update/runtime injection across HMR and live-reload configurations.

* **Chores**
  * Added a patch changeset and CODEOWNERS entry for the React plugin.

## Checklist

### DIFF
--- a/packages/rspeedy/plugin-react/src/entry.ts
+++ b/packages/rspeedy/plugin-react/src/entry.ts
@@ -62,6 +62,9 @@ export function applyEntry(
     const entries = chain.entryPoints.entries() ?? {}
     const isLynx = environment.name === 'lynx'
     const isWeb = environment.name === 'web'
+    const { hmr, liveReload } = environment.config.dev ?? {}
+    const enabledHMR = isDev && !isWeb && hmr !== false
+    const enabledLiveReload = isDev && !isWeb && liveReload !== false
 
     chain.entryPoints.clear()
 
@@ -116,7 +119,7 @@ export function applyEntry(
           import: imports,
           filename: mainThreadName,
         })
-        .when(isDev && !isWeb, entry => {
+        .when(enabledHMR, entry => {
           const require = createRequire(import.meta.url)
           // use prepend to make sure it does not affect the exports
           // from the entry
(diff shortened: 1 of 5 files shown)


packages/rspeedy/plugin-react/src/entry.ts
```diff
@@ -62,6 +62,7 @@ export function applyEntry(
     const entries = chain.entryPoints.entries() ?? {}
     const isLynx = environment.name === 'lynx'
     const isWeb = environment.name === 'web'
+    const enabledHMR = isDev && !isWeb && environment.config.dev?.hmr !== false
```
upupming: Do we also disable HMR in the standalone lazy bundle? (Since it does not work)

HuJean: After this pr, we need to add a demo for lazy bundle to clarify current behavior

colinaaa: We should also handle `dev.liveReload`.

packages/rspeedy/plugin-react/src/entry.ts
```diff
@@ -137,7 +138,7 @@ export function applyEntry(
         })
         // in standalone lazy bundle mode, we do not add
         // other entries to avoid wrongly exporting from other entries
-        .when(isDev && !isWeb, entry => {
+        .when(enabledHMR, entry => {
```
colinaaa, line 143: This should be different with the previous condition.

See comment at https://github.com/lynx-family/lynx-stack/pull/1885#discussion_r2426469734

colinaaa (commented):
It seems like we already have a similar PR: https://github.com/lynx-family/lynx-stack/pull/1885

PR #2036, opened 2025-12-23
fix: Cannot read properties of undefined (reading '__elements')

upupming:
@coderabbitai summary

## Checklist


upupming:
Fixed by #2041

PR #2048, opened 2025-12-24
fix: undefined factory error when loading a standalone lazy bundle after hydration

upupming:
## Summary by CodeRabbit

* **Bug Fixes**
  * Prevented an undefined factory error when loading a standalone lazy bundle after hydration by skipping first-screen snapshot updates for lazy-loaded bundles.

* **Tests**
  * Updated snapshot-patch tests to validate behavior when lazy bundles are loaded after hydration and to assert empty patch state in that scenario.

Fix a regression introduced by https://github.com/lynx-family/lynx-stack/pull/1899

## Checklist

### DIFF
--- a/packages/react/runtime/src/snapshot.ts
+++ b/packages/react/runtime/src/snapshot.ts
@@ -140,9 +140,13 @@ export let snapshotCreatorMap: Record<string, (uniqId: string) => string> = {};
 if (__DEV__ && __JS__) {
   snapshotCreatorMap = new Proxy(snapshotCreatorMap, {
     set(target, prop: string, value: (uniqId: string) => string) {
-      // `__globalSnapshotPatch` does not exist before hydration,
-      // so the snapshot of the first screen will not be sent to the main thread.
-      if (__globalSnapshotPatch) {
+      if (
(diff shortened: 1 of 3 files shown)


packages/react/runtime/src/snapshot.ts
```diff
@@ -140,9 +140,13 @@ export let snapshotCreatorMap: Record<string, (uniqId: string) => string> = {};
...
     set(target, prop: string, value: (uniqId: string) => string) {
-      // `__globalSnapshotPatch` does not exist before hydration,
-      // so the snapshot of the first screen will not be sent to the main thread.
-      if (__globalSnapshotPatch) {
+      if (
+        // `__globalSnapshotPatch` does not exist before hydration,
+        // so the snapshot of the first screen will not be sent to the main thread.
+        __globalSnapshotPatch
+        // `prop` will be `https://example.com/main.lynx.bundle:__snapshot_835da_eff1e_1` when loading a lazy bundle after hydration.
+        && !prop.includes(':')
```
HuJean, line 148: it is for lazy bundle？

upupming: Yes

PR #2059, opened 2025-12-26
chore: Add tracing start/stop tools and enhance global switch management

KananSu:
- Introduced `StartTracing` and `EndTracing` tools to the devtool-mcp-server, enabling trace event collection control via MCP server.
  - `StartTracing`: Starts trace event collection, checks and enables debug mode for Android, and provides user feedback based on result.
  - `EndTracing`: Stops trace event collection, reads trace data from stream, saves it to a temporary file, and returns the file path or error.
- Registered the new tracing tools in the MCP server tool list.
- Extended global switch management:
  - Added support for `enable_debug_mode` global key.
  - Ensured both `enable_devtool` and `enable_debug_mode` are enabled when connecting.
  - Updated response types and internal logic to use `{ message: ... }` instead of `{ global_value: ... }` for global switch responses.
- Added `offCDPEvent` method for event listener cleanup.
- Updated event map to support `Tracing.tracingComplete`.

## Summary by CodeRabbit

* **New Features**
  * Added StartTracing and EndTracing tools to begin/end trace collection and save trace files.
  * Clients now auto-enable debug mode on connect to simplify tracing.
  * Added ability to detach CDP event listeners to prevent duplicate notifications.

* **Bug Fixes**
  * Improved trace error messages, timeout handling, streaming, and cleanup for more reliable tracing.

* **Documentation**
  * Added a changeset documenting the new tracing tools and behavior.

## Checklist

- [x] Tests not required.
- [x] Documentation not required
- [x] Changeset added.


packages/mcp-servers/devtool-mcp-server/src/tools/Tracing/StartTracing.ts
```diff
@@ -0,0 +1,80 @@
...
+    );
+
+    if (result.error) {
+      const msg = result.error.message;
+
+      if (
+        msg.indexOf('Failed to get trace controller') >= 0
+        || msg.indexOf('Not implemented:') >= 0
+        || msg.indexOf('Tracing not enabled') >= 0
+        || msg.indexOf('Failed to start tracing') >= 0
```
colinaaa: Use `String.prototype.includes` instead.

packages/mcp-servers/devtool-mcp-server/src/tools/Tracing/StartTracing.ts
```diff
@@ -0,0 +1,80 @@
...
+      ) {
+        response.appendLines(
+          'Start Trace failed: Tracing functionality is not supported in the current version. Please integrate the Lynx development version (with -dev suffix) to enable tracing. For more information, visit: https://lynxjs.org/en/guide/start/integrate-lynx-dev-version.html',
```
colinaaa: Instead of adding `Start Trace failed` to the response, we should throw an error.

packages/mcp-servers/devtool-mcp-server/src/tools/Tracing/StartTracing.ts
```diff
@@ -0,0 +1,80 @@
...
+        'enable_debug_mode',
+      );
+      if (!value) {
+        await connector.setGlobalSwitch(
+          params.clientId,
+          'enable_debug_mode',
+          true,
+        );
+        response.appendLines(
+          'Start Trace failed: Please restart the app to enable tracing functionality.',
```
colinaaa: ditto. Directly throw an error

packages/mcp-servers/devtool-mcp-server/src/tools/Tracing/EndTracing.ts
```diff
@@ -0,0 +1,136 @@
...
+    const sendIOReadMessage = (stream: number): Promise<IOReadResponse> => {
+      return connector.sendCDPMessage<IOReadResponse>(
+        params.clientId,
+        -1,
+        'IO.read',
+        { handle: stream, size: 1024 * 1024 },
+      );
+    };
+    const readStreamData = async (stream: number): Promise<Buffer[]> => {
+      const dataChunks: Buffer[] = [];
```
colinaaa: Instead of caching all the data in memory, we should using streaming API:

```ts
        const sourceStream = new ReadableStream<Uint8Array>({
          async pull(controller) {
            try {
              const message = await connector.sendCDPMessage<IOReadResponse>(
                params.clientId,
                -1,
                'IO.read',
                { handle: streamHandle, size: 1024 * 1024 }
              );

              if (message.data) {
                controller.enqueue(Buffer.from(message.data, 'base64'));
              }

              if (message.eof) {
                controller.close();
              }
            } catch (err) {
              controller.error(err);
            }
          },
        });

        const fileWritable = Writable.toWeb(fs.createWriteStream(tempFilePath));

        await sourceStream.pipeTo(fileWritable);
```

packages/mcp-servers/devtool-mcp-server/src/tools/Tracing/StartTracing.ts
```diff
@@ -0,0 +1,80 @@
...
+      }
+    }
+    const config = {
+      recordMode: 'recordContinuously',
+      includedCategories: ['*'],
+      enableSystrace: true,
+      bufferSize: 200 * 1024,
+      JSProfileInterval: 1,
+      enableCompress: true,
+    };
```
colinaaa: Maybe we should add these to the option of `Tracing_start`

KananSu: Add `JSProfileInterval`, `JSProfileType`, and `enableSystrace` to the `Tracing_start` options.  The other options can remain at their default values, as they typically don't need customization for most tracing scenarios.

packages/mcp-servers/devtool-mcp-server/src/tools/Tracing/EndTracing.ts
```diff
@@ -0,0 +1,120 @@
...
+
+        response.appendLines(
+          `Trace completed successfully, trace file path: ${tempFilePath}`,
+        );
+      } catch (error) {
+        response.appendLines(
+          `Trace completed failed, ${
+            error instanceof Error ? error.message : String(error)
+          }`,
+        );
```
colinaaa: Instead of appending an error message to response, we should reject the promise.

packages/mcp-servers/devtool-mcp-server/src/tools/Tracing/EndTracing.ts
```diff
@@ -0,0 +1,120 @@
...
+      const fileWritable = Writable.toWeb(writeStream);
+
+      await sourceStream.pipeTo(fileWritable);
+    };
+    const { promise, resolve } = Promise.withResolvers<void>();
+
+    const timer = setTimeout(() => {
+      response.appendLines(
+        'Loading trace data timeout, please try again later!',
+      );
```
colinaaa: ditto.

packages/mcp-servers/devtool-mcp-server/src/tools/Tracing/StartTracing.ts
```diff
@@ -0,0 +1,87 @@
...
+        msg.includes('Failed to get trace controller')
+        || msg.includes('Not implemented:')
+        || msg.includes('Tracing not enabled')
+        || msg.includes('Failed to start tracing')
+      ) {
+        throw new Error(
+          'Tracing functionality is not supported in the current version. Please integrate the Lynx development version (with -dev suffix) to enable tracing. For more information, visit: https://lynxjs.org/en/guide/start/integrate-lynx-dev-version.html',
+        );
+      } else {
+        throw new Error('Trace command error');
```
colinaaa: Maybe we should also include the `msg` in the thrown error.
```suggestion
        throw new Error(`Trace command error: ${msg}`);
```

packages/mcp-servers/devtool-mcp-server/src/tools/Tracing/StartTracing.ts
```diff
@@ -0,0 +1,87 @@
...
+      'JavaScript profile interval in milliseconds. Set to -1 to disable JavaScript profiling.',
+    ),
+  },
+  annotations: {
+    readOnlyHint: false,
+  },
+  async handler({ params }, response, context) {
+    const connector = context.connector();
+    const client = connector.usbClients.get(params.clientId);
+    if (client && client.info.query.os === 'Android') {
```
colinaaa:
```suggestion
    if (client?.info.query.os === 'Android') {
```

packages/mcp-servers/devtool-mcp-server/src/tools/Tracing/EndTracing.ts
```diff
@@ -0,0 +1,120 @@
...
+  annotations: {
+    readOnlyHint: false,
+  },
+  async handler({ params }, response, context) {
+    const connector = context.connector();
+    const sendIOReadMessage = (stream: number): Promise<IOReadResponse> => {
+      return connector.sendCDPMessage<IOReadResponse>(
+        params.clientId,
+        -1,
+        'IO.read',
```
colinaaa, line 33: Do we also need to close the IO stream? E.g.: sending `IO.close` CDP request. See: https://chromedevtools.github.io/devtools-protocol/tot/IO/#method-close

packages/mcp-servers/devtool-mcp-server/src/tools/Tracing/EndTracing.ts
```diff
@@ -0,0 +1,149 @@
...
+          await sendIOCloseMessage(stream);
+        } catch {
+          void 0;
+        }
+      }
+    };
+    const { promise, resolve, reject } = Promise.withResolvers<void>();
+    promise.catch(() => {
+      void 0;
+    });
```
colinaaa: What's the point of catching the returned `promise`? We should reject the promise so that agents/LLMs would know about this call has failed.
```suggestion
```

packages/mcp-servers/devtool-mcp-server/src/tools/Tracing/EndTracing.ts
```diff
@@ -0,0 +1,149 @@
...
+    const settle = (errorText?: string) => {
+      if (isSettled) return;
+      isSettled = true;
+
+      if (errorText) {
+        reject(new Error(errorText));
+      } else {
+        resolve();
+      }
+    };
```
colinaaa: Is this necessary? Why not calling `resolve` and `reject` directly.

packages/mcp-servers/devtool-mcp-server/src/tools/Tracing/EndTracing.ts
```diff
@@ -0,0 +1,149 @@
...
+        response.appendLines(
+          `Trace completed successfully, trace file path: ${tempFilePath}`,
+        );
+        settle();
+      } catch (error) {
+        settle(
+          `Trace completed failed, ${
+            error instanceof Error ? error.message : String(error)
+          }`,
+        );
```
colinaaa: Maybe we can just reject with `error`


```suggestion
        reject(error)
```

colinaaa (changes requested):

packages/mcp-servers/devtool-mcp-server/src/tools/Tracing/EndTracing.ts
KananSu: The `settle` function is needed to ensure that the Promise is resolved or rejected exactly once and to prevent multiple calls from causing unexpected behavior. Here’s why it’s important:
- Idempotency:
The `settle` function uses a flag (`isSettled`) to make sure that once the Promise is either resolved or rejected, subsequent calls to `settle` do nothing. This prevents multiple `resolve` or `reject` calls on the same Promise, which would have no effect or could cause confusing side effects.
- Centralized control:
It provides a single place to handle the logic of resolving or rejecting the Promise, simplifying the code and making it easier to maintain.
- Timeout and event handling coordination:
In your code, `settle` is called either when the timeout triggers (rejecting with an error) or when the trace completes successfully (resolving). This ensures that whichever happens first settles the Promise, and the other path is ignored.

packages/mcp-servers/devtool-mcp-server/src/tools/Tracing/EndTracing.ts
KananSu: ### Problem Description
```
await connector.sendCDPMessage(
  params.clientId,
  -1,
  'Tracing.end',
);
```
- This call may hang indefinitely if, for example, the device is disconnected or USB communication is slow.
- In such a case, the outer `await tool.handler(...)` does not actually start waiting yet because the internal async operation is stuck or delayed.
- Meanwhile, inside `handler`, a `setTimeout` may trigger a `reject` due to timeout or error.
- Because the outer `await` has not yet bound a rejection handler, Node.js treats this rejection as unhandled, causing an `UnhandledPromiseRejection` error.
### Why add .catch(() => {})?
- Adding `promise.catch(() => {})` registers a rejection handler immediately on the Promise.
- This tells Node.js that the rejection is handled, preventing it from emitting an unhandled rejection warning or error.
- Importantly, this `.catch()` does NOT change the rejected state of the Promise — the Promise remains rejected.
- When the outer `await tool.handler(...)` eventually runs, its `try/catch` will still properly catch the rejection as expected.

PR #2112, opened 2026-01-15
fix: should clear list callbacks when `__DestroyLifetime`

gaoachao:
## Summary by CodeRabbit

* **New Features**
  * Exposed a lightweight native event accessor for testing and integration.

* **Bug Fixes**
  * Fixed a memory leak by clearing list update callbacks when a list is destroyed.

* **Tests**
  * Added tests to verify list callbacks are cleared on destruction.

* **Chores**
  * Prepared a patch release entry for the package.

## Checklist


packages/react/runtime/src/snapshot/list.ts
```diff
@@ -22,6 +22,12 @@ export function snapshotCreateList(
     componentAtIndexes,
   );
   const listID = __GetElementUniqueID(list);
+
+  lynx.getNative().addEventListener('__DestroyLifetime', () => {
```
HuJean: Add a safe check

hzy (changes requested):
You will need a test to avoid regress of this leak

packages/react/runtime/src/snapshot/list.ts
gaoachao: done

Issue #2125, opened 2026-01-22
[Bug]: updateGlobalProps can not update component when the root node is not a Component


HuJean:
### System Info

None

### Details

None

### Reproduce link

None

### Reproduce Steps

None

PR #2126, opened 2026-01-22
fix: increment `vnode._original` to make sure the component is re-ren…

HuJean:
#2125 

## Summary by CodeRabbit

* **Bug Fixes**
  * Ensure components reliably re-render when global properties are updated in edge cases where previous updates could be skipped.

* **Tests**
  * Expanded lifecycle tests to cover more update scenarios, hydration flows, and propagation of global prop changes.

* **Changelog**
  * Added a patch release entry documenting the behavioral fix.

## Checklist

### DIFF
--- a/packages/react/runtime/src/lynx/runWithForce.ts
+++ b/packages/react/runtime/src/lynx/runWithForce.ts
@@ -31,6 +31,9 @@ export function runWithForce(cb: () => void): void {
 
     const c = oldVNode[COMPONENT];
     if (c) {
+      if (vnode[ORIGINAL] != null && oldVNode[ORIGINAL] === vnode[ORIGINAL]) {
+        vnode[ORIGINAL] += 1;
+      }
       c[FORCE] = true;
     } else {
       // mount phase of a new Component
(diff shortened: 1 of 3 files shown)


packages/react/runtime/src/lynx/runWithForce.ts
```diff
@@ -31,6 +31,9 @@ export function runWithForce(cb: () => void): void {
 
     const c = oldVNode[COMPONENT];
     if (c) {
+      if (vnode[ORIGINAL] != null && oldVNode[ORIGINAL] === vnode[ORIGINAL]) {
```
upupming, line 34:
```suggestion
      if (vnode !=== oldVNode && vnode[ORIGINAL] != null && oldVNode[ORIGINAL] === vnode[ORIGINAL]) {
```

upupming: I think we can remove line 11-22 if we do this.

https://github.com/lynx-family/lynx-stack/blob/62ef34ddf7168c8fff8bf2200bbad200ae82e7c6/packages/react/runtime/src/lynx/runWithForce.ts#L11-L22

PR #2138, opened 2026-01-26
fix(webpack/react): deduplicate lazy bundles for same file imported via different paths

### LINKED ISSUE #455: [Bug]: same lazy bundle used multiple times with different rel path import should output single bundle
These dynamic points to the same `Foo.jsx` file, but will produce there different sub directory on lazy bundle output, we should make them to same single output.

```js
// index.tsx
const Foo = lazy(() => import('./Foo.jsx'))
// subdir/index.tsx
const Foo = lazy(() => import('../Foo.jsx'))
// some other places, use alias
const Foo = lazy(() => import('@/Foo.jsx'))
```

Image

See repro at: https://github.com/upupming/rspeedy-mininal-template/pull/1

orionmiz:
(WIP)

When the same file is imported via different relative paths (e.g., `./Foo.jsx`
from index.jsx and `../Foo.jsx` from subdir/importer.jsx), they now produce
a single bundle instead of duplicates.

The fix extends the `asyncChunkName` hook to receive ChunkGroup information,
allowing the ReactWebpackPlugin to normalize chunk names based on resolved
module paths rather than raw import strings.

Closes: https://github.com/lynx-family/lynx-stack/issues/455

@coderabbitai summary

## Checklist

### DIFF
--- a/.changeset/chatty-dryers-find.md
+++ b/.changeset/chatty-dryers-find.md
@@ -1,0 +1,6 @@
+---
+"@lynx-js/template-webpack-plugin": patch
+"@lynx-js/react-webpack-plugin": patch
+---
+
+fix: deduplicate lazy bundles when the same file is imported via different paths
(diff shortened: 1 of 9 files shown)


orionmiz:
self closing in favor of https://github.com/lynx-family/lynx-stack/pull/2961

PR #2143, opened 2026-01-27
feat: support bundle and load css in external bundle

upupming:
## Summary by CodeRabbit

* **New Features**
  * Support for bundling and loading CSS in external bundles, including main-thread stylesheet adoption and flush handling
  * Example updated to ensure external CSS loads reliably; CSS handling now applied unconditionally

* **Bug Fixes**
  * Improved error logging for external bundle load failures

* **Tests**
  * Added tests and fixtures to verify CSS is included and decoded in external bundles

* **Chores**
  * Dependency updates to support CSS serialization and parsing

## Checklist

### DIFF
--- a/packages/rspeedy/plugin-react/src/css.ts
+++ b/packages/rspeedy/plugin-react/src/css.ts
@@ -61,14 +61,12 @@ export function applyCSS(
           const rule = chain.module.rule(ruleName)
 
           removeLightningCSS(rule)
-          removeIgnoreCSS(rule)
 
           // Replace the CssExtractRspackPlugin.loader with ours.
           // This is for scoped CSS.
           rule
             .issuerLayer(LAYERS.BACKGROUND)
             .use(CHAIN_ID.USE.MINI_CSS_EXTRACT)
-            .before(CHAIN_ID.USE.CSS)
             .loader(CssExtractPlugin.loader)
             .end()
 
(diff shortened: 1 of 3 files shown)


luhc228 (dismissed):

packages/rspeedy/lynx-bundle-rslib-config/package.json
```diff
@@ -36,14 +36,17 @@
     "test": "vitest"
   },
   "dependencies": {
+    "@lynx-js/css-serializer": "workspace:*",
     "@lynx-js/runtime-wrapper-webpack-plugin": "workspace:*",
-    "@lynx-js/tasm": "0.0.20"
+    "@lynx-js/tasm": "0.0.26",
+    "css-tree": "^3.1.0"
```
colinaaa: It seems like the `css-tree` package has never been used.

packages/rspeedy/plugin-react/src/css.ts
```diff
@@ -140,6 +142,21 @@ export function applyCSS(
...
+        ) {
+          rule.uses.delete(CHAIN_ID.USE.IGNORE_CSS)
+        }
+      }
+
+      if (!chain.plugins.has(CHAIN_ID.PLUGIN.MINI_CSS_EXTRACT)) {
+        chain
+          .plugin(CHAIN_ID.PLUGIN.MINI_CSS_EXTRACT)
+          .use(CssExtractPlugin)
+      }
```
colinaaa: This should be related with [`output.emitCss`](https://rsbuild.rs/config/output/emit-css) option, which defaults to `true` when `output.target === 'web'`.

Instead of changing the underlying loader rules, we should set `output.target` to `'web'` in the Rslib config.

colinaaa (approved):

PR #2144, opened 2026-01-27
feat: add MainThreadValue

f0rdream:
## Summary by CodeRabbit

* **New Features**
  * Added typed `MainThreadValue` support for creating and hydrating values on the main thread.
  * Added the `useMotionValue` hook for retaining and updating motion values in main-thread interactions.
  * Improved worklet handling of captured object and member values.

* **Documentation**
  * Added usage guidance and examples for motion values and main-thread values.

* **Tests**
  * Added coverage for serialization, hydration, registration, cleanup, and motion value updates.

## Checklist


f0rdream:
Current design follow-up for the MainThreadValue work introduced here:

- RFC #3446 separates the realized main-thread object, its opaque background handle, and rebindable refs.
- Core PR #3453 implements that MainThreadObject model while retaining this work as an unchanged ancestor.
- Motion PR #3455 migrates MotionValue without replacing the real upstream object with a facade.
- Follow-up #3467 adds exact-type initial-payload inspection so Motion can render its first background style without keeping a duplicate WeakMap.

The latest boundary keeps serialization and hydration fields private to ReactLynx. Library code can recognize only its own handle type and read only its creation payload; it still cannot call the realized object from the background runtime or observe later main-thread mutations.

f0rdream:
Superseded by #3477, which implements the current MainThreadObject contract as the standalone ReactLynx core PR. Closing this stale combined design so review stays on the current implementation.

PR #2159, opened 2026-01-30
feat(worklet-runtime): add main thread flush loop detection

Yradex:
Add DEV-only guard that detects and reports re-entrant MainThread flush loops caused by MTS handlers. The guard tracks execution markers and compresses duplicate traces to help identify the source of infinite loops.

example error msg:

main-thread.js exception: Error: [ReactLynx][DEV] MainThread flush loop detected: render executed 257 times without yielding (limit=256). Trace: render <- element:setStyleProperty width <- MainThreadFunction id=835d:f06f6:1 <- event:layoutchange <- render <- element:setStyleProperty width <- MainThreadFunction id=835d:f06f6:1 <- event:layoutchange <- render <- element:setStyleProperty width <- MainThreadFunction id=835d:f06f6:1 <- event:layoutchange <- render <- element:setStyleProperty width <- MainThreadFunction id=835d:f06f6:1 <- event:layoutchange <- render <- element:setStyleProperty width <- MainThreadFunction id=835d:f06f6:1 <- event:layoutchange <- render ...

## Summary by CodeRabbit

* **New Features**
  * Development-only guard to detect and prevent MainThread flush loops caused by re-entrant UI mutation handlers.

* **Tests**
  * Added comprehensive test coverage for flush loop guard behavior and profile utilities.

* **Bug Fixes**
  * Improved error reporting in development mode with better diagnostic messages.

## Checklist

### DIFF
--- a/packages/react/worklet-runtime/src/api/element.ts
+++ b/packages/react/worklet-runtime/src/api/element.ts
@@ -51,7 +51,7 @@ export class Element {
   }
 
   public setStyleProperties(styles: Record<string, string>): void {
-    /* v8 ignore next 6 */
+    /* v8 ignore next 5 */
     if (__DEV__) {
       mainThreadFlushLoopMark(
         `element:setStyleProperties keys=${Object.keys(styles).length}`,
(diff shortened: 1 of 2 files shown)


packages/react/worklet-runtime/src/api/element.ts
```diff
@@ -28,16 +33,19 @@ export class Element {
   }
 
   public setAttribute(name: string, value: unknown): void {
+    mainThreadFlushLoopMark(`element:setAttribute ${name}`);
```
HuJean: This would be kept in production. Can it be wrapped with "DEV" for shaking?

PR #2188, opened 2026-02-03
fix: reduce redundant updates for handlers and gestures

Yradex:
## Summary by CodeRabbit

* **Bug Fixes**
  * Reduced redundant runtime updates when event handlers and gesture objects remain stable across re-renders, decreasing unnecessary native updates
  * Improved gesture removal to reliably clear gesture state on affected elements
  * Prevented unnecessary native updates when prop spreading doesn't semantically change handlers or gestures

## Summary

This PR reduces redundant native updates for stable main-thread handlers and gestures by separating cached background-side objects from the payloads committed to the main thread.

The previous flow mixed commit-time runtime metadata (for example `_execId`) into objects that were also reused for diffing. As a result, a handler or gesture could stay semantically unchanged across rerenders but still look different to the patching layer, which caused unnecessary native updates. This PR moves those paths to copy-on-commit, and also tightens the spread-driven gesture removal flow so cleanup only touches ReactLynx-owned gesture state.

### 1. Copy-on-commit for gestures and worklets

The main issue was that commit-time preparation mutated objects that were later reused for diffing. For stable references, that meant the diff saw runtime-only churn instead of a real semantic change.

Before, commit-time preparation effectively behaved like this:

```ts
baseGesture.callbacks[name] = onPostWorkletCtx(baseGesture.callbacks[name])!;
```

That mutates the cached gesture object itself.

Now the runtime prepares a committed copy instead:

```ts
const committedCallbacks = { ...baseGesture.callbacks };
committedCallbacks[name] = prepareWorkletForCommit(callback)!;

const committed = {
  ...baseGesture,
  callbacks: committedCallbacks,
};
```

This keeps the background-side cached object stable while still attaching `_execId` to the payload sent to the main thread.

### 2. Spread props only clone when commit-time rewriting is actually needed

Stable rerenders through spread props were still expensive because every changed spread was shallow-cloned and scanned, even when it only contained primitive attrs.

The old shape was effectively:

```ts
const committed = { ...spread };
for (const key in committed) {
  // scan and possibly rewrite every key
}
return committed;
```

Now the spread commit path clones lazily:

```ts
let committed;
for (const key in spread) {
  // only clone once we discover a worklet / gesture / timing flag
  committed ??= { ...spread };
}
return committed ?? spread;
```

That means ordinary spread updates keep the original object and avoid unnecessary allocation/scanning, while worklets and gestures still get rewritten correctly when needed.

### 3. Gesture removal from spread props now cleans up the right state

When a spread removed `main-thread:gesture`, the removal happened late in `updateSpread()`, after surviving keys had already been processed. That made cleanup order important.

A problematic transition looked like this:

```ts
{ 'clip-radius': 8, 'main-thread:gesture': g }
-> { 'clip-radius': 8 }
```

`clip-radius` still requires `flatten={false}`, so clearing `flatten` during gesture cleanup was incorrect. This PR fixes that by only clearing ReactLynx-owned gesture state and leaving unrelated spread state alone.

In addition, when `__RemoveGestureDetector` is available, the runtime no longer forcefully writes `gesture = null` itself. That avoids clobbering a user-provided `gesture` attr if detector cleanup is already owned by the native remove API.

### 4. Committed gesture serialization no longer depends on mutating the original object

The gesture commit path now serializes the committed copy instead of relying on the original object's `toJSON` behavior. This makes the runtime more robust when callbacks have been rewritten for commit-time transport and avoids reintroducing `_execId` churn through serialization.

Conceptually, the committed payload now comes from the committed object itself:

```ts
const serialize = () => serializeCommittedGesture(committedGesture);
return {
  ...committedGesture,
  serialize,
  toJSON: serialize,
};
```

This keeps serialization aligned with the object that actually contains the committed callbacks.

### 5. Behavioral boundary of this change

This PR is intentionally scoped to runtime-owned commit and cleanup behavior:

- Stable main-thread handlers and gestures should no longer trigger redundant native patches.
- Spread rerenders should avoid unnecessary cloning when no runtime rewrite is required.
- Removing `main-thread:gesture` from spread props should not clear unrelated state such as `flatten`.
- When detector removal can already be handled by `__RemoveGestureDetector`, ReactLynx should not additionally clobber a user-provided `gesture` attr.

## Checklist


HuJean (dismissed):

f0rdream:
LGTM

HuJean (approved):

PR #2200, opened 2026-02-06
fix(rspeedy/react): alias `use-sync-external-store`

HuJean:
## Summary by CodeRabbit

* **Bug Fixes**
  * Improved module alias resolution to include .js variants for the external store package (with-selector), ensuring consistent resolution.

* **Tests**
  * Added assertions covering .js variant alias resolution in both development and production configs.

* **Chores**
  * Added a changeset entry and updated bundle alias snapshots to reflect adjusted module path extensions.

## Checklist


packages/rspeedy/plugin-react/src/useSyncExternalStore.ts
```diff
@@ -11,19 +11,22 @@ export function applyUseSyncExternalStore(api: RsbuildPluginAPI): void {
     const useSyncExternalStoreEntries = [
       'use-sync-external-store',
       'use-sync-external-store/with-selector',
+      'use-sync-external-store/with-selector.js',
```
upupming, line 14: If we only keep `use-sync-external-store/with-selector.js` and remove `use-sync-external-store/with-selector`, does it work?

HuJean: Can not determine how the user imports it, so these two should be set both.

PR #2224, opened 2026-02-11
fix: properly cleanup __DestroyLifetime listeners and listCallbacks in snapshotDestroyList

gaoachao:
## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed cleanup of list destruction so listeners are removed and no lingering handlers remain.
  * List public methods now return safe default values after a list is removed instead of throwing.

* **Tests**
  * Updated tests to reflect post-removal behavior and the improved listener cleanup.

## Checklist

### DIFF
--- a/.changeset/tender-otters-cheat.md
+++ b/.changeset/tender-otters-cheat.md
@@ -1,0 +1,5 @@
+---
+"@lynx-js/react": patch
+---
+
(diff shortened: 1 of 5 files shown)


DwwWxx (commented):
lgtm

HuJean (approved):

hzy (approved):

Issue #2226, opened 2026-02-12
[Feature]: Display all entries and schema URLs in non-TTY terminals


knullptr:
### What problem does this feature solve?

Currently, plugin-qrcode provides an interactive terminal tool for selecting entries  and schemas in TTY mode:                                                              
```
  ●  Scan with Lynx
  │
  ◆  [QRCode]
  │
  ◆  [current schema]
  │
  ◆  Usage
  │   r  Switch entries
  │   a  Switch schema
  │   h  Help
  │   o  Open page (Require USB connection)
  │   q  Quit
  └
```
However, in non-TTY environments (CI/CD pipelines, AI coding agents, piped output like rspeedy dev | cat), the interactive loop is skipped entirely. Only the first entry's first schema URL is printed, and the QR code is rendered — which is useless in these contexts. This means there is no way to discover all available entries and their schema URLs when running in non-interactive terminals.

### What does the proposed API of configuration look like?

When running in a non-TTY terminal, the plugin should print a structured list of all entries with all their schema URLs:
```
  ●  Open with Lynx
  │
  ◆  entry-1
  │        https: [https schema]
  |        http: [http schema]
  │
  ◆  entry-2
  │        https: [https schema]
  |        http: [http schema]
  └
```

PR #2227, opened 2026-02-12
feat: print all entries and schema URLs in non-TTY mode (#2226)

knullptr:
In non-TTY environments (CI/CD, AI coding agents, piped output), the interactive loop was skipped and only the first entry's QR code was rendered — which is useless in these contexts.

Now `registerConsoleShortcuts` detects non-TTY and prints a structured list of all entries with all schema URLs, buffered into a single write to avoid interleaving with other async output.

## Summary by CodeRabbit

* **New Features**
  * Non-TTY support: plugin now prints all entries and all schema URLs in non-interactive environments and invokes provided callbacks for each URL.
* **Tests**
  * Added test coverage for non-TTY behavior, ensuring output content, callback invocation, and no interactive loop.
* **Chores**
  * Published a patch changeset documenting the non-TTY output behavior.

## Checklist


upupming:
LGTM, thanks!

PR #2277, opened 2026-02-28
feat: gesture callback memo

f0rdream:
Optimize gesture callbacks and relationships to prevent unnecessary gesture registration and rerenders.

## Summary by CodeRabbit

* **New Features**
  * Added a memoized main-thread callback hook to keep gesture callbacks stable while executing the latest logic.

* **Bug Fixes**
  * Prevented unnecessary state/version bumps and avoided duplicate or self-referential gesture linkages.

* **Tests**
  * Added tests for idempotent gesture configuration, composed-gesture deduplication, and memoized-callback behavior.

* **Chores**
  * Added a changeset for a patch release and updated build task dependency.

## Checklist

### DIFF
--- a/packages/lynx/gesture-runtime/src/baseGesture.ts
+++ b/packages/lynx/gesture-runtime/src/baseGesture.ts
@@ -141,8 +141,10 @@ abstract class BaseGesture<
   }
 
   updateConfig = (k: string, v: unknown): this => {
-    this.execId += 1;
-    (this.config as Record<string, unknown>)[k] = v;
+    if ((this.config as Record<string, unknown>)[k] !== v) {
+      this.execId += 1;
+      (this.config as Record<string, unknown>)[k] = v;
+    }
     return this;
   };
 
@@ -150,7 +152,9 @@ abstract class BaseGesture<
     k: keyof typeof this.callbacks,
     cb: GestureCallback<TEvent>,
   ): this => {
-    this.execId += 1;
+    if (!(k in this.callbacks)) {
+      this.execId += 1;
+    }
     // Wrapped callback is compatible with GestureCallback<TEvent> at runtime
     this.callbacks[k] = wrapCallback<TEvent>(
       cb,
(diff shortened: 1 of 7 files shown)


packages/lynx/gesture-runtime/src/utils/useMainThreadMemoizedFn.ts
```diff
@@ -0,0 +1,61 @@
...
+ *
+ * @example
+ * ```tsx
+ * const handleScroll = useMainThreadMemoizedFn((e: MainThread.TouchEvent) => {
+ *   'main thread';
+ *   console.log(count); // Access captured variable
+ * });
+ * ```
+ */
+export function useMainThreadMemoizedFn<T extends noop>(fn: T): T {
```
Yradex, line 27: This hook is implemented here, but it is not exported from the package entry (`src/index.ts`). As a result, users cannot consume it via `@lynx-js/gesture-runtime`. Could you either export it from the public entry or scope this as internal-only (and adjust release notes accordingly)?

packages/lynx/gesture-runtime/src/baseGesture.ts
```diff
@@ -193,13 +197,21 @@ abstract class BaseGesture<
...
-    if (gesture.type === GestureTypeInner.COMPOSED) {
-      this.waitFor = this.waitFor.concat(gesture.toGestureArray());
-    } else {
-      this.waitFor.push(
-        gesture as BaseGesture<BaseGestureConfig, GestureChangeEvent>,
-      );
+    const gestures = gesture.type === GestureTypeInner.COMPOSED
+      ? gesture.toGestureArray()
+      : [gesture as BaseGesture<BaseGestureConfig, GestureChangeEvent>];
+    const existingIds = new Set(this.waitFor.map(g => g.id));
```
Yradex: `externalWaitFor`, `externalSimultaneous`, and `externalContinueWith` now share almost identical normalize + self-filter + dedupe + conditional `execId` update logic. Would you consider extracting a small internal helper (parameterized by relation key) to reduce duplication and future drift risk?

packages/lynx/gesture-runtime/src/utils/useMainThreadMemoizedFn.ts
f0rdream: This should be internal only, and hopefully will ship officially in libs like `reactlynx-use`, not in `@lynx-js/gesture-runtime`.

Yes, I should not expose these concept in release note, great review!

packages/lynx/gesture-runtime/src/baseGesture.ts
f0rdream: Sure

Yradex (approved):

PR #2279, opened 2026-03-01
feat(react): add preact upstream tests for E2E pipeline verification

Huxpro:
## Summary

- Add `packages/react/preact-upstream-tests/` package that runs Preact's own test suite through the real ReactLynx dual-threaded rendering pipeline (BSI → patch → SI → Element PAPI → jsdom)
- Implements generic snapshot factory, BSI shims, Element PAPI dispatch, and pipeline render/commit hook wrapping — all without modifying upstream Preact tests (kept as a git submodule)
- Structured `skiplist.json` (Hermes-inspired) declaratively manages test exclusions with documented reasons
- **Current results**: 203 pass / 0 fail / 107 skip (across 310 tests in 5 files)

## Test plan
- [ ] `pnpm preact:init && pnpm test` passes in `packages/react/preact-upstream-tests/`
- [ ] CI workflow runs successfully
- [ ] No regressions in existing `@lynx-js/react` tests

## Summary by CodeRabbit

* **New Features**
  * Added a comprehensive Preact upstream E2E test suite with dual-mode runs (compiled and no-compiler), test scripts, and workspace integration.
  * Added a grouped test coverage dashboard and reporting tool.
  * Added flexible skiplist-driven test filtering and targeted SKIPLIST_ONLY execution.

* **Documentation**
  * Added detailed README and skiplist guidance for running and maintaining the test suite.

* **API**
  * Added applyProp helper for test env property handling.
  * Added transform plugin for JSX/transform integration.


packages/react/preact-upstream-tests/package.json
```diff
@@ -0,0 +1,25 @@
+{
+  "name": "@lynx-js/preact-upstream-tests",
+  "version": "0.0.0",
+  "private": true,
+  "scripts": {
+    "preact:init": "cd ../../.. && git submodule update --init packages/react/preact-upstream-tests/preact",
```
upupming, line 6: Because this package is picked up by the root Vitest project globs, introducing a mandatory manual `preact:init` step here also makes `pnpm test` depend on local submodule state. On a fresh checkout this will fail before the reviewer even gets to this package. Could we either keep this suite opt-in at the root level or make the submodule bootstrap automatic/self-checking?

packages/react/preact-upstream-tests/setup-compiled.js
```diff
@@ -0,0 +1,109 @@
...
+
+// --- Wrapper transparency shim ---
+// The compiler wraps text expressions in <wrapper> elements. Upstream Preact
+// tests assert on innerHTML and expect bare text. Patch the innerHTML getter
+// to strip <wrapper>/<\/wrapper> tags so these tests pass.
+
+const _innerHTMLDesc = Object.getOwnPropertyDescriptor(Element.prototype, 'innerHTML');
+Object.defineProperty(Element.prototype, 'innerHTML', {
+  get() {
+    return _innerHTMLDesc.get.call(this).replace(/<\/?wrapper>/g, '');
```
upupming, line 35: This makes the compiled suite report green by rewriting what the assertion reads rather than by fixing the structural difference. If `<wrapper>` is a real output difference in compiled mode, masking it at `innerHTML` level seems risky because it hides exactly the class of regressions this harness is supposed to surface.

packages/testing-library/testing-environment/src/lynx/applyProp.ts
```diff
@@ -0,0 +1,73 @@
...
+  if (key === 'id') {
+    __SetID(el, (value ?? '') as string);
+    return;
+  }
+  if (key === 'htmlFor') {
+    __SetAttribute(el, 'for', value as string | null);
+    return;
+  }
+  if (key.startsWith('data-')) {
+    __AddDataset(el, key.slice(5), (value ?? '') as string);
```
upupming, line 51: I think this changes `data-*` removal semantics: when `value` becomes `null`/`false`, this still goes through `__AddDataset(..., "")`, so the key is retained with an empty string instead of being removed. That diverges from both DOM behavior and the runtime spread path, which clears missing dataset entries by rebuilding the dataset object.

upupming (commented):
Thanks for putting this together — I think the overall goal here is valuable. Pulling upstream Preact tests into the Lynx pipeline should give us a much stronger semantic regression signal than only testing our own fixtures.

My main concern is that a few of the current adaptations weaken that signal or broaden the surface area more than necessary:

- the suite now depends on manual submodule initialization, which makes root-level test execution more fragile on a fresh checkout;
- the compiled-mode `innerHTML` patch appears to hide a real structural output difference instead of surfacing it as a known gap;
- `applyProp()` looks useful as a harness helper, but exposing it from `@lynx-js/testing-environment` feels like a layering leak, and the current `data-*` handling seems to diverge from removal semantics.

I think this is very close to being a strong testing harness, but I’d feel more comfortable if we kept the bridge code narrowly scoped, avoided patches that rewrite observed output, and made the suite opt-in or self-bootstrapping from the repo root.

packages/testing-library/testing-environment/src/index.ts
```diff
@@ -14,6 +14,7 @@ import { GlobalEventEmitter } from './lynx/GlobalEventEmitter.js';
 export { initElementTree } from './lynx/ElementPAPI.js';
 export type { LynxElement } from './lynx/ElementPAPI.js';
```
upupming, line 15: `applyProp()` looks useful for this harness, but exporting it from `@lynx-js/testing-environment` feels like it expands that package from "Node.js implementation of Lynx APIs" into "DOM/Preact-to-Lynx adapter helpers". Since this helper is not itself a Lynx API, would it make sense to keep it private to the upstream-test harness (or at least under a clearly internal testing-library surface) instead of making it part of the testing-environment public API?

PR #2297, opened 2026-03-03
feat: remove stale gestures when diff

f0rdream:
## Summary by CodeRabbit

* **New Features**
  * Exposed a gesture-detector removal API and added an interactive Gesture Lifecycle example app with config and build scripts.

* **Bug Fixes**
  * Remove stale gesture detectors when gestures are removed, replaced, or reordered to prevent lingering state and incorrect handlers.

* **Tests**
  * Expanded unit and integration tests covering gesture lifecycle, deduplication, diffs, and detector wiring.

* **Style**
  * Added themed stylesheet for the gesture example UI.

## Checklist

### DIFF
--- a/packages/react/runtime/__test__/gesture/processGesture.test.ts
+++ b/packages/react/runtime/__test__/gesture/processGesture.test.ts
@@ -1,0 +1,271 @@
+import { afterEach, beforeEach, describe, expect, it, vi } from 'vitest';
+
+import { processGesture } from '../../src/gesture/processGesture.js';
+
+vi.mock('@lynx-js/react/worklet-runtime/bindings', () => ({
+  onWorkletCtxUpdate: vi.fn(),
+}));
+
+function createSerializedGesture(id: number) {
+  return {
+    id,
+    type: 0,
+    callbacks: {
+      onUpdate: {
+        _wkltId: 'bdd4:dd564:2',
+      },
+    },
+    __isSerialized: true,
+  };
+}
+
+function createSerializedComposedGesture(gestures: ReturnType<typeof createSerializedGesture>[]) {
+  return {
+    type: -1,
+    gestures,
+    __isSerialized: true,
+  };
+}
+
+describe('processGesture', () => {
+  let setAttribute: ReturnType<typeof vi.fn>;
+  let setGestureDetector: ReturnType<typeof vi.fn>;
+  let removeGestureDetector: ReturnType<typeof vi.fn>;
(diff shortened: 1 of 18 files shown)


packages/react/runtime/src/gesture/processGesture.ts
```diff
@@ -58,24 +140,61 @@ export function processGesture(
...
-    const baseGesture = gesture as BaseGesture;
-    const oldBaseGesture = oldGesture as BaseGesture | undefined;
+  // On update, remove old detectors first to avoid stale callbacks.
+  for (const oldBaseGesture of oldBaseGesturesById.values()) {
+    removeGestureDetector(dom, oldBaseGesture.id);
+  }
+
+  for (const [index, baseGesture] of uniqBaseGestures.entries()) {
+    const oldBaseGesture = oldBaseGesturesById.get(baseGesture.id)
+      ?? uniqOldBaseGestures[index];
```
Yradex: [P1] This fallback can reuse the same old gesture twice when a composed gesture both preserves one id and inserts/reorders another. For example, with `old=[1,2]` and `new=[2,3]`, the first new gesture matches old `2` via `oldBaseGesturesById.get(2)`, but the second then falls back to `uniqOldBaseGestures[1]`, which is still old `2` because matched entries are never removed from the fallback pool. That means `getGestureInfo()` can pair both new detectors with the same old callbacks, and during first-screen hydration `onWorkletCtxUpdate()` will call `_hydrateCtx(new-2, old-2)` and `_hydrateCtx(new-3, old-2)`. I reproduced this locally with a temporary vitest. The matching here needs to stay one-to-one: once an old gesture has been consumed by an id match, it should not remain eligible for the positional fallback.

f0rdream: Good point, will change that

packages/react/runtime/__test__/gesture/processGesture.test.ts
```diff
@@ -0,0 +1,271 @@
+import { afterEach, beforeEach, describe, expect, it, vi } from 'vitest';
+
+import { processGesture } from '../../src/gesture/processGesture.js';
+
+vi.mock('@lynx-js/react/worklet-runtime/bindings', () => ({
+  onWorkletCtxUpdate: vi.fn(),
+}));
```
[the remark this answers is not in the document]
f0rdream: Resolved

Yradex (approved):

PR #2360, opened 2026-03-19
feat: support createElement

HuJean:
## Summary by CodeRabbit

* **New Features**
  * Full support for React.createElement with improved handling for string vs. component elements and runtime snapshot fallback.

* **Examples / Documentation**
  * Added a new interactive example app with styling, build config, and usage samples.

* **Tests**
  * Large suite of snapshot tests covering element creation, hydration, events, and update scenarios.

* **Types**
  * Runtime creators split into ScrollView and Frame typed APIs.

## Checklist


packages/react/runtime/src/snapshot.ts
```diff
@@ -123,6 +127,33 @@ export const snapshotManager: {
   ]),
 };
 
+export function createRuntimeSnapshot(type: string): void {
```
Yradex: This runtime snapshot definition does not carry any CSS scope metadata (`cssId` / `entryName`). `SnapshotInstance.ensureElements()` only calls `__SetCSSId(...)` when those fields are present, so dynamically created host nodes will diverge from compiled JSX nodes in scoped-CSS bundles: the element is created, but its scoped class rules never match. Can we propagate the current bundle scope into runtime snapshots as well?

packages/react/runtime/src/snapshot.ts
```diff
@@ -293,6 +324,9 @@ export interface SerializedSnapshotInstance {
 
 const DEFAULT_ENTRY_NAME = '__Card__';
 const DEFAULT_CSS_ID = 0;
+const COMPILED_SNAPSHOT = '__snapshot';
+
+export const isCompiledSnapshot: (type: string) => boolean = (type: string) => type.startsWith(COMPILED_SNAPSHOT);
```
Yradex: I think `startsWith("__snapshot")` is too narrow here. This codebase already documents compiled snapshot ids like `entryName:__snapshot_xxx` and `https://example.com/main.lynx.bundle:__snapshot_xxx` for lazy / standalone bundles. Those would now be treated as runtime tags, so a missing compiled snapshot would silently fall back to `createRuntimeSnapshot(type)` instead of surfacing the expected missing-snapshot error. It would be safer to detect `__snapshot` after the optional prefix rather than only at the beginning of the whole string.

packages/react/runtime/src/snapshot/snapshot.ts
```diff
@@ -673,7 +709,21 @@ export class SnapshotInstance {
 
     if (typeof key === 'string') {
       // for more flexible usage, we allow setting non-indexed attributes
-      (this.__extraProps ??= {})[key] = value;
+      const oldValue = (this.__extraProps ??= {})[key];
+      this.__extraProps[key] = value;
+      // dynamic snapshot should update __extraProps to element
+      if (this.__snapshot_def.spreadIndex !== undefined && !isDirectOrDeepEqual(oldValue, value)) {
```
Yradex: This new main-thread branch now consumes string-key props through `spreadIndex`, but the background path still does not mirror that conversion. `BackgroundSnapshotInstance.setAttribute(key, value)` continues to push raw string-key values into `snapshotPatch`, bypassing `setAttributeImpl` / `transformSpread`. That means event handlers, refs and other non-JSON-safe props on `createElement()`-generated host nodes will still be serialized incorrectly during hydration/update (for example functions become `null` once the patch is JSON-serialized). Could we make the background string-key path produce the same spread payload that this branch expects?

examples/react/src/App.tsx
```diff
@@ -33,8 +55,17 @@ export function App() {
               ? <image src={reactLynxLogo} className='Logo--react' />
               : <image src={lynxLogo} className='Logo--lynx' />}
           </view>
-          <text className='Title'>React</text>
-          <text className='Subtitle'>on Lynx</text>
+          <DynamicTag
```
Yradex: I’d suggest keeping the React example as simple and declarative as possible. This PR is about adding `createElement` support, but the main example also serves as the facade demo for new users. Replacing straightforward JSX with `DynamicTag` here makes the entry example harder to scan and may unintentionally suggest that this indirection is the preferred style. It would be clearer to keep the main demo in plain JSX and showcase `createElement` in a separate focused example or a small isolated section.

packages/react/runtime/src/snapshot.ts
```diff
@@ -55,6 +58,7 @@ export function clearPage(): void {
 }
 
 export const __DynamicPartChildren_0: [DynamicPartType, number][] = [[DynamicPartType.Children, 0]];
+export const __DynamicPartListChildren_0: [DynamicPartType, number][] = [[DynamicPartType.ListChildren, 0]];
```
upupming: Since we add this, we can simple snapshot generation by swc.

Change:

https://github.com/lynx-family/lynx-stack/blob/b5ebec8745c2000c4c8c01f99714b70724bd80c6/packages/react/transform/__test__/fixture.spec.js#L196-L199

to something like this:

https://github.com/lynx-family/lynx-stack/blob/b5ebec8745c2000c4c8c01f99714b70724bd80c6/packages/react/transform/crates/swc_plugin_snapshot/tests/__swc_snapshots__/lib.rs/basic_component.js#L9

We can do this in future PRs.

packages/react/runtime/src/snapshot.ts
```diff
@@ -123,6 +127,33 @@ export const snapshotManager: {
   ]),
 };
 
+export function createRuntimeSnapshot(type: string): void {
+  const isListHolder = type === 'list';
+  snapshotManager.values.set(type, {
```
upupming: I believe we should only set once. We should add `if (!snapshotManager.values.has(type))` guard here.

packages/react/runtime/src/snapshot/snapshot/snapshot.ts
```diff
@@ -673,7 +709,21 @@ export class SnapshotInstance {
...
+      if (this.__snapshot_def.spreadIndex !== undefined && !isDirectOrDeepEqual(oldValue, value)) {
+        const spreadIndex = this.__snapshot_def.spreadIndex;
+        this.__values ??= [];
+        this.__values[spreadIndex] ??= {};
+        const oldSpread = this.__values[spreadIndex] as Record<string, unknown>;
+        this.__values[spreadIndex] = {
+          ...oldSpread,
+          __spread: true,
+          [key]: value as unknown,
+        };
```
upupming: Why do we need `...oldSpread`. Can we add a test to show attribute remove is okay here?

examples/react/src/App.tsx
HuJean: get

packages/react/runtime/src/snapshot.ts
HuJean: I think can add a key of snapshot rather than using type only

packages/react/runtime/lepus/index.js
```diff
@@ -22,6 +23,7 @@ export function createElement(type, props, children) {
   for (i in props) {
     if (i == 'key') key = props[i];
     else if (i == 'ref') ref = props[i];
+    else if (eventRegExp.test(i) && typeof props[i] === 'function') normalizedProps[i] = 1;
```
upupming: Can we just omit `eventRegExp.test(i)` (Not sure if it is slow), I think all functions are event callbacks now.

packages/react/runtime/src/snapshot/snapshot/definition.ts
```diff
@@ -150,3 +167,50 @@ export function createSnapshot(
...
+        if (__JS__ && !__DEV__) {
+          return;
+        }
+        /* v8 ignore stop */
+        updateSpread(ctx, index, oldValue as Record<string, unknown>, 0);
+      },
+    ],
+    slot: isListHolder ? __DynamicPartListSlotV2_0 : __DynamicPartSlotV2_0,
+    isListHolder,
+    spreadIndex: 0,
```
upupming: Is this always 0? If so can we avoid this

packages/react/runtime/lepus/index.js
HuJean: This API itself is not called much, so it has little impact.

packages/react/runtime/src/snapshot/snapshot/definition.ts
HuJean: for cloneElement in future

upupming (dismissed):

Yradex (approved):
Approved from the runtime/createElement side. I rechecked the latest head and the previously discussed createElement spread/key handling issues look resolved.

The remaining CodSpeed Performance Analysis failure appears to come from the spread handling logic now being included in the runtime bundle. That increases the JavaScript bundle size, which in turn increases JS loading time in the benchmark. I do not see it as a functional correctness blocker for this PR.

upupming (approved):

PR #2363, opened 2026-03-20
fix: compile flex numbers in style objects to string values

MoonfaceX:
## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed handling of numeric `flex` values in inline styles to convert them to string literals.

* **Tests**
  * Added test coverage for numeric flex value transformation in inline style objects.

## Checklist

### DIFF
--- a/packages/react/transform/__test__/css/inline-style.spec.js
+++ b/packages/react/transform/__test__/css/inline-style.spec.js
@@ -50,4 +50,17 @@ describe('Parse Inline Style', () => {
       await formatMessages(result.warnings, { kind: 'warning', color: false }),
     ).toMatchInlineSnapshot(`[]`);
   });
+
+  it('should compile flex number in style object to string value', async () => {
+    const result = await transformReactLynx(
+      `\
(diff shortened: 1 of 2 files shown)


packages/react/transform/__test__/css/inline-style.spec.js
```diff
@@ -50,4 +50,17 @@ describe('Parse Inline Style', () => {
       await formatMessages(result.warnings, { kind: 'warning', color: false }),
     ).toMatchInlineSnapshot(`[]`);
   });
+
+  it('should compile flex number in style object to string value', async () => {
+    const result = await transformReactLynx(
+      `\
+const style = flag ? { flex: 1, width: '100px' } : { flex: 2 };
```
HuJean, line 57: Most values ​​cannot be covered during compilation. If value must be a string, it should be handled strictly at runtime.

MoonfaceX: agree

PR #2402, opened 2026-03-30
feat: emit debug metadata assets

upupming:
## Summary
- add opt-in debug metadata asset emission to the ReactLynx rsbuild and webpack pipeline
- expose template `beforeEncode` context so custom plugins can attach `debugMetadataUrl` and related metadata before encode
- add a `react-node-index` example that uploads the generated debug metadata and records git metadata
- keep runtime `nodeIndex` fields for client compatibility

## Verification
- `pnpm turbo build`
- `pnpm vitest -u`
- `cargo test`


.changeset/ten-lions-accept.md
```diff
@@ -0,0 +1,6 @@
+---
+'@lynx-js/react-rsbuild-plugin': patch
+'@lynx-js/react-webpack-plugin': patch
+---
+
+Add `enableUiSourceMap` option to enable UI source map generation and debug-metadata asset emission.
```
luhc228, line 6: how about using `enableUISourceMap` instead of `enableUiSourceMap`

packages/webpack/template-webpack-plugin/src/LynxDebugMetadataPlugin.ts
```diff
@@ -0,0 +1,250 @@
...
+              JSON.stringify(
+                createDebugMetadataAsset(
+                  uiSourceMapRecords,
+                ),
+                null,
+                2,
+              ),
+            ),
+          );
+          args.intermediateAssets.push(debugMetadataAssetName);
```
luhc228, line 117: Should the `main-thread.js`, `debug-info.json` and `background.js` in `args.intermediateAssets`

packages/webpack/react-webpack-plugin/src/loaders/main-thread.ts
```diff
@@ -8,6 +8,8 @@ import type { LoaderDefinitionFunction } from '@rspack/core';
 import { getMainThreadTransformOptions } from './options.js';
 import type { ReactLoaderOptions } from './options.js';
 
+const UI_SOURCE_MAP_RECORDS_BUILD_INFO = 'lynxUiSourceMapRecords';
```
luhc228: Can this variable use that defined in template-webpack-plugin?

.changeset/ten-lions-accept.md
upupming: This is auto genereated by rust, the same as `preserveJsx`: 

https://github.com/lynx-family/lynx-stack/blob/codex/node-index-asset/packages/react/transform/index.d.ts#L586-L601

packages/webpack/template-webpack-plugin/src/LynxDebugMetadataPlugin.ts
upupming: No need because they have been hard-coded

colinaaa (approved):

gaoachao (approved):

PR #2441, opened 2026-04-09
feat: simplify implement of hooks for main-thread runtime

HuJean:
## Summary by CodeRabbit

* **New Features**
  * Added dedicated hooks entrypoints, including a main-thread–optimized hooks runtime and a centralized internal constants entry.

* **Tests**
  * Improved test setup to initialize environment state and added layered alias tests for build-time resolution.

* **Refactor**
  * Reorganized public export surface for clearer separation between runtime internals and public entry points.

* **Chores**
  * Updated package export/type mappings, added release metadata, and prepared minor package bumps.

## Checklist


Yradex (changes requested):
`useId()` currently recreates `mask = [0, 0]` on every call, so it always returns `P0-0`.

That means multiple `useId()` calls in the same tree will collide, which breaks any logic relying on stable unique IDs. The new test only covers a single call, so this can still pass with green CI.

Could we make the generated id persistent across hook calls instead of rebuilding the counter each time?

Yradex (dismissed):
The follow-up commits addressed the `useId` issue I called out earlier, and I don't see any remaining blocking problems in the latest revision.

packages/react/runtime/src/hooks/mainThread.ts
```diff
@@ -0,0 +1,212 @@
...
+// LICENSE file in the root directory of this source tree.
+
+/**
+ * Implements hooks in main thread.
+ * This module is modified from preact/hooks
+ *
+ * internal-preact/hooks/dist/hooks.mjs
+ */
+
+import { options } from 'preact';
```
upupming, line 12: Can we avoid import preact in main thread?

packages/react/runtime/src/hooks/react.ts
```diff
@@ -123,7 +122,7 @@ function useStateWithProfile<S>(
   return [state, tracedSetState];
 }
 
-const useState: typeof usePreactState = isProfiling
+const useState: typeof usePreactState = (__BACKGROUND__ && isProfiling)
```
upupming, line 125: It is strange to make `(__BACKGROUND__ && isProfiling)` together, is it expected?

packages/rspeedy/lynx-bundle-rslib-config/test/external-bundle.test.ts
```diff
@@ -704,7 +704,6 @@ describe('pluginReactLynx', () => {
           "preact/compat/server$": "<ROOT>/node_modules/<PNPM_INNER>/@lynx-js/internal-preact/compat/server.mjs",
           "preact/debug$": "<ROOT>/node_modules/<PNPM_INNER>/@lynx-js/internal-preact/debug/dist/debug.mjs",
           "preact/devtools$": "<ROOT>/node_modules/<PNPM_INNER>/@lynx-js/internal-preact/devtools/dist/devtools.mjs",
-          "preact/hooks$": "<ROOT>/node_modules/<PNPM_INNER>/@lynx-js/internal-preact/hooks/dist/hooks.mjs",
```
upupming, line 707: Should `preact/hooks$` be aliased to ours?

packages/react/runtime/src/hooks/react.ts
HuJean: Do not need profile in main thread

packages/rspeedy/lynx-bundle-rslib-config/test/external-bundle.test.ts
HuJean: There has test in plugin-react-alias

upupming (approved):

Yradex (approved):

PR #2457, opened 2026-04-14
fix: prevent memory leaks by explicitly disposing WASM contexts and d…

PupilTong:
…etaching event listeners during component teardown

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevented memory leaks by ensuring event listeners and related resources are fully released when views are removed.
  * Improved disposal ordering to guarantee native bindings and background work are cleaned up, reducing use-after-free and hanging shutdown scenarios.

## Checklist

### DIFF
--- a/.changeset/fix-wasmcontext-memory-leak.md
+++ b/.changeset/fix-wasmcontext-memory-leak.md
@@ -1,0 +1,5 @@
+---
+"@lynx-js/web-core": patch
+---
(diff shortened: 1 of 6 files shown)


PupilTong:
@coderabbitai reconsider your comments with latest PR

Sherry-hue (approved):

PR #2483, opened 2026-04-20
feat(css): enable sourcemaps and remap diagnostics

luhc228:
In this commit, @lynx-js/css-serializer generates cssMap with bundle-CSS line and column locations preserved on style rules and declarations. These locations are written into tasm.json as part of the encoder input, so TASM can report CSS diagnostics using coordinates that correspond to the bundled CSS output.

Those bundle-space coordinates are then remapped in template-webpack-plugin through main.css.map, allowing emitted webpack warnings to point back to the original source CSS files. The cssMap itself is still grouped by cssId and consumed as the style AST for template encoding; its preserved loc fields are what make later diagnostic remapping accurate.

image

## Summary by CodeRabbit

* **New Features**
  * CSS source maps enabled by default in build output
  * CSS diagnostics are remapped to original sources and emitted as compilation warnings
  * CSS input handling now accepts raw strings or { content } objects

* **Bug Fixes**
  * Improved precision of CSS diagnostic messages via source-map resolution
  * Duplicate CSS warnings are deduplicated before emission

* **Tests**
  * Added tests and updated snapshots for source-location preservation and diagnostics

* **Chores**
  * Updated peer compatibility for template plugins

## Checklist


colinaaa (dismissed):

PupilTong (approved):

packages/webpack/template-webpack-plugin/src/LynxEncodePlugin.ts
```diff
@@ -226,11 +234,54 @@ export class LynxEncodePluginImpl {
...
+            } else {
+              webpackWarning.loc = {
+                start: {
+                  line: diagnostic.line,
+                  column: diagnostic.column,
+                },
+              };
+            }
+
+            compilation.warnings.push(webpackWarning);
```
upupming, line 281: How about external bundle? Do we need report this warnings?

colinaaa (approved):

packages/webpack/template-webpack-plugin/src/LynxEncodePlugin.ts
luhc228: Of course we need. But `@lynx-js/tasm` will not check the CSS in customSections. @zhongyr is supporting now.

upupming (approved):
LGTM, btw does it work for scoped css too?

PR #2504, opened 2026-04-22
refactor: replace static component catalog JSON files with an automat…

PupilTong:
…ed TypeScript generator script

## Summary by CodeRabbit

* **Chores**
  * Updated package exports to reference built artifacts and added build scripts and repo build config.
  * Added linter/ESLint ignore entries and a changeset placeholder file.
* **New Tools**
  * Added an automated catalog-generation tool to produce component schemas.
* **Breaking/Behavioral**
  * Removed multiple published catalog schemas and narrowed/changed public prop typings for many UI components.

## Checklist


Sherry-hue:
Could we extract this tool into a separate package? 
Perhaps it would also be needed when generating the catalog for custom components.

Sherry-hue:
It would be even better if tests.

Sherry-hue (approved):

PR #2514, opened 2026-04-23
fix: CSS source map line offsets when wrapping extracted CSS with `@cssId`

luhc228:
## Summary by CodeRabbit

* **Bug Fixes**
  * Corrected CSS source map line offsets when extracted CSS is wrapped with cssId metadata so reported source positions are accurate.

* **New Features**
  * Added a public API to process, resolve and deduplicate Tasm CSS diagnostics, plus related diagnostic types and exports for richer build-time reporting.

* **Tests**
  * Added unit and integration tests for source-map offset behavior and the diagnostics processing pipeline.

* **Chores**
  * Added changeset metadata to publish a patch release.

## Checklist

### DIFF
--- a/packages/webpack/template-webpack-plugin/src/index.ts
+++ b/packages/webpack/template-webpack-plugin/src/index.ts
@@ -19,6 +19,12 @@ export type {
 } from './LynxTemplatePlugin.js';
 export { LynxEncodePlugin } from './LynxEncodePlugin.js';
 export type { LynxEncodePluginOptions } from './LynxEncodePlugin.js';
+export { processTasmCSSDiagnostics } from './cssDiagnostics.js';
+export type {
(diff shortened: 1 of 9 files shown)


packages/webpack/template-webpack-plugin/src/index.ts
```diff
@@ -19,6 +19,12 @@ export type {
 } from './LynxTemplatePlugin.js';
 export { LynxEncodePlugin } from './LynxEncodePlugin.js';
 export type { LynxEncodePluginOptions } from './LynxEncodePlugin.js';
+export { processTasmCSSDiagnostics } from './cssDiagnostics.js';
```
luhc228, line 22: This is for reuse in other packages.

colinaaa (approved):

PR #2530, opened 2026-04-27
feat: add playground native preview

Sherry-hue:
## Summary by CodeRabbit

* **New Features**
  * Shareable dev-bundle flow with shortened bundle URL, QR, and copy-to-clipboard feedback.
  * In-memory payload service for creating short payload URLs and a dev endpoint to fetch the bundle URL.
  * URL-driven globalProps support for overriding demo data.

* **Improvements**
  * Improved QR error reporting and quieter placeholder behavior.
  * Better QR/preview styling and more robust base64url encoding/decoding.

* **Chores**
  * Dev start now builds the bundle before launching the dev server.

* **Removals**
  * Two demo pages removed from the playground.

## Checklist

Add a native preview QR code and pass the data to A2UIRender via globalProps:

image


PupilTong (dismissed):

Huxpro (commented):
## Review

### 1. `__rspeedy_url` returns the wrong port when the configured port is already in use

`rsbuild.config.ts:27`:
```ts
const PORT = Number(process.env.PORT ?? 3000);
```

This is evaluated at module load time, but rsbuild auto-increments the port if it's already taken (e.g. 3000 → 3001). The `/__rspeedy_url` endpoint then returns a URL with port 3000 while the server is actually on 3001 — native preview will fail to connect.

The actual listening port should be read from the server instance at request time rather than from the static config value.

### 2. Dead `didLogRef` effect in `App.tsx`

`lynx-src/App.tsx:120-124`:
```tsx
const didLogRef = useRef(false);
useEffect(() => {
  if (didLogRef.current) return;
  didLogRef.current = true;
}, [effectiveData, globalProps, globalPropsData, initData, rawInitData]);
```

This sets a flag once and does nothing else — looks like debug logging was removed but the scaffolding was left behind. Should be deleted.

### 3. `copyToClipboard` uses deprecated `document.execCommand('copy')`

`DemosPage.tsx:31-47`: The comment says "some lint configs treat `navigator` as a Node-only API" but this code only runs in the browser. `navigator.clipboard.writeText()` is the standard API supported in all major browsers. `document.execCommand('copy')` is deprecated and will eventually be removed.

```tsx
async function copyToClipboard(text: string): Promise<boolean> {
  try {
    await navigator.clipboard.writeText(text);
    return true;
  } catch {
    return false;
  }
}
```

Sherry-hue:
> ## Review
> ### 1. `__rspeedy_url` returns the wrong port when the configured port is already in use
> `rsbuild.config.ts:27`:
> 
> ```ts
> const PORT = Number(process.env.PORT ?? 3000);
> ```
> 
> This is evaluated at module load time, but rsbuild auto-increments the port if it's already taken (e.g. 3000 → 3001). The `/__rspeedy_url` endpoint then returns a URL with port 3000 while the server is actually on 3001 — native preview will fail to connect.
> 
> The actual listening port should be read from the server instance at request time rather than from the static config value.
> 
> ### 2. Dead `didLogRef` effect in `App.tsx`
> `lynx-src/App.tsx:120-124`:
> 
> ```tsx
> const didLogRef = useRef(false);
> useEffect(() => {
>   if (didLogRef.current) return;
>   didLogRef.current = true;
> }, [effectiveData, globalProps, globalPropsData, initData, rawInitData]);
> ```
> 
> This sets a flag once and does nothing else — looks like debug logging was removed but the scaffolding was left behind. Should be deleted.
> 
> ### 3. `copyToClipboard` uses deprecated `document.execCommand('copy')`
> `DemosPage.tsx:31-47`: The comment says "some lint configs treat `navigator` as a Node-only API" but this code only runs in the browser. `navigator.clipboard.writeText()` is the standard API supported in all major browsers. `document.execCommand('copy')` is deprecated and will eventually be removed.
> 
> ```tsx
> async function copyToClipboard(text: string): Promise<boolean> {
>   try {
>     await navigator.clipboard.writeText(text);
>     return true;
>   } catch {
>     return false;
>   }
> }
> ```

grate! These issues have been fixed.

PupilTong (approved):

PR #2544, opened 2026-04-29
feat(react): add experimental element template runtime path

Yradex:
@coderabbitai summary

## Overview

- Adds the first experimental Element Template runtime path for ReactLynx, gated behind `experimental_useElementTemplate`.
- Wires rsbuild and webpack so the flag switches React runtime aliases, transform options, and template encoding to the Element Template path without changing the default Snapshot lane.
- Adds an `examples/react-element-template` app plus bundle analysis coverage so the experimental path has a concrete integration target while the implementation continues to evolve.

## Key Points

- The default ReactLynx path remains Snapshot-based. Element Template is opt-in through:

  ```ts
  pluginReactLynx({
    experimental_useElementTemplate: true,
  })
  ```

- When the flag is enabled, the main-thread JSX/runtime aliases resolve to `@lynx-js/react/element-template` entries, and the transform receives `elementTemplate` options instead of Snapshot options for the relevant bundle.
- The new runtime path creates main-thread Element Template instances, serializes them with `__SerializeElementTemplate`, and sends the serialized result to the background thread for hydration and later patching.
- The empty changeset is intentional: this PR wires an incomplete experimental path for continued development, and the release-facing changeset should be added once the Element Template capability is ready to ship as a coherent user-visible feature.

## Runtime / Build Contract

- `@lynx-js/react` now exposes experimental Element Template entries:

  ```text
  @lynx-js/react/element-template
  @lynx-js/react/element-template/internal
  @lynx-js/react/element-template/jsx-runtime
  @lynx-js/react/element-template/jsx-dev-runtime
  ```

- `ReactWebpackPlugin` collects Element Template definitions from module build info and writes them into template encode data under `elementTemplate`.
- The template plugin initializes `encodeData.elementTemplate` as an object so producers can add compiled templates and encoders can consume the same field consistently.
- Element Template native calls keep identity and ordering explicit through template ids, attribute slots, element slots, and serialized root instances; Snapshot patch buffers and `SnapshotInstance` internals are not part of the ET contract.

## Checklist


HuJean (approved):

examples/react-element-template/lynx.config.js
```diff
@@ -0,0 +1,27 @@
+import { pluginQRCode } from '@lynx-js/qrcode-rsbuild-plugin';
+import { pluginReactLynx } from '@lynx-js/react-rsbuild-plugin';
+import { defineConfig } from '@lynx-js/rspeedy';
+
+const enableBundleAnalysis = !!process.env['RSPEEDY_BUNDLE_ANALYSIS'];
+
+export default defineConfig({
+  plugins: [
+    pluginReactLynx({
+      experimental_useElementTemplate: true,
```
upupming, line 10: Shall we just use `examples/react` and set a env for this like `RSPEEDY_BUNDLE_ANALYSIS` in the above. Since the example source code is just the same.

packages/react/runtime/src/element-template/client/root.ts
```diff
@@ -0,0 +1,74 @@
+// Copyright 2024 The Lynx Authors. All rights reserved.
+// Licensed under the Apache License Version 2.0 that can be found in the
+// LICENSE file in the root directory of this source tree.
+// import { createContext, createElement } from 'preact/compat';
+// import { useState } from 'preact/hooks';
+// import type { Consumer, FC, ReactNode } from 'react';
```
upupming, line 6: Shall we remove this?

packages/rspeedy/plugin-react-alias/src/index.ts
```diff
@@ -87,15 +98,21 @@ export function pluginReactAlias(options: Options): RsbuildPlugin {
           gte(version, '0.111.9999')
             ? resolve('@lynx-js/react/compat')
             : Promise.resolve(null),
+          elementTemplate
+            ? resolve('@lynx-js/react/element-template')
+            : Promise.resolve(null),
+          elementTemplate
+            ? resolve('@lynx-js/react/element-template/internal')
+            : Promise.resolve(null),
```
upupming, line 106: Do we have any test to ensure `@lynx-js/react/element-template/internal` and `@lynx-js/react/internal` have same exports.

upupming: For example, user cannot use `Suspense` now but ts check is okay.

```tsx
File: ./src/App.tsx:51:47-54
  × ESModulesLinkingError: export 'Suspense' (imported as 'Suspense') was not found in '@lynx-js/react' (possible exports: Children, Component, Fragment, PureComponent, __etSlot, cloneElement, createContext, createElement, createRef, default, forwardRef, isValidElement, lazy, memo, root, useCallback, useContext, useDebugValue, useEffect, useErrorBoundary, useId, useImperativeHandle, useLayoutEffect, useMemo, useReducer, useRef, useState, useSyncExternalStore)
    ╭─[51:46]
 49 │                 columnNumber: 17
 50 │             }, this)),
 51 │             __etSlot(1, /*#__PURE__*/ _jsxDEV(Suspense, {
    ·                                               ────────
 52 │                 fallback: /*#__PURE__*/ _jsxDEV(_et_835da_f87c4_4, {}, void 0, false, {
 53 │                     fileName: "/Users/bytedance/projects/lynx-stack-2/examples/react-element-template/src/App.tsx",
    ╰────
Import traces (entry → error):
  ./src/index.tsx
  ./src/App.tsx ×

```

upupming (approved):

PR #2565, opened 2026-05-07
fix(web-core): forward keyboard event properties across threads

KealanAU:
keydown/keyup/keypress were falling through createCrossThreadEvent with no key/code/keyCode/shiftKey etc — stripped by omission. Adds keyboard branch to MinimalRawEventObject and createCrossThreadEvent, with tests for keydown and keyup.

@coderabbitai summary

## Checklist


KealanAU:
Deleting branch as actioning in https://github.com/lynx-family/lynx-stack/pull/2594

PR #2571, opened 2026-05-07
refactor(a2ui): React renderer + headless <A2UI> component

PupilTong:
## Summary

The React surface that consumes the message buffer + surface processor.

- `<A2UI>` — all-in-one component owning a `MessageProcessor` per mount, subscribing to a developer-supplied `MessageStore`, processing tail messages each render, rendering the most recent surface.
- `<A2UIRenderer>` / `NodeRenderer` — lower-level building blocks for consumers that want manual control over surface lifecycle.
- `<A2UIProvider>` + `useAction` / `useDataBinding` / `useCatalog` — internal context + hooks the catalog components reach for.

Catalog migration:
- Built-ins (`Text`, `Button`, `Card`, `Column`, `List`, `Row`, `CheckBox`, `RadioGroup`, `Image`, `Divider`) move from `core/A2UIRender` + global `componentRegistry` to the new `react/A2UIRenderer`. Side-effect re-exports (`<Name>.ts`, `all.ts`) and the global registry are dropped — every consumer composes via `defineCatalog([...])`.
- `src/index.ts` is the new public surface; `package.json` exports trimmed (`./core`, `./chat`, `./catalog/all` removed).

Cleanup: delete `src/core/`, `src/chat/`, `src/utils/` (replaced by the new layered design).

Playground:
- `lynx-src/App.tsx` switches from `BaseClient` + `A2UIRender` to `<A2UI messageStore={...} catalogs={...}>` with the per-component manifest tuple form.
- New `examples/io-mock/` and `examples/io-sse/` show how to push raw protocol messages into a `MessageStore` from a developer-owned IO module.
- `lynx-src/tsconfig.json` enables `resolveJsonModule` for the catalog manifest imports.

## Stacked on

PR-α: [#2572 — feat(a2ui): add headless message buffer and surface processor](https://github.com/lynx-family/lynx-stack/pull/2572).

The first commit on this branch is PR-α's content; once that lands and main is rebased into this branch, only the React-layer commit remains. Reviewers should focus on commit 2 (\`refactor(a2ui): React renderer + headless <A2UI> component\`) — that's the PR-β scope.

## Test plan

- [ ] \`pnpm -F @lynx-js/a2ui-reactlynx test\` — 46/46 pass
- [ ] \`pnpm -F a2ui-playground build:lynx\` — clean build
- [ ] Open the playground demos page; click "Render" on each scenario; the lynx surface renders as expected.

## Summary by CodeRabbit

* **New Features**
  * Introduced `A2UI` React component for rendering agent messages with flexible catalog composition
  * Added catalog builder API supporting dynamic component composition and tree-shaking
  * Added playground examples demonstrating IO integration patterns for agent communication

* **Documentation**
  * Updated catalog README with component integration examples
  * Added playground examples documentation covering integration patterns


packages/genui/a2ui-playground/examples/io-sse/sseAgent.ts
```diff
@@ -0,0 +1,231 @@
...
+          const batch = queue.shift();
+          if (batch && batch.length > 0) {
+            for (const msg of batch) {
+              msg.messageId ??= messageId;
+              store.push(msg);
+            }
+          }
+          await new Promise((r) => setTimeout(r, MESSAGE_PROCESS_DELAY));
+        }
+        isProcessing = false;
```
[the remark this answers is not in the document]
PupilTong: Fixed in 09402788 — split out a non-paced `drainQueue()` helper that `succeed()` awaits before `cleanup()`. So if `complete` fires while `flush` is sleeping on `MESSAGE_PROCESS_DELAY`, every queued batch is still pushed into the store before the SSE connection closes. `fail` and `abort` keep the original drop-the-tail behaviour because the consumer asked to bail.

packages/genui/a2ui-playground/package.json
```diff
@@ -6,7 +6,7 @@
   "scripts": {
     "build": "rsbuild build",
     "build:lynx": "rspeedy build",
-    "dev": "pnpm run build:lynx && rsbuild dev",
+    "dev": "rsbuild dev",
```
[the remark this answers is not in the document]
PupilTong: Fixed in 09402788 — restored `"dev": "pnpm run build:lynx && rsbuild dev"` to match `origin/main`'s post-#2575 state. The rebase had reverted to the older `rsbuild dev` only.

packages/genui/a2ui/src/store/MessageProcessor.ts
```diff
@@ -22,33 +22,39 @@ function isObject(value: unknown): value is Record<string, unknown> {
...
-      if (this.listener) {
-        this.listener({ message, resolve });
-      } else {
-        console.warn('No host listener attached!');
+      if (this.eventListeners.size === 0) {
         resolve([]);
+        return;
       }
+      for (const cb of this.eventListeners) cb({ message, resolve });
     });
```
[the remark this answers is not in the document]
PupilTong: Fixed in 09402788 — `dispatch()` now hands each listener its own one-shot `resolve` wrapper. The outer Promise resolves once every listener settles, with the first non-empty response wins (falling back to `[]` if every listener yielded nothing). No more racing on the shared resolver.

packages/genui/a2ui/src/store/MessageProcessor.ts
```diff
@@ -437,20 +438,15 @@ export class MessageProcessor {
...
-            targetId: surface?.rootComponentId ?? surfaceId,
-            messageId: (message as { messageId?: string }).messageId,
-          });
-        }
+        this.emitUpdate({
+          type: 'deleteSurface',
+          surfaceId,
+          targetId: surface?.rootComponentId ?? surfaceId,
+          messageId: (message as { messageId?: string }).messageId,
+        });
```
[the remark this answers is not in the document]
PupilTong: Fixed in 09402788 — `deleteSurface` now derives `messageId` the same way `beginRendering` does (`?? \`surface:${surfaceId}\``), so consumers that key lifecycle state by message id always see a non-empty key.

packages/genui/a2ui-playground/examples/README.md
```diff
@@ -0,0 +1,70 @@
+# A2UI playground examples
```
Sherry-hue, line 1: Can we merge `createSseAgent` and `createMockAgent` into a single interface `createAgent`?

This should include `initialMessages`, `actions`, and `sse`.

packages/genui/a2ui-playground/examples/README.md
```diff
@@ -0,0 +1,70 @@
...
+`createMockAgent(store, opts)` returns a driver that pushes a fixed
+initial stream into the store and serves canned responses to user
+actions. Used by the playground's `lynx-src/App.tsx` to exercise demos
+without a real agent.
+
+```ts
+const store = createMessageStore();
+const agent = createMockAgent(store, { initialMessages, actionMocks });
+agent.start(); // streams initial messages into the buffer
+agent.onAction(action); // pushes the canned response to a user action
```
Sherry-hue, line 25: Can onAction be directly used as a parameter of createMockAgent?

packages/genui/a2ui/package.json
```diff
@@ -3,91 +3,76 @@
...
+      "types": "./dist/catalog/index.d.ts",
+      "default": "./dist/catalog/index.js"
     },
     "./catalog/Text": {
-      "types": "./dist/catalog/Text.d.ts",
-      "default": "./dist/catalog/Text.js"
+      "types": "./dist/catalog/Text/index.d.ts",
+      "default": "./dist/catalog/Text/index.js"
     },
+    "./catalog/Text/catalog.json": "./dist/catalog/Text/catalog.json",
```
Sherry-hue, line 27: why export it

packages/genui/a2ui/package.json
```diff
@@ -3,91 +3,76 @@
   "version": "0.0.0",
```
Sherry-hue, line 3: README and AGENTS should update

packages/genui/a2ui/src/catalog/README.md
```diff
@@ -1,17 +1,38 @@
...
+constant. A top-level array referencing every built-in defeats
+tree-shaking — every consumer of such an aggregate would bundle every
+component, even the nine you don't use. Composition is per-component, and
+the cost is visible at the import site.
 
 ## The minimum a renderer needs
 
-If you only need to render, names alone are enough. Pass bare components —
+If your app only renders, names alone are enough. Pass bare components —
 the protocol name comes from `displayName ?? component.name`:
```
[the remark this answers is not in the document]
PupilTong: Fixed in 5535e772 — added a callout under "The minimum a renderer needs" warning that production minifiers rewrite `function` names and recommending either explicit `displayName` or the manifest tuple form. Also added `MyChart.displayName = 'MyChart'` to the custom-component example with a comment explaining why.

packages/genui/a2ui/src/react/A2UIRenderer.tsx
```diff
@@ -1,12 +1,215 @@
...
+  const data = snapshot.value;
+  const status = snapshot.status;
+  const error = snapshot.error;
+
+  if (status === 'pending' && data === undefined) {
+    return renderFallback?.() ?? <DefaultLoading id={resource.id} />;
+  }
+
+  if (status === 'error') {
+    return renderError?.(error) ?? <text>Error: {String(error)}</text>;
```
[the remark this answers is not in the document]
PupilTong: Fixed in 5535e772 — switched from `??` to a ternary so an explicit `null` from `renderFallback`/`renderError` correctly suppresses the built-in placeholder/error UI.

packages/genui/a2ui/src/react/useDataBinding.ts
```diff
@@ -1,5 +1,217 @@
...
+
+  const signalValue = useSyncExternalStore(
+    subscribe,
+    getSnapshot,
+    getSnapshot,
+  );
+
+  const currentValue = path
+    ? (signalValue ?? initialValue)
+    : (initialValue ?? fallbackValue);
```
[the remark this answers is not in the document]
PupilTong: Fixed in 5535e772 — the `path` branch now also chains through `fallbackValue`: `signalValue ?? initialValue ?? fallbackValue`. Controlled props with a stale/uninitialized binding now respect the caller-provided default. (CodeRabbit's earlier auto-confirmation was premature — the fix wasn't actually applied yet.)

packages/genui/a2ui/src/store/MessageProcessor.ts
```diff
@@ -22,33 +22,69 @@ function isObject(value: unknown): value is Record<string, unknown> {
...
+          resolve: (value) => {
+            if (called) return;
+            called = true;
+            if (!hasResponse) {
+              hasResponse = true;
+              firstResponse = value;
+            }
+            tryResolve();
+          },
+        });
```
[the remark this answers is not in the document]
PupilTong: Acknowledged — a never-resolving listener can hold the promise indefinitely. The current contract is callback-style: `onEvent` listeners MUST call `resolve` exactly once. `<A2UI>`'s internal listener calls `resolve([])` at the top of its handler so it can never deadlock; user-defined listeners that don't follow the contract are out of scope for this PR. Adding a per-listener timeout would change the dispatch latency contract and force consumers to handle a separate "listener timed out" failure mode. Filing a follow-up issue makes more sense than expanding scope here.

packages/genui/a2ui/src/store/MessageProcessor.ts
```diff
@@ -22,33 +22,69 @@ function isObject(value: unknown): value is Record<string, unknown> {
...
+        let called = false;
+        cb({
+          message,
+          resolve: (value) => {
+            if (called) return;
+            called = true;
+            if (!hasResponse) {
+              hasResponse = true;
+              firstResponse = value;
+            }
```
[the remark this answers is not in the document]
PupilTong: Fixed in 5535e772 — added an `isMeaningfulResponse` predicate that filters out `null`, `undefined`, empty arrays, and empty objects. The first **meaningful** response wins now, so `<A2UI>`'s no-op `resolve([])` listener can't shadow a real response from another subscriber.

packages/genui/a2ui-playground/examples/README.md
PupilTong: The split is intentional — these are **examples**, not shipped APIs from `@lynx-js/a2ui-reactlynx`. Each demonstrates a different IO transport: `createMockAgent` for in-process canned responses (used by the demos page), `createSseAgent` for streaming over Server-Sent Events. A merged `createAgent({ initialMessages, actions, sse })` would conflate distinct lifecycles (timed playback vs network connection) and force every consumer to bundle SSE plumbing they may not use. Happy to add a one-liner in the README clarifying that each is a starting point developers copy-and-adapt.

packages/genui/a2ui-playground/examples/README.md
PupilTong: It can — the current shape is `createMockAgent(store, { initialMessages, actionMocks, delayMs })`, where `actionMocks` is the lookup table for canned responses. The driver returned by `createMockAgent` exposes an `onAction(action)` method that resolves the mock for that action's name and pushes the response messages into the store. So consumers wire the playground's `<A2UI onAction={...}>` to `agent.onAction`, like `onAction={(a) => void agent.onAction(a)}`. If you want `onAction` as a ctor-time callback instead, that's an easy follow-up — happy to do it in a separate PR if useful.

packages/genui/a2ui/package.json
PupilTong: The `./catalog/<Name>/catalog.json` subpath exports let consumers import the per-component manifest with `import textManifest from '@lynx-js/a2ui-reactlynx/catalog/Text/catalog.json'` and pair it with the component via the tuple form: `defineCatalog([[Text, textManifest], ...])`. The manifest carries the JSON-Schema fragment the extractor produced so `serializeCatalog` can announce it to the agent during handshake. Without these subpath exports, the manifests would be unreachable from outside the package. Documented in [`src/catalog/README.md`](packages/genui/a2ui/src/catalog/README.md).

packages/genui/a2ui/package.json
PupilTong: Will do — both `packages/genui/a2ui/README.md` and `AGENTS.md` still reflect the pre-refactor `BaseClient` + `<A2UIRender>` API. I'll update them to the new `<A2UI messageStore={...} catalogs={...}>` flow in a follow-up commit on this PR.

PupilTong: Now done in 453b10a1:

- [`README.md`](packages/genui/a2ui/README.md) rewritten around the new `<A2UI>` + `MessageStore` + `defineCatalog` flow. Documents the 3-layer split (`./store`, `./react`, `./catalog`), the bare-vs-tuple catalog forms with the production-minification warning, custom components with `displayName`, and the new `onAction` / `wrapSurface` props. Removed all `BaseClient` / `A2UIRender` / `Conversation` / global-registry references.
- [`AGENTS.md`](packages/genui/a2ui/AGENTS.md) rewritten around the same 3-layer architecture: store layer (buffer + processor + resource + signals), React layer (`<A2UI>` + hooks + renderer), catalog layer. Updated the build instructions, schema generation, and "adding a new catalog component" checklist to match the post-refactor file layout (`exports` map entries for both the component and its `catalog.json` manifest).

packages/genui/a2ui-playground/examples/README.md
Sherry-hue: Since you said "these are just examples," then what is the purpose of adding these two APIs? The playground doesn't use them.

PupilTong: Half-right — the playground **does** use `createMockAgent` ([lynx-src/a2ui/App.tsx:240,264](packages/genui/a2ui-playground/lynx-src/a2ui/App.tsx)) to drive the demos. It's `createSseAgent` that the playground doesn't consume.

The intent for `createSseAgent` was a **reference template** developers can copy when wiring up a real SSE-backed agent — same role as `createMockAgent` for non-streaming/canned cases. If you'd rather drop reference-only templates from the playground's `examples/` folder, two options:

1. Delete `examples/io-sse/` entirely from this PR and let consumers fish it out of the docs / a future package.
2. Keep it but add a one-liner in `examples/README.md` clarifying that `createMockAgent` is what the playground actually runs and `createSseAgent` is a template-only reference.

Which do you prefer? I'll apply your call in the next commit.

PupilTong: Done in c204e523 — deleted `examples/io-sse/sseAgent.ts` and dropped its section from `examples/README.md`. The playground now ships only the example it actually consumes (`createMockAgent`). The SSE flow can land in a separate PR alongside a real SSE-backed demo when there's one to validate it against.

Sherry-hue (approved):

PR #2583, opened 2026-05-08
feat(web-core): make event coords relative to lynx-view top-left

PupilTong:
## Summary

- Web Platform's `layoutchange` and pointer events (`touch*` / `mouse*` / `click`) currently emit **viewport-relative** coordinates, which only matches native Lynx semantics when the host `<lynx-view>` happens to sit at viewport origin. This PR makes those payloads **lynx-view-relative** so handlers see the same values they'd see in native Lynx regardless of where the lynx-view is embedded.
- A single `ResizeObserver` per `LynxViewInstance` observes the host `<lynx-view>` and caches its `getBoundingClientRect().left/top`. The cache is the **only** measurement point — no synchronous pre-warm, no re-measurement at event-handler time. The observer is disconnected on dispose.
- `createCrossThreadEvent` takes the cached offset and subtracts it from every viewport-relative coordinate it emits — `layoutchange.detail.{top,left,right,bottom}`, `touch.detail.{x,y}` and the per-touch `clientX/Y`/`pageX/Y` inside `touches`/`targetTouches`/`changedTouches`, `mouse.{x,y,clientX,clientY,pageX,pageY}`, `click.detail.{x,y}`. `screenX/screenY` are screen-relative (different reference frame) and untouched.
- AGENTS.md gets a new "Coordinate normalization" sub-bullet under Event Handling.

## Test plan

Three new e2e cases were added under `web-core-e2e` (existing `api-bindlauoutchange` left untouched as a regression baseline):

- [ ] `api-bindlayoutchange-lynx-view-relative` — shifts the lynx-view via `page.evaluate` to (200, 200) and asserts `bindLayoutChange` handler sees `top===50, left===50, bottom===150, right===150` (would be 250/250/350/350 viewport-relative without the fix).
- [ ] `api-bindtap-lynx-view-relative` — same offset; clicks at viewport (300, 300) → expects `bindtap` handler to see `e.detail.{x,y}` in `(0, 200)` (lynx-view-relative ~100; would be ~300 viewport-relative).
- [ ] `api-bindtouchstart-lynx-view-relative` — same shape with `bindtouchstart`, also asserts `e.touches[0].clientX/Y` are shifted; skipped on Firefox project (no touch emulation).
- [ ] `pnpm build` at repo root (`tsc --build`) passes — composite TS project type-checks all consumers of the extended `WASMJSBindingInjectedHandler` interface and the new `createCrossThreadEvent` signature.
- [ ] No new screenshot snapshots; all assertions are `toHaveCSS`.
- [ ] Did not run the full Playwright suite locally (env is macOS, runners are Ubuntu) — relying on CI for cross-browser coverage.

## Summary by CodeRabbit

* **Bug Fixes**
  * Positional event coordinates (layout-change, tap/click, mouse, touch) now report positions relative to the host view, fixing incorrect targeting when a view is offset or transformed.

* **New Features**
  * boundingClientRect queries now return host-view–relative coordinates and use a per-view cached rect with rate-limited invalidation to handle transforms efficiently.

* **Tests**
  * E2E tests added for layout-change, tap, touchstart, transformed view, and boundingClientRect with an offset host view.

* **Documentation**
  * Docs updated with coordinate-normalization details and new commit guidance.

### DIFF
--- a/.changeset/web-core-layoutchange-relative-coords.md
+++ b/.changeset/web-core-layoutchange-relative-coords.md
@@ -1,0 +1,14 @@
+---
+"@lynx-js/web-core": patch
+---
+
+Add lynx-view-relative coordinates to positional event payloads (matching native Lynx semantics) while preserving viewport/document coordinates for Web interop, and switch the `boundingClientRect` UI method to lynx-view-relative.
+
+The lynx-view's rect is cached by a new `BoundingClientRectService` (one per `LynxViewInstance`). The cache is invalidated by `transitionend`/`animationend` on the lynx-view itself (filtered to ignore descendants bubbling through) and by an idle-callback path throttled to at most one invalidation per 240 ms. This picks up CSS `transform`s and similar drifts that `ResizeObserver` would not catch, while bounding the cost when many events read the rect in a tight loop and avoiding event-modification feedback loops.
+
+Coordinate model:
+
+- `x`/`y` (top-level on `mouse*`; `detail.x`/`detail.y` on `click`/`touch*`; per-touch on `touches`/`targetTouches`/`changedTouches`) — lynx-view-relative (Lynx parity).
+- `clientX`/`clientY`, `pageX`/`pageY` (top-level on `mouse*`/`click`; per-touch alongside the added `x`/`y`) — viewport- and document-relative, unchanged from the underlying DOM event (Web interop).
+- `layoutchange.detail.{top,left,right,bottom}` is lynx-view-relative.
+- The `boundingClientRect` UI method (`SelectorQuery#fields({rect: true})`, `NodesRef.invoke('boundingClientRect')`) returns lynx-view-relative coordinates.
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -248,6 +248,10 @@ These instructions were generated through comprehensive analysis and testing of 
 
 Remember: This is a complex, multi-language monorepo. Always allow extra time for builds and tests, and follow the exact command sequences provided.
 
+## Commit conventions
+
+- Do **not** add `Co-Authored-By:` (or other AI-attribution) footers to commit messages. Authorship is tracked via the git author field; trailers like these add noise to the history.
+
 ## Submit knowledge updates
 
 When you learn new patterns or best practices while working on the `lynx-stack` project, you should update or create one or more ".github/*.instructions.md" files, adding natural language instructions to the file(s).
--- a/packages/web-platform/web-core-e2e/tests/reactlynx.spec.ts
+++ b/packages/web-platform/web-core-e2e/tests/reactlynx.spec.ts
@@ -2126,6 +2126,94 @@ test.describe('reactlynx3 tests', () => {
       const target = page.locator('#target');
       await expect(target).toHaveCSS('background-color', 'rgb(0, 128, 0)'); // green
     });
+    // Inject CSS for the lynx-view BEFORE the bundle loads, so the offset
+    // (or transform) is in place by the time the first layoutchange fires.
+    // Applying the offset after `goto` would race the initial layoutchange,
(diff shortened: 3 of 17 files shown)


Dugyu (changes requested):
The current cached `getBoundingClientRect()` only updates on resize.
`<lynx-view>` can move via CSS transforms or layout changes, so cached offsets may become stale.
Consider tracking position changes (e.g., `MutationObserver` or `requestAnimationFrame`) to keep view-relative coordinates accurate.

Dugyu:
As written, the cached offset can become stale while the host view is moving, which would make the emitted “lynx-view-relative” coordinates drift from the actual visual position.

I agree with avoiding synchronous `getBoundingClientRect()` calls directly inside the event hot path, but I don’t think `ResizeObserver` alone is sufficient as the long-term invalidation mechanism here.

Could we consider a position-aware refresh strategy instead? For example:

* refreshing at most once per animation frame when events are emitted,
* or broadening invalidation sources beyond resize-only updates.

That would preserve the “avoid per-event measurement” goal while keeping coordinates correct when the host view moves without resizing.

Dugyu:
One higher-level concern: once we normalize all event coordinates into `<lynx-view>` local space, handlers lose direct access to viewport-space coordinates.

That matches native Lynx semantics, but it also removes information that is often useful on Web (global overlays, cross-view drag/drop, viewport alignment, etc.).

Would it make sense to preserve both coordinate spaces instead of fully replacing one with the other? For example:

- local/view-relative coordinates for Lynx parity,
- viewport/page coordinates alongside them for Web interoperability.

```ts
e.detail = {
  x, y,                 // local
  clientX, clientY,     // viewport
  pageX, pageY,         // page
}
```

That would also make the reference frame explicit instead of overloading existing fields with different semantics on different platforms.

Dugyu:
Did a quick validation with the `lynx-ui` Popover demos on top of this PR using the demo branch:

`lynx-family/lynx-stack#demo/pr-2583-web-coords`

Compared recordings:

0.20.3

[web-coords-popover-20-3.webm](https://github.com/user-attachments/assets/7a1852a3-6d47-4d59-981f-1221dc69a74c)

PR workspace version

[web-coords-popover-pr-2583.webm](https://github.com/user-attachments/assets/2f7feb6f-818c-4f9a-9e94-22a3e401d079)

Observations:

* Previous `0.20.3` behavior:
  * bundle loading flickers heavily
  * `CustomAnchor` and `OffsetAdjustment` examples never render correctly

* PR branch behavior:
  * noticeable improvement overall
  * after the first browser load, both examples usually render
  * however, after refresh they may disappear again
  * the first popover position still looks unstable / slightly misaligned

So this does seem to improve the coordinate issue in practice, but there may still be some remaining offset drift or stale-position cases when the host `<lynx-view>` moves.

Test environment:

* demo branch: `demo/pr-2583-web-coords`
* browser-based validation with `lynx-ui` Popover examples

Dugyu (dismissed):

Dugyu:
After the latest invalidation updates (`transitionend` / `animationend`) in commit `6e4cae0`, part of the scroll-related positioning issue seems improved.

From local validation:

* if the rendered page height does not exceed the viewport height, scrolling no longer appears to affect popover positioning
* the previous large drift during normal page scroll seems mostly resolved in that case

Previous

[web-coords-scroll-drift.webm](https://github.com/user-attachments/assets/c518f3ec-0f3d-4fdc-bb71-59facbae36bd)

Now(`6e4cae0`)

[web-coords-commit-6e4cae0.webm](https://github.com/user-attachments/assets/9b6e407d-6b94-49c3-ad53-a4d34e2713f5)

Sherry-hue (dismissed):

Sherry-hue (approved):

PR #2589, opened 2026-05-09
feat(react): support element template events

Yradex:
## Summary by CodeRabbit

* **New Features**
  * Experimental Element Template direct-event support: queued delivery, controlled flush on hydration, and reliable background↔main-thread event synchronization.
  * Improved attribute-slot preparation and normalization for more consistent hydration and native template creation.

* **Tests**
  * Broad test coverage added for event registration/dispatch, queued delivery/flush semantics, hydration edge cases, compiled templates, and teardown behavior.

* **Chores**
  * Marked as no package release required (internal/experimental).

## Overview

This PR adds the Element Template event implementation across transform output, runtime attribute-slot handling, background hydration/update behavior, and native bridge wiring. It keeps the work behind the existing experimental Element Template path while making direct event attributes flow through the ET runtime instead of being treated as ordinary static attributes.

## Key Points

- Adds an ET attr-plan side table so transformed templates can identify event attribute slots without changing the serialized template shape for non-event attrs.
- Prepares event attribute slots before main-thread create and background hydrate/update paths, including uid-based event values and handler registration.
- Wires ET native event dispatch through the runtime bridge and covers direct-event subtree migration behavior.
- Organizes event-focused runtime coverage and compiled fixtures, and includes an empty changeset because this remains unpublished experimental ET internals.

## Checklist

- [x] Changeset added (or not required).

## Verification

- `git diff --check github/main..stack/element-template`
- `pnpm --dir packages/react/runtime run test:et`
- `cargo test --test element_template --test element_template_contract`

### DIFF
--- a/packages/react/runtime/src/element-template/background/instance.ts
+++ b/packages/react/runtime/src/element-template/background/instance.ts
@@ -2,7 +2,7 @@
 // Licensed under the Apache License Version 2.0 that can be found in the
 // LICENSE file in the root directory of this source tree.
 
-import { clearAttributeSlotEventHandlers, prepareAttributeSlots as prepareRawAttributeSlots } from './attr-slots.js';
+import { prepareAttributeSlots as prepareRawAttributeSlots } from './attr-slots.js';
 import { globalCommitContext, markRemovedSubtreeForCurrentCommit } from './commit-context.js';
 import { isElementTemplateHydrated } from './commit-hook.js';
 import { backgroundElementTemplateInstanceManager } from './manager.js';
@@ -81,12 +81,7 @@ export class BackgroundElementTemplateInstance {
     this.nodeType = isBuiltinRawTextTemplateKey(type) ? 3 : 1;
     backgroundElementTemplateInstanceManager.register(this);
     if (initialAttributeSlots) {
-      const preparedSlots = prepareRawAttributeSlots(
-        this.type,
-        this.instanceId,
-        initialAttributeSlots,
-        isElementTemplateHydrated(),
-      );
+      const preparedSlots = prepareRawAttributeSlots(this.type, this.instanceId, initialAttributeSlots);
       this.attributeSlots = preparedSlots;
       if (preparedSlots !== initialAttributeSlots) {
         this.rawAttributeSlots = initialAttributeSlots;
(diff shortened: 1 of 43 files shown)


packages/react/runtime/src/element-template/prop-adapters/event.ts
```diff
@@ -0,0 +1,78 @@
...
+  return eventHandlerByEventValue.get(eventValue);
+}
+
+export function publishEvent(eventValue: string, data: unknown): void {
+  if (dispatchEvent(eventValue, data)) {
+    return;
+  }
+  if (queuePendingEvents) {
+    pendingEvents.push([eventValue, data]);
+  }
```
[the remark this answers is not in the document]
Yradex: Not for now.

packages/react/runtime/src/element-template/runtime/render/render-opcodes.ts
```diff
@@ -89,12 +95,35 @@ export function renderOpcodesIntoElementTemplate(
...
 
-        const elementRef = createElementTemplateWithHandle(
-          concreteTemplateKey,
-          null,
-          attributeSlots ?? null,
-          elementSlots ?? null,
-        );
+        const attrPlan = __etAttrPlanMap[concreteTemplateKey];
+        let elementRef: ElementRef;
+        if (attrPlan === undefined) {
```
HuJean, line 97: Maybe createElementTemplateWithHandle and createElementTemplateWithReservedHandle can be a function and check attrPlan inner?

packages/react/runtime/src/element-template/background/instance.ts
```diff
@@ -283,18 +288,40 @@ export class BackgroundElementTemplateInstance {
...
     // Hydration binds this object to a template that already exists on the main
     // thread; future updates must treat it as created without emitting create.
     this.hasEmittedCreate = true;
   }
 
+  prepareAttributeSlotsForNative(): void {
+    if (!this.rawAttributeSlots) {
+      return;
+    }
+    this.attributeSlots = prepareRawAttributeSlots(
```
HuJean, line 305: Do attributeSlots and rawAttributeSlots has same content ?

Yradex: No. `rawAttributeSlots` saves the original attributes(event functions), and `attributeSlots` saves the prepared attributes(eventValue like `1:0`).

packages/react/runtime/src/element-template/runtime/render/render-opcodes.ts
Yradex: Good point about the duplication. For now I’d prefer not to move attrPlan lookup into the create helper though, because `handle.ts` should stay as the low-level handle/native-ref primitive and not know about prop adapter state.

What I can do instead is remove `createElementTemplateWithHandle` and make all callers explicitly reserve a handle, then call `createElementTemplateWithReservedHandle`. That keeps the attrPlan preparation in the render/materialization layer, while avoiding the duplicated create helper.

When we support MTS later, I'll extract `attrPlan` check and `createElementTemplate` into another function because that's where a real duplication would appear.

HuJean (approved):

gaoachao (approved):

upupming (approved):

PR #2590, opened 2026-05-09
fix(react): clear removed snapshot prop refs

Yradex:
## Summary by CodeRabbit

* **Bug Fixes**
  * Cleared transient snapshot child property references when snapshot subtrees are removed, preventing deleted items from being retained.

* **Tests**
  * Added tests covering cleanup behavior for removed children, including nested arrays and cyclic reference scenarios.

[Review Change Stack](https://app.coderabbit.ai/change-stack/lynx-family/lynx-stack/pull/2590)

## Summary

This fixes a main-thread snapshot retention path where removed child subtrees could still be strongly referenced by compiler-generated transient child props. Snapshot JSX uses `$*` props as staging references for named children; when `SnapshotInstance.removeChild()` detached a child, the structural tree and native elements were cleaned up, but those `$*` props could still point at the removed snapshot subtree.

The practical failure mode is that a deleted list holder or list item subtree can remain reachable from the root props chain after it is removed. That keeps old snapshot instances alive even though they are no longer part of the rendered tree.

## Key points

- Clear transient `$*` owner props during removal. Direct refs such as `{ $0: removedChild }` are deleted from `props`, so the owner no longer keeps a hard reference to the removed subtree after `removeChild()`.
- Handle compiled array child shapes recursively. When a transient prop stores children as arrays, only removed snapshot entries are cleared, including nested arrays:

  ```js
  // before removal
  owner.props.$0 = [keptChild, [removedChild]];

  // after removeChild(removedChild)
  owner.props.$0 = [keptChild, [undefined]];
  ```

- Clean transient props inside the removed subtree as it is torn down. This breaks references held by the deleted subtree itself, including component/list structures that had their own compiled child staging props.

## Runtime Contract

This only affects compiler-owned transient child props whose keys start with `$`. Normal user props, public runtime APIs, serialized snapshot shape, native patch operations, and list update protocols are unchanged.

The cleanup is tied to `SnapshotInstance.removeChild()`: removed snapshot instances and descendants are collected into a removal set, then `$*` prop values are cleared if they directly reference any removed snapshot or contain one inside an array. Existing array identity is preserved, and non-removed entries keep their current identity and order.

## Checklist

### DIFF
--- a/packages/react/runtime/src/snapshot/snapshot/snapshot.ts
+++ b/packages/react/runtime/src/snapshot/snapshot/snapshot.ts
@@ -41,6 +41,62 @@ export const snapshotInstanceManager: {
   },
 };
 
+function isRemovedSnapshot(
+  value: unknown,
+  removedSnapshots: WeakSet<object>,
+): boolean {
+  return (typeof value === 'object' || typeof value === 'function') && value !== null
+    && removedSnapshots.has(value);
+}
+
+function clearRemovedSnapshotsFromArray(
+  values: unknown[],
+  removedSnapshots: WeakSet<object>,
+  seen = new WeakSet<object>(),
+): boolean {
+  if (seen.has(values)) {
+    return false;
+  }
+  seen.add(values);
+
+  let changed = false;
+  values.forEach((value, index) => {
+    if (isRemovedSnapshot(value, removedSnapshots)) {
+      values[index] = undefined;
+      changed = true;
+    } else if (Array.isArray(value) && clearRemovedSnapshotsFromArray(value, removedSnapshots, seen)) {
+      changed = true;
+    }
+  });
+  return changed;
+}
+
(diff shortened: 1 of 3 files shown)


packages/react/runtime/src/snapshot/snapshot/snapshot.ts
```diff
@@ -41,6 +41,62 @@ export const snapshotInstanceManager: {
...
+
+  const removedSnapshots = new WeakSet<object>();
+  traverseSnapshotInstance(removedChild, snapshot => {
+    removedSnapshots.add(snapshot);
+  });
+
+  for (const key of Object.keys(props)) {
+    // The compiler stores named children in transient `$N` props. Once a child
+    // subtree is removed, those staging refs must not keep the old snapshots alive.
+    if (!/^\$\d+$/.test(key)) {
```
upupming:
```suggestion
    if (key[0] === '$') {
```

upupming (approved):

HuJean (approved):

PR #2591, opened 2026-05-09
fix(react): avoid retaining transformed worklet ctx

Yradex:
## Summary by CodeRabbit

* **Bug Fixes**
  * Improved memory management for nested worklets: transformed nested worklets now use weak references so original list-item contexts are not retained, reducing memory footprint.

* **Tests**
  * Added tests confirming nested worklet contexts use weak references and that hydration correctly converts first-screen entries into hydrated function metadata.

[Review Change Stack](https://app.coderabbit.ai/change-stack/lynx-family/lynx-stack/pull/2591)

## Summary

- Transforming a nested worklet already binds the callable to a shallow copy of the nested worklet context, but the transformed function also kept the original context through strong internal metadata.
- That strong back-reference can make a cached transformed worklet function retain the original list-item worklet context after the item is clicked, because the `workletCache` value points back to the context it was meant to avoid retaining.
- This PR switches that internal metadata to a `WeakRef`, so hydration can recover a still-live first-screen context without extending the lifetime of list-item worklet contexts.

## Key Points

- The root cause is the extra strong context edge added after nested worklet transformation:

  ```ts
  obj[key] = registeredWorklet.bind({ ...subObj });
  obj[key].ctx = subObj;
  ```

  The bound function no longer needs the original object for execution, but that metadata kept the original object alive through any cache that retained the transformed function.

- The transformed function now stores the original nested worklet context through weak metadata:

  ```ts
  obj[key].ctxRef = new WeakRef(subObj);
  ```

  This preserves first-screen context recovery while avoiding a cache value that strongly retains the original context.

- Hydration consumes the weak metadata from transformed first-screen functions:

  ```ts
  const firstScreenValue = fn.ctxRef?.deref();
  ```

  This does not change public ReactLynx APIs, serialized native payloads, or the worklet registration contract.

- The new runtime test covers the retain-cycle-sensitive shape directly: a transformed nested worklet function should expose weak context metadata, should not expose the old strong context metadata, and the weak reference should point to the original nested worklet context while it is still reachable.

## Checklist

### DIFF
--- a/.changeset/weak-worklet-context.md
+++ b/.changeset/weak-worklet-context.md
@@ -1,0 +1,7 @@
+---
+"@lynx-js/react": patch
+---
+
+Avoid retaining transformed nested worklet contexts through a strong `.ctx` reference.
+
+Nested worklets transformed by the worklet runtime now store the original worklet context through `ctxRef` instead of a strong `ctx` property, preventing cached transformed worklet functions from keeping list-item worklet contexts alive. Hydration still falls back to the legacy `.ctx` shape for compatibility.
--- a/.changeset/weak-worklet-context.md
+++ b/.changeset/weak-worklet-context.md
@@ -2,6 +2,6 @@
 "@lynx-js/react": patch
 ---
 
-Avoid retaining transformed nested worklet contexts through a strong `.ctx` reference.
+Avoid retaining transformed nested worklet contexts after worklet transformation.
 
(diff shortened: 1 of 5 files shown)


Yradex:
@coderabbitai review

The PR description and changeset have been updated to align the stated scope with the implementation. Please re-review the latest head.

Yradex:
@coderabbitai review

The previous compatibility requirement came from stale PR wording, not from the intended supported contract of this change. The latest PR description and changeset now describe the actual scope: weak internal metadata for context recovery without adding an extra fallback path. Please re-review with that scope.

HuJean (approved):

PR #2592, opened 2026-05-09
fix(react): retain offscreen worklet ctx refs

Yradex:
## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed worklet callback retention to keep event, ref, gesture, and spread handlers alive on the main thread before snapshot elements are materialized, ensuring proper attachment during DOM updates.

[Review Change Stack](https://app.coderabbit.ai/change-stack/lynx-family/lynx-stack/pull/2592)

## Overview

Main-thread worklets used by events, refs, gestures, and spread props can be recorded on snapshot values before the snapshot has concrete `__elements`. This happens for offscreen list items and similar deferred materialization paths.

Previously, lifecycle retaining was coupled to `onWorkletCtxUpdate()`, which only runs when an element is available. That meant an offscreen main-thread worklet ctx could miss the lifecycle `addRef` step before the later DOM attach/update path, so the background worklet ctx was allowed to be released too early.

This PR separates lifecycle retaining from element update side effects. Snapshot update paths now retain main-thread worklet ctx values as soon as they observe them, while `onWorkletCtxUpdate()` remains responsible only for hydration and delayed-worklet behavior that actually needs an element.

## Key Points

- Root cause: lifecycle `addRef` was behind the `snapshot.__elements` guard through `onWorkletCtxUpdate()`. Offscreen snapshots skipped that path until materialization, which was too late for retaining the background ctx reference.
- Fix: introduce a minimal `retainWorkletCtx(worklet)` helper and call it from event, ref, gesture, and spread update paths before returning for missing `__elements`.
- Gesture handling retains callback worklets when the snapshot value updates, then suppresses a second retain when `processGesture()` later performs the real DOM update for the same value.
- `JsFunctionLifecycleManager` now deduplicates repeated `addRef()` calls for the same object identity, so repeated `componentAtIndex` access cannot inflate the ref count for the exact same retained object.

## Runtime Contract

The lifecycle retain step is now independent from element-specific update work:

```ts
// Retain can happen before native elements are materialized.
if (workletType === 'main-thread') {
  retainWorkletCtx(worklet);
}

// Element-side hydration, delayed execution, and DOM binding still require elements.
if (snapshot.__elements) {
  onWorkletCtxUpdate(worklet, oldWorklet, isMainThreadHydrating, element);
}
```

For gestures, the retained unit is each serialized base gesture callback. The DOM update path keeps using `processGesture()` for detector setup, but passes `retainCallbacks: false` when the callbacks have already been retained by the snapshot update path.

This does not change the serialized worklet shape, native gesture detector contract, transform output, or public component API. Missing `_execId` remains a no-op for lifecycle retaining, and legacy first-screen hydration / delayed-worklet behavior stays in `onWorkletCtxUpdate()`.

## Compatibility / Boundaries

- The dedupe is object-identity based: retaining the same object twice is ignored, while distinct objects that share an `execId` still contribute separate lifecycle references.
- The fix is limited to main-thread worklet lifecycle retention for event, ref, gesture, and spread paths. It does not introduce debug logging, dump fields, or example-page changes.
- Existing compatibility branches for old main-thread ref values and old gesture cleanup APIs are preserved.

## Checklist

### DIFF
--- a/packages/react/runtime/src/snapshot/snapshot/gesture.ts
+++ b/packages/react/runtime/src/snapshot/snapshot/gesture.ts
@@ -1,7 +1,7 @@
 // Copyright 2025 The Lynx Authors. All rights reserved.
 // Licensed under the Apache License Version 2.0 that can be found in the
 // LICENSE file in the root directory of this source tree.
-import { processGesture } from '../gesture/processGesture.js';
+import { processGesture, retainGestureWorkletCtx } from '../gesture/processGesture.js';
 import type { GestureKind } from '../gesture/types.js';
 import { isMainThreadHydrating } from '../lifecycle/patch/isMainThreadHydrating.js';
 import type { SnapshotInstance } from '../snapshot/snapshot.js';
@@ -13,12 +13,19 @@ export function updateGesture(
   elementIndex: number,
   workletType: string,
 ): void {
+  const value = snapshot.__values![expIndex] as GestureKind;
+  if (workletType === 'main-thread') {
+    retainGestureWorkletCtx(value);
+  }
+
   if (!snapshot.__elements) {
     return;
   }
-  const value = snapshot.__values![expIndex] as GestureKind;
 
   if (workletType === 'main-thread') {
-    processGesture(snapshot.__elements[elementIndex]!, value, oldValue as GestureKind, isMainThreadHydrating);
+    processGesture(snapshot.__elements[elementIndex]!, value, oldValue as GestureKind, isMainThreadHydrating, {
+      domSet: false,
+      retainCallbacks: false,
+    });
   }
 }
--- a/packages/react/runtime/src/worklet-runtime/jsFunctionLifecycle.ts
+++ b/packages/react/runtime/src/worklet-runtime/jsFunctionLifecycle.ts
@@ -14,6 +14,7 @@ import { isSdkVersionGt } from './utils/version.js';
 class JsFunctionLifecycleManager {
   private execIdRefCount = new Map<number, number>();
   private execIdSetToFire = new Set<number>();
+  private retainedObjects = new WeakSet<object>();
   private willFire = false;
   private registry?: FinalizationRegistry<number> = undefined;
 
@@ -22,6 +23,10 @@ class JsFunctionLifecycleManager {
   }
 
   addRef(execId: number, objToRef: object): void {
+    if (this.retainedObjects.has(objToRef)) {
+      return;
+    }
+    this.retainedObjects.add(objToRef);
     this.execIdRefCount.set(
       execId,
       (this.execIdRefCount.get(execId) ?? 0) + 1,
--- a/.changeset/retain-offscreen-worklet-ctx.md
+++ b/.changeset/retain-offscreen-worklet-ctx.md
@@ -1,0 +1,5 @@
+---
+"@lynx-js/react": patch
+---
+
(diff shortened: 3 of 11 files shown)


packages/react/runtime/src/snapshot/snapshot/gesture.ts
```diff
@@ -13,12 +13,19 @@ export function updateGesture(
   elementIndex: number,
   workletType: string,
 ): void {
+  const value = snapshot.__values![expIndex] as GestureKind;
+  if (workletType === 'main-thread') {
```
f0rdream, line 17: Are there extra concern on early retain?

What will happen if `!snapshot__elements` happens, and processGesture does not happen?

packages/react/runtime/src/worklet-runtime/jsFunctionLifecycle.ts
```diff
@@ -22,6 +23,10 @@ class JsFunctionLifecycleManager {
   }
 
   addRef(execId: number, objToRef: object): void {
+    if (this.retainedObjects.has(objToRef)) {
+      return;
```
f0rdream, line 27: Codex said this like:

```
const bg = runOnBackground(() => {
  'background only';
  console.log('background');
});

const onTapA = () => {
  'main thread';
  bg();
};

const onTapB = () => {
  'main thread';
  bg();
};

return (
  <>
    <view main-thread:bindtap={onTapA} />
    <view main-thread:bindtap={onTapB} />
  </>
);

```
Step by step:

bg is one JS function handle object, something like:
```
const bgHandle = { _jsFnId: 1 };
```

onTapA and onTapB are two different main-thread worklet ctx objects, but both capture the same nested bgHandle.

Conceptually:
```
ctxA = { _wkltId: 'onTapA', bg: bgHandle };
ctxB = { _wkltId: 'onTapB', bg: bgHandle };
```
When these worklet ctxs are committed to the main thread, background gives them different exec ids:

```
ctxA._execId = 100;
ctxB._execId = 101;
```
First time ctxA is transformed/run, transformWorkletInner() finds the nested JS function handle:
```
subObj._execId = ctxA._execId; // 100
manager.addRef(100, subObj);
```
With the PR’s WeakSet<object> dedupe, bgHandle is added to retainedObjects.

Later ctxB is transformed/run. It finds the same nested object identity:
```
subObj._execId = ctxB._execId; // 101
manager.addRef(101, subObj);
```
But now retainedObjects.has(bgHandle) is true, so addRef() returns early.

Result: execId=101 was never ref-counted or registered with FinalizationRegistry for that nested handle.

When ctxB/bgHandle is eventually no longer reachable, there is no finalizer registration for (bgHandle, 101), so removeRef(101) is not called for that retain. The background execIdMap entry for 101 can stay alive longer than intended, or indefinitely if no other registered object releases it.

The bug is not “dedupe exists”; dedupe is useful. The bug is the key. The same object can be reused under multiple _execIds, so the dedupe key needs to be effectively:
```
(object identity, execId)
```
not just:

```
object identity
```

packages/react/runtime/src/snapshot/snapshot/gesture.ts
f0rdream: I understand this now, this should be by desgin

f0rdream (approved):

HuJean (approved):

PR #2634, opened 2026-05-14
feat(template-webpack-plugin): parallelize TASM encode in a shared worker pool

upupming:
## Summary

- Move template encoding into a process-wide `tinypool` worker pool so multi-entry builds encode in parallel (`cpus().length - 1` threads) instead of serializing on the main thread.
- Watch-mode rebuilds reuse warm workers across compiles — the pool is a static field on `LynxEncodePlugin`, lazily spawned on first encode.
- `Buffer.from(uint8Array)` rewraps the postMessage-demoted buffer back to a real `Buffer` so downstream consumers (rspack's `WasmHash`) can still call `.copy()` on it.
- The worker file URL always resolves to the built `lib/worker/encode.js`, so the static pool initializer works both from `src/` (in-tree tests) and `lib/` (installed consumers).

## Test plan

- [x] `pnpm exec rstest run` in `packages/webpack/template-webpack-plugin` — 346 / 346 pass (including 2 new tests in `test/encode-worker-pool.test.ts`)
- [x] `pnpm build` in `examples/react` succeeds with the new pool path
- [x] New tests assert `pool.completed += N` for N-entry builds and that rebuild does not spawn new workers

## Summary by CodeRabbit

* **New Features**
  * Enabled parallel TASM template encoding for multi-entry builds.
  * Optimized watch-mode rebuilds with encoder worker reuse for faster iteration.

* **Chores**
  * Tightened Node.js version requirement to `^18.14 || >=19.4`.

[Review Change Stack](https://app.coderabbit.ai/change-stack/lynx-family/lynx-stack/pull/2634)


packages/webpack/template-webpack-plugin/src/LynxEncodePlugin.ts
```diff
@@ -49,6 +60,16 @@ export class LynxEncodePlugin {
...
   static BEFORE_EMIT_STAGE = 256;
+  /**
+   * Shared TASM encode worker pool: multiple entries (and multiple
+   * `LynxEncodePlugin` instances in the same process) share its worker
+   * slots for parallel encode; watch-mode rebuilds keep the same workers
+   * warm. One core is reserved for the main thread / OS.
+   */
+  static encodePool: Tinypool = new Tinypool({
+    filename: pathToFileURL(ENCODE_WORKER_PATH).href,
+    maxThreads: Math.max(1, cpus().length - 1),
```
colinaaa: Maybe we can use `os.availableParallelism()` instead.

Or we may want a more complex strategy like:

https://github.com/lynx-family/lynx-stack/blob/189be6c36e3d5fa0acf9077409c97515270ec606/packages/web-platform/playwright-fixtures/src/playwright.common.ts#L18

colinaaa (dismissed):

colinaaa (approved):

PR #2642, opened 2026-05-15
feat: lynx unified debug metadata standard

upupming:
Each Lynx entry now emits a single `debug-metadata.json` containing everything needed for reverse symbolication: JS / CSS source maps, lepusNG bytecode debug info, UI source map, and git + build context. The dev-server exposes sub-field queries via `?field=…&filename=…`.

Highlights:

- New `@lynx-js/debug-metadata` (zero-dep schema + resolver) and `@lynx-js/debug-metadata-rsbuild-plugin` (emitter + dev-server middleware).
- JS / CSS `sourceMappingURL` trailers rewritten to `debug-metadata.json?field=source-map`.
- `compilerOptions.templateDebugUrl` repointed at `?field=bytecode-debug-info` for the C++ runtime; `sourceContent.config.debugMetadataUrl` exposes the base endpoint URL.
- `debug-info.json` is no longer written to disk.
- Auto-registered by Rspeedy — zero user config.

## Summary by CodeRabbit

* **New Features**
  * Per-entry unified debug-metadata.json with dev-server ?field=... queries; template debug URLs now point to this endpoint and legacy debug-info.json is no longer emitted
  * New zero-dependency debug-metadata schema package and middleware; plugin auto-registered by Rspeedy with no user config
* **Refactor**
  * Debug-metadata emission moved to a dedicated plugin; sourceMappingURL trailers rewritten to the unified endpoint; template plugin wiring updated
* **Documentation**
  * README and schema docs for debug-metadata and query resolvers
* **Tests**
  * Extensive unit and integration tests added for collectors, resolvers, rewriter, middleware
* **Chores**
  * Package/version and peer-range updates plus snapshot refreshes

[Review Change Stack](https://app.coderabbit.ai/change-stack/lynx-family/lynx-stack/pull/2642?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


packages/rspeedy/plugin-debug-metadata/package.json
```diff
@@ -0,0 +1,62 @@
...
+  "dependencies": {
+    "@lynx-js/debug-metadata": "workspace:^"
+  },
+  "devDependencies": {
+    "@lynx-js/template-webpack-plugin": "workspace:*",
+    "@microsoft/api-extractor": "catalog:",
+    "@rollup/plugin-typescript": "^12.3.0",
+    "@rsbuild/core": "catalog:rsbuild",
+    "picocolors": "^1.1.1",
+    "terminal-link": "^4.0.0",
```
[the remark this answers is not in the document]
upupming: Not applicable to this package — `rslib build` inlines `picocolors` and `terminal-link` into the published `dist/index.js` (verified by grepping the bundle), and `publishConfig.exports` points consumers at `dist/`. So a downstream `npm install @lynx-js/debug-metadata-rsbuild-plugin` does not need `picocolors` / `terminal-link` as separate runtime deps — they ship inlined.

The in-workspace `lib/` build (which the top-level `exports` points at for monorepo dev) is non-bundled and does need them, but pnpm workspace install pulls them via `devDependencies` already.

Moving them to `dependencies` would only cause downstream consumers to install a redundant second copy on disk, on top of the one already inlined into `dist/`. Leaving as-is.

packages/rspeedy/plugin-debug-metadata/src/LynxDebugMetadataPlugin.ts
```diff
@@ -0,0 +1,253 @@
...
+      return entryPathMapCache
+    }
+
+    compiler.hooks.thisCompilation.tap(this.name, compilation => {
+      const templateHooks = this.options.LynxTemplatePlugin
+        .getLynxTemplatePluginHooks(
+          compilation,
+        )
+
+      applySourceMappingURLRewriter(compiler, compilation)
```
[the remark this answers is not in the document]
upupming: Skipping for this PR by design — production source-map serving will be handled by Slardar (which uploads `.map` files + reverse-symbolicates on its side) in a follow-up, and open-source consumers will pick their own static host strategy. The dev-server middleware here is intentionally the only consumer of `?field=source-map` URLs for now.

Will document the dev-vs-prod scope when the Slardar integration lands.

packages/rspeedy/plugin-debug-metadata/src/collectors/artifacts.ts
```diff
@@ -0,0 +1,119 @@
...
+
+    for (const chunk of chunkGroup.chunks) {
+      for (const file of chunk.files) {
+        if (!isMappableAsset(file)) continue
+        if (seen.has(file)) continue
+        seen.add(file)
+
+        const mapPath = `${file}.map`
+        const map = readSourceMap(compilation, mapPath)
+        if (!map) continue
```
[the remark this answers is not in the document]
upupming: Custom `output.sourceMapFilename` is not a configuration path we currently support in Rspeedy / Lynx — assuming `${file}.map` is the established convention for this stack, and we have no in-tree users overriding it. If a real use case shows up we can revisit; for now closing as out-of-scope.

packages/rspeedy/plugin-debug-metadata/src/LynxDebugMetadataPlugin.ts
```diff
@@ -0,0 +1,253 @@
...
+          }
+          const debugMetadataAssetName = path.posix.format({
+            dir: args.intermediate,
+            base: this.options.debugMetadataAssetName,
+          })
+          compilation.emitAsset(
+            debugMetadataAssetName,
+            new RawSource(JSON.stringify(asset, null, 2)),
+          )
+          args.intermediateAssets.push(debugMetadataAssetName)
```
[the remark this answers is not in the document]
upupming: Intentional — `debug-metadata.json` is supposed to be cleaned up from production bundles. The file is a build-time artifact for reverse-symbolication services (Slardar uploads it during release; open-source consumers do their own upload) — shipping it inside the final user-facing bundle would leak internal debug data and bloat downloads.

`templateDebugUrl` / `debugMetadataUrl` in tasm.json point at the URL where Slardar (or the user's static host) serves the uploaded copy, not the local file. Dev / debug / rsdoctor builds keep the file on disk so the dev-server middleware can serve it locally.

Added a code comment documenting why the `intermediateAssets.push` is there.

packages/rspeedy/plugin-debug-metadata/src/source-mapping-url-rewriter.ts
```diff
@@ -0,0 +1,107 @@
...
+  let decodedBasename: string
+  try {
+    decodedBasename = decodeURIComponent(encodedBasename)
+  } catch {
+    decodedBasename = encodedBasename
+  }
+  const dir = lastSlash >= 0 ? encodedPath.slice(0, lastSlash + 1) : ''
+  const newUrl =
+    `${dir}${DEBUG_METADATA_ASSET_NAME}?field=source-map&filename=${
+      encodeURIComponent(decodedBasename)
```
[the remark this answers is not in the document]
upupming: Fixed by removing the customization option entirely (commit incoming). `LynxDebugMetadataPlugin` no longer takes `debugMetadataAssetName`; the asset name is hardcoded to `debug-metadata.json` in every layer (plugin emitter, rewriter, template `templateDebugUrl`, middleware). Single source of truth via the existing `DEBUG_METADATA_ASSET_NAME` constant. Same posture as #14 (`output.sourceMapFilename`): not a configuration path we support in Rspeedy / Lynx.

packages/rspeedy/plugin-debug-metadata/src/source-mapping-url-rewriter.ts
```diff
@@ -0,0 +1,107 @@
...
+    }`
```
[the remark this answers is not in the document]
upupming: Fixed. The rewriter now passes the bundler-relative `mapAssetPath` (`${jsAssetName}.map`) and writes it into the query as `path=<encoded>` instead of `filename=<basename>`. The resolver matches by exact `path`, so two entries emitting `app/index.js.map` and `vendor/index.js.map` get disambiguated correctly. +1 test (`uses full bundler-relative mapAssetPath so same-basename across entries disambiguates`).

packages/rspeedy/plugin-debug-metadata/src/source-mapping-url-rewriter.ts
```diff
@@ -0,0 +1,110 @@
...
+
+  const noHash = originalUrl.split('#', 1)[0]!
+  const encodedPath = noHash.split('?', 1)[0]!
+  const lastSlash = encodedPath.lastIndexOf('/')
+  const encodedBasename = encodedPath.slice(lastSlash + 1)
+  if (!encodedBasename) return undefined
+  const dir = lastSlash >= 0 ? encodedPath.slice(0, lastSlash + 1) : ''
+  const newUrl = `${dir}${DEBUG_METADATA_ASSET_NAME}?field=source-map&path=${
+    encodeURIComponent(mapAssetPath)
+  }`
```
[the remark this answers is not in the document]
upupming: Fixed by restricting the rewriter to JS in `compilation.entrypoints[*].chunks` (initial chunks only). Async / dynamic-import chunks are no longer rewritten — their original `.map`-sibling `sourceMappingURL` stays intact, so they keep working against the file on disk. Async-template metadata lives at `<intermediateRoot>/async/<filename>/debug-metadata.json` which is a separate concern; if the URL-rewrite-against-metadata story extends to async chunks later we can revisit.

packages/rspeedy/plugin-debug-metadata/src/middleware.ts
```diff
@@ -0,0 +1,349 @@
...
+    const outputPath = compiler.outputPath
+      ?? compiler.options?.output?.path
+      ?? ''
+    const publicPathPrefix = extractPublicPathPrefix(
+      compiler.options?.output?.publicPath,
+    )
+    let stripped = pathname
+    if (publicPathPrefix && stripped.startsWith(publicPathPrefix)) {
+      stripped = stripped.slice(publicPathPrefix.length)
+    }
```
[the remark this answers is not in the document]
upupming: Fixed. Strip only when the URL equals the prefix exactly or has a `/` immediately after it. +1 test (`/assets2/...` no longer matches `/assets` prefix).

packages/webpack/template-webpack-plugin/src/LynxTemplatePlugin.ts
```diff
@@ -789,22 +785,34 @@ class LynxTemplatePluginImpl {
...
+      compiler.options.output.publicPath,
+      {
+        chunk: mainThreadChunk,
+        hash: compilation.hash,
+        contentHash: mainThreadChunk?.contentHash ?? {},
+      },
+    );
+    if (resolvedPublicPath && resolvedPublicPath !== '/') {
+      debugMetadataUrl = joinPublicPath(resolvedPublicPath, debugMetadataPath);
+      templateDebugUrl = `${debugMetadataUrl}${debugMetadataDebugInfoQuery}`;
```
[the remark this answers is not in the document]
upupming: We use `"@lynx-js/template-webpack-plugin": minor` a breaking change to avoid this

packages/rspeedy/plugin-debug-metadata/src/LynxDebugMetadataPlugin.ts
```diff
@@ -0,0 +1,335 @@
...
+    const asset = compilation.getAsset(assetName)
+    if (!asset) continue
+    const before = asset.source.source().toString()
+    const after = rewriteTrailerToAbsoluteUrl(
+      before,
+      debugMetadataUrl,
+      `${assetName}.map`,
+    )
+    if (after === undefined) continue
+    compilation.updateAsset(assetName, new RawSource(after), asset.info)
```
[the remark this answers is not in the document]
upupming: Fixed in ee31b890e. After `compilation.updateAsset` writes the rewritten JS to disk, the tap now also mirrors `after` back into `args.encodeData`:

- `args.encodeData.manifest[name]` is overwritten with the rewritten string for background entries (manifest was snapshotted before `beforeEncode` fired).
- `args.encodeData.lepusCode.root` / `chunks[i]` are replaced with a fresh `{...chunk, source: newSource}` so the encoder's later `source.source()` reads the rewritten trailer.

So the encoded template binary now embeds the same metadata-endpoint URL the on-disk JS carries — runtime stack traces sourced from inlined script no longer point at stale sibling maps.

packages/rspeedy/plugin-debug-metadata/src/collectors/artifacts.ts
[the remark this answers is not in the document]
upupming: Fixed in 1d2c5be55. `collectArtifacts` now always emits a JS artifact and only omits the `source-map` debug source when no sibling `.map` exists. The `beforeEmit` lepusNG attachment still finds a main-thread artifact to prepend `bytecode-debug-info` to, so `?field=bytecode-debug-info` works in dev builds with `output.sourceMap: false`. CSS keeps the existing skip-when-no-map behavior — without a source-map source it carries no other debug payload.

.changeset/debug-metadata-rsbuild-plugin.md
```diff
@@ -0,0 +1,15 @@
+---
+"@lynx-js/debug-metadata-rsbuild-plugin": minor
+"@lynx-js/debug-metadata": minor
+"@lynx-js/rspeedy": minor
+"@lynx-js/template-webpack-plugin": minor
+"@lynx-js/react-webpack-plugin": patch
```
luhc228: I think we should split to another changeset. 

> New `@lynx-js/debug-metadata` schema package

and 

> New `@lynx-js/debug-metadata-rsbuild-plugin` emits the file and serves `?field=…` queries in dev.

do not belong to @lynx-js/react-webpack-plugin

.changeset/debug-metadata-rsbuild-plugin.md
```diff
@@ -0,0 +1,15 @@
...
+---
+
+Add unified `debug-metadata.json` per Lynx entry.
+
+- New `@lynx-js/debug-metadata` schema package (zero-dep).
+- New `@lynx-js/debug-metadata-rsbuild-plugin` emits the file and serves `?field=…` queries in dev.
+- JS `//# sourceMappingURL=` and tasm `templateDebugUrl` repointed at the new endpoint.
```
luhc228, line 12: Should upgrade [lynx-devtool](https://github.com/lynx-family/lynx-devtool) to support parse `lynx-metadata.json`?

packages/rspeedy/plugin-debug-metadata/src/collectors/git.ts
```diff
@@ -0,0 +1,87 @@
...
+
+/**
+ * Capture git state for `cwd`. Returns `null` when `cwd` is not inside
+ * a git worktree (or git is unavailable), in which case the caller
+ * should omit the `meta.git` field entirely.
+ */
+export function collectGitMetadata(cwd: string): GitMetadata | null {
+  const commit = tryRunGit(cwd, ['rev-parse', 'HEAD'])
+  if (!commit) return null
+  const branch = tryRunGit(cwd, ['rev-parse', '--abbrev-ref', 'HEAD'])
```
luhc228: I think commit is enough. Branch will be deleted when PR was merged.

packages/rspeedy/plugin-debug-metadata/src/collectors/git.ts
```diff
@@ -0,0 +1,87 @@
...
+    ?? 'unknown'
+  const rootDir = tryRunGit(cwd, ['rev-parse', '--show-toplevel'])
```
luhc228, line 74: when we need rootDir?

packages/tools/debug-metadata/test/resolve.test.ts
```diff
@@ -0,0 +1,225 @@
...
+  mappings: '',
+};
+
+const SAMPLE: DebugMetadataAsset = {
+  artifacts: [
+    {
+      kind: 'main-thread',
+      filename: 'main-thread.js',
+      path: '.rspeedy/main/main-thread.js',
+      tasmSection: ['lepusCode', 'root'],
```
luhc228, line 30: Why we need tasmSection?

.changeset/debug-metadata-rsbuild-plugin.md
upupming: Sure, will do it in the future

packages/rspeedy/plugin-debug-metadata/src/collectors/git.ts
upupming: Sure, removed

packages/rspeedy/plugin-debug-metadata/src/collectors/git.ts
upupming: For UI node from lazy bundle or common libs, we are hard to find how it was built. Add rootDir to let AI quick locate the actual path to build.

packages/tools/debug-metadata/test/resolve.test.ts
upupming: In the future, customSections will be used for fetchBundle, they run bytecode on main thread too. Use this section to locate the DebugMetadataAsset for any customSections quickly for bytecode remapping

.changeset/debug-metadata-rsbuild-plugin.md
luhc228: Should we add some notes in the changelog?

luhc228 (dismissed):

packages/rspeedy/plugin-debug-metadata/src/pluginLynxDebugMetadata.ts
```diff
@@ -0,0 +1,102 @@
...
+        // entry plumbing — emitting our metadata there would overwrite
+        // the lynx one with web-asset paths, and rewriting web JS
+        // trailers would point them at a metadata file whose artifacts
+        // don't include the web asset paths. Web JS source maps work
+        // fine via the default `.map` sibling; skip cleanly.
+        const isLynx = environment.name === 'lynx'
+          || environment.name.startsWith('lynx-')
+        if (!isLynx) return
+
+        if (!exposed) return
```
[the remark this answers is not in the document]
upupming: We considered fail-fast here but intentionally keep the silent skip.

Unlike `pluginLynxConfig` — which is added only as part of a DSL plugin chain, so `LynxTemplatePlugin` is always exposed by the time it runs — `pluginLynxDebugMetadata` is auto-registered for *every* Rspeedy build via `applyDefaultPlugins`. A lynx environment without a DSL plugin exposing `LynxTemplatePlugin` is legitimate: pure-core builds, and several of our internal DSLs (not only ReactLynx). Throwing there breaks those builds — it fails a number of `@lynx-js/rspeedy` core tests that build a lynx env without any DSL plugin.

Gating the throw on a known DSL-plugin registry (the approach `pluginLynxConfig` can afford) would require enumerating every DSL — react, ttml, and internal/future ones — which cannot be kept complete in this package. So we skip silently rather than risk a false-positive hard error.

packages/rspeedy/plugin-debug-metadata/src/pluginLynxDebugMetadata.ts
```diff
@@ -0,0 +1,102 @@
...
+function devServerOrigin(api: RsbuildPluginAPI): string | undefined {
+  if (!api.context.devServer) return
+  const port = api.context.devServer.port
+  const { assetPrefix } = api.getNormalizedConfig().dev
+  if (typeof assetPrefix !== 'string') {
+    const errorMsg = 'dev.assetPrefix is not string, skip printing QRCode'
+    // Rspeedy will normalized dev.assetPrefix to string
+    throw new Error(errorMsg)
+  }
+  return new URL(assetPrefix.replaceAll('<port>', String(port))).origin
```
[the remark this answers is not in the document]
upupming: Good catch — fixed in aaa27f15d. `devServerOrigin` now returns `origin + pathname` (trailing slash trimmed) instead of just `.origin`, so a non-root `dev.assetPrefix` (e.g. `https://host:<port>/assets/`) resolves the same route as the rewritten `.map` `sourceMappingURL`. The root-path case is unchanged.

packages/rspeedy/plugin-debug-metadata/src/pluginLynxDebugMetadata.ts
```diff
@@ -0,0 +1,103 @@
...
+const PLUGIN_NAME = 'lynx:debug-metadata'
+
+function devServerOrigin(api: RsbuildPluginAPI): string | undefined {
+  if (!api.context.devServer) return
+  const port = api.context.devServer.port
+  const { assetPrefix } = api.getNormalizedConfig().dev
+  if (typeof assetPrefix !== 'string') {
+    const errorMsg = 'dev.assetPrefix is not string, skip printing QRCode'
+    // Rspeedy will normalized dev.assetPrefix to string
+    throw new Error(errorMsg)
```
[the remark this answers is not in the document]
upupming: Fixed in eac7e0bac. `devServerOrigin` now returns `undefined` (instead of throwing) when `dev.assetPrefix` is not a string. `dev.assetPrefix: false` is left as `false` by `pluginDev`, so the dev URL is now skipped gracefully — the same way the core sourcemap plugin returns early for `assetPrefix === false`.

packages/rspeedy/plugin-debug-metadata/src/LynxDebugMetadataPlugin.ts
```diff
@@ -0,0 +1,355 @@
...
+            uiSourceMap: createUiSourceMap(uiSourceMapRecords),
+            meta: {
+              ...(git ? { git } : {}),
+              rspeedy,
+            },
+          }
+          const debugMetadataAssetName = path.posix.format({
+            dir: args.intermediate,
+            base: DEBUG_METADATA_ASSET_NAME,
+          })
```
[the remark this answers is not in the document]
upupming: Fixed in eac7e0bac. The per-template `intermediate` is now normalized with `.replace(/\\/g, "/")` before it feeds `path.posix.format` and the `intermediateByEntryKey` map (which `beforeEmit` reuses), so the emitted asset name and `debugMetadataUrl` stay forward-slashed on Windows and match the dev-server/browser request. This mirrors the existing `rootName.replace(/\\/g, "/")` normalization a few lines below.

packages/webpack/template-webpack-plugin/src/LynxTemplatePlugin.ts
```diff
@@ -789,22 +785,34 @@ class LynxTemplatePluginImpl {
...
-        debugInfoPath,
-        compiler.options.output.publicPath,
-      ).toString();
+    const mainThreadAssetName = assetsInfoByGroups.mainThread[0]?.name;
+    const mainThreadBasename = mainThreadAssetName
+      ? path.posix.basename(mainThreadAssetName.replace(/\\/g, '/'))
+      : 'main-thread.js';
+    const debugMetadataDebugInfoQuery = `?field=bytecode-debug-info&filename=${
+      encodeURIComponent(mainThreadBasename)
+    }`;
```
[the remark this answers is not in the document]
upupming: We looked into this and are keeping the basename query for now. Each template (the main card and every dynamic component) emits its own `debug-metadata.json`, and `findArtifact` resolves within that single container. Within one container the artifacts are the template's own main-thread and background chunks, whose basenames never collide (`main-thread.js` vs `background.js`, or `<name>-react__main-thread.js` vs `<name>-react__background.js` for lazy bundles). The bytecode lookup uses the lepus root chunk basename, which uniquely identifies the main-thread artifact in that container. The `[name]/index.js` collision would require two same-basename artifacts in the *same* container, but those source-map filenames come from the fixed Lynx intermediate chunk names, not user `output.filename`, so it cannot happen in practice. Happy to switch to path-based matching (like `findSourceMap`) if a real collision shows up.

packages/rspeedy/plugin-debug-metadata/src/release-banner.ts
```diff
@@ -0,0 +1,40 @@
...
+}
+
+export function getReleaseRuntime(): string {
+  return `(function () {
+  'use strict';
+  try {
+    throw new Error(${RELEASE_DEFINE});
+  } catch (e) {
+    e.name = 'LynxGetSourceMapReleaseError';
+    if (typeof _SetSourceMapRelease === 'function') {
```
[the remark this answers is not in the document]
upupming: Thanks — we are intentionally not coupling this banner to the web-core API here. The release banner implements the native Lynx engine's source-map release convention (a thrown `Error` whose message/stack the engine reads), and it is only injected for the `lynx` rsbuild environment — the `web` environment is gated out (`isLynx` check in `pluginLynxDebugMetadata`). Lynx-for-Web (`web-core`) is a separate runtime with its own `_SetSourceMapRelease(errInfo) => errInfo?.release` shape; driving its release from this native-oriented banner is the wrong layer. On web-core the call is simply a harmless no-op (release stays unset) rather than a regression, so we are leaving the banner native-only for now.

colinaaa (dismissed):

.changeset/debug-metadata-asset-prefix-path.md
```diff
@@ -0,0 +1,8 @@
+---
+"@lynx-js/debug-metadata-rsbuild-plugin": patch
+---
+
+Harden dev `debugMetadataUrl` derivation from `dev.assetPrefix`:
```
upupming: This changeset should be removed since `@lynx-js/debug-metadata-rsbuild-plugin` is a new plugin and its changeset is included in `.changeset/debug-metadata-rsbuild-plugin.md`

.changeset/debug-metadata-release-banner.md
```diff
@@ -0,0 +1,5 @@
+---
+"@lynx-js/debug-metadata-rsbuild-plugin": minor
```
upupming: This changeset should be removed since `@lynx-js/debug-metadata-rsbuild-plugin` is a new plugin and its changeset is included in `.changeset/debug-metadata-rsbuild-plugin.md`

.changeset/debug-metadata-windows-intermediate.md
```diff
@@ -0,0 +1,5 @@
+---
+"@lynx-js/debug-metadata-rsbuild-plugin": patch
```
upupming: This changeset should be removed since `@lynx-js/debug-metadata-rsbuild-plugin` is a new plugin and its changeset is included in `.changeset/debug-metadata-rsbuild-plugin.md`

packages/rspeedy/plugin-debug-metadata/src/collectors/artifacts.ts
```diff
@@ -0,0 +1,127 @@
...
+  } catch {
+    return undefined
+  }
+}
+
+/**
+ * Unique key used by remapping services to match a map.
+ *
+ * - JS assets use `chunk.hash` so the value matches the
+ *   `__SOURCEMAP_RELEASE__` banner `SlardarWebpackPlugin` bakes into
```
upupming:
```suggestion
 *   `__DEBUG_METADATA_RELEASE__` banner
```

packages/rspeedy/plugin-debug-metadata/src/constants.ts
```diff
@@ -0,0 +1,10 @@
+// Copyright 2026 The Lynx Authors. All rights reserved.
+// Licensed under the Apache License Version 2.0 that can be found in the
+// LICENSE file in the root directory of this source tree.
+
+/**
+ * Default name of the debug metadata asset emitted alongside each template.
+ *
+ * @public
+ */
+export const DEBUG_METADATA_ASSET_NAME = 'debug-metadata.json'
```
upupming, line 10: Can we export this in npm pkg exports to avoid repeat define this var.

upupming (commented):
I should review the changeset to avoid duplicates and make code clean

packages/rspeedy/plugin-debug-metadata/src/pluginLynxDebugMetadata.ts
```diff
@@ -0,0 +1,99 @@
...
+import type { CompilerHandle } from './middleware.js'
+
+const PLUGIN_NAME = 'lynx:debug-metadata'
+
+function devServerOrigin(api: RsbuildPluginAPI): string | undefined {
+  if (!api.context.devServer) return
+  const port = api.context.devServer.port
+  const { assetPrefix } = api.getNormalizedConfig().dev
+  if (typeof assetPrefix !== 'string') return
+  const url = new URL(assetPrefix.replaceAll('<port>', String(port)))
```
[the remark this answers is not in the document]
upupming: Fixed in 3186f6099. `devServerOrigin` now resolves `dev.assetPrefix` against a placeholder base, so a path-only public path like `/assets/` no longer throws — the placeholder origin is dropped and the trailer stays relative (`/assets/debug-metadata.json`); absolute prefixes keep their origin, and an underivable prefix returns `undefined` to skip the rewrite.

packages/rspeedy/core/test/plugins/sourcemap.plugin.test.ts
```diff
@@ -28,7 +28,15 @@ describe('sourcemap.plugin', () => {
       const rspeedy = await createStubRspeedy({})
       const config = await rspeedy.unwrapConfig()
       expect(config.devtool).toBe(false)
-      expect(SourceMapDevToolPlugin).not.toBeCalled()
+      expect(SourceMapDevToolPlugin).toBeCalledWith(
```
HuJean, line 31: why not called before?

packages/rspeedy/plugin-react/test/sourcemap.test.ts
```diff
@@ -166,6 +166,40 @@ describe('Sourcemap', () => {
...
+      // The release is `debugmetadata:` + the chunk hash; the bare hash is the
+      // source-map artifact `key` reverse-resolution locates the container by.
+      expect(mtRelease, 'main-thread should declare a release').toMatch(
+        /^debugmetadata:[0-9a-f]+$/,
+      )
+      expect(btRelease, 'background should declare a release').toMatch(
+        /^debugmetadata:[0-9a-f]+$/,
+      )
+      // The injected runtime registers the release with the Lynx engine.
+      expect(mainThread).toContain('_SetSourceMapRelease')
```
HuJean, line 195: it's injected by debug-metadata-rsbuild-plugin both dev and prod?

packages/rspeedy/core/test/plugins/sourcemap.plugin.test.ts
upupming: Because `output.source` is `false` on prod by default, now we change it to `true` for lynx for generating debug metadata's sourcemap section on build.

packages/rspeedy/plugin-react/test/sourcemap.test.ts
upupming: Yes we will inject it both dev and prod, it will help lynx devtool use the stackframe with release id to call our remap backend to get original sources.

HuJean (approved):

luhc228 (approved):

PR #2645, opened 2026-05-15
feat(react): support Element Template refs

Yradex:
## Summary by CodeRabbit

* **New Features**
  * Added element template ref support with proper hydration handling and lifecycle management
  * Implemented selector-based ref proxies for deferred ref operations

* **Tests**
  * Comprehensive test coverage for element template ref behavior during hydration and updates
  * Added tests for ref cleanup ordering and error handling

[Review Change Stack](https://app.coderabbit.ai/change-stack/lynx-family/lynx-stack/pull/2645?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

## Overview

- Adds ordinary ref support for the Element Template runtime, covering transform metadata, direct refs, spread refs, hydration/update behavior, and cleanup.
- Keeps the native-visible slot value serializable by sending stable ref markers while the actual user ref values stay in ET-owned adapter state.

## Key Points

- Transform now emits ET attr-plan metadata for direct `ref` and spread `ref` without putting user ref values into Template Definition or native payloads.
- Runtime prepares direct and spread ref slots as handle/slot markers, flushes attach/detach at commit boundaries, and delays selector proxy UI operations until hydrate has a stable handle.
- Hydration and update compare native-visible markers rather than raw JavaScript ref identities, so stable refs do not get duplicate callbacks while changed refs still flush side effects.
- Removed subtree and destroy paths clear retained ref state and queued operations through the ET adapter boundary.

## Checklist

### DIFF
--- a/packages/react/runtime/src/core/ref.ts
+++ b/packages/react/runtime/src/core/ref.ts
@@ -1,0 +1,166 @@
+// Copyright 2026 The Lynx Authors. All rights reserved.
+// Licensed under the Apache License Version 2.0 that can be found in the
+// LICENSE file in the root directory of this source tree.
+
+import type { NodesRef, SelectorQuery } from '@lynx-js/types';
+
+export type RefCleanup = (() => void) | void;
+export type RefCallback<T> = ((ref: T | null) => RefCleanup) & {
+  _unmount?: RefCleanup;
+};
+export interface RefObject<T> {
+  current: T | null;
+}
+export type OrdinaryRef<T> = RefCallback<T> | RefObject<T>;
+
+type FunctionPropertyNames<T> = {
+  [K in keyof T]: T[K] extends (...args: unknown[]) => unknown ? K : never;
+}[keyof T];
+
+export type ForwardableNodesRefMethod = Exclude<FunctionPropertyNames<NodesRef>, 'exec'>;
+export type RefProxyForwardedMethods<TProxy> = {
+  [K in ForwardableNodesRefMethod]: (...args: Parameters<NodesRef[K]>) => TProxy;
+};
+
+type RefTask = (nodesRef: NodesRef) => SelectorQuery;
+
+export function assertValidRef<T>(value: unknown): OrdinaryRef<T> {
+  if (
+    typeof value === 'function'
+    || (typeof value === 'object' && value !== null && 'current' in value)
+  ) {
+    return value as OrdinaryRef<T>;
+  }
+  throw new Error(
+    `Elements' "ref" property should be a function, or an object created `
+      + `by createRef(), but got [${typeof value}] instead`,
+  );
+}
+
+export function normalizeRefValue<T>(value: unknown): OrdinaryRef<T> | null | undefined {
+  if (value === null || value === undefined) {
+    return value;
+  }
+  return assertValidRef<T>(value);
+}
+
+export function applyOrdinaryRef<T>(
+  ref: OrdinaryRef<T>,
+  value: T | null,
+): void {
+  try {
+    if (typeof ref === 'function') {
+      const cleanup = ref._unmount;
+      const hasCleanup = typeof cleanup === 'function';
+      if (hasCleanup) {
+        cleanup();
+      }
+      ref._unmount = undefined;
+
+      if (!hasCleanup || value !== null) {
+        const nextCleanup = ref(value);
+        if (typeof nextCleanup === 'function') {
+          ref._unmount = nextCleanup;
+        }
+      }
+    } else {
+      ref.current = value;
+    }
+  } catch (error) {
+    lynx.reportError(error as Error);
+  }
+}
+
+// Keeps the Snapshot/ET ordinary ref ordering shared without owning backend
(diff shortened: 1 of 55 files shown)


packages/react/runtime/src/snapshot/lifecycle/ref/delay.ts
```diff
@@ -58,64 +50,31 @@ function runDelayedUiOps(): void {
...
-        };
+    return this.createProxy();
+  }
 
-        return forward(prop as ForwardableNodesRefMethod);
-      },
-    }) as RefProxy;
+  protected createProxyTarget(): RefProxy {
+    return new RefProxy(this.refAttr);
   }
```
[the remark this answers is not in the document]
Yradex: @coderabbitai I do not think this recurses. `setTask` is a TypeScript `private` method, not a runtime `#private` field, so it still exists on the prototype. `SelectorRefProxy.createProxy()` checks `prop in target` before the forwarded NodesRef-method branch, so accessing `.setTask` on the proxy should go through `Reflect.get(...)` and call the real method.

Could you re-check this with that guard considered? Returning `this` from `createProxyTarget()` would change the current proxy/task ownership semantics, so I would avoid that unless there is a concrete failing chain.

packages/react/runtime/src/core/ref.ts
```diff
@@ -0,0 +1,166 @@
...
+  }
+  return assertValidRef<T>(value);
+}
+
+export function applyOrdinaryRef<T>(
+  ref: OrdinaryRef<T>,
+  value: T | null,
+): void {
+  try {
+    if (typeof ref === 'function') {
```
HuJean, line 52: Has same logic in snapshot mode?

Yradex: Yes. See https://github.com/lynx-family/lynx-stack/pull/2645/changes#diff-ca25deeb8888e78750bf9089a63cd316475f4ff3043389c0b3cf8238c281b890R35

packages/react/runtime/src/element-template/background/manager.ts
```diff
@@ -14,7 +14,7 @@ export const backgroundElementTemplateInstanceManager: {
   clear(): void;
 } = {
   nextId: 0,
-  values: new Map<number, BackgroundElementTemplateInstance>(),
+  values: /*#__PURE__*/ new Map<number, BackgroundElementTemplateInstance>(),
```
HuJean: The __PURE__ is useless cause backgroundElementTemplateInstanceManager must be used?

Yradex: Yes.

packages/react/runtime/src/element-template/background/hydrate.ts
```diff
@@ -133,53 +145,59 @@ function hydrateInstance(
     if (!serializedSlot && !backgroundSlot) {
       continue;
     }
-    hydrateElementSlot(instance, slotId, serializedSlot ?? []);
+    if (!hydrateElementSlot(instance, slotId, serializedSlot ?? [])) {
```
HuJean, line 148: return hydrateElementSlot(instance, slotId, serializedSlot ?? []);

packages/react/runtime/src/element-template/background/destroy.ts
```diff
@@ -1,18 +1,27 @@
...
+import { __root } from '../runtime/page/root-instance.js';
 
 export function destroyElementTemplateBackgroundRuntime(): void {
   resetElementTemplateHydrationListener();
-  resetElementTemplateCommitState();
-  // Destroy is the only place that may discard delayed removed subtrees instead
-  // of letting the Snapshot-aligned timer tear them down later.
   cancelElementTemplateRemovedSubtreeCleanup();
+
+  render(null, __root as unknown as ContainerNode);
```
HuJean, line 18: clear vnode tree？

packages/react/runtime/src/element-template/prop-adapters/spread.ts
```diff
@@ -52,7 +61,6 @@ export function prepareSpreadAttrSlot(
 
     if (
       spreadValue === undefined
-      || key === 'ref'
       || key.endsWith(':ref')
```
HuJean, line 64: foo:ref is same with main-thread:ref ?

Yradex: No. It's not supported.

packages/react/runtime/src/element-template/background/destroy.ts
Yradex: Yes. It's used to call ref cleanups.

packages/react/runtime/src/element-template/background/hydrate.ts
Yradex: It's in a for loop, so we can't just return here. We don't support multiple roots though.

HuJean (approved):

gaoachao (approved):

PR #2651, opened 2026-05-16
feat(react): route ET background tree via __slotIndex

upupming:
## Summary

- Background ET tree now lives on `__slotIndex` set by the Lynx Preact fork from `$N` props; the `BackgroundElementTemplateSlot` wrapper instances are gone
- Main-thread renderer drops the legacy `children` slot-array fallback in favor of `$N` only
- `__etSlot` runtime helper removed; SWC plugin emits `$N` named children directly

## Test plan

- `pnpm test:et` — 427 passing, 100% coverage

## Summary by CodeRabbit

* **New Features**
  * Added support for sparse element slots for more efficient conditional slot content.

* **Refactor**
  * Slot model simplified to per-instance slot indexing instead of wrapper nodes, improving slot handling and performance.
  * Template rendering and hydration now use named placeholder props for slot content; snapshots and tests updated to match the new behavior.

* **Tests**
  * Extensive fixture and test updates to reflect sparse-slot and placeholder prop behavior.

[Review Change Stack](https://app.coderabbit.ai/change-stack/lynx-family/lynx-stack/pull/2651?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


packages/react/transform/crates/swc_plugin_element_template/lowering.rs
```diff
@@ -136,27 +99,25 @@ where
...
+      let (slot_index, expr) = match dynamic_child {
+        DynamicElementPart::Slot(expr, idx) | DynamicElementPart::ListSlot(expr, idx) => {
+          (idx, expr)
+        }
+      };
       rendered_attrs.push(JSXAttrOrSpread::JSXAttr(JSXAttr {
         span: DUMMY_SP,
-        name: JSXAttrName::Ident(IdentName::new("children".into(), DUMMY_SP)),
+        name: JSXAttrName::Ident(IdentName::new(format!("${}", slot_index).into(), DUMMY_SP)),
         value: Some(JSXAttrValue::JSXExprContainer(JSXExprContainer {
```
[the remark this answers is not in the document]
upupming: This is correct for upstream Preact, but the project uses `@lynx-js/internal-preact` (currently `10.29.1-20260424024911-12b794f`) which has explicit `$N` named-children support. In `node_modules/.pnpm/@lynx-js+internal-preact@…/dist/preact.mjs`, `diffElementNodes` does:

```js
for (i in newProps) {
  value = newProps[i];
  if (i == 'children') {
    newChildren = value;
  } else if (typeof i == 'string' && i[0] == '$') {
    (newChildren ??= [])[+i.slice(1)] = value;
    hasNamedChildren = true;
  } else if (...) { ... } else if (...) { setProperty(dom, i, value); }
  ...
}
```

So `$N` props land in `newChildren`, never in `setProperty`. `diffChildren` is then called with `_slotIndex = true` so each child receives its `__slotIndex` from its array position (`preact.mjs:1331-1343` / `:1258-1259`), and the background `insertBefore` keys the insertNode op off that field (`background/instance.ts:178-194`).

End-to-end coverage on the background path is in:

- `runtime/background/render/preact.test.tsx`
- `runtime/background/update/compiled-fixtures.test.tsx` (6 tests: keyed moves / removals / re-add for `compiled-keyed-list` and `compiled-keyed-subtree-list`)
- `runtime/background/event/compiled-fixtures.test.tsx` (8 tests, including hydration + event dispatch through `$N`-mounted subtrees)

These tests run `transformReactLynx` against real JSX with `experimental_useElementTemplate`, dynamically import the compiled output, and execute it through Preact's actual background diff via `setupBackgroundElementTemplateDocument()`. If `$N` were dropped as DOM attributes, none of these subtrees would mount or hydrate; all 16 pass.

.changeset/et-multi-slots.md
```diff
@@ -0,0 +1,5 @@
+---
+
+---
+
+No package release is required because this change continues the unpublished Element Template runtime path behind the experimental `experimental_useElementTemplate` flag. The SWC `swc_plugin_element_template` lowering now emits `$0`, `$1`, ... named-slot JSX props instead of `__etSlot(N, ...)` wrapper calls, the background runtime routes inserts/removes off `child.__slotIndex` directly, and the legacy `BackgroundElementTemplateSlot` wrapper + `__etSlot` runtime component are removed. Only internal runtime, transform, fixtures, and tests change — public APIs, package exports, and release-facing defaults are unchanged.
```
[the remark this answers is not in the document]
upupming: Intentional — this PR continues the unpublished `experimental_useElementTemplate` runtime path that's gated off in release defaults, so no public package should bump.

Matches the existing precedent in `.changeset/empty-et-event-implementation.md` (from #2589, which used the identical `---\n\n---` empty-frontmatter form with a body explaining why no release is required). The `@changesets/cli` is lenient about an empty package list when there's body text — it skips the entry and doesn't bump anything, which is the desired behavior here.

packages/react/runtime/src/element-template/background/instance.ts
```diff
@@ -165,39 +185,27 @@ export class BackgroundElementTemplateInstance {
...
-
-    if (silent) {
+    if (!this.canEmitPatch()) {
       return;
     }
+    // beforeChild may live in another slot — fall back to no anchor so the
+    // child appends at the slot's tail. Same-slot reference is honored.
+    const beforeId = beforeChild && beforeChild.__slotIndex === slotId
+      ? beforeChild.instanceId
+      : 0;
```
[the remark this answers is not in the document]
upupming: Good catch — fixed in b295b6e6.

The op stream only encodes `(parent, slotId, child, beforeId)`, so when `beforeChild.__slotIndex !== child.__slotIndex` the main thread had to fall back to append-tail (`beforeId = 0`). I now treat the cross-slot anchor as `null` for the local sibling-chain mutation as well, so `syncMultiSlotChildren(this, slotId)` filters into the same order both sides agree on.

Added a regression test in `runtime/background/instance.test.ts`:

```ts
it('keeps elementSlots[slotId] in lock-step with append-tail when anchor is cross-slot', () => {
  const parent = new BackgroundElementTemplateInstance('view');
  const slot0Tail = new BackgroundElementTemplateInstance('text');
  slot0Tail.__slotIndex = 0;
  const crossSlotAnchor = new BackgroundElementTemplateInstance('text');
  crossSlotAnchor.__slotIndex = 1;
  parent.appendChild(slot0Tail);
  parent.appendChild(crossSlotAnchor);

  const incoming = new BackgroundElementTemplateInstance('text');
  incoming.__slotIndex = 0;
  parent.insertBefore(incoming, crossSlotAnchor);

  expect(parent.elementSlots[0]).toEqual([slot0Tail, incoming]);
  expect(parent.lastChild).toBe(incoming);
});
```

Before the fix this asserted `[incoming, slot0Tail]` (chain order driven by the literal anchor) while main thread saw `[slot0Tail, incoming]` from the append-tail op — exactly the divergence you described.

packages/react/runtime/src/element-template/internal.ts
```diff
@@ -53,5 +53,4 @@ export type { Options } from 'preact';
 // export { transformToWorklet } from '../worklet/call/transformToWorklet.js';
 // export { registerWorkletOnBackground } from '../worklet/hmr.js';
 // export { loadWorkletRuntime } from '@lynx-js/react/worklet-runtime/bindings';
-export { __etSlot } from './runtime/components/slot.js';
 export { __etAttrPlanMap, adaptEventAttrSlot, adaptSpreadAttrSlot } from './runtime/template/attr-slot-plan.js';
```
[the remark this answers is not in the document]
upupming: This was an intentional cleanup, not an oversight — `experimental_useElementTemplate` is unpublished and gated off in release defaults, so there are no end-user bundles in the wild whose `__etSlot` import we need to honor.

The cleanup is end-to-end consistent:

- **No emitter ships `__etSlot` anymore.** The transform crate's `slot.rs` helper (which generated `__etSlot(N, ...)` calls) is deleted, and `lowering.rs` emits `$N` JSX attrs in every target now (`LEPUS` and `JS`). Verified via `cargo test -p swc_plugin_element_template` — none of the 33 snapshot outputs contain `__etSlot`.
- **No consumer references it.** `grep -rn "__etSlot" packages/react/runtime packages/react/transform packages/react/testing-library packages/webpack` returns zero matches after this PR. The legacy `runtime/components/slot.test.ts` and the legacy `__etSlot` assertions in `background-adapter.et.test.tsx` were removed in the same commit (95636592) that deleted the runtime component.
- **The aliased test entry (`@lynx-js/react/internal` → `src/element-template/internal.ts`) just stops re-exporting a function nothing calls.** Loading the file does not throw; the only failure mode would be a build that imports `__etSlot` by name, which no current emitter produces.

Keeping a dead export would carry the contract forward and force the next refactor to delete it again. Per the conversation that prompted this cleanup, removing it is the deliberate outcome.

packages/react/runtime/src/element-template/background/instance.ts
```diff
@@ -221,45 +235,29 @@ export class BackgroundElementTemplateInstance {
...
-          parent.instanceId,
-          slotId,
-          child.instanceId,
-          collectElementTemplateSubtreeHandleIds(child),
-        );
-        // The removed JS object graph may outlive the detach until GC, so keep
-        // it pending and tear it down on the Snapshot-aligned delayed boundary.
-        markRemovedSubtreeForCurrentCommit(child);
-      }
+    const slotId = child.__slotIndex;
```
Yradex, line 224: I don't think syncing only `child.__slotIndex` is enough here. For a moved existing child, the slot index can already point to the destination slot by the time `removeChild()` runs, so the previous source slot can keep a stale reference.

Can we either track the slot the child was attached under, or resync both/all affected slot arrays after a move?

packages/react/runtime/__test__/element-template/runtime/background/instance.test.ts
```diff
@@ -1056,108 +1029,136 @@ describe('BackgroundElementTemplateInstance Shadow State', () => {
...
+    // Preact may pick a cross-slot DOM cursor as the anchor when a node moves
+    // between slots. The op emits `beforeId = 0` (append-tail), so the local
+    // sibling chain must also append-tail; otherwise filtering the chain by
+    // __slotIndex would yield [incoming, slot0Tail] while main thread sees
+    // [slot0Tail, incoming].
+    parent.insertBefore(incoming, crossSlotAnchor);
+
+    expect(parent.elementSlots[0]).toEqual([slot0Tail, incoming]);
+    expect(parent.lastChild).toBe(incoming);
   });
```
Yradex, line 1626: Can we add coverage for moving an already-attached child between slots? The important assertions are that the destination slot gets the child and the old source slot no longer contains it. It would be good to cover both adjacent and non-adjacent slot changes.

packages/react/runtime/src/element-template/runtime/render/render-to-opcodes.ts
```diff
@@ -252,11 +252,33 @@ function renderCompiledEtHostVNode(vnode, props, context, opcodes) {
...
+      continue;
+    }
+    (namedSlotChildren ??= [])[slotId] = props[name];
+    hasNamedSlotChildren = true;
+  }
+
+  if (hasNamedSlotChildren) {
+    renderEtSlotArray(namedSlotChildren, context, vnode, opcodes);
+  } else {
+    // Legacy ET host nodes still encode dynamic element slots in `children`,
```
Yradex: Do we still need the legacy `children` slot-array fallback here? If ET hosts are now expected to use `$N` props only, I think this fallback can be removed so the renderer has a single input contract.

packages/react/runtime/__test__/element-template/runtime/render/render-to-opcodes.et.test.jsx
```diff
@@ -59,6 +59,46 @@ describe('Element Template renderToOpcodes', () => {
...
+      'first',
+      __OpSlot,
+      2,
+      __OpText,
+      'third',
+      __OpEnd,
+    ]);
+  });
+
+  it('prefers $N named props over legacy children when both are present', () => {
```
Yradex: If we remove the legacy `children` fallback, this coverage should go away too. The renderer tests should focus on the `$N` input shape instead of keeping the old slot-array contract alive.

packages/react/runtime/__test__/element-template/fixtures/hydrate/background-hydrate/_shared.tsx
```diff
@@ -501,12 +485,6 @@ export function runCaseByName(name: string): unknown {
     backgroundElementTemplateInstanceManager.nextId = 0;
 
     const rootInstance = new BackgroundElementTemplateInstance('root');
-    const slot0 = new BackgroundElementTemplateSlot();
-    slot0.setAttribute('id', 0);
-    rootInstance.appendChild(slot0);
-    const slot1 = new BackgroundElementTemplateSlot();
-    slot1.setAttribute('id', 1);
```
Yradex, line 508: This fixture name looks stale now. The setup no longer creates any existing background slot entry, so it is not really testing “existing slots” anymore. Can we either rename it or materialize an actual empty/sparse background slot?

packages/webpack/react-webpack-plugin/test/background-loader.test.ts
```diff
@@ -75,9 +75,10 @@ describe('background loader', () => {
...
       '"@lynx-js/react/element-template/jsx-runtime"',
     );
-    expect(result.code).toMatch(
-      /"@lynx-js\/react\/element-template(?:\/internal)?"/,
-    );
     expect(result.code).not.toContain('"@lynx-js/react/internal"');
+    // Multi-slot lowering no longer needs an `__etSlot` runtime call, so the
+    // ET path is now identified by the `_et_` template id prefix rather than
+    // by an `@lynx-js/react/element-template[/internal]` import.
+    expect(result.code).toMatch(/_et_[a-z0-9]+_/);
```
Yradex, line 85: This test name does not quite match the assertion anymore. It no longer checks for an ET runtime import; it checks that ET transform happened via the generated template id. Maybe rename it, and add a negative assertion for `__etSlot` / element-template internal imports.

packages/react/runtime/src/element-template/background/instance.ts
Yradex: Seems
`root.render(<_et_host $2={<_et_child key="moved" />} />)`
`root.render(<_et_host $0={<_et_child key="moved" />} />)`
results in host slots 
`[ [ '_et_child:3:0' ], <1 empty item>, [ '_et_child:3:0' ] ]`
and the node is not removed from `elementSlots[2]`

Yradex: How about 
```
root.render(<_et_host $1={<_et_child key="moved" />} />)
root.render(<_et_host $0={<_et_child key="moved" />} />)
```

upupming: Tried both shapes against current `main` and added them as regression coverage in `f6712915`. Snapshotting the host's `elementSlots` plus the child `instanceId` after each render:

**`$2` → `$0`**
```
after $2: [<empty>, <empty>, [_et_child#3]]
after $0: [[_et_child#4]]
```

**`$1` → `$0`**
```
after $1: [<empty>, [_et_child#3]]
after $0: [[_et_child#4]]
```

The child's `instanceId` changes (`#3` → `#4`), so Preact is unmounting and fresh-mounting rather than moving. The Lynx Preact fork's `findMatchingIndex` requires `oldVNode._slotIndex === newSlot` for keyed reuse (`preact.module.js:750` / `:758`), which #2664 introduced to disable cross-slot vnode reuse. With that, a same-key child cannot migrate across `$N` indices via `root.render` — the previous source slot is cleared by the regular unmount path, not by any source-slot resync.

`elementSlots` is also a pure getter over the linked-list children + their `__slotIndex` (`instance.ts:64-72`), so there is no shadow state that could leak across re-renders. The defensive case (a child whose `__slotIndex` is mutated between slots without going through Preact) is still covered by the existing `BackgroundElementTemplateSlot Children` unit tests in `instance.test.ts`.

packages/react/runtime/__test__/element-template/runtime/background/instance.test.ts
upupming: Both the adjacent and non-adjacent shapes are already covered:

- `instance.test.ts` `BackgroundElementTemplateSlot Children` describe:
  - `should clear the previous slot index when partId changes after attachment` — 0 → 1 (adjacent)
  - `should move slot children to the new slot index when partId changes` — 0 → 3 (non-adjacent)
  - plus `should detach a moved child from the old slot shadow state when silent reparenting` and `should detach a moved child from the old parent slot shadow state` for the silent / cross-parent variants
- `preact.test.tsx` adds end-to-end coverage that drives the same scenario through Preact diff (`f6712915`): `$2 → $0` and `$1 → $0`. In both cases the assertion is the same one you described — destination slot has the child, source slot is empty.

packages/react/runtime/src/element-template/runtime/render/render-to-opcodes.ts
upupming: Dropped in `f9d07200`. Tracing the path end-to-end:

- `swc_plugin_element_template/lowering.rs:104-118` only emits `$N` attrs and `children: vec![]` for compiled ET hosts.
- `renderCompiledEtHostVNode` now consumes `$N` exclusively (`render-to-opcodes.ts`).
- Snapshot's `children` handling (`renderToOpcodes/index.ts:348-411`) is *siblings semantics* — children inherit the parent's `slotIndex`. ET hosts have no equivalent: the host *is* the slot container, every child needs an explicit slot index. The old fallback interpreted `children` as a positional slot array, which doesn't match snapshot either — so it was both unreachable from the compiler and inconsistent with the parallel system.

The two legacy renderer tests that exercised the children-as-slot-array shape are dropped; renderer coverage on the `$N` contract is preserved (`emits slot opcodes for ET host nodes using $N named props`, `skips empty $N slots and renders only the populated indices`). Full ET suite stays at 100% coverage.

packages/react/runtime/__test__/element-template/runtime/render/render-to-opcodes.et.test.jsx
upupming: Done in `f9d07200` together with the fallback removal — the two `children`-shape tests are gone and replaced with `$N`-shape coverage (`emits slot opcodes for ET host nodes using $N named props`, `skips empty $N slots and renders only the populated indices`).

packages/react/runtime/__test__/element-template/fixtures/hydrate/background-hydrate/_shared.tsx
upupming: Updated in `0d29bf26` to match the name. The fixture now parks a child on the background instance at slot 1 while the serialized payload only ships slot 0, which actually exercises the `Math.max(serializedSlots.length, instance.elementSlots.length)` extension in `hydrate.ts` — the loop visits the background-only slot and synthesizes a create + insert:

```
stream: [
  1, 2, "child", null, [], [],   // createTemplate(child)
  3, -1, 1, 2, 0,                // insertNode(parent=-1, slot=1, child=2, beforeId=0)
]
```

packages/webpack/react-webpack-plugin/test/background-loader.test.ts
upupming: Renamed in `9e6b43cb` to `compiles ET background output to Preact JSX with multi-slot template ids`, which matches both assertions (Preact JSX runtime + multi-slot template id generation). Negative imports now also cover `"@lynx-js/react/element-template"` so the wrapper-path runtimes are fully fenced off. Skipped the `__etSlot` negative assertion since `__etSlot` was fully removed in `470db2ff` and is no longer in any runtime — asserting its absence would be a no-op.

packages/react/runtime/src/element-template/background/instance.ts
```diff
@@ -231,50 +211,32 @@ export class BackgroundElementTemplateInstance {
...
+        collectElementTemplateSubtreeHandleIds(child),
+      );
+      // The removed JS object graph may outlive the detach until GC, so keep
+      // it pending and tear it down on the Snapshot-aligned delayed boundary.
+      markRemovedSubtreeForPostDispatchTeardown(child);
+    } else {
+      if (!isElementTemplateHydrated()) {
+        // Pre-hydration commits have already exposed refs to user effects, so
+        // a local slot removal must detach them even though no native patch exists.
         child.queueRefCleanupForSubtree();
```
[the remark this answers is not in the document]
upupming: Confirmed and addressed in 6f559bdf. Pulled the trailing unconditional `queueRefCleanupForSubtree` into each `if`/`else` arm so it runs exactly once. The pre-hydration arm now also calls it before any `tearDown()` (which wipes `rawAttributeSlots` and would silently zero the cleanup). Added a callback-ref regression test in `instance.test.ts`.

packages/react/runtime/__test__/element-template/runtime/background/instance.test.ts
```diff
@@ -534,11 +567,10 @@ describe('BackgroundElementTemplateInstance', () => {
       const ref = vi.fn(() => cleanup);
       __etAttrPlanMap.view = [0, adaptRefAttrSlot];
       const parent = new BackgroundElementTemplateInstance('view');
-      const slot = new BackgroundElementTemplateSlot();
-      slot.setAttribute('id', 0);
-      parent.appendChild(slot);
+      parent.setAttribute('id', 0);
+      parent.appendChild(parent);
```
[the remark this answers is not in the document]
upupming: Fixed in `bd3b36c2`. The `parent.appendChild(parent)` (and the stray `setAttribute('id', 0)` next to it) were leftover from the `BackgroundElementTemplateSlot` migration — originally `parent.appendChild(slot)` followed by `slot.setAttribute('id', 0)`. Neither is needed for the hydrated-removal ref-cleanup scenario the test covers, so both are dropped. The remaining setup is just `parent → child` with a callback ref on `child`, then `parent.removeChild(child)` post-hydration.

packages/react/runtime/__test__/element-template/runtime/background/instance.test.ts
```diff
@@ -64,35 +64,22 @@ describe('BackgroundElementTemplateInstance', () => {
...
     parent.markMaterializedByHydration();
     child.markMaterializedByHydration();
     globalCommitContext.ops = [];
     child.parentNode?.removeChild(child);
 
-    expect(slot.childNodes).toEqual([]);
+    expect(parent.childNodes).toEqual([]);
     expect(child.parentNode).toBeNull();
-    expect(parent.elementSlots[0]).toEqual([]);
+    expect(parent.elementSlots[0]).toBeUndefined();
```
Yradex, line 81: This file still has a few stale `BackgroundElementTemplateSlot` leftovers even though the runtime no longer exposes that abstraction: the import at the top, the `BackgroundElementTemplateSlot Children` describe block, and similar references in `hydrate.test.ts`, `commit-context.test.ts`, background-hydrate `_shared.tsx`, and the adapter test name. Can we clean these up so the tests describe the new `__slotIndex` model instead of keeping the old slot-wrapper wording around?

After updating this, please run a strict review over the follow-up diff too, since this PR has had a few rounds of test/contract cleanup already.

packages/react/runtime/__test__/element-template/debug/alog.test.ts
```diff
@@ -104,6 +104,19 @@ describe('ElementTemplate alog helpers', () => {
...
+  it('skips sparse element slots when printing the background tree', () => {
+    const root = new BackgroundElementTemplateInstance('root');
+    const child = new BackgroundElementTemplateInstance('view');
+    child.__slotIndex = 1;
+    root.appendChild(child);
+
+    const output = printElementTemplateTreeToString(root);
+
+    expect(output).toContain(`view#${child.instanceId}`);
+    expect(output).toContain(`elementSlots[1]: [${child.instanceId}]`);
```
Yradex, line 115: `elementSlots` is a getter now, so the assignments in the previous test (`card.elementSlots[0] = ...` / `[1] = ...`) only mutate a temporary array returned by the getter. The test still passes because `appendChild(text)` already makes slot 0 visible, but those lines no longer set up real background state. Can we drop them, and set `text.__slotIndex` explicitly if the test needs a non-zero/sparse slot?

After updating this, please run a strict review over the follow-up diff too, since this PR has had a few rounds of test/contract cleanup already.

packages/react/runtime/src/element-template/background/instance.ts
```diff
@@ -71,6 +61,16 @@ export class BackgroundElementTemplateInstance {
     return nodes;
   }
 
+  get elementSlots(): BackgroundElementTemplateInstance[][] {
```
Yradex, line 64: `elementSlots` is now a derived getter, so each read walks the child sibling list and allocates a new grouped array. `hydrateInstance` reads it for `length`, reads it again inside the slot loop, and then `hydrateElementSlot` reads it once more for the same slot.

Can we cache `const backgroundElementSlots = instance.elementSlots` once in `hydrateInstance`, use that for `slotCount` / `backgroundSlot`, and pass the slot children into `hydrateElementSlot`? That keeps the new derived model, but avoids repeated regrouping during hydration.

After updating this, please run a strict review over the follow-up diff and check whether there are any other repeated `elementSlots` getter reads with the same issue.

packages/react/runtime/src/element-template/background/instance.ts
```diff
@@ -91,7 +91,7 @@ export class BackgroundElementTemplateInstance {
     }
   }
 
-  emitCreate(): void {
+  emitCreate(elementSlots: BackgroundElementTemplateInstance[][]): void {
```
Yradex: Passing `elementSlots` into `emitCreate` avoids recomputing the derived getter during recursive create, but it also makes the normal method easy to call with stale or unrelated slot data. Most call sites now have to spell `instance.emitCreate(instance.elementSlots)`, which leaks an internal performance detail to every caller.

Can we keep `emitCreate()` as the safe default API, and move the precomputed-slot path into a private/internal helper used by the recursive create flow? That would keep the optimization while keeping regular callers responsible only for “emit create”, not for supplying the correct slot snapshot.

packages/react/runtime/src/element-template/background/instance.ts
```diff
@@ -231,52 +211,35 @@ export class BackgroundElementTemplateInstance {
...
+      // it pending and tear it down on the Snapshot-aligned delayed boundary.
+      markRemovedSubtreeForPostDispatchTeardown(child);
+      child.queueRefCleanupForSubtree();
+    } else {
+      // Pre-hydration commits and post-hydration unmaterialized subtrees have
+      // both exposed refs to user effects (via the pre-hydration flush), so a
+      // local removal must detach them even though no native patch exists.
+      // Run the ref cleanup before any `tearDown()` below — tearDown clears
+      // `rawAttributeSlots`, which `queueRefCleanupForSubtree` depends on.
+      child.queueRefCleanupForSubtree();
```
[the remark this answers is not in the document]
upupming: Fixed in `3b887b35`. Gated the cleanup by the same predicate that `setAttribute` uses to decide whether to publish the attach — `!isElementTemplateHydrated() || !child.needsMainThreadCreate()` — so callback / object refs only see a detach when a corresponding attach actually fired. The hydrated-unmaterialized-tear-down-before-insert path now just runs `tearDown()` and skips `queueRefCleanupForSubtree`. Regression test in `instance.test.ts`: `does not detach refs that never attached on a post-hydration unmaterialized subtree`.

packages/react/runtime/src/element-template/background/instance.ts
```diff
@@ -175,39 +175,19 @@ export class BackgroundElementTemplateInstance {
...
-        );
-      }
+    if (silent || !this.canEmitUpdatePatch()) {
       return;
     }
 
-    if (silent) {
-      return;
-    }
+    const beforeId = (beforeChild && beforeChild.__slotIndex === child.__slotIndex) ? beforeChild.instanceId : 0;
```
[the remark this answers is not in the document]
upupming: The interleaved shape `[slot1, slot0, slot1]` is not actually reachable here. The Lynx Preact fork that drives `BackgroundElementTemplateInstance` keeps the linked-list children sorted by `__slotIndex` by construction:

- When a host vnode has `$N` props, the fork builds `newChildren = [$0, $1, ...]` in numeric order and walks them with `slotIndex === true` in `diffChildren` (preact.module.js:1338-1349). Each child gets `_slotIndex = i` (its index in `newChildren`), and the DOM insertion uses the next not-yet-processed old child as the anchor, so children land in slot order.
- `#2664` already forbids cross-slot keyed reuse — `findMatchingIndex` requires `oldVNode._slotIndex === newSlot` (preact.module.js:750 / :758). A keyed child cannot migrate between `$N` indices on re-render; the source slot unmounts and the destination slot fresh-mounts. (We covered this end-to-end in `preact.test.tsx` with the `$2 → $0` / `$1 → $0` cases on commit `dbb90ba7`.)

Together those two invariants mean the linked list is always `[slot0 children..., slot1 children..., slot2 children..., ...]` with no interleaving. So when `beforeChild.__slotIndex !== child.__slotIndex` reaches this line, `beforeChild` is necessarily the first child of a *later* slot, which means `child` is being inserted at the tail of its own slot — `beforeId = 0` (append) is the correct destination for the main-thread slot list. There is no later same-slot sibling whose id we are throwing away.

The earlier behavior (using `beforeChild.instanceId` unconditionally) is what was actually broken: the runtime saw a `referenceId` belonging to a different slot, which `runtime/patch.ts` resolves to a node outside the destination slot. The same-slot guard fixes that and the append-fallback is the right answer for the cross-slot case.

packages/react/runtime/src/element-template/background/instance.ts
upupming: To keep it simple, I just simple it by iterate the linked list on demand.

Yradex (approved):

gaoachao (approved):

colinaaa (approved):

PR #2683, opened 2026-05-20
feat(create-rspeedy): print fullscreen URL hint in dev server output

KealanAU:
Add a pluginLynxFullscreenHint Rsbuild plugin to the React-TS template that wraps server.printUrls and appends a '∟ No nav' entry with ?fullscreen=true under each Lynx bundle URL. Tapping the variant opens the bundle in LynxExplorer with the in-app nav chr
CleanShot 2026-05-20 at 16 03 22@2x
ome stripped.

## Summary by CodeRabbit

* **New Features**
  * Dev server now prints an extra "∟ Fullscreen" hint for Lynx bundle URLs (adds ?fullscreen=true). Enabled by default; can be disabled via a new optional fullscreen config.
  * QR output now includes a fullscreen variant in the rotation, and the dev console shortcut switches to the fullscreen flow (opens with in-app chrome stripped).
* **Tests**
  * Added tests for fullscreen hint behavior, query‑preservation, parsing fallback, and multi‑entry handling.
* **Chores**
  * Added a changeset entry for a patch release.

[Review Change Stack](https://app.coderabbit.ai/change-stack/lynx-family/lynx-stack/pull/2683?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

## Checklist

- [* ] Tests updated (or not required).
- [* ] Changeset added, and when a BREAKING CHANGE occurs, it needs to be clearly marked (or not required).


KealanAU:
I did this as a template, but I can change to instead put this in dev.plugin.ts

packages/rspeedy/create-rspeedy/template-react-ts/lynx.config.ts
```diff
@@ -1,9 +1,40 @@
-import { defineConfig } from '@lynx-js/rspeedy'
+import { defineConfig, type RsbuildPlugin } from '@lynx-js/rspeedy'
 
 import { pluginQRCode } from '@lynx-js/qrcode-rsbuild-plugin'
 import { pluginReactLynx } from '@lynx-js/react-rsbuild-plugin'
 import { pluginTypeCheck } from '@rsbuild/plugin-type-check'
 
+const pluginLynxFullscreenHint = (): RsbuildPlugin => ({
```
luhc228: Hello, @KealanAU . Thank you for your contribution. Could you please implement this feature in pluginQRCode? So we can reuse this in other project easier.

KealanAU: @luhc228  Thanks so much for having a look and your feedback. I've just refactored to do that. Please give me a heads up if there are any issues. Also feel free to push your own changes if you think i've missed anything :)

.changeset/feat-lynx-fullscreen-hint.md
```diff
@@ -0,0 +1,7 @@
+---
+"@lynx-js/qrcode-rsbuild-plugin": patch
+---
+
+feat(qrcode-rsbuild-plugin): print fullscreen URL hint in dev server output
+
+Append a `∟ No nav` entry with `?fullscreen=true` under each Lynx bundle URL printed by the dev server. Tapping the variant opens the bundle in LynxExplorer with the in-app navigation chrome stripped. Disable via the new `fullscreen: false` option.
```
luhc228: No Nav -> FullScreen would be better.

luhc228: Will this url have qr code printed?

luhc228: We modified the url by [`schema`](https://github.com/lynx-family/lynx-stack/blob/60bdcd4abb3689d809a04c0338cb55c0fcf2e825/examples/react/lynx.config.js#L13) option before. I am not sure which design is better

KealanAU: Yes, fair. I'd missed the part about schema. I've updated and allowed it in as a second option now in the schema toggle
https://github.com/user-attachments/assets/e896fb63-5266-4735-b4df-bcdef1522671

packages/rspeedy/plugin-qrcode/src/index.ts
```diff
@@ -98,15 +115,35 @@ export function pluginQRCode(
   options?: PluginQRCodeOptions,
 ): RsbuildPlugin {
   const defaultPluginOptions = {
-    schema: (url) => ({ http: url }),
+    schema: (url) => ({ nav: url }),
```
luhc228: How about using `fullscreen` instead of `nav`?

packages/rspeedy/plugin-qrcode/src/index.ts
```diff
@@ -98,15 +115,35 @@ export function pluginQRCode(
...
+    fullscreen: true,
```
luhc228: I prefer the default value of fullscreen to be `false` because not all apps need fullscreen, and changing it would also break the previous behavior.

examples/react/lynx.config.js
```diff
@@ -7,12 +7,7 @@ const enableBundleAnalysis = !!process.env['RSPEEDY_BUNDLE_ANALYSIS'];
 export default defineConfig({
   plugins: [
     pluginReactLynx(),
-    pluginQRCode({
-      schema(url) {
-        // We use `?fullscreen=true` to open the page in LynxExplorer in full screen mode
-        return `${url}?fullscreen=true`;
-      },
-    }),
+    pluginQRCode(),
```
upupming: Shall we set `fullscreen: true` here since its default value is false

KealanAU: @upupming , Sure I've added that. it was mainly to unlock for downstream - https://vyui.dev/

luhc228 (approved):

PR #2694, opened 2026-05-22
feat(react): support ET typed page update flow

Yradex:
## Summary by CodeRabbit

* **New Features**
  * Added reload versioning system to track template refresh cycles.
  * Introduced typed element template support for improved page rendering control.
  * Added page data reset capability for `updatePage` operations.

* **Tests**
  * Extended test coverage for element template lifecycle, reload behavior, and page data updates.

[Review Change Stack](https://app.coderabbit.ai/change-stack/lynx-family/lynx-stack/pull/2694?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

## Overview

Element Template now owns the page/update lifecycle through typed ET primitives instead of relying on legacy page append semantics. This change aligns the typed element protocol, page root handling, immediate `updatePage`, and `reloadTemplate` flow so main-thread create, background hydrate, and later update events all use the same ET handle and reload-version boundaries.

## Key Points

- Adds a `createTypedElement` update command so the main thread can create typed native nodes through `__CreateTypedElementTemplate(type, attributes, slots, uid, options)`.
- Replaces legacy ET page handling with a typed page container created as `__CreateTypedElementTemplate("page", null, null, "0", null)`, then inserts or removes React roots through page slot `0`.
- Implements immediate ET `updatePage` and `reloadTemplate` support by sharing initData merge/reset logic, rebuilding main/background ET state on reload, and resetting template/event/runtime state at the reload boundary.
- Carries `reloadVersion` through hydrate and update commit payloads so stale events from before a reload are ignored while the previous array-shaped hydrate payload remains accepted.

## Runtime Contract

The update stream can now carry typed creates:

```ts
[
  ElementTemplateUpdateOps.createTypedElement,
  handleId,
  type,
  attributes,
  elementSlots,
  options,
]
```

`elementSlots` are resolved from handle ids to `ElementRef[][]` before native create. For options, this phase only resolves list-specific `options.listChildren` handle references; generic nested option refs remain deferred.

Hydrate/update payloads can now be wrapped as:

```ts
{
  instances: SerializedEtNode[],
  reloadVersion?: number,
}
```

The background listener still accepts the old array payload for compatibility. Serialized typed nodes are modeled in the protocol, but hydrate matching remains compiled-node only for now; typed roots are reported and dropped until typed hydrate support lands.

## Checklist


packages/react/runtime/src/core/lynx-page-data.ts
```diff
@@ -0,0 +1,15 @@
...
+import { isEmptyObject } from '../utils.js';
+
+export function applyUpdatePageData(data: unknown, options?: Pick<UpdatePageOption, 'resetPageData'>): void {
+  if (options?.resetPageData) {
+    lynx.__initData = {};
+  }
+
+  if (typeof data == 'object' && data !== null && !isEmptyObject(data)) {
+    lynx.__initData ??= {};
+    Object.assign(lynx.__initData, data);
```
HuJean, line 13: Need to confirm whether this would trigger an update. I remember it used to be a new reference.

packages/react/runtime/src/element-template/native/index.ts
```diff
@@ -44,6 +45,7 @@ function init(): void {
     lynxCoreInject.tt.publishEvent = publishEvent;
     lynxCoreInject.tt.publicComponentEvent = publicComponentEvent;
     lynxCoreInject.tt.updateCardData = updateCardData;
+    lynxCoreInject.tt.onAppReload = reloadBackground;
```
HuJean, line 48: Maybe in the future we could add a common init.

Yradex: Sure

HuJean (approved):

PR #2715, opened 2026-05-26
fix(template-webpack-plugin): always inline lazy bundle background

upupming:
## Summary

A lazy bundle (`appType: "DynamicComponent"`) should use `lynx.requireModule` instead of `lynx.requireModuleAsync` to load its bts. Otherwise it will fail to get exports when `installChunk`.

### Fix

For a lazy bundle (`appType === 'DynamicComponent'`), the bts is **always inlined** regardless of `inlineScripts`. The option still applies to `card` templates.

## Tests

- **Artifact level** (`template-webpack-plugin`): new `inline-scripts/lazy-bundle-split` case — a split lazy bundle with a non-matching `inlineScripts` regex; asserts every split background chunk is inlined (`requireModule`, no `requireModuleAsync`). Updated `external` / `inline-fn` / `inline-object` / `inline-regex` which previously asserted the externalized (broken) shape.
- **Execution level** (`plugin-react`): new `lazy.test.ts` case builds a real lazy bundle via the rspeedy pipeline with `inlineScripts: false`, then executes the generated `app-service.js` through a `tt`/`lynx` mock and asserts the bundle is available synchronously. Verified to fail without the fix (reproduces `lynx.requireModuleAsync(".../__background.js")`, the production failure shape).

## Notes

- Behavior change to the public `inlineScripts` option: it is now a no-op for lazy bundles (card behavior unchanged).

## Summary by CodeRabbit

* **Bug Fixes**
  * Lazy bundle background chunks are no longer externalized by the inline-scripts matcher; they are always inlined into the service bundle to prevent runtime load failures.

* **Tests**
  * Added regression tests covering lazy-bundle inlining and split-bundle scenarios.
  * Updated existing tests to assert synchronous background module loading and validate the inlined service bundle.

[Review Change Stack](https://app.coderabbit.ai/change-stack/lynx-family/lynx-stack/pull/2715?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/.changeset/inline-lazy-bundle-bts.md
+++ b/.changeset/inline-lazy-bundle-bts.md
@@ -1,0 +1,7 @@
+---
+"@lynx-js/template-webpack-plugin": patch
+---
+
+Always inline a lazy bundle's background (bts) chunk.
+
+A lazy bundle (`appType: "DynamicComponent"`) runs its background synchronously when the bundle is required, so its bts must be inlined into `app-service.js`. Previously a non-matching `inlineScripts` matcher could externalize it via `lynx.requireModuleAsync`, leaving the module unavailable at `installChunk` time and breaking the bundle. The bts of a lazy bundle is now always inlined regardless of `inlineScripts`; the option still applies to card templates.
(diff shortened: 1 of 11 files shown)


upupming:
@codex review
@coderabbitai review

colinaaa (approved):

PR #2755, opened 2026-05-29
feat(rspeedy): lower let/const to var in output for QuickJS parse speed

upupming:
## Summary

Convert every `let`/`const` in the build output to `var` (QuickJS parses `var` faster). Two halves:

- **User source** — add the SWC `transform-block-scoping` pass to both the background and main-thread layers, prepended **on top of** the existing target baseline. It is deliberately kept **outside** `getESVersionEnvInclude` (which mirrors exactly what `jsc.target` lowered, and is locked by a guard test), since `jsc.target` keeps `let`/`const`.
- **Bundler-generated code** — set rspack `output.environment.const = false`, so the runtime/wrapper code rspack emits also uses `var`.

This is an **intentional output change** (var instead of let/const), applied as a rspeedy-core default for all builds.

## Verification

- Direct SWC check with the exact env config: `const a = 1; let b = 2` → `var a = 1; var b = 2`.
- Unit tests: `transform-block-scoping` present in both layers' `env.include`; `output.environment.const === false`; snapshots updated. The `getESVersionEnvInclude` ↔ `jsc.target` equivalence guard still passes (block-scoping is added separately).
- `examples/react` builds clean.

## Test plan

- [ ] `turbo build` (example builds) + `turbo api-extractor`
- [ ] Vitest (rspeedy core + react plugin)
- [ ] `changeset status` shows `@lynx-js/rspeedy` + `@lynx-js/react-rsbuild-plugin` minor

## Performance validation (empirical)

Measured `var` vs `const`/`let` on the **real Lynx engine (PrimJS, via `benchx_cli`)** — same machine, var and const/let bundles run **interleaved (N=30)** to cancel timing drift; walltime read from the perfetto trace via `trace_processor`:

| walltime metric | const/let (min / median) | var (min / median) | var faster |
| --- | --- | --- | --- |
| `executeLoadedScript` — chunk top-level execution (the `LoadScript` region) | 28.98 / 30.18 ms | **28.56 / 29.88 ms** | −1.5% / −1.0% |
| `prepareAndEvalScript` — prepare + eval | 7.35 / 7.76 ms | **7.15 / 7.57 ms** | −2.8% / −2.5% |

Plus isolated **parse+compile** on QuickJS (PrimJS's parent engine; `std.evalScript(compile_only)`, 1.2 MB minified, identical SWC codegen, N=120, reproduced twice): `var` is **~3–5% faster (min)**.

So in **walltime** — what actually affects on-device load time — `var` is faster across parse, eval, and full chunk execution (~1–5%). Modest but consistent.

### On the CodSpeed check

CodSpeed flags `…_LoadScript −7%`, but that is an **instruction-count** number, not walltime. Those `*_LoadScript` benchmarks come from `benchmark/react/plugins/pluginScriptLoad.mjs`, which injects `Codspeed.startBenchmark()/stopBenchmark()` around each chunk's top-level execution (gated on `typeof Codspeed !== "undefined"`, so they only exist under CodSpeed). CodSpeed counts **instructions** in that region; the **same region measured locally in walltime** (`executeLoadedScript`, table above) is *faster* for `var`. In other words `var` runs slightly **more instructions but in less time** — the instruction-count proxy diverges from real time here. The reported run was also cross-environment (base/head on different CPUs and memory, with an implausible `fib +28%` swing), which further reduces its reliability. The flagged regressions can be acknowledged on CodSpeed.

## Summary by CodeRabbit

* **New Features**
  * Build output now defaults to using var instead of const/let across runtime and main/background layers to improve QuickJS parsing performance.
  * The build toolchain now applies a block-scoping transform by default; configs can still override to retain let/const.

* **Tests**
  * Added and updated tests verifying lowering behavior and that user overrides for environment/config options are respected.


packages/rspeedy/core/src/plugins/output.plugin.ts
```diff
@@ -38,6 +45,7 @@ export function pluginOutput(options?: Output): RsbuildPlugin {
               },
             },
             config,
+            lowerToVar,
```
colinaaa: Instead of brute force overriding the user-provided config, should we make `lowerToVar` as the default?

E.g.:

```js
mergeRsbuildConfig(lowerToVar, config)
```

colinaaa: Or we may add `tools.rspack.output.environment.const: false` to `defaults.ts`.

colinaaa (dismissed):

packages/rspeedy/core/src/plugins/output.plugin.ts
upupming: Fixed in 562f2a2 — moved `lowerToVar` to be the first arg of `mergeRsbuildConfig` so user-provided `tools.rspack.output.environment.const` overrides it. Added a regression test (`user can opt out of const/let lowering`).

colinaaa (approved):

PR #2784, opened 2026-06-03
feat(template): enable syncXElementRegistry in page config by default

upupming:
## Summary

Emit `syncXElementRegistry: true` in the generated page config by default. This lets the Lynx runtime sync the XElement registry during page setup, so the open-source `<input>` / `<textarea>` resolve to the new XElement implementation (`x-input-ng` / `x-textarea-ng`) instead of the deprecated legacy elements (which currently surface a runtime warning: `A deprecated <input> is created!`).

The flag is added alongside the existing always-on runtime page-setup flags (`useNewSwiper`, `enableNewIntersectionObserver`, `enableNativeList`) in `LynxTemplatePlugin`'s `sourceContent.config`.

## Background

The `syncXElementRegistry` page config option was added to the template assembler / Lynx core in lynx-family/lynx#6773. This PR makes rspeedy's build output opt into it by default so downstream pages pick up the new element implementation without manual config.

## Changes

- `packages/webpack/template-webpack-plugin/src/LynxTemplatePlugin.ts` — add `syncXElementRegistry: true` to the runtime page config.
- Update the `defaults/config` snapshot.
- Changeset (`minor` for `@lynx-js/template-webpack-plugin`).

## Notes

- The upstream default is `false`; this PR flips it on for rspeedy output. The runtime ignores the key on versions that don't support it, consistent with how the sibling flags are handled.
- Dependent packages don't need manual changeset entries: `react-rsbuild-plugin` (the user-facing consumer) gets its internal dep range updated automatically (`updateInternalDependencies: patch`), and the peer dependents stay within their already-declared `^0.12.0` ranges (`onlyUpdatePeerDependentsWhenOutOfRange: true`).

## Test

`pnpm rstest run cases` in `template-webpack-plugin` → 345 passed.

## Release Notes

* **Chores**
  * Registry synchronization is now enabled by default during page setup. The `<input>` and `<textarea>` elements have been updated to use newer implementations in place of deprecated legacy versions.


MoonfaceX:
strong lgtm

luhc228 (approved):

PR #2794, opened 2026-06-04
fix(genui-a2ui-prompt): serialize build:api after build to avoid rspack cache race

upupming:
## Summary

`pnpm turbo api-extractor` can flake on `code-style-check` / `test-api` with:

```
thread 'tokio-N' panicked at crates/rspack_storage/src/filesystem/db/transaction/mod.rs:53:7:
Transaction already in progress by process M:rslib-node in directory '.../packages/genui/a2ui-prompt/node_modules/.cache/rspack/<hash>/.temp'
Aborted (core dumped)
[ELIFECYCLE] Command failed with exit code 134.
```

The race: `@lynx-js/genui#api-extractor` (in `packages/genui/turbo.json`) depends on both `//#build` *and* `@lynx-js/genui-a2ui-prompt#build:api`. The build chain triggers `genui-a2ui-prompt#build`, and at the same time turbo schedules `genui-a2ui-prompt#build:api`. Both scripts are literally `rslib build`, both produce the same rspack content hash, and both write to the same `node_modules/.cache/rspack/<hash>/.temp` directory. rspack's own filesystem-storage transaction lock catches the conflict and aborts the second process with `SIGABRT` (exit 134).

Fix: add `dependsOn: ["build"]` to `build:api` in `packages/genui/a2ui-prompt/turbo.json` so turbo serializes the two. `build` runs first, releases the rspack transaction on exit; `build:api` then runs against a warm content-addressed cache.

This is independent from the api-extractor lock TOCTOU fix in #2780 — that one fixed the JSON lockfile race in `run-api-extractor.mjs`; this one is rspack's internal cache directory, surfaced by turbo's task graph rather than my lock script.

Scoped to just `genui-a2ui-prompt` because that's the only package where I have a panic trace. `openui`/`a2ui`/`a2ui-catalog-extractor` share the same `build`/`build:api` shape but use `tsc` not `rslib`, so they don't hit the rspack lock. If they start flaking we can extend the same dep.

## Test plan

- [ ] `code-style-check` (which runs `pnpm turbo api-extractor`) passes on this PR.
- [ ] No regression in `test-api` (which runs `pnpm turbo api-extractor -- --local`).
- [ ] Local: `pnpm turbo api-extractor --filter=@lynx-js/genui` still produces the genui api reports without panic.

## Summary by CodeRabbit

* **Chores**
  * Updated build task dependencies to enforce correct execution order within the build pipeline.


upupming:
Self-note from running this on PR #2793:

`test-api` flips to a different failure on this branch: `@lynx-js/genui-a2ui:build:api` runs `tsc` and errors with `Cannot find module '@lynx-js/genui-a2ui-prompt' or its corresponding type declarations`.

Cause: the `a2ui-prompt:build:api → build` dep I added pushes a2ui-prompt's dist generation later in the schedule, but `a2ui:build:api` still has `dependsOn: []`, so it starts immediately and runs `tsc` before a2ui-prompt's `.d.ts` is on disk.

So this fix is only half right — it kills the same-package rspack race in a2ui-prompt, but breaks the cross-package consumption order (a2ui imports from a2ui-prompt). The proper fix needs to cascade across packages too — either:
- add `^build:api` to every genui-sub `build:api` so they wait on upstream sub builds, or
- drop `build:api` entirely and rewire `genui#api-extractor.dependsOn` to use the standard `build` task.

Pausing this PR; will rework before un-drafting.

gaoachao (approved):

PR #2800, opened 2026-06-04
feat(openui): add more catalog components

jonathanfilbert:
## Summary

Expands the OpenUI component catalog with 7 new primitives spanning inputs and media, and wires them into the library, renderer styles, and playground so they're previewable end-to-end. Follow-ups wire the v0.5 `ActionPlan` shape through the catalog and renderer, gate primary actions on form validation, forward named form state with supported steps, and explicitly report unsupported runtime-dependent steps in `OpenUiRenderer` instead of silently dropping them.

### Demo
https://github.com/user-attachments/assets/f4e16759-fc39-4c26-96fe-26bae0186a43

### New components

| Component | Category | Backed by | Notes |
| --- | --- | --- | --- |
| `CheckBox` | inputs | `@lynx-js/lynx-ui` `Checkbox` | Local visual state on tap; optional `name` writes selected values into form state for submitted payloads. |
| `RadioGroup` | inputs | `@lynx-js/lynx-ui` `Radio`/`RadioGroupRoot` | Single-select; optional `name` writes the selected value into form state for submitted payloads. |
| `Slider` | inputs | `@lynx-js/lynx-ui` `Slider` | Continuous value bound to a `value` prop; optional `name` writes raw numeric values into form state, with `max` defaulting to 100. |
| `TextField` | inputs | `@lynx-js/lynx-ui` `Input` | Single-line text input with optional `placeholder`/`label`; optional `name` writes user-entered values into form state. |
| `Image` | data-display | `<image>` | Maps `fit` → `mode` (`contain`→`aspectFit`, `cover`→`aspectFill`, …); 6 size variants. |
| `Icon` | data-display | inline text + font | 28 Material icon ligatures; font is base64-embedded in `renderer.css`. |
| `Loading` | data-display | `<view>` skeleton | `inline` (single bar) and `block` (two-bar placeholder). |

### Playground coverage

Three new `openui-scenarios` exercise the new primitives:

- **settings-panel** — `Card` + `Image`/`Icon`/`CheckBox`/`Loading` mix.
- **sound-settings** — `Slider`s for volume/bass/treble inside a `Card`.
- **pizza-order** — `TextField`, `RadioGroup`, and `CheckBox` inside a form layout with stable field names for form-bound submit payloads. Extended with `Apartment Number` (`number` variant) and `Building Access Code` (`obscured` variant) fields to exercise the `ActionPlan` path with form-bound state.

Each scenario ships with the raw OpenUI source and the corresponding parsed JSON for the renderer.

### Drive-by fix

- `a2ui-playground/lynx-src/openui/App.tsx`: added `width: '100%'` to the root `scroll-view` so wide content (e.g. the new settings panel) lays out full-width instead of being clipped to the inner column.

## Release Notes

* **New Features**
  * Added seven new components: CheckBox, Icon, Image, Loading, RadioGroup, Slider, and TextField.
  * Introduced form field binding and validation support for input components.
  * New "Inputs" component category to organize input-related components.
  * Support for sequential action steps, enabling more complex user interactions.
  * Three new interactive sample scenarios for testing.

* **Bug Fixes**
  * Fixed scroll view sizing to properly fill available width and height.


gaoachao:
Review notes from my pass on the latest PR head:

- **[P1] ActionPlan steps are silently dropped in `OpenUiRenderer`.** In `packages/genui/openui/src/core/renderer.tsx`, the ActionPlan switch only handles `ToAssistant` and `OpenUrl`. `Run`, `Set`, and `Reset` currently fall through without doing anything, so generated OpenUI v0.5 actions such as `@Set` state updates or `@Run` query/mutation invalidations will appear to fire but have no effect. Please either implement these steps here as well, or explicitly report unsupported step types instead of silently ignoring them.

- **[P1] New input components do not propagate user-entered values into ActionPlan/form state.** `TextField`, `CheckBox`, `RadioGroup`, and `Slider` only keep local component state. They call `triggerAction(next, undefined, action)`, but the ActionPlan path in the renderer uses the step's `message` and only attaches state already present in the form context. Since these inputs never write to the form store, final actions such as the pizza-order submit button will not receive the actual address/radio/slider/checkbox values the user selected. These components should either register/update form fields via the existing form state hooks or include the changed value in the emitted ActionPlan event in a well-defined way.

- **[P2] Local input state is not synchronized when props update.** The new components initialize local state from `props.value` once (`useState(props.value...)`) but do not sync on later prop changes. Because the OpenUI renderer preserves component instances by stable statement keys during incremental/streaming rendering, values that arrive or change after the first render can leave the UI showing stale state. Consider syncing with `useEffect` or moving these components onto the shared form/store state path.

- **[P2] `Slider` requires `max` even though the implementation documents/defaults it to 100.** The schema has `max: z.number()` while the component code and catalog describe a default max of `100`. Omitting `max` will be rejected before the default can apply; this should likely be `max: z.number().optional()`.

jonathanfilbert:
Hey @gaoachao , thank you for the reviews!

Summary of changes:
- `OpenUiRenderer` **no longer silently drops unsupported ActionPlan steps**. `Run`, `Set`, `Reset`, and unknown step types now report an explicit warning, with a comment explaining why this lightweight renderer path does not execute them directly.
- `TextField`, `CheckBox`, `RadioGroup`, and `Slider` now support optional stable `name` values and write user-entered values into the existing form state hooks, so **submitted ActionPlan payloads can include the selected/input values**.
- **Input components now sync updated `props.value` into local/form state** while the field is still untouched, avoiding stale UI during incremental rendering without overwriting active user edits.
- `Slider.max` is now optional to match the documented/default `100` behavior.
- Slider action emission preserves decimal precision by sending the raw value string.

I also **updated the pizza-order scenario with stable field names** so it exercises the form-bound ActionPlan path.

> Review notes from my pass on the latest PR head:
> 
> * **[P1] ActionPlan steps are silently dropped in `OpenUiRenderer`.** In `packages/genui/openui/src/core/renderer.tsx`, the ActionPlan switch only handles `ToAssistant` and `OpenUrl`. `Run`, `Set`, and `Reset` currently fall through without doing anything, so generated OpenUI v0.5 actions such as `@Set` state updates or `@Run` query/mutation invalidations will appear to fire but have no effect. Please either implement these steps here as well, or explicitly report unsupported step types instead of silently ignoring them.
> * **[P1] New input components do not propagate user-entered values into ActionPlan/form state.** `TextField`, `CheckBox`, `RadioGroup`, and `Slider` only keep local component state. They call `triggerAction(next, undefined, action)`, but the ActionPlan path in the renderer uses the step's `message` and only attaches state already present in the form context. Since these inputs never write to the form store, final actions such as the pizza-order submit button will not receive the actual address/radio/slider/checkbox values the user selected. These components should either register/update form fields via the existing form state hooks or include the changed value in the emitted ActionPlan event in a well-defined way.
> * **[P2] Local input state is not synchronized when props update.** The new components initialize local state from `props.value` once (`useState(props.value...)`) but do not sync on later prop changes. Because the OpenUI renderer preserves component instances by stable statement keys during incremental/streaming rendering, values that arrive or change after the first render can leave the UI showing stale state. Consider syncing with `useEffect` or moving these components onto the shared form/store state path.
> * **[P2] `Slider` requires `max` even though the implementation documents/defaults it to 100.** The schema has `max: z.number()` while the component code and catalog describe a default max of `100`. Omitting `max` will be rejected before the default can apply; this should likely be `max: z.number().optional()`.

gaoachao (approved):

PR #2806, opened 2026-06-05
feat(react): support Element Template typed lists

Yradex:
## Overview

This PR adds Element Template support for typed `list`.

The core design is to keep ET list behavior aligned with Snapshot from the native-visible side: native still receives list callbacks and `update-list-info` diffs, while ET owns template holder refs, list item refs, and typed template hydration internally. Normal list updates are incremental diffs, not repeated full `listChildren` snapshots.

## Why

`list` cannot be treated like ordinary ET children:

- list item roots are logical list records, not visible `elementSlots`;
- native asks JS for components through list callbacks such as `component-at-index`;
- list updates must scale to large/infinite children, so pushing the full child list on every render is too expensive;
- hydration can reconcile a serialized list tree against a different background list tree, similar to Snapshot-visible behavior.

This PR adds the ET-specific list state layer needed to bridge those contracts.

### 1. Transform and render shape

A typed ET `<list>` keeps its logical item roots in list-specific data instead of normal element slots.

- Runtime opcodes collect list item root handles separately from normal children.
- List item platform metadata is extracted per item and stored with the item record.
- Typed list serialization uses `options.listChildren`; typed list `elementSlots` remain empty.
- Invalid ET list children are treated as internal contract violations in DEV rather than adding hot-path production checks.

### 2. Initial main-thread creation

During first render, ET creates the list item roots first, then creates the typed list holder.

The list holder is registered with an `ETListState` keyed by the holder handle id. Each item is registered as an `ETListItemRecord` keyed by item handle id, containing the item ref, template key, platform info, attachment state, and move bookkeeping.

After the holder is created, ET flushes the initial Snapshot-shaped `update-list-info` for the list. This keeps native creation, callback registration, and list diff delivery in a deterministic order.

### 3. Native list callbacks

Typed list callbacks are installed on the list holder:

- `component-at-index`
- `component-at-indexes`
- `enqueue-component`

The callbacks close over the mutable `ETListState`, so they do not need to be recreated for every list mutation.

When native asks for an item, ET inserts the item root ref into the list holder template slot through the ET slot APIs and returns the native unique id from `__GetElementUniqueID(item.ref)`. When native enqueues an item, ET detaches the corresponding ref from the holder.

The attached move path keeps same-list moves correct even when native callback and enqueue ordering is not deterministic.

### 4. Incremental updates

Background list mutations are represented as typed list patch ops:

- `insertTypedListItem`
- `removeTypedListItem`
- `updateTypedListItem`

Main thread resolves those ops through the ET registry, records the change in a pending list update, and flushes one native `update-list-info` diff for the list.

The pending update tracks the pre-update item order, inserted item ids, removed item ids, and affected subtree handles. On flush, it derives:

- `insertAction` for inserted or moved items;
- `removeAction` using old indexes from the baseline order;
- `updateAction` for in-place platform info / type changes.

Removed subtree handles are released only after the native list diff has been written, so native sees the removal before JS tears down the registry state.

### 5. Hydration

Hydration treats typed list children as list-specific state instead of normal slots.

`hydrateListInstance` binds the existing list holder handle, normalizes transient native attributes, and reconciles serialized `options.listChildren` with the background child nodes.

Important behavior:

- child count mismatch is valid and reconciled;
- type mismatch is handled as remove plus insert replacement;
- reorder, insert, remove, and platform info changes go through the same typed list patch path as normal updates;
- list-specific data stays in `options.listChildren`, not `elementSlots`.

This aligns the user-visible behavior with Snapshot without copying Snapshot's internal recycle-pool implementation.

### 6. Destroy and reload

List state is explicitly destroyed when ET subtrees are released, native reload happens, or MTS destroy runs. Late native callbacks after destroy become safe no-ops and do not mutate removed refs.

Nested typed list states are also destroyed when their holder subtrees are released.

## Reviewer Notes

Key files:

- `packages/react/runtime/src/element-template/runtime/list/list.ts`
  - ET list state, item records, native callbacks, pending diff collection, destroy/reload handling.
- `packages/react/runtime/src/element-template/runtime/patch.ts`
  - typed list patch op application and native `update-list-info` flushing.
- `packages/react/runtime/src/element-template/background/hydrate.ts`
  - typed list hydration and reconciliation.
- `packages/react/runtime/src/element-template/runtime/render/render-opcodes.ts`
  - first-render list item collection and typed list state creation.
- `packages/react/transform/crates/swc_plugin_element_template/lib.rs`
  - transform-side typed list shape.

The intended reviewer mental model is:

1. ET still materializes list items as typed template roots.
2. Native still interacts with list through Snapshot-style callbacks and diffs.
3. Runtime state maps ET handle ids to holder/item refs.
4. Updates are reduced to one incremental `update-list-info` per render batch.
5. Hydration reconciles list children through the same patch machinery.

## Validation

Ran the focused ET/runtime validation after the final coverage fix:

- `pnpm -C packages/react/runtime test:et`
- `pnpm -C packages/react/runtime build`
- `pnpm exec dprint check --allow-no-files -- <touched runtime files>`
- `pnpm exec eslint --max-warnings=0 --no-warn-ignored -- <touched runtime files>`
- `git diff --check -- packages/react/runtime`

ET runtime coverage is back to 100% for statements, branches, functions, and lines.

## Release Notes

* **New Features**
  * Added typed list support with incremental item operations (insert, remove, update) for dynamic list rendering.
  * Added list item platform information handling to customize list behavior per item.

* **Tests**
  * Extended test coverage for typed list hydration, patching, and lifecycle management.
  * Added tests for list item mutations and platform info updates.


packages/react/runtime/src/element-template/runtime/list/list.ts
```diff
@@ -0,0 +1,828 @@
...
+        listHandleId: state.listHandleId,
+        listID,
+        cellIndex,
+        itemHandleId: item.uid,
+        referenceHandleId: referenceItem?.uid ?? null,
+      });
+    }
+  }
+
+  const sign = __GetElementUniqueID(item.ref);
```
DwwWxx, line 629: 这里从 item.ref 获取 sign 是否合理？我理解通过 item.ref 获取到的是 TemplateElement 的 sign，但它代表的其实不是使用 "list-item" 标签创建出来的 element

Yradex: That's the current contract. The native ElementTemplate layer will handle the conversion itself.

HuJean (approved):

gaoachao (approved):

PR #2857, opened 2026-06-17
feat: support cloneElement

HuJean:
## Summary by CodeRabbit

* **New Features**
  * Expanded `cloneElement` snapshot cloning to cover more scenarios (including repeated and runtime-created clones), with warnings when compiled snapshot children can’t be replaced.
  * Added `preactCloneElement` as an additional export for Preact-compat cloning.
  * Updated the React example UI with a dedicated clone panel and related styling.
* **Bug Fixes**
  * Improved handling of `key`, `ref`, and `children`, including background cloning patching and hydration.
* **Tests**
  * Updated/expanded snapshot and end-to-end coverage for cloning and `createElement` children-in-props behavior.
* **Documentation**
  * Refreshed published API/export docs for `preactCloneElement`.

## Checklist

### DIFF
--- a/packages/react/runtime/__test__/snapshot/children.test.jsx
+++ b/packages/react/runtime/__test__/snapshot/children.test.jsx
@@ -143,14 +143,18 @@ describe('children api', () => {
         <view>
           <view
             class="a"
-            id="init"
+            id="test"
           />
           <view
             class="a"
             id="test"
           />
-          <view />
-          <view />
+          <view
+            id="test"
+          />
+          <view
+            id="test"
+          />
         </view>
       </page>
     `);
--- a/packages/react/runtime/src/snapshot/lynx/element.ts
+++ b/packages/react/runtime/src/snapshot/lynx/element.ts
@@ -2,12 +2,62 @@
 // Licensed under the Apache License Version 2.0 that can be found in the
 // LICENSE file in the root directory of this source tree.
 
-import { createElement as createElementBackground } from 'preact/compat';
+import { cloneElement as cloneElementBackground, createElement as createElementBackground } from 'preact/compat';
 
-import { createElement as createElementMainThread } from '@lynx-js/react/lepus';
+import { cloneElement as cloneElementMainThread, createElement as createElementMainThread } from '@lynx-js/react/lepus';
+
+import { getCloneSnapshotInfo, getCloneSnapshotType, isCompiledSnapshot } from '../snapshot/utils.js';
 
 type CreateElement = typeof import('preact/compat').createElement;
 type CreateElementParams = Parameters<CreateElement>;
+
+type CloneElement = typeof import('preact/compat').cloneElement;
+type CloneElementParams = Parameters<CloneElement>;
+
+function splitProps(
+  props: unknown,
+  rest: CreateElementParams[2][],
+  initialKey: string | undefined = undefined,
+): {
+  key: string | undefined;
+  children: CreateElementParams[2][];
+  spreadProps: Record<string, unknown>;
+} {
+  let key = initialKey;
+  let children = rest;
+  let spreadProps: Record<string, unknown> = {};
+  if (props && typeof props === 'object') {
+    spreadProps = props as Record<string, unknown>;
+    if ('key' in spreadProps) {
+      const { key: keyValue, ...propsWithoutKey } = spreadProps;
+      key = keyValue as string;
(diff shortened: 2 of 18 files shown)


packages/react/runtime/src/snapshot/lynx/element.ts
```diff
@@ -40,9 +85,79 @@ export const createElement =
...
+      return cloneElementBackground(vnode, props, ...rest);
+    }
+    if (!props && rest.length === 0) {
+      // no props, no children. clone directly
+      const _baseCloneElement = __BACKGROUND__ ? cloneElementBackground : cloneElementMainThread as CloneElement;
+      return _baseCloneElement(vnode, props, ...rest);
+    }
+    const preProps = vnode.props as unknown as {
+      values?: Record<string, unknown>[];
+      $0?: CreateElementParams[2];
```
upupming, line 111: It seems `$0` is only used when in `!isCompiledSnapshot(type)` branch. Cast preProps here will be unsafe.

packages/react/runtime/src/snapshot/lynx/element.ts
```diff
@@ -40,9 +85,79 @@ export const createElement =
...
+    };
+    const preValues = preProps.values ?? [];
+    const resolvedProps = splitProps(
+      props,
+      rest,
+      vnode.key as string | undefined,
+    );
+    const { key, spreadProps } = resolvedProps;
+    let { children } = resolvedProps;
```
upupming, line 120: What if user uses `$0`, it seems be dismissed now

```
const clone = cloneElement(original, {
      className: 'b',
      $0: <text>Props</text>,
    });
```

packages/react/runtime/src/snapshot/lynx/element.ts
```diff
@@ -40,9 +85,79 @@ export const createElement =
...
+      const preSpread = preValues[cloneSpreadIndex] ?? {};
+      values[cloneSpreadIndex] = {
+        ...preSpread,
+        ...spreadProps,
+        __spread: true,
+      };
+    }
+
+    if (children.length > 0) {
+      console.warn('cloneElement from compiled snapshot with children is not supported');
```
upupming, line 155: If a user uses `$x` and just match the holes/slots in the snapshot, shall we support it? But it is too deep into our snapshot mechanism.

packages/react/runtime/__test__/snapshot/children.test.jsx
```diff
@@ -143,14 +143,18 @@ describe('children api', () => {
         <view>
           <view
             class="a"
-            id="init"
+            id="test"
```
upupming, line 146: Is this change as expected?

packages/react/runtime/src/snapshot/lynx/element.ts
```diff
@@ -2,13 +2,63 @@
...
+} {
+  let key = initialKey;
+  let children = rest;
+  let spreadProps: Record<string, unknown> = {};
+  if (props && typeof props === 'object') {
+    spreadProps = props as Record<string, unknown>;
+    if ('key' in spreadProps) {
+      const { key: keyValue, ...propsWithoutKey } = spreadProps;
+      key = keyValue as string;
+      spreadProps = propsWithoutKey;
```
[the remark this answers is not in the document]
upupming: Is there any test against different kind of `key`?

upupming (dismissed):

packages/react/runtime/__test__/snapshot/children.test.jsx
HuJean: yes

packages/react/runtime/src/snapshot/lynx/element.ts
```diff
@@ -40,9 +85,79 @@ export const createElement =
...
+/**
+ * export to users, and in framework would use preact cloneElement directly
+ */
+export const cloneElement =
+  (function(vnode: CloneElementParams[0], props: CloneElementParams[1], ...rest: CloneElementParams[2][]) {
+    const type = vnode.type;
+    if (typeof type !== 'string') {
+      // clone component by preact api
+      return cloneElementBackground(vnode, props, ...rest);
+    }
```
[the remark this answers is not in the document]
HuJean: For component, background and main-thread is same.

Yradex (approved):

PR #2911, opened 2026-07-01
perf(benchmark): bump benchx_cli to the mimalloc release (benchx_cli-202607011810)

upupming:
## What

Bump the benchmark's `benchx_cli` binary from `benchx_cli-202606120920` to **`benchx_cli-202607011810`** (one line in `packages/lynx/benchx_cli/scripts/build_unix.sh`, which downloads the pinned release from lynx-community/benchx_cli).

## Why

`benchx_cli-202607011810` is the first release that links **mimalloc** as the benchmark process allocator ([lynx-community/benchx_cli#7](https://github.com/lynx-community/benchx_cli/pull/7)).

The CodSpeed simulation profiles for these benchmarks show allocator-induced variance from glibc's `ptmalloc` — e.g. `malloc_consolidate` (free-chunk coalescing) appearing as a multi-percent self-time spike in some runs but not others, depending on heap state. mimalloc's sharded thread-local free lists give a uniform, history-independent malloc/free fast path and avoid on-critical-path consolidation, so the numbers should be more reproducible (lower walltime jitter, fewer phantom path differences in simulation).

The override is verified active in the release build (on Linux via strong-vs-weak symbols; on macOS via `MI_OSX_ZONE` + `MI_OSX_INTERPOSE`, confirmed at runtime with `MIMALLOC_SHOW_STATS`).

## Notes

- Only `build_unix.sh` changes — the benchmark runs on Linux/macOS only, there is no Windows benchx_cli build to bump.
- No checksum to update; the script fetches the release asset by tag.

## Test

- The release `benchx_cli-202607011810` has both assets published (`benchx_cli_Linux_x86_64.tar.gz`, `benchx_cli_Darwin_arm64.tar.gz`), so the download resolves.
- CI's benchmark job will exercise the new binary end to end.

## Summary by CodeRabbit

* **Chores**
  * Updated the bundled CLI release used during builds to a newer version.

### DIFF
--- a/packages/lynx/benchx_cli/scripts/build_unix.sh
+++ b/packages/lynx/benchx_cli/scripts/build_unix.sh
@@ -13,7 +13,7 @@ if [ ! -f "package.json" ] || ! grep -q '"name": "benchx_cli"' package.json; the
   exit 1
 fi
 
-LOCKED_VERSION="benchx_cli-202606120920"
+LOCKED_VERSION="benchx_cli-202607011810"
 
 # Check if binary is up to date
 if [ -f "./dist/bin/benchx_cli" ] && [ -f "./dist/bin/benchx_cli.version" ]; then


upupming:
⚠️ **Do not merge yet.** The CodSpeed report shows a ~6x regression across 152 benchmarks (both Simulation and WallTime). Root cause: benchx_cli's release is a debug build (is_debug defaults true), so the newly-linked mimalloc inherited MI_DEBUG=2 + -O0 and ran as a slow debug allocator, replacing glibc's always-optimized malloc.

Fix in flight: lynx-community/benchx_cli#10 builds mimalloc optimized + non-debug (MI_DEBUG=0, NDEBUG, -O2; verified locally, mimalloc now reports `debug level : 0`). Once that merges and a new benchx_cli release is cut, I'll bump `LOCKED_VERSION` here to the fixed release and the regression should disappear.

HuJean (approved):

PR #2920, opened 2026-07-06
fix(rspeedy/react): only enable Yarn PnP resolution when running under the PnP runtime

upupming:
Fixes #2539

## Problem

`rspeedy dev` / `rspeedy build` fails with:

```
error   Cannot find module '@lynx-js/react/jsx-runtime'
    at .../@lynx-js/react-alias-rsbuild-plugin/dist/index.js
```

whenever a stray Yarn PnP manifest (`.pnp.cjs`) exists in **any ancestor directory** of an npm/pnpm project — e.g. left over from a past Yarn Berry install in `$HOME` or a parent folder.

## Root cause

`createLazyResolver` hardcodes `enablePnp: true` (introduced in #1964 to support Yarn PnP). With `enablePnp: true`, Rspack's `ResolverFactory` walks up from the resolve directory looking for a `.pnp.cjs`; once found, resolution goes through the PnP manifest **without falling back to `node_modules`**. A leftover manifest that doesn't know the project's dependencies then makes every resolution fail.

Minimal repro (reproduces on latest Rspeedy 0.15.1 / Rspack 2.0.8, not just the 0.14.3 reported in the issue):

```bash
mkdir demo && cd demo
echo '{"name":"parent-app"}' > package.json
corepack use yarn@4.9.1        # Yarn 4 defaults to the PnP linker → generates .pnp.cjs

npx -y create-rspeedy app --template react-ts
cd app && npm install
npm run build                  # ❌ Cannot find module '@lynx-js/react/jsx-runtime'
rm ../.pnp.cjs && npm run build  # ✅ works
```

Isolated to the resolver itself:

```js
import { rspack } from '@rspack/core'
const { ResolverFactory } = rspack.experiments.resolver
// with a stray .pnp.cjs in an ancestor directory:
new ResolverFactory({ conditionNames: ['import'], enablePnp: true })
  .sync(projectDir, '@lynx-js/react/jsx-runtime')   // ❌ Cannot find module
new ResolverFactory({ conditionNames: ['import'], enablePnp: false })
  .sync(projectDir, '@lynx-js/react/jsx-runtime')   // ✅ node_modules/@lynx-js/react/runtime/jsx-runtime/index.js
```

## Fix

[...]

PR #2922, opened 2026-07-06
test(web-platform): regenerate server-e2e snapshots for scroll-view fading-edge keyframes

upupming:
## Problem

`test-rstest` currently fails on every PR with 3 snapshot mismatches:

```
packages/web-platform/web-core-e2e/server-tests/server-e2e.test.ts
  × executeTemplate should run lepusCode.root from basic-element-x-refresh-view-demo.web.bundle
  × executeTemplate should run lepusCode.root from basic-element-x-overlay-ng-demo.web.bundle
  × executeTemplate should run lepusCode.root from basic-scroll-view.web.bundle
```

Confirmed on at least two independent PR branches today (#2920 and the `feat/react-first-update-data-sync-timing` branch, via Codecov test analytics).

## Root cause

d7a98bd8f (#2898) moved the scroll-view fading-edge `@keyframes` out of the shadow-root html template — `topFading` / `botFading` were renamed to `scrollViewTopFading` / `scrollViewBotFading` and now live in the `@supports`-gated section of `scroll-view.css` — so they no longer appear in the SSR output of `executeTemplate`. The server-e2e snapshots were not regenerated.

The mismatch only surfaces on a **cold build**: with a warm turbo cache the SSR output still comes from pre-#2898 artifacts and the tests pass, which is presumably why it slipped through. Reproducible locally with:

```bash
pnpm turbo build --filter=@lynx-js/web-core-e2e... --force
cd packages/web-platform/web-core-e2e
pnpm rstest run server-tests/server-e2e.test.ts   # 3 failures
```

## Fix

Regenerated the snapshots (`rstest run server-tests/server-e2e.test.ts -u`). The diff is pure deletions: the 5 stale `topFading`/`botFading` keyframes blocks. All 17 server-e2e tests pass against freshly built (`--force`) artifacts.

## Checklist

- [x] Changeset added, and when a BREAKING CHANGE occurs, it needs to be clearly marked (or not required). — test-only change, no changeset needed

### DIFF
--- a/packages/web-platform/web-core-e2e/server-tests/__snapshots__/server-e2e.test.ts.snap
+++ b/packages/web-platform/web-core-e2e/server-tests/__snapshots__/server-e2e.test.ts.snap
@@ -429,37 +429,6 @@ exports[`executeTemplate should run lepusCode.root from basic-element-x-overlay-
                       ::-webkit-scrollbar {
                         display: none;
                       }
-
-                      @keyframes topFading {
-                        0% {
-                          box-shadow: transparent 0px 0px 0px 0px;
-                        }
-                        5% {
-                          box-shadow: var(--scroll-view-bg-color) 0px 0px
-                            var(--scroll-view-fading-edge-length)
-                            var(--scroll-view-fading-edge-length);
-                        }
-                        100% {
-                          box-shadow: var(--scroll-view-bg-color) 0px 0px
-                            var(--scroll-view-fading-edge-length)
-                            var(--scroll-view-fading-edge-length);
-                        }
-                      }
-                      @keyframes botFading {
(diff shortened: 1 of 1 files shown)


upupming:
Note on the CodSpeed check failure: this PR only changes a test snapshot file, so it cannot affect runtime performance. The two flagged regressions (`slower fib(20)` -9.36%, `002-hello-reactLynx__main-thread-processData` -5.83%) are both WallTime-mode measurements — comparing this run against the latest `main` run shows 164 unchanged benchmarks with a handful of ±5-25% WallTime swings in both directions, which is measurement noise.

PupilTong (approved):

PR #2928, opened 2026-07-06
feat: support async external ReactLynx bundle

upupming:
## Summary

Enable consuming ReactLynx (and its component libraries) as an **async external bundle** — fetched at runtime via `lynx.fetchBundle` / `lynx.loadScript` and mounted as a Promise — instead of a synchronous global. This lets a host embed a pinned ReactLynx runtime and load app/component bundles built against it separately.

The async-loading model breaks four sync assumptions across the toolchain; each is fixed in its own package (separate changesets):

### `@lynx-js/lynx-bundle-rslib-config` — `commonjs2` output
Rslib's `cjs` format defaults to `commonjs-static`, which copies named exports one-by-one at startup. When an external is async, the entry becomes an async module whose exports are only available behind a Promise, so that copy reads `undefined`. Switched to `commonjs2` (whole `module.exports` assignment) so a sync entry exports the same namespace and an async entry exports a Promise consumers await.

### `@lynx-js/externals-loading-webpack-plugin` — async subpath externals
`async: true` externals with a subpath `libraryName` (e.g. `['ReactLynx', 'React']`) now pick subpath segments **after** the mounted namespace Promise resolves, instead of reading them synchronously off the pending Promise.

### `@lynx-js/react-refresh-webpack-plugin` — ESM HMR runtime injection
The refresh runtime helpers were injected via `ProvidePlugin`, whose dependency edge does not participate in async-module handling. Under async externals `@lynx-js/react/refresh` becomes an async module, so `isComponent` / `flush` resolved to a pending Promise and `shouldBind` crashed at every module footer. The runtime is now an ESM module injected as a **real import** by the loader, so the importing module awaits it through the normal async-module machinery.

### `@lynx-js/cache-events-webpack-plugin` — async-entry + main-thread caching
The event-caching runtime is keyed on `RuntimeGlobals.startup`, which async-external entries don't require (their startup is a plain `__webpack_require__(entry)`). The plugin now requires `startup` for async entries so their startup becomes a wrappable, Promise-returning function, and:
- **Background**: caches the same `tt` / performance / `globalThis` events as chunk-split entries.
- **Main thread**: caches the first-screen `renderPage` call and replays it against the real `renderPage` once the async ReactLynx runtime has loaded (otherwise the page element is never created and the screen stays blank); shims `processData` as a synchronous passthrough.

## Example
Adds the `examples/react-externals` async variant (`*.react-async.ts` configs) demonstrating the full flow.

## Tests
- `lynx-bundle-rslib-config`: asserts `commonjs2` output (`module.exports =` assignment, no static per-name copy).
- `externals-loading-webpack-plugin`: async-subpath case.
- `react-refresh-webpack-plugin`: HMR snapshots updated for the ESM-import injection.
- `cache-events-webpack-plugin`: async-external-entry integration + main-thread `renderPage`/`processData` unit tests.

## Not verified here
Runtime rendering on a real device (LynxExplorer) — the mechanisms are verified via built-artifact inspection and unit/integration tests, but end-to-end first-screen rendering + cross-thread firstScreen timing should be confirmed on-device.

## Summary by CodeRabbit

* **New Features**
  * Added async externals support via `{ libraryName, async }`, including Promise-based loading for nested subpaths, and React async example build/dev/preview workflows.
* **Bug Fixes**
  * Prevented blank pages for async externals by ensuring main-thread caching waits for async startup readiness and replays the first-screen `renderPage`.
  * Fixed async subpath resolution timing and resolved React Fast Refresh crashes with async ReactLynx externals.
* **API / Configuration**
  * Updated `setupListTransformer` to accept `{ isMainThread }` context for thread-specific cache event setup.

### DIFF
--- a/.changeset/cache-events-async-entry.md
+++ b/.changeset/cache-events-async-entry.md
@@ -1,0 +1,9 @@
+---
+"@lynx-js/cache-events-webpack-plugin": minor
+---
+
+Cache and replay native calls for async-external entries on both threads.
+
+An async-external entry renders startup as a plain `__webpack_require__(entry)` that the event-caching runtime (keyed on `RuntimeGlobals.startup`) never hooked, so calls made while the external ReactLynx bundle was still loading were lost. The plugin now requires `startup` for such async entry chunks: the background thread caches the same `tt` / performance / `globalThis` events as the chunk-split path, and the main thread caches the first-screen `renderPage` and replays it once loaded so the page is not left blank.
+
+`setupListTransformer` now receives a second `{ isMainThread }` argument (it runs once per thread), so custom cache events can be added to a single thread — e.g. `(setupList, { isMainThread }) => isMainThread ? setupList : [...setupList, myEvent]`. Existing single-argument transformers are unaffected.
--- a/.changeset/fix-async-externals-subpath.md
+++ b/.changeset/fix-async-externals-subpath.md
@@ -1,0 +1,5 @@
+---
+"@lynx-js/externals-loading-webpack-plugin": patch
+---
+
+Fix `async: true` externals with a subpath `libraryName` (e.g. `['ReactLynx', 'React']`) resolving to `undefined`. The generated `promise` external now picks all subpath segments after the mounted namespace promise resolves, instead of reading them synchronously off the pending promise.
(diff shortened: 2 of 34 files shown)


upupming:
@coderabbitai review

luhc228 (approved):

PR #2937, opened 2026-07-08
feat(qrcode-rsbuild-plugin): add showQRCode option to registerConsoleShortcuts

lampeggiante:
The fix

Some integrations always launch the Lynx page through a custom link flow and never use the scan path. In those cases, the ASCII QR block printed by `registerConsoleShortcuts` is just terminal noise — but today there is no way to suppress only the QR while keeping URL output and the interactive shortcuts.

The shared shortcut runtime now accepts an internal `showQRCode` option. When `showQRCode` is `false`, `registerConsoleShortcuts` still prints URL(s), keeps entry/schema switching and custom shortcuts working, but skips rendering the ASCII QR code block. Default remains `true`, so existing behavior is unchanged.

`@lynx-js/qrcode-rsbuild-plugin` (private): `registerConsoleShortcuts` `Options` now accept `showQRCode?: boolean`; the initial print and the in-loop reprint both gate `showQRCode(value)` behind `options.showQRCode ?? true`. Unit tests added for both branches; existing behavior unchanged when the flag is omitted.

Demo

This change is intended for downstream consumers that want to preserve URL output and interactive console shortcuts while suppressing the QR block. No public `pluginQRCode` option surface is changed here; this PR teaches the shared shortcut runtime the new field.

Testing

Ran locally with:

- `npx vitest run test/shortcuts.test.ts`

Result: 7 tests passed.
- existing 5 tests unchanged
- +1 new case: renders the QR block by default
- +1 new case: skips the QR block when `showQRCode` is `false`

Summary by CodeRabbit

New Features
- Added an internal `showQRCode` option to the shared console shortcut runtime.

Bug Fixes
- Allows downstream integrations to suppress the QR ASCII block without forking shortcut behavior.

Tests
- Added coverage for both `showQRCode` branches (default / false).

Chores
- Fixed test ordering / isolation so the new `showQRCode` tests do not leak into the existing interactive shortcut case.

## Summary by CodeRabbit

* **New Features**
  * Added a new `showQRCode` option for console shortcuts, letting the QR code display be turned off while keeping URL output and interactive behavior intact.
  * QR code display now remains on by default, so existing behavior is unchanged unless disabled.

* **Tests**
  * Added coverage for the new option, including default QR rendering and the no-QR setting.

## Summary by CodeRabbit

* **New Features**
  * Added a new `showQRCode` option for console shortcut registration.
  * When disabled, the app still prints the URL and supports the interactive switching flow, but skips the ASCII QR code.

* **Bug Fixes**
  * QR code display now respects the new setting consistently during initial output and later updates.

* **Tests**
  * Added coverage for both default QR display and the hidden-QR behavior.

### DIFF
--- a/.changeset/qrcode-showQRCode.md
+++ b/.changeset/qrcode-showQRCode.md
@@ -1,0 +1,7 @@
+---
+"@lynx-js/qrcode-rsbuild-plugin": patch
+---
+
+Add `showQRCode` option to `registerConsoleShortcuts`.
+
+When passed `showQRCode: false`, the shortcut runtime still prints URL(s) and keeps the interactive schema/entry switching, but skips rendering the ASCII QR code. This lets embedders that always launch via a deep link (or wrap the plugin with their own connection flow — e.g. `@byted-lynx/hdt-rsbuild-plugin`) suppress the QR block without forking the shortcut loop. Default remains `true`, so existing behavior is unchanged.
--- a/packages/rspeedy/plugin-qrcode/src/shortcuts.ts
+++ b/packages/rspeedy/plugin-qrcode/src/shortcuts.ts
@@ -16,6 +16,17 @@ interface Options {
   entries: string[]
   schema: CustomizedSchemaFn
   port: number
+  /**
+   * Whether to render the ASCII QR code in the terminal.
+   *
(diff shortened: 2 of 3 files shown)


upupming (approved):

upupming:
@lampeggiante Thanks for your contribution!

PR #2954, opened 2026-07-10
feat(rslib): add enableJsBytecode option, off by default in development

upupming:
## Summary

Adds an `enableJsBytecode` option to `@lynx-js/lynx-bundle-rslib-config` that controls whether main thread chunks of an external bundle are tagged with the `JsBytecode` encoding (i.e. compiled to bytecode by the encoder), and turns it off by default in development.

- **New option** `EncodeOptions.enableJsBytecode` (`defineExternalBundleRslibConfig(config, { enableJsBytecode })`) and `ExternalBundleWebpackPluginOptions.enableJsBytecode`.
- **Development default**: `false` when `NODE_ENV` is `'development'`, `true` otherwise. When disabled, main thread chunks are encoded as plain JavaScript source — faster encoding, and the emitted section stays readable for debugging.
- **`'tasm'` target only**: for the `'web'` target the `JsBytecode` tag never triggers bytecode compilation; it only routes main thread chunks to the `lepusCode` slot, so it is always kept there regardless of the option.

## Why `NODE_ENV` and not rspack `mode`?

`rslib build` always compiles with rspack mode `production` (verified empirically — `NODE_ENV=development` still yields `mode: 'production'`), so `compiler.options.mode` cannot signal development here. `NODE_ENV` matches the existing development signal in this package (the `minify` default of `DEFAULT_EXTERNAL_BUNDLE_LIB_CONFIG`), and is read at encode time so it also works when set right before a build.

## Tests

- `enableJsBytecode: false` → main thread section decodes as plain JS string.
- `NODE_ENV=development` with no option → plain JS string (new dev default).
- `NODE_ENV=development` with explicit `enableJsBytecode: true` → bytecode.
- Existing tests keep asserting the production default stays bytecode.

All 25 package tests pass; API report (`etc/lynx-bundle-rslib-config.api.md`) regenerated.

## Summary by CodeRabbit

* **New Features**
  * Added an `enableJsBytecode` option to control JsBytecode compilation for main-thread chunks in external bundles.
  * Defaults are now environment-aware: disabled in development and enabled otherwise.
  * Web bundle behavior preserves JsBytecode routing information even if the option is not enabled.
* **Documentation**
  * Updated the public API to include `enableJsBytecode` in the relevant options types.
* **Tests**
  * Added coverage verifying output changes across explicit `enableJsBytecode` settings and `NODE_ENV` values.

### DIFF
--- a/.changeset/external-bundle-js-bytecode-option.md
+++ b/.changeset/external-bundle-js-bytecode-option.md
@@ -2,4 +2,4 @@
 "@lynx-js/lynx-bundle-rslib-config": minor
 ---
 
-Add an `enableJsBytecode` option to `EncodeOptions` (and `ExternalBundleWebpackPluginOptions`) that controls whether main thread chunks are compiled to JsBytecode in the emitted external bundle. It defaults to `false` when `NODE_ENV` is `'development'` — keeping main thread chunks as plain JavaScript source for faster encoding and easier debugging — and `true` otherwise. The option only affects the `'tasm'` target; for the `'web'` target the `JsBytecode` tag is routing-only and always kept.
+Add an `enableJsBytecode` option to control compiling main thread chunks to JsBytecode. Defaults to `false` when `NODE_ENV` is `'development'`, `true` otherwise.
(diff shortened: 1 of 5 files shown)


upupming:
@coderabbitai review

luhc228 (approved):

PR #2963, opened 2026-07-11
RFC(react): root.hydrate() + mainThreadRender — dual-thread rendering as ownership, two verbs, and boundaries

Huxpro:
# RFC: Dual-thread rendering as ownership, two verbs, and boundaries

> Layer 2 of the IFR opt-in/opt-out design. Stacked on the L1 branch (`claude/lynx-ifr-opt-design-osbqod`, which ships the `<Background>` boundary) — this PR's diff contains only the L2 work. The PR body doubles as the RFC.

## Motivation

ReactLynx's startup pipeline is configured today by `firstScreenSyncTiming: 'immediately' | 'jsReady' | 'manual'`, plus `markFirstScreenSyncReady()` for the `'manual'` mode. Three problems, in increasing order of depth:

1. **"First screen" is the wrong concept.** It names a *moment* (the first paint), while the thing being configured is a *mechanism* (which thread renders, and when UI ownership transfers). It has a gap with IFR — a product-outcome name — and no connection to the dual-thread model that actually explains the behavior.
2. **"Timing" is the wrong dimension.** The enum hardcodes three schedules of one underlying action instead of exposing the action. `'manual'` + `markFirstScreenSyncReady()` already leaks the truth: what users need is the *verb*, and "timing" should degenerate into ordinary program order.
3. **The knobs don't compose.** Whole-app configuration lives in build config while the interesting decisions (wait for data, wait for an animation, bail after a timeout) live in application code.

## The model

Reading the runtime (`calledByNative.ts`, `firstScreenSync.ts`), the three timings differ in exactly one thing: **when UI ownership transfers from the main thread to the background thread.**

- Before the handover, the main thread owns the UI: native data updates re-render synchronously from snapshots; background render output is invisible and background events are queued.
- After the handover, the background thread owns the UI and drives it through patches.
- **Hydration *is* the handover** — both the mechanism (diff + ID binding) and the ownership-transfer transaction.

So the API should be: **ownership + two verbs (`render`, `hydrate`) + boundaries**, with "first screen" removed from the engineering vocabulary (IFR stays as the product narrative). Retroactively, L1's `<Background>` already follows this scheme.

### `root.hydrate(): Promise<void>` — the handover as a first-class verb

```ts

root.render(<App />);
await root.hydrate();   // resolves when the handover completes
```

- Resolves when the handover completes on the calling thread — on the background thread, after the main thread applied **and acked** the hydration patch; on the main thread, right after the patch is applied.
- Idempotent: repeated calls return the same promise.
- In the automatic presets it is an awaiter; in a held handover it is also the trigger (a background-thread call is forwarded to the main thread, same as `markFirstScreenSyncReady`).

### `root.render(jsx, { hydrate: false })` — splitting the two verbs

```ts
root.render(<App />, { hydrate: false });

const data = await fetchInitialData();
store.prime(data);            // pre-handover state folds into the hydration patch
await root.hydrate();         // first patch carries real data — no skeleton flash
```

`{ hydrate: false }` holds the handover **without any build configuration**: entry code runs on both threads *before* the native `renderPage` call, so the main thread makes its hold-or-sync decision with the flag already in place. "Timing" is now just program order:

```ts
// bounded wait — hydration gating is also background-interactivity gating
await Promise.race([dataReady, timeout(300)]);
await root.hydrate();
```

### `pluginReactLynx({ mainThreadRender: false })` — the config surface shrinks to a compiler fact

The only startup knob that remains build-time is the one the compiler needs to know: whether the main thread renders the first screen at all. `mainThreadRender: false` produces an empty first frame that background hydration fills in — semantically a root-level `<Background>`. Behavior lives in code; build config only encodes what the build may assume (and, in a follow-up, what it may strip from `main-thread.js`).

### Compatibility mapping

| Today | In this model |
|---|---|
| `'immediately'` (default) | automatic handover; unchanged default behavior |
| `'jsReady'` | automatic handover signaled at background render; unchanged |
| `'manual'` + `markFirstScreenSyncReady()` | held handover; `root.hydrate()` triggers it (mark is now `@deprecated`) |
| — | `root.render(jsx, { hydrate: false })`: held handover without build config, composable with any preset |
| — | `root.hydrate()`: awaitable completion in *all* modes |
| `enableSSR` must be `'jsReady'` (runtime throw) | dissolves naturally: hydrate is always an explicit background act |

No protocol changes: the existing `rLynxFirstScreenSyncReady` / `rLynxFirstScreen` / `rLynxChange` handshake and the two-bit state machine in `firstScreenSync.ts` are reused as-is; the new surface is the verb-shaped skin plus a completion promise resolved at the existing ack points (`tt.ts` hydration callback, `updateMainThread` on `isHydration`).

## Semantics that are now documented guarantees

1. **Deferring `hydrate()` defers all background-driven UI**: pre-handover, background renders are invisible and events are queued — every "wait" should be bounded. This is why the awaitable/composable shape (works with `Promise.race`) is a prerequisite for safe use, not a nicety.
2. **`hydrate()` ships the *current* background tree, not the initial one**: pre-handover `setState` folds into the hydration patch (the prime-then-hydrate pattern).
3. **Holding is not free**: each pre-handover data update re-renders the main thread and diffs; `'immediately'` remains the default fast-path for apps that don't need the hold.

## Element Template backend

The ET backend performs the handover automatically. `root.hydrate()` resolves on the background thread when the hydration update is dispatched; a held handover (`{ hydrate: false }`) and main-thread `hydrate()` throw with explicit errors. `mainThreadRender: false` is supported.

## Future layers (not in this PR)

- **`<MainThread>` islands**: first-frame opt-in under a background-first default — requires the compiler to preserve the structural spine on the main thread (L3, same machinery as `<Background>` children-DCE).
- **Per-boundary hydration**: `root.hydrate()` is the root boundary's hydrate; boundaries can grow their own handles (selective hydration).
- **Ownership islands**: subtrees permanently owned by the main thread (MTS state, worklets) — lifting Lynx's per-function thread placement (`'main thread'` / `'background only'` directives) to per-subtree.

## Implementation

**Runtime (`@lynx-js/react`)**
- `src/core/hydration.ts` (new): per-thread hydration-completion deferred.
- `src/lynx-api.ts`: `Root.hydrate`, `RootRenderOptions`, `markFirstScreenSyncReady` deprecated.
- `src/snapshot/lifecycle/event/firstScreenSync.ts`: `disableAutoHydrate` / `isHydrationHeld` on the existing state machine.
- `src/snapshot/lynx/calledByNative.ts`: hold-aware `renderPage` / `updatePage`; `__MAIN_THREAD_RENDER__` gate.
- `src/snapshot/lynx/tt.ts` + `src/snapshot/lifecycle/patch/updateMainThread.ts`: resolve the deferred at the existing ack/apply points.
- Element Template: `client/root.ts`, `native/main-thread-api.ts`, `background/hydration-listener.ts`.

**Build**
- `pluginReactLynx({ mainThreadRender })` → `ReactWebpackPlugin` → `DefinePlugin` `__MAIN_THREAD_RENDER__` (same plumbing as `__FIRST_SCREEN_SYNC_TIMING__`).

**Tests** — full runtime suite green with 100% coverage thresholds (snapshot 94 files / core 12 / element-template 76): hydrate as awaiter (completion at ack, not at diff), held handover via render options (suppresses the `jsReady` auto-signal, keeps main-thread synchronous updates), manual-preset trigger from both threads, idempotency, `mainThreadRender: false` empty-first-frame + hydration fill-in, and the pre-hydration re-render skip. `react-webpack-plugin` (87 tests) and `react-rsbuild-plugin` suites pass (one pre-existing web-platform failure is environmental — missing wasm-bindgen artifact — and reproduces without this diff).

**Benchmark** — vitest bench (`mainThreadRender.bench.jsx`, wired into CodSpeed via the package `bench` script): dual-thread startup with a 200-item tree is ~equal with `mainThreadRender` on vs off (1.10x, within noise) — consistent with the model: total work is conserved; turning IFR off moves cost from the first frame to the hydration patch, it doesn't reduce it.

## Checklist

https://claude.ai/code/session_01LyHx91XciQDJunC9fj6bTe

---
_Generated by [Claude Code](https://claude.ai/code/session_01LyHx91XciQDJunC9fj6bTe)_


packages/react/runtime/src/lynx-api.ts
```diff
@@ -99,18 +176,29 @@ export const root: Root = {
...
     }
   },
+  hydrate: (): Promise<void> => {
+    if (isHydrationHeld()) {
+      if (typeof __BACKGROUND__ !== 'undefined' && __BACKGROUND__) {
+        // The handover happens on the main thread, forward the request to it.
+        lynx.getNativeApp().callLepusMethod(LifecycleConstant.firstScreenSyncReady, {});
+      } else {
+        onFirstScreenSyncReady();
+      }
```
upupming, line 198: This only mark hydrate is ready but cannot ensure hydration is performed. How can we wait the hydration complete?

upupming (commented):
Can `.hydrate()` be used with different `firstScreenTiming`? 
Or we assume user should always use `immediately` and call `.hydrate` differently to achieve the same goal?

packages/react/runtime/src/lynx-api.ts
Huxpro: Good question — the mark/trigger on lines 192–199 only *starts* the handover; the thing you await for completion is the `Promise<void>` returned on line 200 (`getHydrationPromise()`). It resolves when hydration has actually been **performed**, not just marked ready, wired to the existing ack points:

- **Background thread** — resolved in the `callLepusMethod(LifecycleConstant.patchUpdate, …)` ack callback in `snapshot/lynx/tt.ts` (`markHydrationCompleted()`), i.e. after the main thread has applied **and acked** the hydration patch.
- **Main thread** — resolved in `snapshot/lifecycle/patch/updateMainThread.ts` once the hydration patch has been applied (`if (patchOptions.isHydration) markHydrationCompleted()`).

So the wait is just:

```ts
await root.hydrate(); // resolves once hydration has actually completed
```

`rootHydrate.test.jsx › resolves when the hydration handover completes` pins the distinction down: right after the `firstScreen` diff the promise is still pending, and it only resolves once the `rLynxChange` patch is applied + acked — so "ready" and "complete" are deliberately separate, and the returned promise tracks the latter. (This is also the gap that made `markFirstScreenSyncReady()` insufficient, which is why it's deprecated in favor of `root.hydrate()`.)

---

PR #2971, opened 2026-07-13
fix(web-core): map textarea tag to x-textarea

Huxpro:
`__CreateElement` looks up `LYNX_TAG_TO_HTML_TAG_MAP`, which had no
`textarea` entry, so `textarea` fell through to a bare HTML `<textarea>`
instead of the `x-textarea` custom element registered by
`@lynx-js/web-elements`. Because the element was never `x-textarea`, the
Lynx event forwarding (`input`/`focus`/`blur`) never wired up, so typed
input never bridged to the framework thread — `bindinput`/`@input`
bindings (e.g. Vue's `v-model`) silently did nothing. `input` already
worked because it was mapped to `x-input`.

## Changes

- **`ts/constants.ts`** — add `textarea: 'x-textarea'` to `LYNX_TAG_TO_HTML_TAG_MAP`
  so `__CreateElement('textarea', …)` produces an `x-textarea` element. The
  runtime event tables already bridge `x-textarea`'s `lynxinput`/`lynxfocus`/
  `lynxblur` events, so no further wiring is needed.
- **`src/constants.rs`** — add the same `("textarea", "x-textarea")` entry to the
  parallel Rust map, which drives CSS type-selector rewriting, so `textarea`
  selectors in Lynx stylesheets are rewritten to `x-textarea` and keep matching
  the rendered element.
- **Tests** — `element-apis.spec.ts` covers the `input`/`textarea` tag mapping at
  element creation time; `style_info_decoder.rs` gets a `textarea` type-selector
  rewrite test mirroring the existing `frame` selector test.
- **Changeset** — publishes the fix as a `@lynx-js/web-core` patch.

Note: the wasm `create_element` has no tag allow-list (it stores whatever mapped
tag the TS layer passes), so no wasm-side allow-list change is required beyond the
two maps above.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed Lynx `textarea` elements so `input`, `focus`, and `blur` events are forwarded correctly.
  * Updated `textarea` rendering to use the `x-textarea` custom element while preserving Lynx tag identity and selector matching.
* **Tests**
  * Added element API tests for `input` → `x-input` and `textarea` → `x-textarea`.
  * Added a decoder test confirming `textarea` CSS type selectors are rewritten to match `x-textarea` output.

### DIFF
--- a/.changeset/web-core-textarea-mapping.md
+++ b/.changeset/web-core-textarea-mapping.md
@@ -1,0 +1,9 @@
+---
+"@lynx-js/web-core": patch
+---
+
+Map the Lynx `textarea` tag to the `x-textarea` custom element when creating elements on web.
+
+`__CreateElement` looked up `LYNX_TAG_TO_HTML_TAG_MAP`, which had no `textarea` entry, so `textarea` fell through to a bare HTML `<textarea>` instead of the `x-textarea` custom element registered by `@lynx-js/web-elements`. Because the element was never `x-textarea`, none of the Lynx event forwarding (`input`/`focus`/`blur`) was wired up, so typed input never bridged to the framework thread — any `bindinput`/`@input` binding (e.g. Vue's `v-model`) silently did nothing. `input` already worked because it was mapped to `x-input`.
+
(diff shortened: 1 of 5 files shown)


PupilTong:
please also update the constants.rs

PupilTong (approved):

PR #2977, opened 2026-07-14
RFC(react): the 0.0 first screen — root &lt;Background&gt; auto-strips every component body (L4)

Huxpro:
# RFC: the 0.0 first screen — root `<Background>` auto-strips every component body

> Layer 4 of the IFR opt-in/opt-out design. **Stacked on the L3 branch (`claude/lynx-ifr-l3-bundling`, #2970)** — this PR's diff contains only the L4 work. The PR body doubles as the RFC.

## The thesis

On the 0.0 → 1.0 first-screen scale that this design builds:

- **1.0** — default IFR: every component renders on the main-thread first frame.
- **per-subtree opt-out** — `<Background>` (L1, #2965) + `'background only'` (L3, #2970): a boundary defers a subtree, and its render logic is stripped from the main-thread bundle.
- **0.0** — nothing renders on the main thread; the whole tree hydrates from the background.

L2 (#2963) already reaches 0.0 at *runtime* with `mainThreadRender: false`. But that is a separate switch, and it only *skips execution* — every component's render logic is still bundled into the main thread. This layer closes both gaps: it makes 0.0 **the root case of the same `<Background>` boundary**, and it makes the deferral **compile-time**, so the render logic is physically absent from the main-thread (LEPUS) bundle rather than merely skipped.

The design intent: the exposed API stays user-space `<Background>`. The underlying flag is *just implementation*. A `<Background>` at the render root

```tsx
import { Background, root } from '@lynx-js/react';

root.render(
  // host-element fallback only — component bodies are gone from the main thread
  <Background fallback={<view><text>Loading…</text></view>}>
    <App />
  </Background>,
);
```

declares a 0.0 first screen — the whole first frame is the static `fallback`, and nothing renders on the main thread — which is exactly the condition under which it is safe to empty **every** component render body from the main-thread bundle. No per-component `'background only'` annotation is needed: L3 is the surgical, per-subtree tool; L4 is the whole-program endpoint of the same boundary.

## The mechanism

Two halves, resolved into one loader flag.

**1. The strip (compile-time).** `directive_dce` gains a `strip_all_components` mode on the LEPUS target. It empties the body of every function-like that **returns JSX** — i.e. a component render:

- The discriminator is *returns* JSX, not *contains* JSX, and it does not descend into nested functions. So an entry callback like `runAfterLoadScript(() => root.render(<App/>))` (JSX only as a call **argument**) is left intact, and the generated snapshot `create`/`update` closures and worklet closures (which return element-PAPI arrays / nothing, never JSX) are never mistaken for components.
- Correctness is by construction, from the same pass order L3 relies on — `worklet → snapshot → directive_dce`. The worklet and snapshot passes hoist their definitions to **module scope** *before* the strip empties the bodies, so on the main-thread bundle:

  | part of every component | main-thread bundle |
  |---|---|
  | render logic (bodies, hooks, effects, handlers) | **absent** |
  | logic-only imports reached only by render bodies | **absent** |
  | element (snapshot) definitions | **present** |
  | main-thread-script (worklet) definitions | **present** |
  | the root `<Background>`'s static `fallback` | **present** |

**2. The trigger.** The rsbuild plugin decides `stripAllComponents` two ways — *"identify root `<Background>` to light it up automatically" **and** "keep an internal explicit switch as the fallback"*, both, as agreed:

- **Auto-detection** (the ergonomic path): when `experimental_stripAllComponents` is left `undefined`, the plugin scans the entry sources for a root-level `<Background>` — a `<Background>` imported from `@lynx-js/react` sitting directly in a `.render(...)` call. A `<Background>` nested *inside* the app (a partial opt-out) never appears in the entry's `.render(...)`, so it is correctly ignored.
- **The internal switch** (the fallback): `experimental_stripAllComponents: true | false` forces or forbids the strip regardless of detection.

The resolved boolean flows main-thread loader → `directiveDCE.stripAllComponents` → the transform. The background bundle is untouched (the strip is LEPUS-only); it renders and hydrates the full tree.

## The constraint

Because component bodies no longer exist on the main thread, a root `<Background>`'s `fallback` must be composed of **host elements** (`<view>`, `<text>`, …) rather than user components — a `<Loading/>` component in the fallback would render nothing, since its body was stripped. The static host-element fallback renders through the retained snapshot definitions. This is documented on the `<Background>` JSDoc and the option.

## What's in this PR

- **Transform** (`swc_plugin_directive_dce`): the `strip_all_components` mode + the "returns JSX" component discriminator, with Rust unit tests (12 green).
- **Plumbing**: `stripAllComponents` on the react-webpack-plugin main-thread loader; `experimental_stripAllComponents` (`@alpha`) + the root-`<Background>` auto-detection on the rsbuild plugin.
- **Docs**: the `<Background>` JSDoc now documents the root-level 0.0 case and the static-fallback constraint; the option is documented.
- **Real-bundle proof** (`packages/webpack/react-webpack-plugin/test/strip-all-components.test.ts`): builds a dual-thread bundle from a **zero-annotation** root-`<Background>` app and asserts, against the emitted `main__main-thread.js`, that every component's render logic and its logic-only imports are **absent** while the snapshot, worklet, and static fallback definitions are **present** — and that a **control build without the strip keeps the render logic**, proving the strip is the cause. The background bundle keeps everything.
- **Detection unit tests** (`packages/rspeedy/plugin-react/test/stripComponents.test.ts`): the root-`<Background>` scan matches the root case (incl. multi-line and subpath imports), ignores a nested boundary, ignores a non-`@lynx-js/react` `Background`, and honors the explicit override.
- **Changeset**.

## Scope and follow-ups

- **A guardrail** — a build/lint diagnostic when the root `<Background>`'s `fallback` contains a user component (which the strip would blank), turning today's documented constraint into an enforced one.
- **Detection robustness** — the entry scan is a best-effort heuristic for the canonical `root.render(<Background …>)` form; an aliased `Background` import or an indirected render is not detected. The internal switch is the robust override for those cases.

## Checklist

https://claude.ai/code/session_01LyHx91XciQDJunC9fj6bTe

---
_Generated by [Claude Code](https://claude.ai/code/session_01LyHx91XciQDJunC9fj6bTe)_

### DIFF
--- a/.changeset/strip-all-components-root-background.md
+++ b/.changeset/strip-all-components-root-background.md
@@ -1,0 +1,32 @@
+---
+"@lynx-js/react-rsbuild-plugin": minor
+"@lynx-js/react-webpack-plugin": minor
+"@lynx-js/react": patch
+---
+
+Light up a whole-program strip when a root-level `<Background>` is detected, keeping every component's render logic out of the main-thread (LEPUS) bundle.
+
+A `<Background>` at the render root declares a 0.0 first screen: the whole first frame is the static `fallback`, and nothing renders on the main thread. Under that condition it is safe to empty _every_ component render body from the main-thread bundle — no `'background only'` annotation needed. The build detects the root `<Background>` in an entry and turns this on automatically:
+
+```tsx
+import { Background, root } from '@lynx-js/react';
+
+root.render(
+  // host-element fallback only — component bodies are gone from the main thread
+  <Background
+    fallback={
+      <view>
+        <text>Loading…</text>
+      </view>
+    }
+  >
+    <App />
+  </Background>,
+);
+```
+
+The element (snapshot) and main-thread-script (worklet) definitions the first-screen hydration needs are hoisted to module scope before the strip, so they are retained; only the render bodies (and the logic-only modules they reached) are shaken out. The background thread is untouched — it renders and hydrates the full tree.
+
+Auto-detection can be overridden with the internal `experimental_stripAllComponents` option (`true`/`false` forces or forbids the strip regardless of detection). Because component bodies no longer exist on the main thread, a root `<Background>`'s `fallback` must be composed of host elements rather than user components.
+
+Unlike the per-component `'background only'` opt-out, this is the whole-program (0.0) endpoint of the same boundary; unlike a runtime-only switch, the render logic is physically absent from the main-thread bundle rather than merely skipped.
--- a/packages/react/runtime/src/core/background.ts
+++ b/packages/react/runtime/src/core/background.ts
@@ -96,6 +96,24 @@ export interface BackgroundProps {
  * because its main-thread render is a no-op — rendering it on the first screen
  * without a boundary would produce an empty frame.
  *
+ * At the render root, `<Background>` becomes the 0.0 first screen — the whole
+ * first frame is the static `fallback` and nothing renders on the main thread:
+ *
+ * ```tsx
+ * root.render(
+ *   // host-element fallback only, since component bodies are stripped from the
+ *   // main thread
+ *   <Background fallback={<view><text>Loading…</text></view>}>
+ *     <App />
+ *   </Background>,
+ * )
+ * ```
+ *
+ * The build detects this and strips every component render body from the
+ * main-thread bundle automatically — the whole-program endpoint of the same
+ * boundary, with no per-component annotation. The `fallback` must then be
+ * composed of host elements rather than user components.
+ *
  * @public
  */
 export function Background(props: BackgroundProps): ReactNode {
--- a/packages/react/transform/crates/swc_plugin_directive_dce/lib.rs
+++ b/packages/react/transform/crates/swc_plugin_directive_dce/lib.rs
@@ -1,10 +1,10 @@
 use serde::Deserialize;
 use std::fmt::Debug;
 use swc_core::{
-  common::{errors::HANDLER, Span},
+  common::{errors::HANDLER, Span, DUMMY_SP},
   ecma::{
     ast::*,
-    visit::{VisitMut, VisitMutWith},
+    visit::{Visit, VisitMut, VisitMutWith, VisitWith},
   },
 };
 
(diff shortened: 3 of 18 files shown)


Huxpro:
## 📐 Design doc + a correction worth surfacing before this merges

**Design report (Claude Artifact):** https://claude.ai/code/artifact/42ed955c-c004-40ae-8f74-d7fd5b605403
*(private by default — share from the artifact page if reviewers should see it)*

The doc lays out the whole 1.0 → 0.0 first-screen spectrum with **real rspack output** for every tier (all four tiers built from one shared component set, only the `<Background>` position + strip switch change). Writing it surfaced a correctness point that changes how this PR should ship.

---

### The finding: the whole-program auto-strip is not hydration-safe as a default

`root.render(<Background fallback={…}><App/></Background>)` where `App` delegates to components in other modules (i.e. every real app):

| main-thread.js after auto-strip | result |
|---|---|
| `function App() {}` | body emptied ✓ |
| `import { Feed }` in App | **removed** (unused after strip) |
| `./Feed.jsx`, `./Header.jsx` modules | **gone entirely** |
| Feed snapshot / worklet definitions | **shaken out with the module** ✗ |

Emptying `App`'s body severs its references to `<Header/>`/`<Feed/>`; the transform's `shake` pass then drops the now-unused imports, and the child modules — **carrying the snapshot/worklet definitions hydration needs** — leave the main-thread bundle. In production the main thread can't rebuild the real content when the background hands over.

The `strip-all-components` test passes only because its fixture is **self-contained** (`App` itself is the heavy component, referenced directly by `root.render`) — the one shape where the strip happens to work.

### The complete 0.0-B already exists — it's the composition

`root <Background>` **×** per-component `'background only'` (i.e. L3 pushed to the root). Verified against real output:

```
app modules:  ./FeedBg.jsx, ./App.jsx, ./e.jsx   ← all present
Feed snapshot / worklet:  PRESENT
Feed render logic + formatFeed:  stripped
hydration:  complete
```

It works for the same reason L3 works: **structural parents keep their bodies → keep the child references → the modules (and their hoisted definitions) survive.** `'background only'` only empties the leaf logic. Root `<Background>` clears the first screen to the fallback. Nothing to fix here — this is the "0.0 is just a composed effect" that the L3 goal called for.

So the gap isn't 0.0-B; it's the **annotation-free convenience layer** (this PR's `stripAllComponents`) that tries to skip the `'background only'` annotations by emptying *every* body — and severs the references that pin definitions.

---

## 🤝 Handover

### 1 · The PR stack

```
main
 └─ L1  #2965  claude/lynx-ifr-opt-design-osbqod   <Background> component
     ├─ L2  #2963  claude/lynx-ifr-l2-hydration    root.hydrate() + mainThreadRender   (parallel)
     └─ L3  #2970  claude/lynx-ifr-l3-bundling      'background only' compile-time separation
         └─ L4  #2977  claude/lynx-ifr-l4-zero-ifr  root-<Background> auto-detect + stripAllComponents  ← this PR
```

- **L1 #2965** — the `<Background fallback>` boundary component (`packages/react/runtime/src/core/background.ts`, 3 lines: main thread returns `fallback`, background returns `children`). A **runtime** opt-out of the main-thread first screen. No protocol/build change. This is axis-1 (what renders on the first frame). *Verified.*
- **L2 #2963** — `root.hydrate()` + `root.render(jsx, { hydrate: false })` (imperative hydration handover) and the `mainThreadRender` build option (a wholesale runtime 0.0). Parallel branch off L1. *Verified.*
- **L3 #2970** — compose `<Background>` with the `'background only'` directive so the deferred subtree's **render logic leaves the main-thread bundle** while its snapshot + worklet definitions stay (definition-site strip + boundary keeps the reference). This is axis-2 (where the code lives). *Verified cross-module against real bundle.*
- **L4 #2977 (this)** — root-`<Background>` auto-detection + the `stripAllComponents` transform mode (empty every component body). Intended as the annotation-free 0.0. Transform + plumbing + auto-detection are solid and tested; **the whole-program strip has the reference-severing gap above.**

### 2 · The goal (original intent, verbatim)

> **首屏优化 / opt-in-opt-out（起点）:** 在 Lynx 的官方 React 实现中，有一个首屏的优化的思路……思考一下，我们从实现和 API 设计上，如何使得 IFR 这个功能变成 opt-in 或者 opt-out 的？如何设计可以更优雅和更具备组合性？

> **L2:** first screen 这个概念不好……我不满意现在的 timing 设计……我认为我们应该继续强调和发扬双线程 island 的这种可组合性。

> **L3:** IFR 本质上是一种强制默认全量首屏……B 的完全抛弃 IFR，应该只是一种组合出的效果……我们希望这个上层抽象同时是坚固和完备的……也可以通过改变打包策略，彻底的让部分代码不出现在它不该出现的上下文里，避免副作用泄漏的可能。

> **L4:** 我希望把 0.0 的给统一。在我看来，底下的 flag 只是实现，真正暴露的 API 仍然最好是 user-space 的 Background。只不过 Background root 的时候会 trigger 底下运行时可以做到某种 whole-program optimization…… 识别 root-`<Background>` 自动点亮 vs 保留一个内部显式开关兜底 —— 我认同两者都要！

The through-line: IFR is a forced full first screen; make first-screen participation **composable per subtree** so it becomes a 0.0→1.0 dial; keep the exposed API in user space (`<Background>`); make the abstraction **robust and complete** by also removing deferred code from contexts it shouldn't reach (no side-effect leakage). 0.0 should be a *composed* point on the dial, not a separate mode.

### 3 · The gap & the help needed

**Gap:** L4's auto-strip, wired to fire by default on a root `<Background>`, drops cross-module definitions and breaks production hydration for realistic (multi-module) trees. The self-contained test masked it.

**Two decisions / tasks for whoever picks this up:**

1. **Decision (product/API):** demote the auto-strip from a *default* to an **explicit experimental opt-in**? Under that model, a root `<Background>` guarantees only the **0.0 first screen** (runtime; all code retained — safe), and code-stripping stays with the `'background only'` composition. `stripAllComponents` remains as an off-by-default internal switch for the self-contained case. *(This is the safe, ship-now option; needs the maintainer's call.)*

2. **Frontier (engineering) — make the auto-strip reference-safe.** So the annotation-free 0.0 becomes complete: when emptying a component body, **preserve keep-alive references to the child components it rendered** (so their modules + hoisted snapshot/worklet definitions survive), instead of emptying to `{}`. L3 gets this pinning for free from the boundary's reference; the root strip has to synthesize it. Options to explore:
   - In the transform, after emptying a body, emit reference statements (or retain the imports) for every component identifier the JSX referenced — so `shake` keeps them and webpack keeps the modules for their side-effectful `snapshotCreatorMap[id] = …` / `registerWorkletInternal(…)`.
   - Or a plugin-side manifest that pins every module carrying a snapshot/worklet registration.
   - Either way, add a **multi-module** bundle test (root `<Background>` over an `App` that delegates to child-module components) asserting the child snapshot + worklet survive and hydration is buildable — the test this PR is missing.

3. **Nice-to-have guardrails (from L1/L3 follow-ups):** lint a root `<Background>` whose `fallback` contains a user component (blanked by the strip); lint a `'background only'` component used outside a boundary.

**What's already verified and needs no rework:** L1, L2, L3, and the **0.0-B composition** (root `<Background>` × `'background only'`) — all confirmed against real bundle output in the design doc above. The transform's `strip_all_components` mode, the "returns JSX" component discriminator, and the root-`<Background>` auto-detection are correct in isolation; only the *default wiring* of the whole-program strip is in question.

---

PR #3010, opened 2026-07-20
feat(react): mark empty patches with `emptyPatch` flush option

upupming:
## What

When a background commit produces **no element mutations**, tell the main thread about it by passing `emptyPatch: true` to `__FlushElementTree`.

## Why

There are real scenarios (search result pages, for example) where a large share of commits are "no patch": the background thread re-renders, diffing produces zero snapshot ops, yet the main thread still has to walk the whole update path and native still runs a full pipeline.

The engine side now accepts `emptyPatch` on `__FlushElementTree` and, under the unified pixel pipeline, can end the pipeline early for such updates. This PR is the frontend half — it gives native the signal.

## How

The flag is computed on the **background thread**, where the commit's emptiness is already known, and the main thread forwards `flushOptions` untouched.

`emptyPatch: true` is set when both hold:

- the commit carries no operation (`snapshotPatch` is empty / `ops` is empty)
- there is no `delayedRunOnMainThreadData` (a queued `runOnMainThread` task can still mutate elements)

Hydration is never marked: it commits through a separate background entry point (`tt.ts` for the snapshot runtime, the hydration listener for ElementTemplate) that does not set the flag, so the exclusion is structural rather than a runtime condition. This keeps the first-screen pipeline untouched — the `alog` snapshot in `packages/react/testing-library`, which captures the hydrate flush, passes unmodified.

For ElementTemplate the flag is set inside the existing "has a payload to dispatch" branch, so a fully empty commit still dispatches nothing at all rather than gaining a new IPC round-trip.

## Compatibility

Safe on every engine version — SDKs that don't know the option simply don't parse it, with no side effects.

## Test plan

- New `__test__/snapshot/lifecycle/emptyPatch.test.jsx`: a `setState` that doesn't change the rendered output flushes with `emptyPatch: true`; a `setState` that does change the tree flushes without the property.
- `update-timing.test.ts`: empty update envelope asserts `emptyPatch: true`; non-empty asserts the property is absent.
- `element-template-patch.test.tsx`: the three option-only/empty-ops flush assertions now expect `emptyPatch: true`.
- Hydration case asserted explicitly: the hydrate flush never carries `emptyPatch`.
- `pnpm test:snapshot` in `packages/react/runtime`: 95/95 files green, 100% coverage thresholds met.

### DIFF
--- a/packages/react/runtime/src/snapshot/lifecycle/patch/updateMainThread.ts
+++ b/packages/react/runtime/src/snapshot/lifecycle/patch/updateMainThread.ts
@@ -98,14 +98,6 @@ function updateMainThread(
   if (patchOptions.pipelineOptions) {
     flushOptions.pipelineOptions = patchOptions.pipelineOptions;
   }
-  // Hydration drives the first-screen pipeline, leave it untouched.
-  if (
-    !patchOptions.isHydration
-    && !delayedRunOnMainThreadData?.length
-    && patchList.every(({ snapshotPatch }) => !snapshotPatch?.length)
-  ) {
-    flushOptions.emptyPatch = true;
-  }
   __FlushElementTree(__page, flushOptions);
 
   if (flowIds) {
--- a/packages/react/runtime/src/snapshot/lifecycle/patch/updateMainThread.ts
+++ b/packages/react/runtime/src/snapshot/lifecycle/patch/updateMainThread.ts
@@ -98,8 +98,10 @@ function updateMainThread(
   if (patchOptions.pipelineOptions) {
     flushOptions.pipelineOptions = patchOptions.pipelineOptions;
   }
+  // Hydration drives the first-screen pipeline, leave it untouched.
   if (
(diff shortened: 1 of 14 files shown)


packages/react/runtime/src/snapshot/lifecycle/patch/updateMainThread.ts
```diff
@@ -98,6 +98,14 @@ function updateMainThread(
   if (patchOptions.pipelineOptions) {
     flushOptions.pipelineOptions = patchOptions.pipelineOptions;
   }
+  // Hydration drives the first-screen pipeline, leave it untouched.
+  if (
+    !patchOptions.isHydration
+    && !delayedRunOnMainThreadData?.length
+    && patchList.every(({ snapshotPatch }) => !snapshotPatch?.length)
+  ) {
+    flushOptions.emptyPatch = true;
```
HuJean: Maybe can calculate in bts？

HuJean (approved):

PR #3011, opened 2026-07-20
feat(react): render multiple shared-context pages through the classic root.render

upupming:
> **Draft** — the runtime mechanism is complete and verified on device, and a worked example is included. The larger RFC discussed internally (generating the per-page glue from the build plugin) is not written yet. Opening early for design feedback on the shape of the API and the context-switching mechanism.

## Problem

`root.render` is bound to the module-level `__root`, and so is the rest of the render state — `__globalSnapshotPatch`, the commit task map, the instance registry, the delayed event buffers. That is fine while one JS context renders one page.

It stops being fine under the shared-context model, where several pages load their common dependencies once onto `globalThis` and share a single ReactLynx runtime. Then:

- component trees overwrite each other, because every root renders into the same `__root`;
- two roots' snapshot patches merge into a single native stream, because the patch buffer is one buffer;
- hydration rewrites background ids to each page's main-thread ids, which collide in a shared instance registry;
- only the first page runs the module-level bootstrap, so later pages never get their `tt` handlers injected — their events fall through to lynx_core's fallback and throw `cannot read property '-2:0:' of undefined`.

## Approach

Give each root a private "register file" of that state. Every module that owns per-root state registers a slot (`registerContextSlot`) that can save its module-level variables into a `RootContext` and load them back; switching roots is a register swap. Hot paths keep reading the same module-level variables, so **the classic single-root path is unchanged and pays no overhead** — it simply always runs on the default context.

Ambient state alone is not enough, because Preact defers commits to a microtask and its rerender queue may interleave components of different roots. Two things close that gap:

- every `BackgroundSnapshotInstance` is stamped with its owning context at construction;
- an `options.renderComponent` hook re-establishes the owner's context before each component re-render, where the diff, the patch recording and the commit all run synchronously.

Async native acks capture their own root's task map and instance registry in closures, so a callback arriving while another root is current still operates on the right one. The per-root `lynx` channel is resolved lazily at send time (`contextLynx()`), never captured at module load — environments like the testing library swap `globalThis.lynx` when switching threads, and an eager capture goes stale.

## API

**No new public API, and nothing changes for normal apps.** The whole mechanism is gated behind the `__MULTI_ROOT_RENDER_CONTEXT__` compile-time define — exposed as the alpha `experimental_multiRootRenderContext` option on `ReactWebpackPlugin` / `pluginReactLynx`, default `false`. Normal builds strip the machinery entirely (slot registration, the `renderComponent` hook, per-root channel selection all compile away), so the classic single-page path keeps its original code verbatim in every non-define arm. The shared external runtime bundle is NOT built with it by default either — users who need the mode build their own `react-umd` with `experimental_multiRootRenderContext` on.

User code keeps the classic call:

```js
import { root } from '@lynx-js/react'

root.render(<App />)
```

What changes is one line of per-page glue ahead of it — hand-written today, to be generated by the build plugin:

```js
// Bind the shared runtime to this page's own bridge objects.
if (__BACKGROUND__) {
  root.__experimentalBindRenderContext({ lynx, lynxCoreInject })
}
```

`__experimentalBindRenderContext` is an experimental hook that exists on `root` only when the runtime is built with the flag; it binds the shared runtime to the page currently being loaded, and `root.render` then renders into that page's own root. Incoming native calls on that page's `tt` operate on that root only, and the root's outgoing patches (`rLynxChange`, `rLynxFirstScreenSyncReady`) reach that page's view only. **No native-side change is required** — native already dispatches through per-page bridge objects, so object identity is the view discriminator. On the main thread each page has its own VM, so the glue is a no-op there and the same entry works on both threads. Calling it with `undefined` clears the binding; without the glue (or with the flag off), `root.render` takes the classic singleton path untouched.

`createRoot` / `ReactLynxRoot` are not exported from the public entry — the primitive stays internal, reachable only through the bind hook.

## Example

`examples/react-multi-root-render-context` is a runnable end-to-end setup: two pages (`pageA` / `pageB`) share one JS context through `@lynx-js/external-bundle-rsbuild-plugin`. The ReactLynx runtime is built locally as an external bundle with `experimental_multiRootRenderContext: true`, a shared module (`example-shared`) holds a mutable store, and each page stays at ~9 kB with everything externalized. Clicking +1 on one page is visible from the other; an `initCount` probe asserts the shared bundle evaluates exactly once per context.

## Also fixed

Shared-context lifecycle bugs that only surface once a second page actually runs:

- per-root cleanup tasks and the context-global `ctx not found` listener are no longer torn down when one page of a shared context is destroyed;
- `firstScreen` re-asserts its own context after draining Preact's shared rerender queue, which may render another root's components — otherwise it could hydrate the wrong root's tree.

## Verification

Single-page output keeps its original code verbatim in every non-define arm; the whole multi-root machinery compiles away (`__MULTI_ROOT_RENDER_CONTEXT__` / `contextLynx` / `switchRootContext` leave zero trace in a default build of `examples/react`). The residual size cost is **+8 bytes** on the 94 kB `examples/react` bundle (94143 → 94151, same-toolchain builds of `main` and this branch) — one declared class field plus two local captures kept for readability.

Full runtime suite green: 108 files / 811 tests, plus the testing-library suite (159 tests) — which caught a real bug: it swaps `globalThis.lynx` per thread switch, so the context channel must be resolved lazily, not captured. 100% coverage on the new files, api-extractor report unchanged, and the `experimental_multiRootRenderContext` passthrough is covered on the rspeedy side.

New tests cover the isolation properties directly, including one that hydrates the *same* first-screen payload into two roots so both end up with identical main-thread snapshot ids — the exact collision a shared registry cannot survive — then shows events and patches still routing per page. The bind hook has its own tests: two sequential page binds render without clobbering, clearing the binding restores the classic path, and the first-screen handshake goes through the bound page's channel.

Verified end to end on device with two pages in one shared context group: load page A, load page B, interact with both, close page B, page A stays fully interactive.

## Not in this PR

- Generating the page glue from the build plugin, and whether an element-scoped `createRoot` ever becomes public — RFC material.
- `root.render`'s remaining side effects and the full "no module-level mutable state" refactor.
- Per-view `renderPage`/`updatePage` on the main thread, which is only meaningful once a page hosts more than one root.


.changeset/reactlynx-create-root.md
```diff
@@ -0,0 +1,16 @@
+---
+"@lynx-js/react": patch
+"@lynx-js/react-webpack-plugin": patch
+"@lynx-js/react-rsbuild-plugin": patch
+"@lynx-js/react-umd": patch
+---
+
+Support rendering multiple shared-context cards through the classic `root.render`, behind a build-time opt-in.
```
upupming: Changeset should not describe the whole implement detail, we should just use simple short sentences.

packages/react-umd/rslib.config.ts
```diff
@@ -12,7 +12,9 @@ export default defineExternalBundleRslibConfig({
     },
   },
   plugins: [
-    pluginReactLynx(),
+    // The shared external bundle serves several cards in one JS context, so
+    // it is built with multi-card roots on.
+    pluginReactLynx({ experimental_multiCardRoots: true }),
```
upupming: Please do not enable this in react-umd by default, users that need this mode should build their react-umd by their own

packages/react/runtime/__test__/snapshot/utils/envManager.ts
```diff
@@ -76,6 +77,9 @@ export class EnvManager {
   }
 
   resetEnv(): void {
+    // Multi-root tests may leave a non-default context current; the resets
+    // below all target the default context's (module-level) state.
```
upupming: Remove comments

packages/react/runtime/lazy/internal.js
```diff
@@ -24,6 +24,7 @@ export const {
   __DynamicPartSlotV2,
   __DynamicPartSlotV2_0,
   __DynamicPartMultiChildren,
+  __bootstrapCard,
```
upupming: How about `__experimantalBootstrapCard`

packages/react/runtime/src/core/thread-function-call/main-thread.ts
```diff
@@ -17,6 +18,17 @@ export type RunOnMainThread = <R, Fn extends (...args: any[]) => R>(fn: Fn) => (
 
 export let delayedRunOnMainThreadData: RunWorkletCtxData[] = [];
 
+registerContextSlot({
```
upupming: Shound't we wrap this in a `if (__MULTI_CARD__)`

packages/react/runtime/src/core/runtime-destroy.ts
```diff
@@ -2,9 +2,24 @@
...
 // LICENSE file in the root directory of this source tree.
 
+import { registerContextSlot } from '../root-context.js';
+
 type DestroyTask = () => void;
 
-const destroyTasks = new Set<DestroyTask>();
+// Per-root: destroying one card must not run/clear another card's cleanup
+// tasks in a shared background context.
+let destroyTasks = new Set<DestroyTask>();
```
upupming: remove comments please

packages/react/runtime/src/core/runtime-destroy.ts
```diff
@@ -2,9 +2,24 @@
...
+
+registerContextSlot({
```
upupming: Can we move all `registerContextSlot` to a single file and maintain them there will be more easier and do not affect normal reactlynx work flow

upupming (commented):
There are bundle size increase in relative ci check, it is because our tree shaking about `registerContextSlot` is not deep enough.

packages/react/runtime/src/snapshot/lifecycle/contextSwitchHook.ts
```diff
@@ -0,0 +1,27 @@
...
+import { hook } from '../../utils.js';
+
+const onRenderComponentHook = <T extends unknown[]>(
+  old: ((...args: T) => void) | undefined,
+  ...args: T
+) => {
+  const component = args[1] as { [PARENT_DOM]?: { __rootCtx?: RootContext } } | undefined;
+  const ctx = component?.[PARENT_DOM]?.__rootCtx;
+  if (ctx) {
+    switchRootContext(ctx);
```
upupming, line 19: Will this make diff too slow?

PR #3166, opened 2026-07-23
perf(react): optimize runtime hot paths for render, hydrate and destroy

upupming:
## Summary

Micro-optimizations in the ReactLynx runtime's hottest per-element paths, driven by CodSpeed simulation flamegraphs (dominated by QuickJS interpretation of runtime JS), local perfetto/walltime profiling of `benchmark/react`, and a V8 `--cpu-prof` probe of the background-thread first-screen render:

- **Snapshot instance fields are `declare`d instead of emitted as class fields**: with `useDefineForClassFields`, every declared field was defined on every instance up front — 14 defines per `BackgroundSnapshotInstance` and 17 per `SnapshotInstance`, most immediately overwritten by the constructor or never set at all. This was the single largest constructor cost (`add_property`/`add_shape_property`/`js_clone_shape` in the CodSpeed simulation flamegraph).
- **`SnapshotInstance` / `BackgroundSnapshotInstance` constructor**: one `Map.get` instead of `Map.has` + `Map.get`.
- **`ensureElements`**: only enter `__pendingListUpdates.runWithoutUpdates` (closure alloc + try/finally) when the instance actually has `__values`; hoist `isSlotV2` out of the child loop.
- **hydrate (main thread + background)**: when before/after children match pairwise by type — the common hydration case — hydrate in place instead of running `diffArrayLepus`/`diffArrayAction` (which allocate a keyed map, `Set`s and tuple arrays per node). The fast path is behaviorally identical: with pairwise-equal types the diff is provably empty and `diffArrayLepus` pairs children in order. Background hydrate also materializes `childNodes` only when the snapshot has slots, and the per-node values closure is skipped when there are no values.
- **`traverseSnapshotInstance`**: walks sibling links instead of materializing a `childNodes` array per node. All existing callbacks only detach the *visited* node, and both the first child and each next sibling are captured before the corresponding callback can run.
- **`BackgroundSnapshotInstance.setAttributeImpl`**: report `needUpdate` via a module-level out-var instead of allocating a `{needUpdate, valueToCommit}` object per dynamic part per commit.
- **main-thread `useReducer`**: state updates are dropped on the main thread anyway, so share one no-op dispatch function instead of allocating a closure per hook state.
- **renderToOpcodes**: skip the options-hook dispatch for element (string-typed) vnodes in production without devtools — no consumer does anything for them there (`mainThreadImpl` only tracks the current component, `profileHooks` only traces components). Development keeps the full dispatch for component stacks and devtools.
- **profile hooks**: install the per-vnode DIFF/DIFF2/DIFFED hooks directly instead of via the variadic `hook()` helper (which allocates a rest-args array per invocation), cache `ReactLynx::diff::X` / `ReactLynx::render::X` label strings per component type, and cache the render profiling wrapper per render function instead of allocating a closure every component render.
- **`updateSpread`**: gate the event-name regex behind cheap `includes('bind'/'catch')` checks and allocate the `dataset` object lazily.
- **`createTextNode`**: the `data` setter Preact writes through now lives on a subclass prototype instead of a per-instance `Object.defineProperty`.

## Local walltime results (benchx_cli, Darwin arm64, median of 100 rounds, vs main)

| benchmark | main | this PR | delta |
|---|---|---|---|
| 003 componentAtIndex reuse | 0.92 ms | 0.35 ms | **−62%** |
| 007-four-layer-views-hydrate | 25.46 ms | 11.58 ms | **−55%** |
| 003-hello-list-hydrate | 2.32 ms | 1.18 ms | **−49%** |
| 007-four-layer-views-destroyBackground | 6.09 ms | 3.12 ms | **−49%** |
| 002-hello-reactLynx-hydrate | 0.41 ms | 0.22 ms | **−45%** |
| 003-hello-list-destroyBackground | 0.51 ms | 0.33 ms | **−34%** |
| 013-attrs-runtime renderMainThread | 4.21 ms | 3.21 ms | **−24%** |
| 007 renderMainThread | 95.9 ms | 73.6 ms | **−23%** |
| 003 renderMainThread | 4.58 ms | 3.50 ms | **−24%** |
| 008 renderMainThread | 14.68 ms | 11.37 ms | **−23%** |
| 002 renderMainThread | 0.95 ms | 0.78 ms | **−18%** |
| renderBackground (002/003/007/008/013) | | | −2%…−9% |
| background.js LoadScript | | | −8%…−18% |

No case regressed. A Node V8 `--cpu-prof` probe of a 6965-element background first render (relevant for the web-platform runtime): mount 3.24 ms → 2.02 ms (−38%), unmount −40%.

## Validation

- `packages/react/runtime` vitest: 107 files / 798 tests pass, unchanged snapshots.
- `benchmark/react` perfetto + `detectLeak.mjs`: no FiberElement leaks in any case.

## Notes

- One observable (non-functional) change: profile trace labels are cached per component type, so a component that mutates `displayName` *between renders* keeps its first label. Similarly, `component.render` stays wrapped by the profiling wrapper instead of being restored after every render. Both only affect builds with `performance.profile: true`.
- `declare`d optional fields (`__extraProps`, `__listItemPlatformInfo`, …) are no longer own properties initialized to `undefined`; all runtime accesses are truthiness/`undefined` checks, so reads behave identically. No `in`/`hasOwnProperty` checks against snapshot instances exist in the runtime.
- Remaining known costs are outside this repo: renderMainThread is ~⅔ native FiberElement ops, and the bulk of renderBackground is inside the `@lynx-js/internal-preact` diff (`constructNewChildrenArray`/`diffElementNodes` ≈ 20% of background mount in the V8 probe) — potential follow-up in that repo.

## CodSpeed-flagged regressions: root cause (raw callgrind analysis)

CodSpeed flags 7 regressions (e.g. `003/006/012 background.js_LoadScript` −13%, `007 serializeRoot` −6.4%, `more faster fib(20)` −7.3%). Diffing the raw Valgrind artifacts (`codspeed-valgrind-tmp-*`) of this PR's run against main's run shows these are **allocator warm-up displacement, not new cost**:

- Per-process totals (authoritative per-part `totals:` from callgrind) **improve ~19–21% across every affected case**, e.g. 006-static-raw-text whole process Ir 122.2M → 97.1M (−20.5%); per region: renderMainThread −39.9%, hydrate −53.7%, destroy −15.2%, bg LoadScript +16.7%.
- Of the +949k Ir in 006's `background.js_LoadScript`, **+852k is `mi_mem_is_zero`** (zero on main) plus the rest of mimalloc's fresh-page path (`mi_arenas_page_alloc_fresh`, `mi_find_free_page`, …): the earlier phases now allocate far less, so the OS-page warm-up that main paid inside the (previously much slower) render/hydrate phases moves into whichever region first touches fresh pages.
- The same mechanism swings benchmarks in the lucky direction too (`011 background.js_LoadScript` −15.2%, `010` −14.5%), and explains `more faster fib(20)` −7.3% — a pure-`fib` region that executes no runtime code at all.
- The per-case bundles only carry this PR's source diff (module graph unchanged, ~+1.2% text size uniformly), and local walltime on real hardware shows the same regions **faster**, not slower (LoadScript −8…−18%, serializeRoot −1.8%).

### Local reproduction (x86_64 valgrind, same binary/bundle/cache config as CI)

Running case 006 under CodSpeed's valgrind fork (`3.26.0-0codspeed3`, CI cache geometry) in an x86_64 container reproduces CI within ~1% per region (renderMainThread −39.9% both, hydrate −56.0% vs −53.7%, destroy −15.7% vs −15.2%) — and settles the LoadScript question:

| `mi_mem_is_zero` inside `background.js_LoadScript` | CI base | CI head | local base | local head |
|---|---|---|---|---|
| self Ir | 0 | 851,982 | **851,982** | 851,982 |

The identical 851,982-instruction fresh-page zeroing cost exists in every run; it merely lands in different regions depending on incidental allocation patterns of the preceding phases (the local base bundle differs from CI main only in build-environment bytes, and that alone moves it). The flagged LoadScript delta is this fixed cost drifting across region boundaries, not new work.

### Follow-up: benchmark-infra noise source identified

Local function-level analysis shows every flagged delta reduces to the same constant: mimalloc's fresh-OS-chunk commit costs exactly **851,982 Ir of `mi_mem_is_zero`** per event, and these events drift between benchmark regions as allocation patterns shift (e.g. local head has one inside `background.js_LoadScript` and one inside `renderBackground`; subtracting it, `renderBackground` is net −154k Ir with every function improving). Additionally, ~50% of the remaining `renderMainThread` simulation cost is mimalloc bookkeeping in a padded/canary build (`mi_verify_padding`, `mi_page_decode_padding`, `mi_ptr_encode_canary`, ~287 Ir per malloc) — JS interpretation is only ~7%. Rebuilding `benchx_cli`'s mimalloc in release configuration (no padding/canary, eager-commit or pre-warmed arenas) would both shrink and stabilize every simulation metric; tracked as a follow-up in lynx-community/benchx_cli.

### Methodology & anti-overfitting

Per the valgrind micro-optimization guidance: the deterministic **Ir count** (per-benchmark callgrind `totals:`) is the judge for every change, not walltime; each change was chosen by comparing Ir across alternative implementations. To guard against overfitting a single benchmark's shape, the wins were confirmed on an orthogonal case — **013-attrs-runtime** (spread / dynamic-attribute path, structurally unlike 006's static raw text): renderMainThread −24.4%, hydrate −11.3%, destroyBackground −30.2%, ~−12% overall (raw CI callgrind). All changes live in generic runtime hot paths (instance field layout, per-vnode hook dispatch, tree traversal) rather than any case-specific branch, and no benchmark case or test was modified to move numbers.

### `background.js_LoadScript` regression — definitive self-cost split

Separating self-Ir (calls' inclusive cost excluded) in case 006's `background.js_LoadScript` region, base vs head built identically under the same local x86_64 CodSpeed-valgrind:

| | base | head | delta |
|---|---|---|---|
| total self-Ir | 6,591,710 | 6,679,886 | +88,176 (+1.34%) |
| allocator self-Ir (`mi_*`/malloc/free) | 5,902,241 | 5,987,273 | +85,032 (**96% of the delta**) |
| real runtime/JS self-Ir | 689,469 | 692,613 | **+3,144 (+0.5%, noise; incl. one −6,960 win)** |

The runtime code actually evaluated while loading the background bundle is **flat**. The CI-reported −13…−16% comes entirely from the fixed 851,982-Ir `mi_mem_is_zero` fresh-page cost whose region placement is non-deterministic across builds (CI main happened to place it outside LoadScript; CI head placed it inside). Built identically, base and head place it the same and it cancels. **No runtime regression to fix; the residual is benchx_cli allocator warm-up placement, addressable only in the bench harness (mimalloc eager-commit / arena pre-warm).**

### The flagged regressions vs the measurement noise floor

Comparing **two clean `main` commits** (924106b → 85e194c) — neither touches `packages/react/runtime`, the SWC transform, or the web-core bench — the CodSpeed diff still reports:

| benchmark | base | head | change | note |
|---|---|---|---|---|
| `basic-performance-div-1000` (WallTime) | 5.6 ms | 7.2 ms | **−21.99%** | **zero relevant code change** |

So this benchmark swings ±22% run-to-run on identical code. That is the noise floor. Categorizing this PR's flagged regressions against it:

- `basic-performance-div-1000` (−8.6%) — **runner noise**, proven above (main-vs-main −22% on the same benchmark).
- `transform 1000 view elements` (−20% WT / −10.8% Sim) — the SWC **Rust** transform bench; this PR contains no Rust changes.
- `more faster fib(20)` (−7.3%) — a pure `fib` region that executes no runtime code.
- `007 …serializeRoot` (−6.4%) — `JSON.stringify(__root)`; `toJSON`/serialization is untouched → allocator page placement.
- `*/background.js_LoadScript` ×3 (−13…−16%) — the only category attributable to this PR: reducing allocation in `renderMainThread`/`hydrate` lowers the process's committed-page high-water-mark, so the one-time fresh-page zeroing (`mi_mem_is_zero`, a fixed 851,982 Ir) lands in the next region that touches new memory. It is the accounting shadow of the `renderMainThread` win (−16M Ir) — same coin, ~20× smaller — and cannot be separated at the runtime level without giving the win back (making instances larger again). The marginal text-node subclass commit, which additionally added two class definitions to every background bundle's module-eval, has been reverted; the definitive fix for the placement noise is arena pre-warm in the benchx_cli mimalloc build (bench harness).

## Summary by CodeRabbit

- **Performance**
  - Improved hydration with a fast path for matching child structures and reduced diff/work allocations.
  - Lowered overhead during background snapshot commits, per-element rendering, and main-thread hooks.
  - Optimized snapshot traversal and JSX spread handling (including event key detection and lazy `data-*` dataset allocation).

- **Developer Experience**
  - Centralized dev-only guidance to clearly report disallowed main-thread state updates.
  - More efficient profiling label generation and render profiling wrapping.

- **Maintenance**
  - Refreshed package changeset metadata for the `@lynx-js/react` patch release.


upupming:
@coderabbitai review

upupming:
@coderabbitai review

PR #3172, opened 2026-07-24
feat(react): render first screen from background thread (enableMTSRendering: false)

upupming:
With `enableMTSRendering: false`, the main thread renders nothing on first screen and the background thread hydrates; the main-thread bundle is reduced to its snapshot + worklet registrations.

- **Runtime**: `__ENABLE_MTS_RENDERING__` gates `renderMainThread` to a no-op (empty first frame).
- **Transform**: `mainThreadDefinesOnly` reduces each module to its registrations (business stripped); `collectMainThreadDefines` exposes them as data. Registrations keep the `ReactLynx` runtime import + kept user imports for worklet bodies.
- **Loader**: reduces every main-thread module except the ReactLynx runtime dependency closure (resolved, not hard-coded); a `sideEffects: true` main-thread rule keeps registrations from otherwise-dropped packages.
- Example `react-disable-mts-rendering` (main-thread root 101.8kB → ~41kB), plus runtime + transform + plugin tests.

## Summary by CodeRabbit

* **New Features**
  * Added optional `enableMTSRendering` configuration for React builds (including rsbuild and webpack plugins).
  * When disabled, the first-screen main-thread render is skipped while required main-thread registrations remain available.
  * Added a React example demonstrating disabled MTS rendering with interaction, event handling, and main-thread scripts.
* **Bug Fixes**
  * Improved handling of main-thread-incompatible module side effects when MTS rendering is disabled.
* **Tests**
  * Added coverage for enabled/disabled rendering, bundle/template output, and side-effect handling.


upupming:
@coderabbitai review

upupming:
@coderabbitai review

PR #3178, opened 2026-07-24
fix: repair main after the Renovate merge wave

upupming:
## Summary

Restore `main` after the Renovate merge wave. Each commit fixes one regression;
all of them are independently reproducible on `main` today.

| Regression | Introduced by |
| --- | --- |
| `clippy` fails: `expected non-macro attribute, found attribute macro 'napi'` | #3160 — lock file maintenance moved `napi-derive` 2.9.3 → 2.16.13, which no longer accepts two stacked field-level `#[napi(...)]` attributes |
| `build:wasm` fails: bindgen schema 0.2.126 vs CLI 0.2.108 | #3160 — same run moved the `wasm-bindgen` crate past the CLI pinned by dotslash |
| `tsc --build` fails: `Cannot find module 'framer-motion/dom'` | #3066 — replaced the `framer-motion` dependency with `motion` without updating the imports |
| `pnpm install --frozen-lockfile` fails | #3141 — `package.json` declares `eslint ^10.7.0` but the merged lockfile still pinned 9.39.5 |
| `react-transform` fails to build for wasm: `napi_register_wasm_v1: symbol multiply defined` | #3160 — the same run moved `napi` 2.10.4 -> 2.16.17 |
| `pnpm eslint .` aborts, then reports 492 problems | #3141 — bumped seven lint packages to new majors without touching `eslint.config.js` |

### Notes on the choices

- **#3141 is reverted rather than migrated.** It changed only `package.json` and
  the lockfile, so none of the work the majors need was done. The last green run
  before it reported 14 warnings and 0 errors; afterwards `pnpm eslint .` aborts,
  and once the four hard failures are fixed it reports 492 problems (176 in
  markdown code fences, 248 `no-unsafe-*` from unresolved types, 68 from new or
  retargeted rules). That is a migration, and it gets its own PR.
- **Exact pins for `wasm-bindgen` and `napi`.** Both are listed in the renovate
  `ignoreDeps`, but that does not cover lock file maintenance, so `cargo update`
  moved them anyway. A caret range lets the same break come back; `=` does not.
  `wasm-bindgen` is upgraded to 0.2.126 on both sides instead of held back.

`peerDependencyRules.allowedVersions` keeps only entries that are demonstrably
harmless, each with the reason inline: the `@ai-sdk/*` → `zod` pair (a legacy
AI SDK line that `@mastra/core` still bundles under an alias) and the unused
`eslint-plugin-import` copy that pnpm's `autoInstallPeers` drags in through
`eslint-import-resolver-typescript`.

## Validation

Run locally on this branch:

- `pnpm dedupe --check`
- `pnpm run build` (`tsc --build`)
- `pnpm --filter @lynx-js/web-core run build:wasm`
- `cargo clippy -p swc_plugin_element_template --all-targets`
- `pnpm dprint check`

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved WebAssembly build reliability by syncing `wasm-bindgen` artifacts and expanding Rust workspace cache invalidation inputs.
  - Fixed React testing library chunk naming issues by replacing problematic `~` characters with `-`.
  - Updated motion animation wiring to use the motion entrypoint.
- **Improvements**
  - Strengthened React/JSX typings for rendered output and deferred list items.
  - Clarified generated template output typing.
- **Tests**
  - Updated motion wrapper tests to mock the motion entrypoint.
- **Chores**
  - Refreshed linting/tooling and dev dependency versions; updated changeset/workflow automation and related build base-ref logic.


upupming:
@coderabbitai review

upupming:
@coderabbitai review

PR #3250, opened 2026-07-27
fix(template-webpack-plugin): keep background sidecar source map when inlining

### LINKED ISSUE #2964: Web platform: emit background-JS source maps as sidecar assets for the .web.bundle (production symbolication)
## Feature request

> _Edited 2026-07-12 to correct two imprecise claims in the original and add an empirical repro. The core ask is unchanged: the web target has no **hidden-sidecar** source map for the background (`app-service.js`) thread._

**Problem.** For the web target, production `app-service.js` (background-thread) stack frames cannot be symbolicated via a **sidecar** source map. In prod, `WebEncodePlugin` inlines the background chunk into a single `'/app-service.js'` string inside the encoded `.web.bundle` and registers the standalone `.js` for deletion (`inlinedAssets` → `compilation.deleteAsset(name)` at `PROCESS_ASSETS_STAGE_REPORT + 1`, `packages/webpack/template-webpack-plugin/src/WebEncodePlugin.ts`), so no sidecar `.js.map` for the shipped background code survives.

**Empirically verified** — `@lynx-js/rspeedy` 0.15.2, production `rspeedy build`, `output.sourceMap.js` set at the config root, inspecting `dist/`:

| `output.sourceMap.js` | `.web.bundle` size | sidecar `.map` in `dist/` |
| --- | --- | --- |
| `source-map` | ~160 kB (unchanged) | **none** |
| `hidden-source-map` | ~160 kB (unchanged) | **none** |
| `inline-source-map` | **~1.59 MB (≈10×)** — map data-URI embedded in the bundle | — (inline) |

So `output.sourceMap.js` **is** honored for the web env (it is not a no-op — `inline-source-map` proves it does generate a map), but the only mode that yields a usable map embeds it **inline** into the shipped `.web.bundle` at ~10× size; both sidecar modes (`source-map`, `hidden-source-map`) emit **no `.map`** for the background code. There is no hidden-sidecar path suitable for a production symbolication service.

Adjacent options that do not cover this:

- `pluginReactLynx({ enableUiSourceMap })` is spread **only for the main thread** (`packages/rspeedy/plugin-react/src/loaders.ts` — `...isMainThread ? { enableUiSourceMap, shake } : {}`) and produces a UI-node → authored-source mapping (emitted as a sidecar `debug-metadata.json`, stripped from prod bundles) with no stack-frame/PC data — it never reaches the background loader and cannot symbolicate background worker frames. _(The original filing said this was "embedded into the bundle"; it is actually a prod-stripped sidecar — corrected.)_
- The unified debug-metadata symbolication is scoped to the native `lynx` env: `pluginLynxDebugMetadata` does `if (!isLynx) return`, with the comment _"Web JS source maps work fine via the default `.map` sibling; skip cleanly."_ — but that assumption breaks precisely because the prod `WebEncodePlugin` inlines the sibling away (per the table above).

**Why it matters.** A production error-reporting / bug-report pipeline (e.g. session-replay console capture + error triage) symbolicates host-page JS fine via ordinary hidden maps, but has no artifact to key a symbolication service on for the app's background-thread frames.

**Proposal.** An opt-in (e.g. `output.sourceMap.emitWebBundleMaps: true`) that re-keys the background chunk's map to the inlined `/app-service.js` module and emits it as a **hidden sidecar** asset next to the `.web.bundle` (no `sourceMappingURL` trailer needed), so `app-service.js`-relative worker frames can be symbolicated in production without the ~10× inline-map bloat.

Repro (three one-line `output.sourceMap.js` config changes + `rspeedy build`, observing `dist/`) available on request.

iPLAYCAFE-dev:
Fixes #2964.

## The mechanism

`compilation.deleteAsset` also deletes everything in `assetInfo.related`, and
`related.sourceMap` is where `SourceMapDevToolPlugin` records the sidecar `.map`.
`WebEncodePlugin` inlines the background chunk into the encoded `.web.bundle` and
then deletes the standalone `app-service.js` — which took its source map with it.

That is why, on the web target:

| `output.sourceMap.js` | sidecar `.map` for background code |
| --- | --- |
| `'source-map'` | none |
| `'hidden-source-map'` | none |
| `'inline-source-map'` | inline only, at ~10x bundle size |

So there was no artifact a production symbolication service could key on for
`app-service.js`-relative worker frames.

## The change

Detach `related.sourceMap` before the delete. The JS is still removed — it lives
inside the template — while its map survives as an ordinary asset. This matches
what rsbuild does for the same configuration, which is what @upupming asked for
in the issue thread.

It is a no-op when no source map was configured: `getAsset(name)?.info.related?.sourceMap`
is undefined and nothing is touched.

## Why `isDebug()` looked like a workaround

Worth recording, because it is what led here. `DEBUG='rspeedy,rsbuild' rspeedy build`
*does* emit the sidecar today — but only because `isDebug()` short-circuits the
whole `inlinedAssets.add(...)` block, so nothing is deleted at all. Measured on a
real project (`@lynx-js/rspeedy` ^0.16.0, identical config, only the env var
differing):

| build | sidecar `.map` | `dist/main.web.bundle` |
| --- | --- | --- |
| `hidden-source-map` | none | 154,708 B · `sha256:c6be08db…` |
| + `DEBUG=` | `background.<hash>.js.map`, 505,863 B | 157,623 B · `sha256:c903d6a0…` |

The two bundles are different artifacts, so a map produced that way does not
describe the bundle a normal build ships — which is the whole requirement for
symbolication. This change makes the sidecar available from an ordinary
production build instead, leaving the emitted bundle untouched.

## Tests

`test/web-encode-plugin-sourcemap.test.ts` drives the plugin through the same
fake-compiler harness as `web-encode-plugin-fetchbundle.test.ts` and pins three
things:

1. the sidecar `.map` survives while the inlined `.js` is still deleted
2. the `related` link is *detached* rather than the deletion being skipped —
   skipping would ship the un-inlined JS as dead weight instead
3. an inlined asset with no source map is deleted unchanged

**I could not run the suite locally**, and would rather say so than imply I did:
this machine is on Node 26 while the repo declares `"node": "^22 || ^24"`, and
`rstest` fails while loading `rstest.config.ts` under it, before reaching any
test. `eslint` does pass on both changed files — the one remaining
`import/no-unresolved` on `@lynx-js/web-core/encode` reproduces identically on
unmodified `main` in my partial (`--filter`) install, so it is environmental and
not from this change. CI should be the authority here.

Happy to adjust the approach — if you would rather gate this behind an explicit
option than change the default, say the word and I will rework it.

## Summary by CodeRabbit

* **Bug Fixes**
  * Preserved background-thread sidecar source maps when inlining JavaScript into `.web.bundle` output.
  * Updated the inlining cleanup behavior to detach related sourcemap metadata so standalone `.map` assets are not removed unintentionally.
* **Tests**
  * Added coverage to verify correct sourcemap cascade behavior for inlined/intermediate assets, including cases with and without related sourcemaps.
* **Chores**
  * Added a patch-level changeset documenting the sourcemap preservation behavior update.

### DIFF
--- a/.changeset/web-encode-preserve-background-sourcemap.md
+++ b/.changeset/web-encode-preserve-background-sourcemap.md
@@ -1,0 +1,20 @@
+---
+'@lynx-js/template-webpack-plugin': patch
+---
+
+Keep the background chunk's sidecar source map when `WebEncodePlugin` inlines the
+chunk into the encoded `.web.bundle`.
+
+`compilation.deleteAsset` also deletes everything in `assetInfo.related`, and
+`related.sourceMap` is where `SourceMapDevToolPlugin` records the sidecar `.map`.
+Because the inlined `app-service.js` is deleted after encoding, its map was
+deleted with it — so on the web target `output.sourceMap.js: 'source-map'` and
+`'hidden-source-map'` emitted no map for background-thread code at all, and only
+`'inline-source-map'` produced one, by embedding it in the shipped bundle at
+roughly 10x the size.
+
+The related link is now detached before the delete, so the JS is still removed
+(it lives inside the template) while its map survives as an ordinary asset —
+matching what rsbuild does for the same configuration.
+
+Fixes #2964.
--- a/packages/webpack/template-webpack-plugin/src/WebEncodePlugin.ts
+++ b/packages/webpack/template-webpack-plugin/src/WebEncodePlugin.ts
@@ -42,6 +42,25 @@ export class WebEncodePlugin {
           stage: Compilation.PROCESS_ASSETS_STAGE_REPORT + 1,
         }, () => {
           inlinedAssets.forEach((name) => {
+            // `deleteAsset` also deletes everything in `assetInfo.related`, and
+            // `related.sourceMap` is where SourceMapDevToolPlugin records the
+            // sidecar `.map`. The JS itself is inlined into the template and
+            // must go, but its source map is the only way to symbolicate
+            // production `/app-service.js` frames, so detach the link first and
+            // let the map survive as an ordinary asset.
+            //
+            // Without this, `output.sourceMap.js: 'source-map'` and
+            // `'hidden-source-map'` both emit no map for the background chunk on
+            // the web target, while `'inline-source-map'` works only by
+            // embedding the map into the shipped `.web.bundle` (~10x size).
+            // rsbuild emits the sidecar in this situation; this makes the web
+            // target behave the same.
+            if (compilation.getAsset(name)?.info.related?.sourceMap) {
+              compilation.updateAsset(name, source => source, info => ({
+                ...info,
+                related: { ...info.related, sourceMap: undefined },
+              }));
+            }
             compilation.deleteAsset(name);
           });
           inlinedAssets.clear();
--- a/packages/webpack/template-webpack-plugin/test/web-encode-plugin-sourcemap.test.ts
+++ b/packages/webpack/template-webpack-plugin/test/web-encode-plugin-sourcemap.test.ts
@@ -13,6 +13,22 @@ interface AssetInfo {
 }
 
 async function drivePlugin(assetsInfo: Map<string, AssetInfo>) {
+  // `lynxRstestConfig` sets `DEBUG=rspeedy` for every webpack test package so
+  // debugging artifacts survive a run. That makes `isDebug()` true, and
+  // WebEncodePlugin then inlines nothing and deletes nothing — the branch under
+  // test never runs. Drop it for this driver only, and put it back afterwards
+  // so the rest of the suite keeps the value the config intends.
+  const debug = process.env['DEBUG'];
+  delete process.env['DEBUG'];
+  try {
+    return await drive(assetsInfo);
+  } finally {
+    if (debug === undefined) delete process.env['DEBUG'];
+    else process.env['DEBUG'] = debug;
+  }
+}
+
+async function drive(assetsInfo: Map<string, AssetInfo>) {
   const deleted: string[] = [];
   let processAssets: (() => void) | undefined;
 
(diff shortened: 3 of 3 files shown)


upupming (dismissed):

upupming:
@iPLAYCAFE-dev LGTM, It seems CI is failed, could you fix it?

upupming (approved):

upupming:
@iPLAYCAFE-dev Thanks! Canary version will be publish in main CI: https://github.com/lynx-family/lynx-stack/actions/runs/30246592526/job/89914941653, try it out.

iPLAYCAFE-dev:
Thanks for the quick review and merge, @upupming! I'll try the canary against the downstream template that hit this and report back with the actual build output — that exercises the real path (`output.sourceMap.js: 'source-map'` on the web target) rather than just the unit level.

One finding from fixing the CI failure that may be worth knowing more generally, since it isn't specific to my test:

`lynxRstestConfig` sets `DEBUG: 'rspeedy'` for every webpack test package ([`rstest-config.ts#L124`](https://github.com/lynx-family/lynx-stack/blob/main/packages/webpack/test-tools/src/rstest-config.ts#L124)), so `isDebug()` is **always true** inside those suites. Any test that means to exercise a `!isDebug()` branch therefore runs the *other* side of the branch and quietly asserts against a no-op — which is exactly what happened here: nothing was inlined, nothing was deleted, and all three assertions saw an empty array.

That makes `WebEncodePlugin`'s asset-deletion path unreachable from tests by default. I opted out locally in the driver and restore the value in a `finally`, so the rest of the suite keeps the config's intent:

```ts
const debug = process.env['DEBUG'];
delete process.env['DEBUG'];
try { return await drive(assetsInfo); }
finally {
  if (debug === undefined) delete process.env['DEBUG'];
  else process.env['DEBUG'] = debug;
}
```

If you'd prefer this be a first-class option on `lynxRstestConfig` (something like `debug: false`) rather than each test poking `process.env`, I'm happy to send that as a follow-up.

upupming:
@iPLAYCAFE-dev It is because most of our test cases need the intermediate js files, i think keep it and opt out when needed is okay here~

iPLAYCAFE-dev:
@upupming I tried the canary against the downstream app as you asked, and I have to report that **this fix does not work** — sorry.

With `output.sourceMap.js: 'source-map'` on the web target:

| build | `.map` assets emitted |
|---|---|
| plugin `0.13.0` (before this PR) | 0 |
| canary `0.14.1-canary-20260727-7795a437` (this PR) | **0 — unchanged** |

The reason is that `updateAsset`'s info updater does not clear `related` on rspack. Reading it back immediately after the call still returns the map name, so `deleteAsset` cascades exactly as before:

```
main/background.c1ec9593.js   relatedAfterUpdate="main/background.c1ec9593.js.map"  mapsLeft=5
main/main-thread.js           relatedAfterUpdate="main/main-thread.js.map"          mapsLeft=4
```

My test could not catch this because its `updateAsset` mock applied the info updater faithfully — the one behaviour rspack does not have. That was my mistake, and it is the part I'd most want reviewed next time.

#3256 replaces the approach: take a reference to the map before the delete and re-emit it afterwards, which depends on nothing but `emitAsset`. Same real build now emits 6 maps including `main/background.<hash>.js.map`. The test there asserts on the final asset set instead, and I verified it fails against this PR's implementation.

#2964 should stay open until #3256 lands.

PR #3254, opened 2026-07-27
feat: set dataset when changed

HuJean:
## Summary by CodeRabbit

* **Bug Fixes**
  * Prevented redundant native dataset updates when spreading `data-*` props without meaningful dataset content changes.
  * Updated equality comparisons to use the correct new/old operand order, improving when attribute and platform-info updates are skipped.
* **New Features**
  * Improved deep equality handling for `undefined` and nested objects, plus clearer circular-structure detection.
* **Tests**
  * Expanded coverage for dataset add/remove/change semantics and circular-reference behavior (including ensuring no unnecessary native attribute calls).
* **Documentation**
  * Refreshed runtime equality guidance to match the updated comparison rules.

## Checklist


packages/react/runtime/src/snapshot/snapshot/spread.ts
```diff
@@ -211,24 +220,24 @@ function updateSpread(
...
-    // collect data regardless of whether it has changed
+  for (const key in oldValue) {
     if (key.startsWith('data-')) {
-      dataset[key.slice(5)] = v;
+      // dataset key removed
+      if (!datasetChanged) {
+        datasetChanged = !(key in newValue);
+      }
+      continue;
     }
```
upupming: This is inside loop now, will it be slower?

Issue #3211, opened 2026-07-26
Playwright: 20 retries is hiding a race in `tests/reactlynx.spec.ts`


upupming:
`playwright.common.ts` sets `retries: isCI ? 20 : 0`. Twenty attempts hides how unreliable the suite actually is, and paying for them is a large part of why a CI round takes as long as it does.

## What a run looks like

Two PRs from 2026-07-25, neither touching anything web-platform:

| PR | failed | flaky | passed |
|---|---|---|---|
| #3181 | 1 | **79** | 452 |
| #3182 | 1 | **65** | 466 |

"flaky" means the test needed at least one retry. So roughly one test in seven does not pass first time. The single failure was a different test in each run — `api-dispose` on chromium in one, `basic-lazy-component-when-needed` on webkit in the other — which is the signature of exhausted retries rather than a real regression.

Every retry line in both runs came from the same file:

```
tests/reactlynx.spec.ts
```

That matters: this is not the suite being broadly unreliable, it is one file with what looks like a systemic race.

## What it costs

`Playwright Web Elements Test (2/2)` is the longest job in a run at **17 min**, against a `globalTimeout` of 20 min — so the suite is running close to the ceiling, and a bad round tips it over into a timeout, which GitHub then reports as `cancelled` and `Done` reads as a failure.

There is a second cost that is easy to miss: a PR reported red for an exhausted-retry test gets re-run, and the whole round is paid again. Over a single day of merging Renovate PRs this accounted for several full CI rounds.

## Why not just lower `retries`

Lowering it today converts the currently-hidden flakiness into hard failures on everyone's PRs. The number is high because the tests need it — so the order has to be the other way round: make `tests/reactlynx.spec.ts` deterministic, then bring `retries` down to something that catches genuine infrastructure blips (2–3) rather than masking races.

Worth deciding as part of that: at 20 retries a test that passes one time in twenty is indistinguishable from one that passes nineteen times in twenty. Both are reported green. Whatever number replaces it should be low enough that "flaky" in the report means something.

## Suggested order

1. Instrument `tests/reactlynx.spec.ts` to find which assertions are racing — the retry counts per test are already in the run output.
2. Fix those.
3. Lower `retries` and let the report speak.

PR #3256, opened 2026-07-27
fix(template-webpack-plugin): re-emit the sidecar map instead of detaching it

### LINKED ISSUE #2964: Web platform: emit background-JS source maps as sidecar assets for the .web.bundle (production symbolication)

iPLAYCAFE-dev:
Follow-up to #3250. **#3250 does not fix #2964** — I verified it against a real build and the `.map` is still gone. Sorry for the noise, @upupming; this replaces that approach.

## What #3250 got wrong

It cleared `related.sourceMap` with `updateAsset` before `deleteAsset`:

```ts
compilation.updateAsset(name, source => source, info => ({
  ...info,
  related: { ...info.related, sourceMap: undefined },
}));
compilation.deleteAsset(name);
```

On rspack the info updater does not clear `related`. Reading it back immediately after the call returns the original value, so the delete cascade still fires:

```
main/background.c1ec9593.js   relatedAfterUpdate="main/background.c1ec9593.js.map"  mapsLeft=5
main/main-thread.js           relatedAfterUpdate="main/main-thread.js.map"          mapsLeft=4
.rspeedy/main/main.css        relatedAfterUpdate=".rspeedy/main/main.css.map"       mapsLeft=3
```

`mapsLeft` drops by one per delete. Setting `related: {}` wholesale does not help either.

## The fix

Take a reference to the map before the delete, put it back after. This depends on nothing but `emitAsset`.

## Verification

Real build, rspeedy 0.16.0 / rspack 2.1.2, web target, `output.sourceMap.js: 'source-map'`, canary `@lynx-js/template-webpack-plugin-canary@0.14.1-canary-20260727-7795a437` overridden into a downstream app:

| build | `.map` assets emitted |
|---|---|
| plugin `0.13.0` (before #3250) | **0** |
| canary with #3250 merged | **0** — unchanged |
| this branch | **6**, including `main/background.<hash>.js.map` |

## Why the old test passed

Its `updateAsset` mock applied the info updater faithfully — the one behaviour rspack does not have — so the test could only ever confirm the intent, never the effect.

The test now models the measured semantics and asserts on the **final asset set** rather than on which names reached `deleteAsset`. An implementation that loses the map through the cascade fails it. I checked that directly: against #3250's implementation the suite reports `failedTests: 1`, `expected false to be true` on `assets.has('background.js.map')`.

The two neighbouring tests still pass under both implementations, which is the point — they guard the opposite error of keeping the inlined JS around.

## Summary by CodeRabbit

- **Bug Fixes**
  - More reliably preserved background/sidecar source maps when encoded chunks are inlined into web bundles.
  - Prevented source-map assets from being accidentally removed during inlined chunk cleanup, improving production stack-trace symbolication.

- **Tests**
  - Added source-map fixtures and enhanced build assertions to confirm inlined outputs are removed while the corresponding `.map` assets remain available.
  - Introduced additional webpack/rspack test coverage for source-map preservation behavior.


.changeset/web-encode-sourcemap-reemit.md
```diff
@@ -0,0 +1,21 @@
+---
+"@lynx-js/template-webpack-plugin": patch
+---
+
+Actually keep the background sidecar source map on the web target.
```
upupming: We should remove this changeset, since we have added in the last PR. You can modify the old changeset if it is not accurate now.

packages/webpack/template-webpack-plugin/test/web-encode-plugin-sourcemap.test.ts
```diff
@@ -12,7 +12,12 @@ interface AssetInfo {
   related?: { sourceMap?: string | undefined };
 }
 
-async function drivePlugin(assetsInfo: Map<string, AssetInfo>) {
+interface Asset {
+  source: string;
+  info: AssetInfo;
+}
+
+async function drivePlugin(assets: Map<string, Asset>, cascade = true) {
```
upupming: Instead mock the behavior of rspack, can we just add a case to `packages/webpack/template-webpack-plugin/test/cases/web`? it will use real rspack build and we can assert the final output.

iPLAYCAFE-dev:
Both done, thanks — the second one is the better test and I should have written it that way to begin with.

## Real build instead of a mock

`test/cases/web/source-map` now builds two entries with `devtool: 'source-map'`. `a` is inlined into the template and deleted; `b` is not.

The assertion names **`a/a.js.map` specifically**. Asserting "some `.map` exists" would pass without the fix, because `b`'s map survives either way — that is the trap the old mock fell into in a different form.

Driving that fixture config through rspack directly, on the two implementations:

| implementation | emitted assets |
|---|---|
| this branch | `a/a.js.map` \| `b/b.js` |
| #3250's detach | `b/b.js` — and the case reports `a/a.js.map was deleted with the asset it belongs to` |

So the case is red against the implementation that is currently on `main`, which is what makes it worth having.

I **deleted** `web-encode-plugin-sourcemap.test.ts` rather than keeping it alongside. Its `updateAsset` mock applied the info updater faithfully, which is the one behaviour rspack does not have — that mock is precisely what let #3250 pass three green tests while changing nothing about the output. Keeping it would keep a test that cannot fail for the reason that matters.

One caveat worth stating: the `cases/` harness does not run to completion in my local environment (the existing `web/all-in-one` and 13 other cases fail here identically, before my change — a local build-artifact issue, `web-core/binary/encode`). So I verified the behaviour by driving the fixture config through rspack directly, as above, and the committed snapshot entries were generated in that same environment. **CI is the authority on the snapshot** — if those two entries differ there, say so and I will take the CI values.

## Changeset

Removed the second one and folded it into `web-encode-preserve-background-sourcemap.md`. I also corrected that changeset's description — it still said the related link "is now detached before the delete", which is the approach that does not work.

iPLAYCAFE-dev:
Applied — and it is a real bug, not a style preference. I hit it while writing this case.

`new URL('.', import.meta.url).pathname` yields `/D:/...` on Windows, a leading-slash path rspack cannot resolve against. Measured on this fixture, same config either way:

| `context` | result |
|---|---|
| `new URL('.', import.meta.url).pathname` | 4 errors — `Can't resolve './a.js'`, no assets emitted |
| `fileURLToPath(new URL('.', import.meta.url))` | resolves — `ASSETS: a/a.js.map \| b/b.js` |

That is why I had to override `context` by hand to verify the fix at all, which I mentioned in the previous comment as "the harness does not run to completion in my local environment". This is a large part of the reason.

**It is not limited to this case.** Every other fixture in this package uses the same line — 7 in `template-webpack-plugin`, 19 across `packages/webpack/*/test/cases`. None of them can run on Windows.

I changed only the new case here, to keep this PR about the source map. Happy to send the sweep as a separate PR if you want it — it is a one-line change per fixture and would make the case suite runnable on Windows contributors' machines. Your call on whether that is worth the churn; you may well consider Linux-only local runs acceptable for this suite.

upupming (dismissed):

upupming:
@iPLAYCAFE-dev Could you fix this error: https://github.com/lynx-family/lynx-stack/actions/runs/30326709170/job/90175061668?pr=3256#step:7:28595

iPLAYCAFE-dev:
@upupming Thanks for running CI — and sorry, that error message is mine and it is misleading. Pushed `84301deb3` to make it legible; here is what I can and cannot yet establish.  ### What CI actually shows  The case has two assertions. Only the **first** fires:  ``` × a/a.js should have been inlined and deleted ```  The sidecar assertion does **not** fire, so `a/a.js.map` is present — the source-map behaviour this PR is about is working on your runner. What fails is that `a/a.js` itself is still in the emitted set.  ### Why that message is not trustworthy  `a/a.js` surviving has two causes, and they want opposite fixes:  1. the encode ran and inlined it, but `deleteAsset` did not take effect — the    regression the case exists to catch; 2. the encode never ran, so `inlinedAssets` was empty and there was nothing to    delete — a broken build wearing the regression's clothes.  The assertion printed the same sentence for both, so the CI log cannot tell us which one you are hitting. That is my mistake in writing the case.  I hit cause (2) reproducing locally — `web-core`'s `binary/encode/encode.js` was absent, the encode threw, and the case then blamed the delete:  ``` × Cannot find module '.../web-core/binary/encode/encode.js'     imported from .../web-core/dist/encode/encodeCSS.js × a/a.js should have been inlined and deleted ```  Your log contains **no** such module error, so your failure is *not* my local one. I did not want to send a patch based on that resemblance.  ### What `84301deb3` changes  Test fixture only — **no change to `WebEncodePlugin`.** The assertion now uses the presence of `a/template.js` to separate the two cases, and every message carries the full emitted asset list so the run stays diagnostic even if that signal is wrong:  - `a/a.js was inlined into a/template.js but NOT deleted; emitted: …` - `a/a.js survives because the encode never ran (no a/template.js) — a broken   build, not a delete regression; emitted: …`  The next run should say which it is, and the asset list should make the cause obvious either way.  ### Where I am stuck  I cannot reproduce your failure locally: building `web-core` here needs the Rust/WASM artifacts that your `Build (Ubuntu)` job produces, and I do not have that toolchain set up. So I would rather not guess at a `WebEncodePlugin` change — if the next run reports "inlined but NOT deleted", that is a real defect in this PR and I will fix it properly; if it reports "the encode never ran", the case needs to stop depending on a fully built `web-core`, and I will rework it that way.  Happy to take direction on which you would prefer.

upupming:
@iPLAYCAFE-dev Seems the error does not go away

iPLAYCAFE-dev:
@upupming The instrument did its job — and it says the problem is my test, not
only the plugin. Thank you for running it again.

## What the run answers

```
× a/a.js was inlined into a/template.js but NOT deleted;
  emitted: .rspeedy/a/tasm.json, a/template.js, a/a.js, a/a.js.map, b/b.js, b/b.js.map
```

`a/template.js` is present, so the encode ran — this is **not** the broken-build
case I hit locally. `a/a.js` survives, so the deletion did not take effect.

## The part I got wrong, and it matters more than the failure

`a/a.js.map` is in that list, and my second assertion did **not** fire. I
previously read that as "the source-map fix works on your runner". **That was
wrong.** A sidecar map can only be collaterally deleted by a delete that
actually happens. Since `a/a.js` was never deleted, the map's survival proves
nothing — **the map assertion in this case is vacuous, and the case as written
does not exercise the fix at all.**

So the honest status of this PR is worse than "one failing test": the test that
is supposed to prove the change is not currently capable of failing for the
right reason. I would rather say that than have it merged on a green tick later.

## What I can and cannot establish

`inlinedAssets` is only populated when `!isDebug() && !isDev && !isRsdoctor()`,
and it is filled from the last `encodeData.manifest` entry plus the lepus/css
chunks. `@lynx-js/test-tools`'s case runner uses `mode: 'none'`, so `isDev`
should be false and the inlining path should be live — which is why I cannot
explain the empty deletion by reading, and I am not going to guess at it a
second time.

I still cannot reproduce locally: a case run needs `web-core`'s Rust/WASM
artifacts that your `Build (Ubuntu)` job produces, and I do not have that
toolchain. I would rather ask than keep using your CI as my test runner.

**Two questions, either of which unblocks me:**

1. Under the case harness (`mode: 'none'`, no `DEBUG`, no `RSDOCTOR`), is
   `WebEncodePlugin`'s inline-and-delete path expected to run at all? If it is
   deliberately inert there, then no case fixture can test this and I should
   move the proof to a unit test over the `processAssets` tap with a synthetic
   compilation — which I can write and verify locally.
2. If it *is* expected to run, do you know what `encodeData.manifest`'s last key
   is for a web target with `output.filename: '[name]/[name].js'`? If it is not
   literally `a/a.js`, `deleteAsset` is silently no-oping on a name that does not
   exist, which would explain every symptom including the surviving map.

## Offer

If (1), I will replace the case fixture with a unit test and drop the fixture
entirely — smaller, faster, actually able to go red, and verifiable on my side
before you spend another CI run.

If you would rather I park this: the plugin change itself
(`re-emit the sidecar instead of detaching it`) is independent of the test, and
I am happy to split the test out so the fix is not held up by my fixture.

upupming:
@iPLAYCAFE-dev It's (1), and the reason is that the harness does set `DEBUG`.

`lynxRstestConfig` hardcodes `env: { DEBUG: 'rspeedy' }` for every webpack test package (`packages/webpack/test-tools/src/rstest-config.ts`), and `isDebug()` treats `'rspeedy'` as a match. So `!isDebug()` is false, `inlinedAssets` stays empty, and the deletion loop is a no-op. That explains every symptom, including the surviving map.

On (2): not a name mismatch. `encodeData.manifest` is keyed by asset name (`Object.fromEntries(assetsInfoByGroups.backgroundThread.map(a => [a.name, ...]))`), so it is literally `a/a.js`.

Note the answer was in the file you deleted — `web-encode-plugin-sourcemap.test.ts` had a comment saying exactly this.

A case fixture can't cover this unless it runs without `DEBUG=rspeedy`, and rstest's env is per-package, so please go with the unit test.

iPLAYCAFE-dev:
@upupming Thank you — that is the mechanism, and I verified it rather than just accepting it:

- `packages/webpack/test-tools/src/rstest-config.ts:124` — `DEBUG: 'rspeedy'`
- `packages/rspeedy/core/src/debug.ts:8-15` — `['rsbuild', 'rspeedy', '*'].some(...)`

So `!isDebug()` is false, `inlinedAssets` stays empty, the deletion loop is a no-op, and the map survives for a reason that has nothing to do with the fix. That accounts for every symptom, and it means my case fixture could never have covered this.

**And you are right that the answer was in the file I deleted.** Its `drivePlugin` helper documented exactly this — that the harness hardcodes `DEBUG=rspeedy`, that it makes the branch under test unreachable — and worked around it by unsetting `DEBUG` for the driver only. I removed a test that already knew the thing I then spent a round asking about, and replaced it with a fixture that structurally cannot work. I should have read what I was deleting before deciding it was redundant.

Pushed `7eb31be`, going with the unit test as you asked:

- restored `web-encode-plugin-sourcemap.test.ts`
- removed `test/cases/web/source-map/*` and reverted the snapshot
- the PR is now just the plugin fix, that test, and the changeset

Two changes inside the restored test, because the fix's mechanism is not the one it was written against — flagging them rather than burying them in a diff:

1. **The driver had no `emitAsset`.** Detaching `related.sourceMap` first does not work on rspack: `updateAsset`'s info updater does not clear `related`, so the cascade still fires. The fix instead takes a reference to the map before the delete and emits it back. The old driver could not observe that at all, so it now models `emitAsset` and a live asset set.

2. **The "related link is detached" case is replaced.** That assertion pinned the mechanism I had to abandon. The replacement pins the outcome — the map is in the **emitted** set — and deliberately also asserts that `related` *still points at the map* and that the map *was* deleted by the cascade, so it cannot pass for the wrong reason. The invariant is now "a developer gets a usable map", not "the link was cleared".

**One thing I could not do, so I am saying it rather than implying otherwise: I have not run this locally.** This machine is on Node 26 and the repo requires `^22 || ^24`, and a sparse checkout could not satisfy the workspace filter (lynx-stack also needs long-path checkout on Windows). I am relying on CI for the run. Given I have already been wrong once in this thread about what a green result meant, please treat my "this should pass" as a hypothesis — if CI disagrees, or if you would rather I set up a matching Node and re-post with a local run before you spend review time, say so and I will.

iPLAYCAFE-dev:
Branch updated to current `main` (`60aa9e2`, was 9 behind, now 0). Two notes, one of which is a correction to my own reasoning.

**CI is waiting on you, not on the branch.** Both workflows on my last two heads sit at `action_required`:

```
completed/action_required  Test
completed/action_required  Code Scanning
```

That is the fork-PR approval gate, so nothing on my side releases them — whenever you next look at this, it needs "Approve and run workflows". Earlier heads on this PR each carried 30 check-runs, so this is not a config change, just the gate re-arming on each new push from the fork.

**The correction:** I updated the branch partly expecting the fresh push to kick CI. It did not, and it could not have — the gate is per-push approval, so a new commit re-arms it rather than clearing it. The update was still worth doing for being 9 behind, but I had the mechanism wrong and would rather say so than let it look like a fix that worked.

Given that, and that I cannot run this locally (Node 26 here vs the repo's `^22 || ^24`), the test changes in `7eb31be` are still unverified by anything but reading. No rush on my account — I would rather you approve the run when it suits you than merge on my say-so.

upupming:
@iPLAYCAFE-dev I ran `7eb31be` locally before approving CI. Two things, and the second matters more.

**1. The test does not pass — 2 of 3 fail.**

```
failedTests: 2, passedTests: 1
× the sidecar .map survives while the inlined .js is still deleted
× the map is restored even though the related link still points at it
  expected false to be true   // live.has('background.js.map')
```

The fixture registers the `related` link but never registers the `.map` as an asset:

```ts
new Map([['background.js', { related: { sourceMap: 'background.js.map' } }]])
```

Your implementation reads the map through `getAsset(mapName)` to get its source. In the driver that returns `undefined`, so `mapSource` is undefined, the `if (mapName && mapSource && ...)` guard short-circuits, and nothing is ever re-emitted. The driver cannot pass against your own fix.

Seeding the map makes it 3/3:

```ts
new Map<string, AssetInfo>([
  ['background.js', { related: { sourceMap: 'background.js.map' } }],
  ['background.js.map', {}],
])
```

**2. With that fixed, the headline test is still green against `main`.**

I applied the fixture fix and swapped `WebEncodePlugin.ts` back to `main` (the #3250 detach you say changes nothing):

```
failedTests: 1, passedTests: 2
× the map is restored even though the related link still points at it
  expected undefined to be 'background.js.map'
```

`the sidecar .map survives while the inlined .js is still deleted` passes. It cannot go red for the right reason.

The cause is that the driver's `updateAsset` is unchanged and still applies the info updater faithfully:

```ts
updateAsset: (name, _source, update) => {
  assetsInfo.set(name, update(assetsInfo.get(name) ?? {}));
},
```

That is the behaviour you identified as the one rspack does not have, and the reason #3250 passed three green tests while changing nothing. Restoring the file restored that flaw too. The only test that fails is the one asserting `related` was *not* cleared, which pins #3250's fingerprint rather than the outcome.

Modelling the measured semantics fixes it:

```ts
const prev = assetsInfo.get(name) ?? {};
// rspack's info updater does not clear `related`
assetsInfo.set(name, { ...update(prev), related: prev.related });
```

With both changes: 2 failures against `main` (including the headline test), 3 passes against this branch.

Minor: the `emitAsset` mock discards `source`, so re-emitting the wrong source would stay green.

One question on scope: `inlinedAssets` also holds the lepus root, lepus chunks, CSS chunks and the `.rspeedy` intermediates, so their maps now survive too. Is that the intent? Could you list the 6 maps your real build emits?

iPLAYCAFE-dev:
@upupming Both correct, both applied verbatim in `f93ec70`. Thank you for running it — you have now done that twice for a change I authored, which is the wrong way round.

I verified your two findings against the code before applying them:

1. `getAsset` returns `undefined` for `background.js.map` because the fixture only ever named it through `related`. My own fix reads the source through `getAsset(mapName)`, so `mapSource` was undefined and `if (mapName && mapSource && …)` short-circuited. **The driver could not pass against the fix it exists to cover** — you are right, and that is a test that proves nothing in either direction.

2. `updateAsset` applying the info updater faithfully is webpack's behaviour, not rspack's.

**The second one is the one worth me sitting with.** I diagnosed #3250 as passing three green tests while changing nothing, because its driver granted `updateAsset` an effect rspack does not give it — and then I restored that same driver and built on it. So my replacement test agreed with the broken approach for exactly the reason the original did. Fixing a vacuous test with a fixture built on the same unmeasured assumption reproduces the defect one layer up, and "the map is restored" going green against `main` is the proof of that.

The rule I keep relearning: a fixture must model what the runtime **measurably does**, not what the API contract says it should. Your `{ ...update(prev), related: prev.related }` is that, stated in two lines.

**On the process, which I would like to fix rather than keep apologising for:** I have pushed three rounds to your PR without running any of them. This machine is on Node 26 and the repo requires `^22 || ^24`, and a sparse checkout could not satisfy the workspace filter — so I have been reasoning about test outcomes instead of observing them, in a thread whose entire subject is a test that looked like it ran and did not. That is not a defensible position to keep taking.

So: I would rather not consume another review cycle on unverified work. If you would like, I will set up a matching Node and full checkout and re-post only once I have a local run with the headline test shown red against `main` and green against the fix. Alternatively, if the change is now small enough that you would rather just take it, say so and I will stop touching it. Your call — you have already spent more of your time on this than I have earned.

iPLAYCAFE-dev:
@upupming I set up a matching toolchain and ran it. `a6ad0f3`. Results below are observed, not reasoned.

**Both directions, measured:**

```
against this PR's WebEncodePlugin.ts   → rstest exit 0   failedTests: 0, passedTests: 3
against upstream main's WebEncodePlugin.ts → rstest exit 1   failedTests: 2, passedTests: 1
  × the sidecar .map survives while the inlined .js is still deleted
      expected false to be true            // live.has('background.js.map')
  × the map is restored even though the related link still points at it
```

The headline case now fails against `main` for the right reason. Your `updateAsset` correction is what made that possible — with the faithful updater it passed either way, exactly as you showed.

**One thing your snippet needed on the way in.** Applied literally:

```ts
assetsInfo.set(name, { ...update(prev), related: prev.related });
```

does not compile in this package — `exactOptionalPropertyTypes: true` rejects an explicit `related: undefined` against an optional property:

```
error TS2379: Argument of type '{ related: {...} | undefined; }' is not assignable
to parameter of type 'AssetInfo' with 'exactOptionalPropertyTypes: true'
```

So it re-attaches only when it was present. `tsc -b` now reports **0 errors in `web-encode-plugin-sourcemap.test.ts`**. It still reports errors in `basic.test.ts`, `lazy-bundle-fetcher.test.ts` and `web-encode-plugin-fetchbundle.test.ts` — all webpack-vs-rspack `Compilation`/`Compiler` type mismatches plus a missing `css-extract-webpack-plugin` build, none of them files this PR touches. Flagging in case that is news; I have not gone near them.

**On the process.** You ran my change twice, and the second time you found that my driver could not pass against my own fix. That should not have been your job. The blocker was that this machine's Node did not satisfy the repo's `^22 || ^24`, so I had been reasoning about test outcomes in a thread whose entire subject is a test that looked like it ran and did not — which is the same error one level up, and I am glad you did not let it through.

That is fixed now: pinned Node 24.18.0, full checkout, `tsc -b …/tsconfig.build.json` then `rstest run`. I will not post another round here without a local red/green pair.

Ready for CI whenever suits you.

iPLAYCAFE-dev:
CI note on the red checks, in case it saves triage time.

`Playwright CSR (1/2) / check` (run 30451906117) finished `1 failed, 34 flaky, 31 skipped, 497 passed`. The single hard failure is `[webkit] tests/reactlynx.spec.ts:1577:5 › apis › api-dispose` at line 1590, `expect(message).toContain('fin')` receiving `[]`. chromium flaked on the same test in the same run, so it does not look webkit-specific.

The same test shows up the same way on branches that cannot plausibly affect it — e.g. `renovate/pnpm-11.x` (run 30391283520) finished `1 failed, 115 flaky` with `api-dispose` in the list. That matches what #3211 describes.

The other two reds look like collateral: `Done` is the aggregator for that job, and CodSpeed is red on most other open PRs right now.

This PR touches only `WebEncodePlugin.ts`, its unit test, and the changeset — nothing under `packages/web-platform/`.

@upupming would you mind re-running just `Playwright CSR (1/2) / check`? If it reproduces I'll dig in rather than leave it to you.

upupming (approved):
Thanks for your contribution!

PR #3262, opened 2026-07-27
feat(web-platform): support `__GetComputedStyleByKey` element PAPI

### LINKED ISSUE #1758: [Feature]: Support `__GetComputedStyleByKey` PAPI on Web Platform

See documentation at https://github.com/lynx-family/lynx-website/pull/372

```js
function __GetComputedStyleByKey(element: Element, key: string): string;
```

Huxpro:
Implements the `__GetComputedStyleByKey` element PAPI on the Web Platform.

```ts
function __GetComputedStyleByKey(element: HTMLElement, key: string): string;
```

Returns the computed value of a CSS property, using dash-notation property names (`background-color`, `font-size`, …).

`__GetComputedStyleByKey` is the PAPI backing `element.getComputedStyle` in main-thread scripts. It is already consumed by `@lynx-js/react` (`worklet-runtime/api/element.ts`) and `@lynx-js/motion` (`polyfill/element.ts`), and both declare it in their PAPI type definitions — but no web-platform implementation existed, so on web the call landed on an undefined global. `ElementPAPIs` had no computed-style member of any name, and the mts realm has no dynamic fallback.

Closes #1758. Supersedes #2033.

## Changes

- **`@lynx-js/web-core`**
  - `ts/types/IElementPAPI.ts` — add `GetComputedStyleByKeyPAPI` and the corresponding `ElementPAPIs` member.
  - `ts/client/mainthread/elementAPIs/` — implement via `getComputedStyle()` and wire it into `createElementAPI`. No extra registration is needed: `LynxViewInstance` `Object.assign`s the whole returned object onto the mts realm.
  - `ts/server/elementAPIs/createElementAPI.ts` — return an empty string during SSR. `ElementPAPIs` is implemented by both the client and the server, so a new required member needs both; SSR has neither layout nor cascade resolution to resolve a computed value against, so this follows the existing stub convention (cf. `__ElementAnimate`).
- **`@lynx-js/testing-environment`** — mirror the implementation so unit tests can exercise the PAPI, and refresh the API report.

## Notes

The client implementation resolves the window via `element.ownerDocument.defaultView` rather than the global `window`, so it uses the element's own window and degrades to an empty string for elements owned by a document with no browsing context.

Once this lands, the `web_lynx` support entry for `lynx-api/main-thread/ElementGetComputedStyle` in `lynx-compat-data` should be flipped from `false` to `true`.

## Checklist

---
_Generated by [Claude Code](https://claude.ai/code/session_01C6vAE45r2jjKXUy7mXZfDw)_

## Summary by CodeRabbit

* **New Features**
  * Added support for retrieving computed CSS values from elements via property name.
  * Works with dash/kebab-case CSS keys (e.g., `margin-top`); unknown keys and non-kebab-case keys return an empty value.
  * Returns an empty value when computed-style lookup isn’t available (including server-side rendering).
* **Tests**
  * Added main-thread and SSR test coverage for the new computed-style lookup behavior.


Huxpro:
Thanks — two CodeRabbit points, both checked and intentionally left as-is.

**1. `lynxTestingEnv.env.window` vs `ele.ownerDocument.defaultView` in the testing-environment bridge**

The ownership invariant holds. Every element the testing environment can produce is created from `lynxTestingEnv.env.window.document` — `ElementPAPI.ts` routes all element creation through it (`__CreatePage`, `__CreateRawText`, `__CreateElement`), and the file uses `ownerDocument` nowhere. Reaching through `lynxTestingEnv.env.window` is this file's established convention, not an oversight.

I also checked the one case that could break the invariant — an element created from the global `document` rather than the env's — and it does not diverge:

```
globalThis.document === lynxTestingEnv.env.window.document   // true
element.ownerDocument === lynxTestingEnv.env.window.document // true (element created globally)
lynxTestingEnv.env.window.getComputedStyle(el)…              // "rgb(255, 0, 0)"
el.ownerDocument.defaultView?.getComputedStyle(el)…          // "rgb(255, 0, 0)"
```

So the two forms are observably identical here, and switching would make this method the only one in the file that bypasses the env window. The web implementation uses `ownerDocument.defaultView` for a different reason: it operates on real DOM elements in a live document, where a caller genuinely can hold an element from a document with no browsing context — that path is covered by a test.

**2. Docstring coverage warning**

This is a generic threshold rather than repo practice: `pureElementPAPIs.ts` has 0 JSDoc blocks across its 28 exported PAPIs, and `IElementPAPI.ts` has 1 across 62 PAPI types. Adding docstrings only to the new members would make them inconsistent with every neighbour. Happy to document the whole surface separately if that is wanted.

---

Huxpro:
@copilot Thanks — the observation is correct, but the proposed fix would break the documented contract, so I've gone the other way and pinned it with tests instead (855d8a03).

**The API is specified as kebab-case only.** From `docs/en/api/lynx-api/main-thread/element-get-computed-style.mdx`:

> **styleName** — A string specifying the property name to be retrieved. The style name must be in [kebab-case](https://developer.mozilla.org/en-US/docs/Glossary/Kebab_case).

And the supported-property list in `lynx-compat-data` (`elements/get-computed-style-supported-properties.json`) is kebab-case throughout — `background-color`, `transform-origin`, `z-index`, … — with zero camelCase entries.

Accepting camelCase on web would make the web platform *more* lenient than the spec. Code written against it would then break on Android/iOS, which is the opposite of what this compatibility layer should do. Confirmed the underlying behavior empirically:

```
getPropertyValue('font-size')       → "16px"
getPropertyValue('fontSize')        → ""
getPropertyValue('background-color')→ "rgb(0, 0, 255)"
getPropertyValue('backgroundColor') → ""
```

So instead of widening the implementation, I added assertions in both suites that camelCase resolves to `''`, so the contract can't drift silently.

**One correction on the evidence, and one real bug it surfaced.** The `react/runtime` call site is not a camelCase requirement — `getComputedStyleProperty` passes `key` straight through, and its test *mocks* `__GetComputedStyleByKey` with a canned return value, so `'fontSize'` there is arbitrary test input rather than a behavioral contract.

The genuine camelCase call sites are in `@lynx-js/motion`:

- `packages/motion/src/polyfill/element.ts:69` — `getStyleProperty(name)`, called as `this.getStyleProperty('backgroundColor')` and similar
- `packages/motion/src/polyfill/element.ts:25` — the `getComputedStyle()` Proxy forwards whatever property name the consumer accesses

Those pass camelCase into an API documented as kebab-case, so they look incorrect against the spec regardless of platform. That is a `@lynx-js/motion` bug rather than something this PR should paper over — worth a separate issue.

---

f0rdream (approved):

PupilTong (approved):

PR #3266, opened 2026-07-27
test: give every case fixture a context that resolves on Windows

iPLAYCAFE-dev:
Split out of #3256 at review request. That PR fixes only the fixture it adds; this is the sweep.

## The bug

`new URL('.', import.meta.url).pathname` yields a **leading-slash** path on Windows — `/D:/GitHub/...` — which rspack cannot resolve against. Every case fixture in `packages/webpack/*` uses that line, so none of them build on a Windows machine.

Measured on `web/all-in-one`, same tree otherwise:

| | result |
|---|---|
| before | **FATAL** — the fixture's own `processAssets` plugin throws, because nothing resolved and `compilation.getAsset()` returned `undefined` |
| after | `resolveErrors=0  assets=2` |

Same for `web/default-export` and `main-thread/mpa`. The failure is loud but misleading: it surfaces as a plugin crash rather than as "cannot resolve", so it does not obviously point at `context`.

## The sweep

28 usages across 26 files, in three shapes:

| shape | count |
|---|---|
| `context: new URL('.', import.meta.url).pathname` | 18 |
| `const __dirname = new URL('.', import.meta.url).pathname` | 5 |
| `dirname(new URL(import.meta.url).pathname)` | 5 |

All become `fileURLToPath(...)`, which is precisely what that helper exists for. `git grep "import.meta.url).pathname"` is now empty under `packages/`.

Three packages are touched — `template-webpack-plugin`, `react-refresh-webpack-plugin`, `externals-loading-webpack-plugin` — and only files under `test/`.

## Notes

- **Test-only**, so the changeset is empty: no package behaviour changes and nothing should be version-bumped. Say the word if you would rather have a patch changeset or none at all.
- I corrected my own count in #3256 along the way: I said 19 there, from a glob that missed `hotCases/`. The real figure is 26 files.
- CI runs on Linux, where `.pathname` happens to be a valid path, so this has never failed there and will not change CI behaviour. The value is that a Windows contributor can run the case suite at all.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved cross-platform path handling in webpack/rspack test fixtures and diagnostics (including Windows).

* **Tests**
  * Added Windows-only rstest coverage.
  * Improved test reliability for environments without native TASM support.
  * Updated CSS source-file/source-map assertions for platform-correct paths.

* **Chores**
  * Expanded CI completion gating to include new test and validation suites.
  * Added Preact dedupe checks and a sherif validation step.
  * Added a fixtures-only changeset with no package behavior/version impact.


upupming:
@iPLAYCAFE-dev Great! Could you also add Rstest for windows in `.github/workflows/test.yml`, I think we forgot to add it before, so we dismissed this bug.

iPLAYCAFE-dev:
@upupming Done in `2e33bdaa9` — and you are right that the missing job is why
this slipped through.

`test-vitest` has had an Ubuntu/Windows pair for a while; `test-rstest` never
got one, so **nothing in CI ever executed the rstest suite on Windows**. The
case fixtures in this PR are only reached by rstest, so the `import.meta.url`
path bug had no job that could observe it.

`test-rstest-windows` mirrors `test-vitest-windows` exactly — same reusable
workflow, same run command as the Ubuntu job byte for byte; only
`needs: build-windows` and the runner differ:

```yaml
  test-rstest-windows:
    needs: build-windows
    uses: ./.github/workflows/workflow-test.yml
    secrets:
      CODECOV_TOKEN: ${{ secrets.CODECOV_TOKEN }}
    with:
      runs-on: lynx-windows-2022-large
      run: >
        pnpm exec rstest run
        -c rstest.config.ts
        --coverage
        --testTimeout=50000
        --hookTimeout=50000
```

It is also added to **`done.needs`**. Without that the aggregate `Done` check
would go green while ignoring the new job — which would have reproduced the
original problem in a new place: a job that exists but gates nothing.

I left `test-rstest` untouched rather than renaming it, so no existing required
check changes name.

I verified the YAML parses, that `needs: build-windows` resolves to a real job,
that the run command matches the Ubuntu one exactly, and that `done.needs`
includes it. I could **not** run the suite locally to confirm it passes on
Windows — the repo requires Node `^22 || ^24` and I had to stand up a Node 24
side-by-side just to get past `ERR_PNPM_UNSUPPORTED_ENGINE`, and a full local
run additionally needs `web-core`'s Rust/WASM artifacts. So this run is the
first real signal; if Windows surfaces further path assumptions in other
fixtures, I am happy to chase them in this PR.

iPLAYCAFE-dev:
Updated to current `main` (`3cc58b6`, was 9 behind).

For the record on the red checks, since they read worse than they were: **nothing failed a test.** Run `30329703102` was `conclusion=cancelled`, so `Build (Windows)` was cancelled, `test-rstest-windows` and `Vitest (Windows)` were skipped, and `Done` exited 1 aggregating that. The five `bundle-size` entries are `neutral`. CodSpeed went red downstream of the same cancellation.

Which matters more than usual here, because this PR is *about* Windows path resolution — the jobs that would actually exercise it are precisely the ones that never ran.

The update re-armed the fork approval gate, same as on #3256:

```
completed/action_required  Test
completed/action_required  Code Scanning
```

So both are waiting on the same click whenever you get to them. No urgency from my side.

iPLAYCAFE-dev:
@upupming — tagging you since my previous note had no mention and you may not have seen it.

Short version: this PR's red checks were a **cancelled run**, not a test failure — `Build (Windows)` cancelled, the two Windows test jobs skipped, `Done` aggregating that into `exit 1`. Since the PR is about Windows path resolution, the jobs that matter never ran.

Branch is now current with `main`, and both this and #3256 are sitting at `action_required` for the fork workflow gate. Whenever it suits you.

iPLAYCAFE-dev:
@upupming this one is still waiting on **Approve and run workflows** — no Actions run has started on it since it was opened, so there is no CI signal either way yet. Whenever you have a moment, could you approve the run?

Happy to rebase or split it if that makes review easier.

upupming:
@iPLAYCAFE-dev There is a regression on Windows, can you fix it also?
image

iPLAYCAFE-dev:
@upupming Fixed in ffcb59e — and thanks for catching it, because one of the six was not a test problem.

**The real one.** `normalizeTasmSourcePath` passed every `file://` source straight to `fileURLToPath`, which throws on win32 when the URL carries no drive letter (and on POSIX when it carries a non-localhost host). The function declares `string | undefined` and every other branch degrades to `undefined`, so the throw escaped an advisory code path and took the whole compilation with it:

```
TypeError: File URL path must be absolute
  at normalizeTasmSourcePath (src/cssDiagnostics.ts:318)
```

It now degrades like the rest of the function, with a patch changeset. The new test uses two URL shapes — drive-less and non-localhost-host — so exactly one is unconvertible per platform, and removing the guard turns the test red on Linux as well as Windows rather than only where you happened to see it.

**The other five** were expectations that hard-coded one platform's path shape:

- `cssDiagnostics.test.ts` ×3 — a `path.resolve` result compared against a POSIX literal; now derived with `resolvePath(CONTEXT, …)`
- `basic.test.ts` ×2 — `` `${context}/a.css` `` built the separator by hand; now `join(context, 'a.css')`

For the record `cssDiagnostics.test.ts` is not touched by this PR — it was already broken on Windows and enabling the job is what surfaced it. Fixing it here because the job is mine.

Verified locally on win32: those two files went 6 failed → 15 passed. Node 26 locally, so CI on 22/24 is the real confirmation.

**One question before you merge.** Now that the fixtures resolve, the `templateCases-cases/web/*` cases actually run, and they produce snapshot entries that are not in `__snapshots__/cases.test.ts.snap` on `main`. I deliberately did **not** commit mine: my checkout is missing `web-core/binary/encode/encode.js`, so anything those runs recorded is untrustworthy. Would you like those snapshots committed from a full build, or left for CI to generate?

upupming:
@iPLAYCAFE-dev 
> my checkout is missing web-core/binary/encode/encode.js

Can you merge the latest main branch?

upupming:
@iPLAYCAFE-dev Two things.

## 1. The Windows failure is in a file your branch does not have

`test-rstest-windows` fails only on `test/encode-worker.test.ts:11`, which builds a path the same broken way (`new URL(...).pathname` → `C:\C:\...`). That file came from #3285, which landed an hour after your last `main` merge:

```
your main merge   3cc58b62   2026-07-29 21:05 (+0800)
#3285             43156341   2026-07-29 22:06 (+0800)
```

CI tests your head merged with current `main`, so it runs a file you cannot see locally. Merge `main`, apply `fileURLToPath` there.

## 2. The `normalizeTasmSourcePath` try/catch is fixing a fixture, not a bug

These source maps are rspack's own (`cssChunk.source.map()` via `LynxEncodePlugin.ts:287`), whose `sources` use `webpack://`. On Windows real paths carry a drive letter, so a `file://` URL there is `file:///C:/...` — convertible. A drive-less `file:///src/app.css` is not something a real build produces.

It is `cssDiagnostics.test.ts:128` on `main`, a hand-written fixture with a POSIX-shaped URL — the same class as the five you already fixed.

The guard also merges two different failures: `if (!sourceFile || !fileExists(sourceFile))` now treats a malformed URL exactly like a missing file, so a real occurrence would surface as a diagnostic with no location and no explanation.

Fix the fixture instead, no production change:

```ts
sources: [pathToFileURL(resolvePath(CONTEXT, 'src/app.css')).href],
```

That test's subject is "path resolves, file is absent", and `fileExists: () => false` already provides it. Then the `try`/`catch`, the new test and `css-diagnostics-file-url-guard.md` can all go.

If you think a real build can emit an unconvertible `file://`, that needs output from an actual build — and the fix would then want to warn, not silently degrade to "file not found".

iPLAYCAFE-dev:
@upupming You're right on both. Reverted in 20a8007 — `src/cssDiagnostics.ts` is byte-identical to upstream again, and the guard test and its changeset are gone.

**On the guard.** What settled it: `git grep "file://"` across this package's `src/` matches the branch itself and nothing else. There is no producer. I inferred a production bug from a hand-written fixture, which is exactly the thing that should never justify a production change.

Your third point stands on its own too, and I'd missed it: returning `undefined` fed `if (!sourceFile || !fileExists(sourceFile))`, so an unconvertible URL became indistinguishable from an absent file. Even if the bug were real, that remedy would have hidden it. The fixture now says what the test is about:

```ts
sources: [pathToFileURL(resolvePath(CONTEXT, 'src/app.css')).href],
```

**On `main`.** Merged. `encode-worker.test.ts:11` got the same `fileURLToPath` treatment, and it is now the only `import.meta.url).pathname` left under `packages/`.

That exposed a second Windows failure, unrelated to paths, and I'd like your call on how I handled it. `prints the TASM encode trace when enabled` calls `encode()` **without** `tasmPkg`, so it exercises the real `@lynx-js/tasm@0.0.48`. That package ships prebuilt binaries for two platforms only — `build/darwin`, `build/linux` — behind its own gate:

```js
process.platform === 'darwin' || (process.platform === 'linux' && process.arch === 'x64')
```

Everywhere else `getEncodeMode()` returns the wasm path, and neither `[TASM Encode Trace]` nor `console.table` appears anywhere in tasm's JS — they come from the native binding. So on Windows the test sees exactly one log line, the `End!` that `encode.ts` writes in its own `finally`. That reads to me as a missing platform capability rather than a bug, and not something an assertion rewrite can reach, so I gated that single test on the same condition tasm uses. The other cases in the file use the mock TASM and still run everywhere. Happy to drop it if you'd rather the job stayed red there, or handle it differently.

Note this also excludes linux-arm64, not just Windows.

**On the missing binary** — merging `main` couldn't have fixed that one, and it's worth knowing why: `packages/web-platform/web-core/.gitignore` has `**/binary/**`, only the `.d.ts` are tracked, and `encode.js` comes from `pnpm build:wasm`. CI builds it, which is why the Windows job never failed those four `templateCases-cases/web/*` cases even though my checkout does.

Verified on win32: cssDiagnostics + basic + encode-worker = **20 passed, 1 skipped, 0 failed**. Whole package = 89 passed, 1 skipped, 4 failed — the 4 being those wasm-dependent cases.

iPLAYCAFE-dev:
@upupming One thing needs you: the runs for `20a8007` came back `action_required`, so CI is waiting on a maintainer approval rather than queued.

```
20a800705  Test           completed/action_required   06:39
ffcb59e2f  Test           completed/failure           03:52   ← ran on its own
```

No rush — just flagging it, since nothing will move until someone clicks approve.

upupming:
Revert looks right — `cssDiagnostics.ts` is byte-identical to upstream, and the fixture now says what the test is about. One thing on the new gate.

`hasNativeTasm` copies a predicate tasm already exports:

```ts
// @lynx-js/tasm/index.d.ts
export function supportNapi(binary?: string): boolean;
```

```ts
import { supportNapi } from '@lynx-js/tasm';
const testWithNativeTasm = supportNapi() ? test : test.skip;
```

If tasm ever ships win32 or linux-arm64 binaries, the copy goes stale silently — it fails by skipping a test that should run, so nothing turns red and nobody notices.

Minor: the commit message says the real tasm here is `0.0.48`, but this package is on `0.0.49` since #3285. Same `supportNapi`, so the gate is correct either way.

iPLAYCAFE-dev:
@upupming Done in 2fae8e1 — and you're right about the failure direction, which is what makes it worth fixing: a stale copy fails by *skipping*, so nothing goes red and nobody finds out.

```ts
import { supportNapi } from '@lynx-js/tasm';
const testWithNativeTasm = supportNapi() ? test : test.skip;
```

I checked it's a real runtime export rather than a `.d.ts` declaration — `index.js` lists `supportNapi` in `module.exports`, and calling it here returns `false` on win32 x64, so the skip now comes from asking tasm instead of from an import failing quietly. Unchanged result: 20 passed, 1 skipped, 0 failed.

On the version: you're right and my commit message was wrong. The declared dependency is `0.0.49` since #3285 — `0.0.48` is what my checkout still has installed, because I merged `main` without reinstalling. I described my `node_modules` and wrote it as if it were the package. Corrected in the new commit message.

One thing still outside my reach: CI for `20a8007` came back `action_required`, so the runs are waiting on a maintainer approval rather than queued. Same will apply to this push.

upupming (approved):
Thanks!

PR #3274, opened 2026-07-28
feat: support experimental_transformBuiltinAttributeNames

HuJean:
## Summary by CodeRabbit

* **New Features**
  * Added an experimental compile-time option to transform Lynx builtin JSX attribute names (camelCase to dash-case) with boolean defaults or custom serializable rules (rename/preserve/mapping-only).
  * Enabled the option through the React build plugins and loader pipeline.
  * Added TypeScript intrinsic typings with React-style camelCased prop aliases.
* **Documentation**
  * Updated benchmark and transform instructions to match the new validation and aliasing requirements.
* **Bug Fixes**
  * Improved Lynx event attribute recognition and standardized React/legacy event prop conversions.
* **Tests**
  * Expanded coverage for default behavior, custom rule precedence, and preservation of spread attributes; updated benchmark cases.

## Checklist


benchmark/react/types/index.d.ts
```diff
@@ -0,0 +1,34 @@
...
+  >;
+};
+
+declare module '@lynx-js/react/jsx-runtime' {
+  export namespace JSX {
+    interface IntrinsicElements {
+      text: CamelCaseIntrinsicElements['text'];
+    }
+  }
+}
```
upupming: So user needs define this manually?

benchmark/react/cases/012-attrs-compile/index.tsx
```diff
@@ -13,13 +13,13 @@ function App() {
...
           text-maxlength='128'
-          enable-font-scaling={false}
+          enableFontScaling={false}
           text-vertical-align='center'
-          tail-color-convert={false}
+          tailColorConvert={false}
           include-font-padding={false}
-          text-fake-bold={false}
+          textFakeBold={false}
           text-selection={false}
```
upupming, line 24: Shall we also add `onClick` example?

benchmark/react/cases/012-attrs-compile/index.tsx
```diff
@@ -13,13 +13,13 @@ function App() {
     <view>
       {Array.from({ length: ELEMENT_COUNT }, () => (
         <text
-          text-maxline='2'
+          textMaxline='2'
           text-maxlength='128'
-          enable-font-scaling={false}
+          enableFontScaling={false}
```
upupming, line 19: Shall we add type check cases in react runtime?

benchmark/react/types/index.d.ts
HuJean: Once it becomes stable, we can consider providing official type extensions.

benchmark/react/cases/012-attrs-compile/index.tsx
HuJean: ok

benchmark/react/cases/012-attrs-compile/index.tsx
HuJean: It is currently compile-time only. We can add tests later once runtime support is available.

upupming (approved):

gaoachao (approved):

PR #3281, opened 2026-07-29
fix(react-webpack-plugin): avoid duplicate async asset wrappers

### LINKED ISSUE #3280: [Bug]: Shared async main-thread asset is wrapped multiple times
## System Info

- OS: Linux
- Node.js: 24.14.1
- pnpm: 11.16.0
- `@lynx-js/react-webpack-plugin`: 0.10.1

## Details

When two main-thread modules dynamically import the same module, Rspack emits one shared async chunk but associates it with multiple chunk groups. `ReactWebpackPlugin` currently prepends its CommonJS wrapper while iterating `compilation.chunkGroups`, so it can update the same emitted JavaScript asset more than once.

The generated shared main-thread asset starts with nested duplicate wrappers:

```js
(function (globDynamicComponentEntry) {
  const module = { exports: {} };
  const exports = module.exports;
  (function (globDynamicComponentEntry) {
    const module = { exports: {} };
    const exports = module.exports;
```

The plugin callback and `processAssets` hook each run once. The duplication happens because the same chunk file is visited once for each chunk group that contains the shared async chunk.

## Minimal reproduction

Reproduction fixture: https://github.com/colinaaa/lynx-stack/tree/codex/fix-react-webpack-wrapper-dedup/packages/webpack/react-webpack-plugin/test/cases/code-splitting/shared-async-chunk

1. Create two modules that both evaluate `import('./shared.js')`.
2. Include both modules in code compiled for the main thread.
3. Build with `ReactWebpackPlugin` and `LynxTemplatePlugin`.
4. Inspect the emitted main-thread JavaScript for the shared async chunk.

The wrapper appears twice even though there is only one physical asset.

## Expected behavior

Each emitted main-thread JavaScript asset should be wrapped exactly once, regardless of how many chunk groups reference its chunk.

## Suggested fix

Track wrapped asset filenames in a `Set` inside the `processAssets` callback and skip files that have already been updated.

colinaaa:
## Summary by CodeRabbit

* **Bug Fixes**
  * Prevented shared asynchronous JavaScript assets from receiving duplicate runtime wrappers when the same async chunk is referenced from multiple chunk groups.
  * Improved correctness for nested shared async chunk loading so the wrapper is emitted only once.
* **Tests**
  * Added/extended coverage for shared and nested shared async chunks, including concurrent loading and verification via generated build artifacts.
  * Added an additional assertion to ensure the shared main-thread wrapper appears exactly once across related scenarios.
* **Documentation**
  * Added contributor guidance to deduplicate emitted assets when iterating chunk groups that may share async chunks.

Fixes #3280.

## Summary

- deduplicate main-thread wrapper injection by emitted asset filename
- add a regression fixture where two import sites share one async chunk
- document the asset-level deduplication requirement for future plugin changes

## Testing

- `pnpm turbo build`
- `pnpm --dir packages/webpack/react-webpack-plugin test`
- `pnpm turbo api-extractor -- --local`
- `pnpm dprint check`
- `pnpm biome check` on the changed source and fixture files
- `pnpm eslint packages/webpack/react-webpack-plugin/src/ReactWebpackPlugin.ts`

## Checklist

- [x] Tests updated
- [x] Documentation updated
- [x] Changeset added

### DIFF
--- a/packages/webpack/react-webpack-plugin/test/cases/code-splitting/shared-async-chunk/index.jsx
+++ b/packages/webpack/react-webpack-plugin/test/cases/code-splitting/shared-async-chunk/index.jsx
@@ -1,0 +1,31 @@
+/// <reference types="vitest/globals" />
+
+import { readFile } from 'node:fs/promises';
+import { resolve } from 'node:path';
+
+import { loadShared as loadFromFirst } from './first.js';
+import { loadShared as loadFromSecond } from './second.js';
+
+it('should load a shared async chunk from multiple import sites', async () => {
+  const [first, second] = await Promise.all([
+    loadFromFirst(),
+    loadFromSecond(),
+  ]);
+
+  expect(first.value).toBe('shared');
+  expect(second.value).toBe('shared');
+});
+
+it('should wrap the shared main-thread asset once', async () => {
+  const tasmJSON = JSON.parse(
+    await readFile(
+      resolve(__dirname, '.rspeedy/lazy-bundle/shared.js/tasm.json'),
+      'utf-8',
+    ),
+  );
+
+  const wrappers = tasmJSON.lepusCode.root.match(
+    /\(function \(globDynamicComponentEntry\) \{/g,
+  );
+  expect(wrappers).toHaveLength(1);
+});
--- a/.changeset/silly-chairs-visit.md
+++ b/.changeset/silly-chairs-visit.md
@@ -1,0 +1,5 @@
+---
+"@lynx-js/react-webpack-plugin": patch
+---
+
+Avoid wrapping shared async main-thread assets more than once.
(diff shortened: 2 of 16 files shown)


packages/webpack/react-webpack-plugin/test/cases/code-splitting/shared-async-chunk/index.jsx
```diff
@@ -0,0 +1,31 @@
...
+    await readFile(
+      resolve(__dirname, '.rspeedy/lazy-bundle/shared.js/tasm.json'),
+      'utf-8',
+    ),
+  );
+
+  const wrappers = tasmJSON.lepusCode.root.match(
+    /\(function \(globDynamicComponentEntry\) \{/g,
+  );
+  expect(wrappers).toHaveLength(1);
```
upupming, line 30: We should also assert in the `same-file-different-paths` case

Image

colinaaa: Added the wrapper-count assertion to `same-file-different-paths`. Rebuilding the pre-fix implementation makes this case report 3 wrappers, while the fix reports 1. I also added `nested-shared-async-chunk` to cover two async parent chunks importing the same child; it reports 2 wrappers before the fix and 1 after it. The full plugin suite passes (94/94).

upupming (approved):

PR #3284, opened 2026-07-29
feat(react): disable IFR with experimental_enableMTSRendering: false

upupming:
## Summary

Adds `pluginReactLynx({ experimental_enableMTSRendering: false })`: the main thread stops compiling business code, and its bundle is assembled from the snapshot and worklet definitions collected while compiling the background. It disables IFR (Instant First-Frame Rendering) to simplify the workflow for meta-frameworks that do not need the dual-thread concept.

- The main-thread entry is a framework entry, so no business module enters the main-thread layer.
- The definitions come out of the background's own transform pass — the snapshot and worklet plugins build the main-thread shape of each registration in the same traversal that emits the background's, so the two cannot drift.
- `@lynx-js/react-webpack-plugin` emits them into the main-thread chunk as a runtime module, and gives every lazy bundle its own assembled section.
- At runtime `renderMainThread()` is a no-op: the main thread creates an empty page and root, and the background hydrates it with a full insert patch. The flag is read behind a `typeof` guard, so a runtime paired with an older build plugin that does not inject it keeps the default behavior.
- Multi-entry builds assemble each main-thread chunk from its own entry's definitions; definitions in shared chunks stay shared.

## Numbers

`main-thread.js` / final bundle, production:

| example | main-thread.js | bundle |
| --- | --- | --- |
| `react` | 45,256 → 42,348 | 97,699 → 94,696 |
| `react-keyboard` | 46,088 → 41,288 | 97,649 → 92,812 |
| `gesture` | 50,243 → 43,937 | 114,087 → 107,796 |
| `tailwindcss` | 105,579 → 66,871 | 220,928 → 177,218 |

`background.js` is byte-identical in both modes.

On device (iPhone 13 Pro), the main thread's first render disappears (`create_vdom` 70.5ms → 0.4ms for `tailwindcss`) and the first paint comes earlier — but that paint is an empty page. Content now arrives with the background's hydrate patch, which lands later than IFR's first paint (`react` ~16ms → ~28ms, `tailwindcss` ~166ms → ~281ms). This mode trades first-screen content latency for bundle size and a single rendering thread; it is not a first-screen optimization.

## Verification

- All 20 examples build; 11 of them verified on device with the mode on (0 console errors, expected element trees).
- A two-entry build with a shared component verified on device: each entry carries its own definitions plus the shared one, 0 errors.
- Definition parity: the snapshot id set in `main-thread.js` is identical in both modes across 8 examples; worklet counts match exactly.
- `mtsRenderingRuntime.test.ts` evaluates the minified `main-thread.js` against stubbed element PAPI and applies a background-shaped patch.
- 15 benchx cases run in a real Lynx runtime; `ReactLynx::renderMainThread` totals 95.9ms with the mode off and never fires with it on, while `FiberElement::Constructor` counts stay identical.

## Limitations

- Main-thread business behavior is gone by design, including `defaultDataProcessor` and main-thread render-time side effects.
- `experimental_useElementTemplate` is not supported yet; the combination fails the build.
- A `runtime: 'shared'` import referenced inside a main-thread function is not supported yet; the transform fails the build.
- Incremental HMR of the assembled bundle is not covered yet.
- The `web` environment receives the same flag but is not verified.

## Summary by CodeRabbit

* **New Features**
  * Added `experimental_enableMTSRendering` option across React build plugins (default enabled), including support for `mainThreadEntries` mapping.
  * When disabled, the system collects and initializes module-time “MTS defines”, including for lazy chunks.
* **Bug Fixes**
  * Improved correctness when main-thread rendering is disabled (root updates, page attribute propagation).
  * Runtime now errors when a required background entrypoint is missing.
* **Tests**
  * Expanded test coverage for option validation, disabled-render behavior, lazy bundles, and MTS defines collection.
  * Added a new disabled-render example and fixtures.


.changeset/main-thread-defines-assembly.md
```diff
@@ -0,0 +1,8 @@
+---
+"@lynx-js/react": patch
+"@lynx-js/react-webpack-plugin": patch
+"@lynx-js/react-rsbuild-plugin": patch
+"@lynx-js/testing-environment": patch
+---
+
+Add `pluginReactLynx({ enableMTSRendering: false })`, which stops compiling business code for the main thread and assembles its bundle from the snapshot and worklet definitions collected while compiling the background.
```
upupming:
```suggestion
Add `pluginReactLynx({ enableMTSRendering: false })`, which disables IFR (Instant First-Frame Rendering) to simplify the workflow for meta-framework that does not need the dual-thread concept.
```

packages/react/runtime/mts-rendering-disabled/index.js
```diff
@@ -0,0 +1,71 @@
+// Copyright 2026 The Lynx Authors. All rights reserved.
+// Licensed under the Apache License Version 2.0 that can be found in the
+// LICENSE file in the root directory of this source tree.
+
+/* global __lynxMainThreadDefines */
+
+import * as ReactLynx from '@lynx-js/react/internal';
+
+const runtime = {
```
upupming:
```suggestion
const simplifiedRuntime = {
```

packages/react/runtime/mts-rendering-disabled/index.js
```diff
@@ -0,0 +1,71 @@
...
+  get __DynamicPartListChildren() {
+    return ReactLynx.__DynamicPartListChildren;
+  },
+  get loadWorkletRuntime() {
+    return ReactLynx.loadWorkletRuntime;
+  },
+};
+
+if (typeof __lynxMainThreadDefines !== 'undefined') {
+  __lynxMainThreadDefines(runtime);
```
upupming:
```suggestion
  __initMainThreadDefines(runtime);
```

packages/react/transform/__test__/main-thread-defines.spec.js
```diff
@@ -0,0 +1,143 @@
...
+    </view>
+  );
+}
+`;
+
+describe('collectMainThreadDefines', () => {
+  it('collects the definitions the main thread needs while compiling the background', async () => {
+    const result = await transformReactLynx(
+      source,
+      options('JS', { collectMainThreadDefines: true }),
```
upupming: It will be very confusing we collect "MainThreadDefines" in background thread. How about rename "MainThreadDefines" to something not related to main thread, such as "snapshotDefines" "mtsDefines". I am not sure if we will add more defines, how can you name this better?

packages/react/transform/crates/swc_plugin_snapshot/lib.rs
```diff
@@ -1434,6 +1507,16 @@ where
...
           }
+          if collecting {
+            snapshot_dynamic_part_def_mt.push(Some(ExprOrSpread {
+              spread: None,
+              expr: Box::new(dynamic_part.to_updater(
+                creator_runtime_expr.clone(),
+                TransformTarget::LEPUS,
+                snapshot_dynamic_part_def.len() as i32,
+              )),
+            }));
```
upupming: Can we resue `ExprOrSpread...` since it is created two times here.

packages/react/transform/src/lib.rs
```diff
@@ -300,6 +317,80 @@ pub struct TransformNodiffOutput {
...
+    compat::es2020::nullish_coalescing(Default::default()),
+    compat::es2021::logical_assignments(),
+    helpers::inject_helpers(unresolved_mark),
+    hygiene_with_config(Config {
+      top_level_mark,
+      ..Default::default()
+    }),
+    fixer(Some(comments)),
+  )
+}
```
upupming, line 345: Do we really need this? I think rspack will transpile our js to es2019 later since we have swc config in`packages/rspeedy/plugin-react/src/loaders.ts`

packages/webpack/react-webpack-plugin/src/MainThreadDefinesRuntimeModule.ts
```diff
@@ -0,0 +1,191 @@
...
+
+export const MAIN_THREAD_DEFINES_BUILD_INFO = 'lynx:main-thread-defines';
+
+export interface MainThreadDefine {
+  kind: 'snapshot' | 'worklet';
+  id: string;
+  code: string;
+}
+
+const PROVIDED_RUNTIME_MEMBERS: ReadonlySet<string> = new Set([
```
upupming: We should not maintain runtime related variables in webpack plugin. This will make it hard to adapt new version of @lynx-js/react when there are new APIs need to expose.

packages/webpack/react-webpack-plugin/src/MainThreadDefinesRuntimeModule.ts
```diff
@@ -0,0 +1,191 @@
...
+  '__DynamicPartSlot',
+  '__DynamicPartSlotV2',
+  '__DynamicPartSlotV2_0',
+  '__DynamicPartListSlotV2',
+  '__DynamicPartChildren',
+  '__DynamicPartChildren_0',
+  '__DynamicPartListChildren',
+]);
+
+const RUNTIME_MEMBER_RE = /\bReactLynx\.([$A-Z_a-z][\w$]*)/g;
```
upupming: Never use regex, it is very fragile.

gaoachao (dismissed):

colinaaa (approved):

upupming:
Closed as in favor of `root.render(__MAIN_THREAD__ ? <Skeleton /> : <App />);` or `<background-only />` (will land in future PR)

Issue #3296, opened 2026-07-30
RFC(react): root `<Background fallback>` as the declarative API over `enableMTSRendering: false`


Huxpro:
# RFC: root `<Background fallback>` as the declarative API over `enableMTSRendering: false`

## Context

#3284 introduces `enableMTSRendering: false`: the main thread stops compiling business code, and its bundle is assembled from the snapshot/worklet definitions collected during the background compilation. As infrastructure this is the strongest implementation of a "0.0 first screen" so far:

- single compilation source — the main-thread shape of each definition is produced in the same traversal as the background's, so the two cannot drift;
- no main-thread business module graph — the reference/tree-shaking hazards that in-place stripping has to fight (see the keep-alive work discussed in #2977) are absent *by construction*;
- the main-thread render traversal disappears entirely (`create_vdom` 70.5ms → 0.4ms in #3284's numbers).

Two gaps keep it from being the full user-facing story:

1. **It is a binary config flag.** How much of an app participates in the first screen is an *app-structure* concern, and the exploration in #2965 / #2977 argues it composes best as a user-space boundary component — with build flags as implementation details, not as the exposed API.
2. **The first paint is an empty page.** #3284 measures content arriving ~2× later than IFR (`react` ~16ms → ~28ms, `tailwindcss` ~166ms → ~281ms) with *nothing* on screen until the background's hydrate patch lands. There is no fallback channel.

## Proposal

Make a root-level `<Background fallback>` the declarative trigger, with `enableMTSRendering: false` as its implementation:

```tsx
import { Background, root } from '@lynx-js/react';

root.render(
  <Background fallback={<view><text>Loading…</text></view>}>
    <App />
  </Background>,
);
```

- A `<Background>` at the render root (detected from the entry, or forced by the internal option) enables the #3284 mode.
- The main-thread framework entry renders the entry's **static host-element fallback** through the assembled snapshot definitions — the first paint becomes a skeleton instead of an empty page, softening the content-latency trade #3284 documents.
- Today's #3284 behavior is exactly the degenerate case `fallback={null}`: the flag becomes a special case of the API rather than a parallel concept.
- A *nested* `<Background>` (#2965) keeps meaning per-subtree runtime deferral. One component; its position on the tree decides how much of the first screen it removes — root position is simply the whole-program endpoint of the same boundary.

## Prior art to harvest (and supersede)

- #2965 — the `<Background fallback>` boundary component. Runtime semantics unchanged by this issue.
[...]

Issue #3297, opened 2026-07-30
Explore(react): component-level `'background only'` on the MTS-defines infrastructure — performance and behavior deltas


Huxpro:
# Explore: component-level `'background only'` on the MTS-defines infrastructure

## Context

#3284's MTS-defines infrastructure collects the main-thread shape of every snapshot/worklet registration during the *background* compilation and assembles it into the main-thread chunk. Today it is wired all-or-nothing: `enableMTSRendering: false` applies it to the whole program.

The per-component form of the same need already exists: `'background only'` (#2970) empties a component's render body on the LEPUS target *in place*, keeping the module-scope definitions for first-screen hydration. That in-place approach carries structural obligations the assembly approach does not:

- The module must stay in the main-thread graph, pinned by references — structural parents keeping their bodies, and for a background-only component that itself delegates to other modules, a synthesized keep-alive (see the fix on branch `claude/lynx-ifr-l4-composability-gaq7q4`, discussed in #2977).
- References the transform cannot see statically (a component resolved through a runtime value) still get severed.
- The module's **top-level statements still execute on the main thread** — only render bodies are emptied, so side-effect isolation is partial.

This issue explores generalizing the #3284 collector from "whole compilation" to "the set of `'background only'` components" under the normal dual-thread mode (`enableMTSRendering: true`), and characterizes the performance and behavior deltas.

## Sketch

For a `'background only'` component:

- Its definitions come from the background collector (assembly), not from its own main-thread compile.
- Its module compiles for the main thread as a **stub shell**: exports keep their names as inert empty functions — note the parent still evaluates `jsx(Feed, …)` as a *value* when it renders on the main thread, even though the `<Background>` boundary never invokes it — while bodies, top-level statements, and in-place definitions are dropped.
- The assembled section for a chunk is `BG-collected defs − MT in-place def ids` (build-time subtraction), or registration is made idempotent at runtime. #3284's `kind:id` collision check stays as the drift alarm.

## What to measure (performance)

- **Main-thread bundle size** vs. today's in-place strip: module shells + no def duplication. Expect modest wins — the render logic is already stripped today; the delta is the module scaffolding and any top-level code.
- **Build time**: background-only modules skip meaningful main-thread compilation.
- **Main-thread first-screen cost**: expected unchanged (structural parents still render; the boundary already skips the subtree at runtime). Verify with the benchx setup from #3284.
[...]

PR #3302, opened 2026-07-30
feat(react): root `<Background fallback>` as the declarative API over `enableMTSRendering: false`

Huxpro:
Implements #3296, on top of #3284 (based on `feat/main-thread-defines-from-background`).

```tsx
import { Background, root } from '@lynx-js/react';

root.render(
  <page>
    <Background fallback={<Skeleton />}>
      <App />
    </Background>
  </page>,
);
```

A `<Background>` declares that its subtree does not participate in the first screen. On the main-thread target the transform **folds the boundary to its `fallback`**, so the deferred subtree's module closure leaves the main-thread bundle while the fallback is compiled for it as ordinary code — it may contain user components, hooks and computed children. The app's element definitions still travel through #3284's assembled definitions, so hydration builds the real tree from them.

## Every permutation, source → MTR → BTR

`tests/bgmatrix/` is 22 cases across three axes — where the boundary sits, how it is imported, what the fallback is — each built under all three `experimental_enableMTSRendering` settings and captured twice: once with the background thread held, so the frame is what the **main thread alone** built (MTR), and once after release (BTR). `bgmatrix.report.mjs` puts the source, both frames and the Lynx `main-thread.js` facts side by side.

```sh
node tests/bgmatrix.build.mjs                 # 66 permutations, both targets
npx playwright test --config=playwright.bgmatrix.config.ts
node tests/bgmatrix.report.mjs                # → bgmatrix-report/index.html
```

**60 of 60 permutations that build paint a first frame on the main thread alone, and every one is correct after hydration.** Under the assembled build, 32 of 40 also drop the deferred code from the main-thread chunk; 2 more keep it because the app asked (`f04` names the deferred component as its own fallback). That leaves **6 — `i03-reexport`, `i04-wrapper`, `i05-dynamic-type`** — which render correctly and ship the deferred code anyway.

That split is the result, and it is not an accident of this implementation. `<Background>` is two mechanisms under one name:

- **A rendering boundary** — the component returns `fallback` on the main thread, `children` on the background thread. Total: it is an ordinary component, so it works at any depth, through any import, behind a conditional, inside a `.map()`, with any fallback.
- **A bundling boundary** — the fold deletes the reference that keeps the deferred subtree in the main-thread graph. Partial, for one reason: *a fold can only delete a reference it can see, in the module where the boundary is written.*

Each failing shape is that sentence in a different disguise. `i03` — the element's name is not resolvable to the runtime export from inside the module. `i05` — the element type is a value, so resolving it means running the program. `i04` — the boundary *is* resolvable, but the reference to the deferred subtree lives at the call site in `children`, a module away, so there is nothing to cut where the reference is. `f05-spread` is the same limit stated about the fallback rather than the child, and it is the one case that refuses the build instead of degrading.

This is the shape React arrived at with Server Components, under the same pressure: `'use client'` is a **module** directive because a bundler splits a graph along import edges. An element can only speak for the one reference written beside it. **An element-level boundary can always express *when* something renders, and only sometimes *where* it ships.**

## Making the loss loud

The repo already ships the module-level half: `import 'background-only'` resolves, in the main-thread layer, to a module that fails the build. `g01` and `g02` are the same app twice — where the fold reaches, and in the `i04` wrapper shape where it cannot:

| case | `'auto'` / `false` | `true` |
|---|---|---|
| `g01-assert-folded` | builds — the module is not in the main-thread graph | build error (classic compiles it there) |
| `g02-assert-not-folded` | **build error**, naming the module | build error |

That does not make `i04` work; it makes `i04` say so, turning 6.5 KB of silent dead weight into a named failure. The rule it enforces is the one the mechanism actually has: **write the boundary where the reference is**, and assert it from the deferred module when the bundle matters.

The general escape — severing the edge rather than asserting about it — wants to be on the import, because that is what a bundler splits on: `import { Feed } from './feed.jsx' with { runtime: 'background' }` would cut `i03`, `i04` and `i05` alike without recognising a `<Background>` at all. Not implemented; the matrix is the argument for it.

## Islands, not a root special case

The fold rewrites each `<Background>` **where it stands**, so nothing about it is specific to the render root:

| layout | main thread compiles | first frame |
|---|---|---|
| boundary at the root | the fallback only | the fallback |
| boundary inside a component | that fallback + everything around it | header + fallback |
| N boundaries, unrelated | every fallback + everything outside them | all of it, interleaved |

They are **not N roots**: they are N holes in one main-thread tree, so the surrounding structure positions each one and a single hydration pass replaces them all independently. What changes with placement is *coverage*, not mechanism.

## Two fixes the matrix turned up

**Detection asked the wrong question.** It looked for a `<Background>` *element*. But its answer decides whether the entry is compiled for the main thread at all, so a miss is not a lost optimization — it is a **blank first screen**, the main thread handed nothing to render. An alias, a namespace import, a re-export and a computed element type all missed, and all four produced exactly that (`i01`, `i02`, `i03`, `i05` under `false`, `sk=0 real=0`). It now asks whether the module *binds* `Background` from the runtime, which is the property the fold itself follows. A false positive costs close to nil: the entry compiles as it would have, and the assembly subtracts everything the main-thread bundle already owns.

**The mode made the main-thread chunk bigger.** It entered through two entry imports — the definitions runtime and the app's entry — so every module those two share (the whole main-thread runtime) was reachable from two concatenation roots, inlinable into neither, and emitted module by module. The bundle this mode exists to shrink came out **~40 KB larger than the classic build**. One generated root importing both, in that order, gets the scope hoisting back. `p10-heavy` (46 KB behind the boundary) now reads −35.9 KB against classic instead of +5 KB.

## Verified end to end through Lynx for Web

Beyond the matrix, `tests/background-island.spec.ts` keeps three baseline cases. The three-island case, main thread alone → after hydration:

- **before**: blue skeleton (island A) · green header (not deferred) · beige skeleton (island C, *inside* that component) · purple skeleton (island B)
- **after**: `real-a` · the same green header, untouched · `real-c` · `real-b`

Holding is scoped to the `lynx-bg` worker. Holding *every* worker stalls the template decode instead — the page goes blank for the wrong reason, and the suite would have been vacuous. That was a real bug in the harness, caught by running the same hold against a case with no `<Background>` at all and seeing it blank too.

## Each definition described once

A fallback compiled for the main thread also sits in the background's module graph, so its definitions were emitted twice: as real code and again through the assembly, once per island. The main-thread loader now records what it owns and the assembly subtracts it — on the three-island fixture, 4 of 7 definitions, and the assembled section drops 4766 → 2097 bytes.

## Notes

- The entry module itself now runs on the main thread — `root.render` has to, for the fallback to exist. Only the *deferred subtree's* closure is dropped.
- The earlier revision carried the fallback as a `root-fallback` metadata define naming a pre-generated static snapshot. That capped fallbacks at static host elements, so it is removed and superseded by the fold — along with the `fallback contains a user component` guardrail, which warned about exactly what now works.
- `bgmatrix-report/` and `dist-bundle/` are gitignored: they are regenerated by the three commands above.

## Relationship to prior art

- **#2965** — landed here as the `<Background>` component; per-subtree runtime semantics unchanged.
- **#2977** — entry detection and the `false | 'auto' | true` shape are ported from it. **Correcting an earlier claim in this PR that the strip is superseded**: it is not. Assembly (zero main-thread graph), this PR's fold (islands only), and the strip (full graph, emptied bodies) are three points on one spectrum.
- **#2963** — not on this branch, so no code interaction yet; `root.hydrate()` stays complementary, moving only *when* the replacement applies.

## Checklist

https://claude.ai/code/session_01CogcVrjRpB4N77xuZwkfK6

---
_Generated by [Claude Code](https://claude.ai/code/session_01CogcVrjRpB4N77xuZwkfK6)_


packages/webpack/react-webpack-plugin/src/MTSDefinesRuntimeModule.ts
```diff
@@ -55,6 +99,11 @@ export function collectMTSDefines<TChunk, TModule>(
   const defines = new Map<string, MTSDefine>();
   for (const define of collected) {
     const key = `${define.kind}:${define.id}`;
+    if (owned?.has(key)) {
```
Huxpro: The `continue` sits *before* the collision check below, so on this path #3284's `kind:id` drift alarm no longer fires.

Concretely: if two background modules emit the same id with different content, and the main thread happens to own that id, the loop skips both copies and the build stays green — the exact drift the check exists to catch, silenced by an unrelated condition.

Subtracting after the check keeps both properties, since the check only reads `defines`:

```js
const seen = defines.get(key);
if (seen !== undefined && seen.code !== define.code) {
  throw new Error(`Two different main-thread definitions share the id ${define.id}.`);
}
if (owned?.has(key)) continue;
defines.set(key, define);
```

or leave the loop untouched and drop the owned keys once it has finished (`for (const key of owned) defines.delete(key)`), which reads closer to the "assembled = all − owned" framing the PR describes.

---

packages/webpack/react-webpack-plugin/src/MTSDefinesRuntimeModule.ts
```diff
@@ -144,12 +193,21 @@ export function createMTSDefinesRuntimeModule(
       }
 
       const { chunkGraph } = compilation;
+      const getModules = (chunk: Chunk) => chunkGraph.getChunkModules(chunk);
+
+      // What this main-thread chunk already carries as real code — the
+      // fallbacks it compiles — so the assembly describes only the deferred
+      // subtrees it does not.
+      const owned = this.chunk
+        ? collectOwnedMTSDefineKeys([this.chunk as Chunk], getModules)
```
Huxpro: The two sides of the subtraction are scoped differently: `owned` comes from this one chunk, while the minuend below is collected over `entrypoint.chunks`. Anything the main thread owns in a *sibling* initial chunk is invisible here, so it gets assembled as well as emitted as real code.

Harmless as the mode stands — with only fallbacks compiled, the main-thread bundle is small enough that it is one chunk in practice. It stops being harmless the moment the main thread compiles business code alongside the fold (`enableMTSRendering: true` + islands, which is where #3297 wants to go): `splitChunks` is on by default, so the entry chunk is routinely not the whole story.

Cheap to make the scopes symmetric now:

```js
const mainThreadEntrypoint = this.chunk
  ? compilation.chunkGraph.getChunkEntryModulesWithChunkGroupIterable /* … */
  : undefined;
// or simply: pass the main-thread entry name in, like `backgroundEntry`,
// and use `compilation.entrypoints.get(mainThreadEntry).chunks`
const owned = collectOwnedMTSDefineKeys(mainThreadChunks, getModules);
```

Passing the main-thread entry name into the constructor next to `backgroundEntry` is probably the least surprising version — the runtime module then names both sides of the subtraction explicitly.

---

packages/webpack/react-webpack-plugin/src/loaders/options.ts
```diff
@@ -258,11 +258,24 @@ export function getMainThreadTransformOptions(
...
+    // is what drops the `children` reference, and with it the deferred
+    // subtree's module closure, from the main-thread bundle.
+    //
+    // Collecting alongside it does not change this output (unlike on the
+    // background target, where it moves the definitions out): it only reports
+    // which definitions this bundle already carries as real code, so the
+    // assembled ones can leave them out.
+    ...(enableMTSRendering === false && {
+      foldBackgroundToFallback: true,
+      collectMTSDefines: true,
```
Huxpro, line 277: The comment above is right that collecting on the LEPUS target does not move the emitted definitions — but `collectMTSDefines` carries a second, unrelated meaning that does take effect here.

In `swc_plugin_worklet`, the collector's presence gates a diagnostic:

```rust
reject_shared_identifiers: self.mts_defs_collector.is_some(),
```

So turning collection on for the main-thread target also turns `import … with { runtime: 'shared' }` from "left to the bundler" into a build failure — inside `<Background>` fallbacks, which is exactly where user components are now allowed to live. There is an existing test asserting the old behaviour for this target:

> `shared runtime imports` › *is left to the bundler when the main thread compiles business code* — `packages/react/transform/__test__/mts-defines.spec.js`

It still passes only because it does not pass `collectMTSDefines`; a fallback that imports a shared-runtime module in a real build would now fail.

Two ways out, both fine:

1. **Split the flag's two jobs.** Have the collector record `kind:id` without the codegen/diagnostic side effects — a `report-only` mode — and use that here. That keeps `reject_shared_identifiers` tied to "the definitions are being *moved*", which is the condition it actually guards.
2. **Accept it and make it explicit.** Gate `reject_shared_identifiers` on the target rather than on the collector's presence, and add a test pinning the new behaviour for fallbacks.

I hit the same fork in #3305 and went with (1) (a two-view collector: full definitions vs. identities only). Happy to hand that piece over if it's useful — the rest of #3305 is superseded by this PR (see the top-level comment).

---

Huxpro (commented):
Reviewed from the #3297 side, where I built the same subtraction and got it wrong in a way this PR gets right. Three inline notes, then one question that matters more than all of them.

## The subtraction here is the sound formulation

Worth stating plainly, because I arrived at the other one:

```
#3305 (mine):  assembled = collect(background-only modules)
this PR:       assembled = collect(every background module) − owned(main thread)
```

Collecting a *subset* makes coverage depend on the subset predicate being complete. It isn't: I can build a tree where a plain, un-annotated component is reachable only through a deferred one, and its definitions end up in neither source — no error, just a first screen missing elements. Subtracting from the *total* cannot have that failure mode: the minuend is everything, and the subtrahend only removes what is provably present as real code elsewhere. Conservation is structural rather than checked.

That also makes the build-time conservation check I was going to propose unnecessary here, which is the better outcome.

Same for the fold: dropping the `children` reference and letting the bundler's reachability analysis do the rest gets the graph cut with no annotation, no export-shape predicate, and no new error surface. I had reached "the module-level marker should be inferred from the boundary, not written by hand" as a conclusion; this reaches it by construction. And "N holes in one main-thread tree, not N roots" is the more accurate framing — I had been calling them roots.

## Inline notes

1. **Drift alarm** — the owned-skip runs before the collision check, so `kind:id` drift stops failing the build when the main thread happens to own the id.
2. **Asymmetric scopes** — `owned` is one chunk, the minuend is `entrypoint.chunks`. Benign today, not benign under `enableMTSRendering: true`.
3. **`collectMTSDefines` has a second meaning** — it also flips `reject_shared_identifiers`, so shared-runtime imports inside fallbacks now fail the build.

(1) and (3) are small; (2) is worth fixing now because #3297 walks straight into it.

## The question: do the deferred subtrees' definitions have to be in the first frame at all?

Today they do — the PR says as much, and the 4766 → 2097 drop comes from subtracting the *fallbacks'* definitions, not from deferring the islands'. So the main-thread chunk still carries the full element vocabulary of every subtree it deliberately does not render on the first screen.

But the boundary already declares that those subtrees do not participate in the first frame, and their definitions are only needed once the background is ready and hydration replaces the fallback. So in principle they could be registered *at that point* instead of shipped in the first-frame chunk — the assembled section becomes a lazily-installed chunk rather than a runtime module, and the first frame carries the shell and nothing else.

That is a different order of magnitude from a subtraction. On the three-island fixture, the 2097 bytes that remain are precisely the deferred vocabulary.

Three things I can't answer from outside this PR:

- **How much is it worth here?** Is the assembled section a meaningful fraction of the main-thread chunk on a real app, or is it noise next to the runtime?
- **Is it lossy?** Deferring registration puts the definitions on the hydration path. If they arrive with the background's first ops there may be no observable cost; if it adds a round trip or a decode before the first replacement, the fallback stays on screen longer and the trade is real. You have the harness — the held-background screenshots would show it directly.
- **Can it just be a switch?** `<Background>` says "not in the first frame"; whether that means *not rendered* or *not even described* could be per-boundary (`fallback` + something like `defer="definitions"`) rather than a global mode. If it's a switch, it need not block anything.

Not asking for it in this PR — it's a follow-up either way. But the answer decides whether #3297 is "extend the fold to `enableMTSRendering: true`" (a few flags) or something with a real payoff behind it, so I'd rather ask before building on top.

## What I'm doing on my side

Closing #3305. Its subtraction is superseded by this one, its module-level `'background only'` stub was the wrong layer for the reason this PR demonstrates, and the two-view collector is the only piece worth keeping — offered above for note (3).

#3297 then becomes: enable this PR's fold + full collection + subtraction under `enableMTSRendering: true`, so an app whose main thread *does* render business code can still have `<Background>` subtrees leave its bundle. That's mostly ungating what is already here, plus note (2), plus a way to call `__initMTSDefines` when the main-thread entry is the app itself rather than the assembled shim. Opening that against this branch.

---

Huxpro:
All three notes fixed in `24dbf9fb`. Note (3) I was mid-verification on when the review landed — my probe had just produced the same result, so it was a live regression, not a latent one.

**(1) Drift alarm.** Correct, and the diagnosis is exact: the skip silenced an alarm about a different question. Took the second suggestion — subtract after the loop, since `assembled = all − owned` is what the code should read like. Two tests pin it: drift still throws when the main thread owns the id, and subtraction is by `kind:id` rather than id.

**(2) Asymmetric scopes.** Fixed as suggested — `mainThreadEntry` passed in next to `backgroundEntry`, both sides read an entrypoint. Agreed it should land now rather than in #3297.

**(3) `collectMTSDefines`' second meaning.** Went with a narrower version of your option 1. Rather than splitting the collector into two views, I moved the gate to the condition it actually guards:

```rust
fn rejects_shared_identifiers(&self) -> bool {
  self.mts_defs_collector.is_some() && self.cfg.target == TransformTarget::JS
}
```

The hazard is *relocation*, not collection. On the background target the definitions are lifted into `__initMTSDefines`, whose scope holds only the runtime, so a shared identifier has no binding there. On the main-thread target they stay where they are and the bundler resolves the import as always. Measured both directions before and after; new test in `mts-defines.spec.js` pins the main-thread side, and the existing background test is untouched.

If your two-view collector has uses beyond this gate I'd still take it — this fix only removes the false positive.

## On deferring the deferred subtrees' definitions

Taking the three questions in order, with what I can measure.

**How much is it worth.** On the three-island fixture the assembled section is 2097 B against a 390 KB main-thread chunk — ~0.5%, noise next to the runtime. But that ratio is the wrong thing to extrapolate: the runtime is constant while the assembled section scales with the *deferred* vocabulary, so the fraction is roughly `deferred UI / (runtime + first-screen UI)`. The fixture defers three trivial components. An app that defers most of itself behind a root `<Background>` — the case this PR started from — is the one where the assembly is nearly the whole non-runtime payload, and there the answer could be large. I don't have a real app here to put a number on it; a benchmark app with a realistic tree would settle it in one build.

**Is it lossy.** I think this is the question that decides it, and I'd expect *yes, somewhat*, for a structural reason: today the definitions are already resident when the first patch arrives, so replacement is a `snapshotPatchApply` and nothing else. Move them onto the hydration path and the first replacement cannot happen until they are installed. Whether that is free depends on whether they can ride the same message as the first ops or need their own delivery — and on lazy bundles it is a `processEvalResultByHost` round trip, which is not free. The held-background harness would show it directly: hold, release, and measure to first replacement with the definitions resident vs. installed on arrival. That is a real experiment, not a thought experiment, and I'd want it before committing to the design.

**Can it be a switch.** Yes, and I think it has to be, for the reason above — the trade is app-shaped. Per-boundary is also the right granularity: `<Background>` already says "not in the first frame", and "not even described" is strictly stronger, so it reads naturally as a modifier on the same boundary rather than a global mode. It also composes with the fold without new machinery: the boundary already identifies the subtree whose definitions would move, which is exactly the set the collector would need to partition.

So: not blocking, and I'd rather it not gate #3297. If #3297 is "ungate the fold + full collection + subtraction under `enableMTSRendering: true`", that is valuable on its own and independent of whether the definitions ship eagerly. Happy to run the latency experiment on this branch's harness if it would help you scope it.

Also noted on #3305 being closed — the framing of subtraction-from-total vs. collecting-a-subset is worth keeping somewhere more durable than a PR comment; I'll fold it into the changeset rationale if it isn't already clear there.

---

Huxpro:
## The fold is one-sided — the background target can fold too

Right now `swc_plugin_background_fallback` only rewrites the LEPUS target: `<Background fallback={F}>{C}</Background>` → `F`. On the background target the boundary survives as a runtime component and `F` stays in the graph. The mirror rewrite — → `C`, dropping `F` — looks sound, and it needs no new machinery on top of what this PR already has.

Filing it here rather than in #3382 because it belongs to the fold, not to `experimental_backgroundIslands`: it applies to the background target under all three `experimental_enableMTSRendering` settings and is orthogonal to that switch.

### It is currently dead weight

Built `multi-island.tsx` under islands and grepped the background chunk:

```json
{ "bgHasSkeletonA": true, "bgHasSkeletonB": true, "bgHasSkeletonC": true, "bgBytes": 429707 }
```

All three fallbacks ship to the background, where nothing ever renders them.

### Why it is sound

**The background never renders the fallback.** `core/background.ts:120` is `if (__MAIN_THREAD__) return props.fallback ?? null; return props.children ?? null;`. Folding to `children` on that target is exactly the runtime semantics, the same way folding to `fallback` is on LEPUS.

**Hydration does not need the fallback's definitions — so there is no mirror assembly to build.** This is the part worth stating explicitly, because the main-thread direction *does* need the deferred subtree's definitions, which is what #3284's assembly exists for. The reverse is free:

- `tt.ts:95` — `const before = JSON.parse(lepusSide) as SerializedSnapshotInstance`. The main-thread first-screen tree arrives as data and is never materialised into `BackgroundSnapshotInstance`, so it never reaches the `throw new Error('BackgroundSnapshot not found: ' + type)` in `backgroundSnapshot.ts:163`.
- Matching is a string compare, `(a, b) => a.type === b.type`.
- The removal path pushes `SnapshotOperation.RemoveChild, before.id, node.id` — ids only.
- The only definition `helper()` reads is `after.__snapshot_def`, i.e. the background's own rendered node, which exists by construction.

The main thread has to *build* the elements, so it needs the definitions. The background only has to name the fallback's nodes to delete them.

**Nothing in the fallback has a live background-side binding today.** The one candidate coupling is a main-thread worklet in the fallback calling `runOnBackground`. It is already broken: `_jsFnId` is assigned when the background renders the node bearing the worklet, and the background never renders the fallback. `BackgroundProps.fallback`'s own doc says as much — "Event handlers and refs inside the fallback are never attached." Folding removes code that is already dead rather than killing anything live.

### It does not enlarge the theory

The boundary statement in this PR's matrix mirrors exactly:

> a deferred module's top level stops running on the main thread **iff** every reference to it lies inside a folded boundary

becomes

> a fallback-only module's top level stops running on the background **iff** every reference to it lies in a `fallback` position

Same three escapes — a component shared between fallback and app, a hoisted binding, a subtree slotted in through props — decided by the same reachability, with no new annotation and no new error form. The `i03`/`i04`/`i05` limits apply unchanged and for the identical reason: a fold can only delete a reference it can see.

### Two things to pin with tests rather than derive

1. **A component shared between a fallback and the app must survive on the background.** If the fold is "delete the `fallback` prop expression and let tree-shaking decide", correctness is automatic — but it needs a fixture so a later implementation cannot regress it silently. Same shape as `f04`, mirrored.

2. **Unverified, and unrelated to this proposal, but surfaced while reading `hydrate`:** the `DynamicPartType.Slot` / `MultiChildren` branch pairs children by index and calls `helper(v1, v2)` with no type comparison, while `SlotV2` / `Children` compares types first. At a boundary the main-thread tree holds the fallback and the background tree holds the real content, so the types differ by construction. I have not established that the V1 path is reachable for a boundary — the assembled section on the root fixture is all `__DynamicPartSlotV2` — so this may be a legacy path only. Worth one case either way.

### Expected size

Modest, and worth saying so: a fallback is usually much smaller than the subtree it stands in for, so the background saves an order of magnitude less than the main thread does. The extra reason to want it is per-render: `<Background fallback={<SkeletonA />}>` allocates a vnode for `SkeletonA` on every render of the enclosing component on the background, only for `Background` to drop it.

---

packages/webpack/react-webpack-plugin/src/MTSDefinesRuntimeModule.ts
Huxpro: Fixed in `24dbf9fb` — took the second variant, dropping the owned keys after the loop rather than short-circuiting inside it. It reads as the `assembled = all − owned` framing, and it keeps the two properties independent: whether the main thread happens to own an id says nothing about whether two background modules disagree on what it means. The code carries that sentence as a comment so the ordering doesn't get "simplified" back later.

---

packages/webpack/react-webpack-plugin/src/MTSDefinesRuntimeModule.ts
Huxpro: Fixed in `24dbf9fb`, with the version you called least surprising: the constructor now takes `(backgroundEntry, mainThreadEntry)` and both sides read `compilation.entrypoints.get(…).chunks`, so the runtime module names both operands of the subtraction explicitly.

Made `mainThreadEntry` required rather than optional — an optional one would let the asymmetry come back silently at a call site that forgets it, which is exactly the failure mode you're describing. #3382 independently arrived at the same shape and flagged the required form as the cleaner of the two.

---

packages/webpack/react-webpack-plugin/src/loaders/options.ts
Huxpro: Good catch — this was a live regression, not a latent one. Fixed in `24dbf9fb`.

Took route (2), and it turns out to land on (1)'s invariant without adding a mode to the collector:

```rust
fn rejects_shared_identifiers(&self) -> bool {
  self.mts_defs_collector.is_some() && self.cfg.target == TransformTarget::JS
}
```

The condition the diagnostic actually guards is "the definitions are being **moved**". Collection happens on both targets; moving happens only on the JS one. So `collecting && target == JS` *is* "moving", stated directly at the one site that asks — where a two-view collector would encode the same fact indirectly, as a mode the caller has to pick correctly. Six call sites went through the helper rather than repeating the conjunction.

The test you asked for is in the same commit, next to the one you quoted:

> `shared runtime imports` › **is still left to the bundler when the main thread only reports ids** — `LEPUS` + `collectMTSDefines` + `foldBackgroundToFallback`, asserting `errors` is empty and `new Foo()` survives.

So the three cases now pin all three answers: `LEPUS` plain (left to the bundler), `JS` + collecting (build error), `LEPUS` + collecting + folding (left to the bundler — this is the `<Background>` fallback path).

Thanks for the #3305 offer — not needed for this, but the two-view collector is still the right shape if collection ever grows a third caller that genuinely needs identities without the codegen.

---

PR #3303, opened 2026-07-30
feat(motion): add multi-entry web examples

f0rdream:
Expose Basic, Gesture, Mini, MotionValue, and Spring as standalone Motion example entries for both Lynx and Web previews. Align the example output naming with the platform-specific multi-entry layout used by lynx-examples.

Follow-up to #3295.

## Summary by CodeRabbit

* **New Features**
  * Expanded the motion examples with dedicated Basic, Gesture, Mini, Motion Value, and Spring demos.
  * Added separate build outputs for each motion example across supported platforms.
* **Improvements**
  * Motion examples now mount consistently and support smoother hot-reload behavior during development.
  * Existing web and Lynx targets remain available.

### DIFF
--- a/examples/motion/lynx.config.js
+++ b/examples/motion/lynx.config.js
@@ -5,15 +5,17 @@ import { defineConfig } from '@lynx-js/rspeedy';
 const enableBundleAnalysis = !!process.env['RSPEEDY_BUNDLE_ANALYSIS'];
 
 export default defineConfig({
-  environments: {
-    lynx: {},
-    web: {},
-  },
   source: {
     entry: {
-      main: './src/index.tsx',
+      basic: './src/Basic/index.tsx',
+      gesture: './src/iOSSlider/index.tsx',
       mini: './src/Mini/index.tsx',
+      'motion-value': './src/MotionValue/index.tsx',
+      spring: './src/Spring/index.tsx',
     },
+  },
+  output: {
+    filename: '[name].[platform].bundle',
   },
   plugins: [
     pluginReactLynx(),
@@ -24,6 +26,10 @@ export default defineConfig({
       },
     }),
   ],
+  environments: {
+    web: {},
+    lynx: {},
+  },
   performance: {
     profile: enableBundleAnalysis,
   },
--- a/examples/motion/src/Basic/index.tsx
+++ b/examples/motion/src/Basic/index.tsx
@@ -1,5 +1,10 @@
 import { animate } from '@lynx-js/motion';
-import { runOnMainThread, useEffect, useMainThreadRef } from '@lynx-js/react';
+import {
+  root,
+  runOnMainThread,
+  useEffect,
+  useMainThreadRef,
+} from '@lynx-js/react';
 import type { MainThread } from '@lynx-js/types';
 
 import './styles.css';
@@ -58,4 +63,10 @@ export default function Basic() {
       </view>
     </view>
   );
+}
+
+root.render(<Basic />);
+
+if (import.meta.webpackHot) {
+  import.meta.webpackHot.accept();
 }
--- a/examples/motion/src/MotionValue/index.tsx
+++ b/examples/motion/src/MotionValue/index.tsx
@@ -1,6 +1,11 @@
 import { motionValue } from '@lynx-js/motion';
 import type { MotionValue } from '@lynx-js/motion';
-import { runOnMainThread, useEffect, useMainThreadRef } from '@lynx-js/react';
+import {
+  root,
+  runOnMainThread,
+  useEffect,
+  useMainThreadRef,
+} from '@lynx-js/react';
 import type { MainThread } from '@lynx-js/types';
 
 import './styles.css';
@@ -74,4 +79,10 @@ export default function Basic() {
       </view>
     </view>
   );
+}
+
+root.render(<Basic />);
+
+if (import.meta.webpackHot) {
+  import.meta.webpackHot.accept();
 }
--- a/examples/motion/src/iOSSlider/index.tsx
+++ b/examples/motion/src/iOSSlider/index.tsx
@@ -8,7 +8,12 @@ import {
   transformValue,
   useMotionValueRef,
 } from '@lynx-js/motion';
-import { runOnMainThread, useEffect, useMainThreadRef } from '@lynx-js/react';
+import {
+  root,
+  runOnMainThread,
+  useEffect,
+  useMainThreadRef,
+} from '@lynx-js/react';
 import type { MainThread } from '@lynx-js/types';
 
 import SunPng from './sun.png';
@@ -137,4 +142,10 @@ export default function Comp() {
       </view>
     </view>
   );
+}
+
+root.render(<Comp />);
+
+if (import.meta.webpackHot) {
+  import.meta.webpackHot.accept();
 }
--- a/examples/motion/src/Spring/index.tsx
+++ b/examples/motion/src/Spring/index.tsx
@@ -1,5 +1,10 @@
 import { animate } from '@lynx-js/motion';
-import { runOnMainThread, useEffect, useMainThreadRef } from '@lynx-js/react';
+import {
+  root,
+  runOnMainThread,
+  useEffect,
+  useMainThreadRef,
+} from '@lynx-js/react';
 import type { MainThread } from '@lynx-js/types';
 
 import './styles.css';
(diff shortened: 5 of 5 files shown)


examples/motion/lynx.config.js
```diff
@@ -5,16 +5,18 @@ import { defineConfig } from '@lynx-js/rspeedy';
...
-    web: {},
-  },
   source: {
     entry: {
-      main: './src/index.tsx',
+      basic: './src/Basic/index.tsx',
+      gesture: './src/iOSSlider/index.tsx',
       mini: './src/Mini/index.tsx',
+      'motion-value': './src/MotionValue/index.tsx',
+      spring: './src/Spring/index.tsx',
```
Huxpro, line 14: Dropping the `main` entry orphans half the example package.

`main` was the only entry pointing at `src/index.tsx` → `src/App.tsx`, and `App.tsx` is the only referrer of `BasicPercent`, `BasicSelector`, `ColorInterception`, `Stagger`, and `Text`. None of those five have their own entry here, and none of them call `root.render` (unlike `Basic`/`Spring`/`MotionValue`/`iOSSlider`/`Mini`). So after this PR the motion examples go from showcasing 10 demos to 5, and `src/index.tsx`, `src/App.tsx`, `src/App.css` plus those five demo folders become dead code that nothing builds and no compiler or CI check will flag.

Two ways out:

- Add the remaining five entries (each demo then needs the same `root.render` + HMR block), or
- Keep `main` alongside the new named entries.

If retiring the aggregated app is intentional, please delete `App.tsx`, `App.css`, `src/index.tsx` and the unreachable demo folders in this same PR so the tree doesn't carry code that can no longer be reached.

---

examples/motion/lynx.config.js
```diff
@@ -5,16 +5,18 @@ import { defineConfig } from '@lynx-js/rspeedy';
 const enableBundleAnalysis = !!process.env['RSPEEDY_BUNDLE_ANALYSIS'];
 
 export default defineConfig({
-  environments: {
-    lynx: {},
-    web: {},
-  },
   source: {
     entry: {
```
Huxpro, line 9: Devtools / debug support is silently dropped along with the `main` entry.

`src/index.tsx` was the only module carrying these side-effect imports:

```js
import '@lynx-js/preact-devtools';
import '@lynx-js/react/debug';
```

None of the five new entries import them, so every motion example now runs without devtools or the React debug build — and `@lynx-js/preact-devtools` becomes an unused `devDependency` in `examples/motion/package.json`.

If that's intended, drop the dependency. Otherwise these need to be present in each entry; a shared `src/bootstrap.ts` holding both imports (and imported first by every entry) would avoid repeating them five times.

---

examples/motion/lynx.config.js
```diff
@@ -5,16 +5,18 @@ import { defineConfig } from '@lynx-js/rspeedy';
...
+      basic: './src/Basic/index.tsx',
+      gesture: './src/iOSSlider/index.tsx',
       mini: './src/Mini/index.tsx',
+      'motion-value': './src/MotionValue/index.tsx',
+      spring: './src/Spring/index.tsx',
     },
   },
+  output: {
+    filename: '[name].[platform].bundle',
+  },
```
Huxpro, line 19: This is a no-op — `'[name].[platform].bundle'` is already the default.

- `DEFAULT_FILENAME` in `packages/rspeedy/core/src/config/defaults.ts:66`
- Confirmed by the resolved-config snapshot in `packages/rspeedy/core/test/plugins/output.plugin.test.ts:815-816`
- Same fallback in `packages/rspeedy/plugin-react/src/entry.ts:126`

So the PR description's "align the example output naming with the platform-specific multi-entry layout" already holds on `main` without this block. Suggest removing it; if it's meant as a deliberate pin against a future default change, a short comment saying so would help, since otherwise it reads as load-bearing config.

---

examples/motion/src/Basic/index.tsx
```diff
@@ -59,3 +64,9 @@ export default function Basic() {
     </view>
   );
 }
+
+root.render(<Basic />);
+
+if (import.meta.webpackHot) {
+  import.meta.webpackHot.accept();
+}
```
Huxpro, line 72: Each demo module is now both a reusable component (`export default`) and an entry with a mount side effect. Importing one of these anywhere mounts a stray root as an import side effect — which is exactly what `src/App.tsx` does today for all of them.

Nothing builds `App.tsx` after this PR (see the comment on `lynx.config.js`), so there's no live breakage, but the trap is latent and this PR replicates the `Mini` precedent four more times.

Consider splitting the bootstrap out — e.g. `src/Basic/entry.tsx`:

```tsx
import Basic from './index.js';

root.render(<Basic />);

if (import.meta.webpackHot) {
  import.meta.webpackHot.accept();
}
```

and pointing the entry at `entry.tsx`. That keeps the demos importable and composable at the cost of five small files, and would make it straightforward to keep an aggregated app around. Same applies to `Spring`, `MotionValue`, and `iOSSlider`.

---

examples/motion/src/MotionValue/index.tsx
```diff
@@ -75,3 +80,9 @@ export default function Basic() {
     </view>
   );
 }
+
+root.render(<Basic />);
```
Huxpro, line 84: Nit: the component in this file is `export default function Basic()` — a copy-paste artifact from `src/Basic/index.tsx`. It was harmless while `App.tsx` aliased it on import, but this line now makes the MotionValue entry read as `root.render(<Basic />)`, which is confusing at the exact place someone lands when opening the demo. Worth renaming to `MotionValueExample` while you're touching the file.

(`src/BasicSelector/index.tsx` has the same `export default function Basic()` issue, though it's outside this diff.)

---

examples/motion/src/iOSSlider/index.tsx
```diff
@@ -138,3 +143,9 @@ export default function Comp() {
     </view>
   );
 }
+
+root.render(<Comp />);
```
Huxpro, line 147: Naming nits, both surfaced by this line:

- The component is `export default function Comp()`, so the entry's mount call reads `root.render(<Comp />)`. `IOSSlider` would be self-describing.
- The entry key for this file is `gesture` (`lynx.config.js`), which matches neither the folder name nor the component, and reads as if it relates to the separate `examples/gesture` package. `ios-slider` would be easier to trace from bundle name back to source.

---

Huxpro (commented):
Reviewed statically against `6031dde` — I did not run a build.

## Overview

`examples/motion` goes from two entries (`main` → `src/index.tsx`, `mini`) to five named ones (`basic`, `gesture`, `mini`, `motion-value`, `spring`), each demo self-mounting via `root.render` with an HMR self-accept guard. `output.filename` is set and the `environments` block moves below `plugins`.

The mechanical change is clean and follows the existing `Mini` precedent. My main concern is that the PR is incomplete as scoped: converting 5 of 10 demos to standalone entries strands the other 5 plus the aggregator app, and the devtools/debug imports disappear with the `main` entry. Details inline.

## Findings

| # | Issue | Where |
|---|---|---|
| 1 | `main` entry removal orphans `App.tsx` and 5 demos (BasicPercent, BasicSelector, ColorInterception, Stagger, Text) | `lynx.config.js` |
| 2 | `@lynx-js/preact-devtools` + `@lynx-js/react/debug` imports silently dropped | `lynx.config.js` |
| 3 | `output.filename` is a no-op — already the rspeedy default | `lynx.config.js` |
| 4 | Demo modules are both component and entry; import side effect mounts a stray root | `src/Basic/index.tsx` et al. |
| 5 | Naming: `MotionValue` exports `Basic`, `iOSSlider` exports `Comp`, entry key `gesture` ≠ folder | `src/MotionValue`, `src/iOSSlider` |

I'd want 1 and 2 resolved before merge; 3–5 are cheap cleanups worth folding in.

## Things that look right

- The `environments` reorder (`web` before `lynx`, placed after `plugins`) matches the convention in `examples/react-lazy-bundle/lynx.config.js` and `examples/gesture/lynx.config.js` — a consistency improvement, not churn.
- **The changeset bot warning is a false alarm.** `@lynx-js/motion-examples` is `private: true` and `.changeset/config.json` sets `privatePackages.version: false`, so no changeset is needed here.
- The CodSpeed regression (`003-hello-list__main-thread-componentAtIndex__reuse`, -7.13% WallTime) can't be caused by an examples-only diff — that's hosted-runner Walltime noise, as CodSpeed's own banner warns.
- No security surface: examples-only, no new dependencies.

## Test coverage

There are no tests here and nothing in `.github/workflows` builds `examples/motion`, so nothing would catch the orphaned entries or the missing devtools imports — they'd only surface when someone opens the example by hand. That matches how the other examples are treated, so not a blocker, but it does raise the cost of findings 1 and 2.

---

PR #3305, opened 2026-07-30
feat(react): component-level 'background only' on the MTS-defines assembly infrastructure

Huxpro:
Explores #3297: generalize #3284's MTS-defines collector from "whole compilation" (`enableMTSRendering: false`) to "the set of `'background only'` modules" under the normal dual-thread mode, behind `pluginReactLynx({ experimental_backgroundOnlyAssembly: true })`.

**#2970 remains the semantic spec of `'background only'`; this upgrades its *implementation*.** The in-place strip stays the default path.

## What changed

**Transform (Rust)**

- `mts_defines.rs` becomes a per-module registry with two attribution views, and the define `kind` stays an extensible enum (room for #3296's root-fallback kind):
  - `defines` — full definitions collected for assembly. `collectMTSDefinesBackgroundOnly` scopes collection to `'background only'` modules (a module-level directive, or every export carrying the component-level directive — evaluated on the parsed AST, so both thread compilations agree by construction).
  - `in_place` — the `kind:id` identities a main-thread compilation registers in place (`collectMTSInPlaceDefines`), recorded without altering its output (unlike full collection, which changes codegen).
- `stubBackgroundOnlyModules` compiles a `'background only'` module for the main thread as a **stub shell**: exports survive as inert empty functions — the parent still evaluates `jsx(Feed, …)` as a value while rendering — and bodies, top-level statements, and in-place definitions are dropped. The stub replaces the module before any other plugin runs, so the main-thread compile skips its snapshot/worklet/JSX work entirely. Detection is conservative: mixed modules, re-exports, classes, and runtime-computed exports stay on the in-place path; type-only exports are ignored.

**Assembly (webpack plugin)**

- `MTSDefinesRuntimeModule` gains subtraction: the assembled section is **BG-collected defs − MT in-place def ids**. Subtraction runs *after* #3284's `kind:id` collision check, which stays as the drift alarm. A shared component (used by both the first-screen and a deferred tree) lives in a non-background-only module, so it stays in place and is never double-registered — the fixtures prove it registers exactly once.
- The runtime's `internal` module **self-registers** the assembled definitions, fenced by the new `__BACKGROUND_ONLY_ASSEMBLY__` define and `__LEPUS__`. The first wiring attempt (an extra main-thread entry import, like the `enableMTSRendering: false` mode) made the entry chunk carry two entry roots, which **broke scope hoisting and fragmented the runtime into ~45 modules (+9 kB minified)** — self-init keeps the chunk at a single entry root, and the define-fence keeps the pinned runtime surface out of every other build (the background bundle is byte-identical).
- The mode is disabled under HMR, where the background compiles with a `MIXED` target; dev builds keep the in-place path on both threads consistently.

## Behavior change (the headline one)

**Module top-level statements of a `'background only'` module no longer run on the main thread.** Today the in-place strip only empties directive function bodies; module-init side effects still execute on the main thread. Under assembly the whole module is a stub shell — stronger isolation, aligned with the directive's stated intent, but breaking for code that (accidentally) relies on main-thread module init.

Transition diagnostic: the transform emits a build warning pointing at the first dropped top-level statement — *"its top-level statements no longer run on the main thread (they still run on the background thread); move side effects the main thread relies on into a module that is not 'background only'"*. The integration test asserts the delta explicitly (marker present in the control build's main thread, absent under assembly, intact on the background).

Also characterized (from #3297's list): reference discipline disappears — definition collection never depends on reference visibility, so runtime-computed component references and delegating background-only components are not hazards on this architecture (the keep-alive patch for the in-place path is structurally unnecessary here); worklets in background-only modules register through assembly (covered by fixtures).

## Parity (in-place vs assembled)

Reusing #3284's parity methodology at three levels, all asserting **identical definition id sets**:

- transform: BG-scoped collection == LEPUS full collection == LEPUS in-place identities, for the same module (`background-only-assembly.spec.js`);
- build: control (in-place) main-thread bundle vs assembly main-thread bundle snapshot-id sets are equal (`background-only-assembly.test.ts`, fixtures adapted from the multi-module strip shape: delegating `App`, named-import `Feed` with a worklet, namespace-import `UI.Card`, logic-only `heavy-format`, shared component);
- example: same check on the minified `examples/react-background-only-assembly` output.

## Numbers (examples/react-background-only-assembly, production)

| | in-place (default) | assembly | delta |
|---|---|---|---|
| main-thread.js | 43,802 B | 44,974 B | **+1,172 B (+2.7%)** |
| background.js | 45,151 B | 45,151 B | byte-identical |
| bundle (gzip) | 43,252 B | 43,810 B | +558 B (+1.3%) |
| build time (3 runs) | ~3.2 s | ~3.2 s | no measurable delta |

The remaining main-thread delta is the assembled section's handoff scaffolding plus the pinned simplified-runtime surface (~20 members the assembled definitions may reach for dynamically — load-bearing against export shaking; on this small example most were otherwise unused). The render logic is already stripped in-place today, so the size win is bounded to module scaffolding and top-level code — on this example the isolation is the point, not the bytes. Build-time savings from skipping main-thread compilation of background-only modules should scale with how much of an app is background-only; invisible at this example's size.

## Not in scope / follow-ups

- Lazy bundles: their main-thread chunks keep the in-place path for background-only modules; per-bundle assembled sections (#3284's `renderLazyMTSDefines`) are not yet wired to subtraction.
- Deciding whether this replaces the in-place strip as *the* `'background only'` implementation — shipped behind `experimental_` for now.
- Trimming the pinned runtime surface to the members the collected definitions actually reference.

## Checklist

- [x] Documentation updated (or not required) — option TSDoc, example app, changeset call-out.
- [x] Changeset added, and when a BREAKING CHANGE occurs, it needs to be clearly marked (or not required) — the behavior change only applies under the new opt-in option and is called out in the changeset.

https://claude.ai/code/session_01MvN9GY55qsuzCbahM2H1Hb

---
_Generated by [Claude Code](https://claude.ai/code/session_01MvN9GY55qsuzCbahM2H1Hb)_


Huxpro:
Closing in favour of #3302's mechanism. Recording why, since the reasoning is the useful part.

## What was wrong with this approach

**The subtraction was built on the weaker of two formulations.**

```
here:   assembled = collect(background-only modules)
#3302:  assembled = collect(every background module) − owned(main thread)
```

Collecting a subset makes coverage depend on the subset predicate being complete, and it isn't. Concretely — verified with a build, not by reading:

```
App (plain) → Feed ('background only') → Leaf (plain, reachable only through Feed)
```

`Feed` is stubbed, so its import edge to `Leaf` leaves the main-thread graph; `Leaf` is not background-only, so the scoped collection never picks it up. Its definitions end up in neither source. No error — just a first screen missing elements. Subtracting from the total cannot fail that way.

This also means the claim in the description that keep-alive (#2977) is "structurally absent" on the assembly architecture was **wrong**. It is present, via a different mechanism: the in-place strip loses the edge to DCE after emptying the body, the stub loses it by dropping the import outright. Same outcome. The control build in that experiment loses `Leaf` too, so this is the pre-existing bug, not a regression — but it is not absent.

**The module-level directive was the wrong layer, and unsound as implemented.**

`'background only'` at module level is meaningless today — `DirectiveDCEVisitor` only visits function bodies and class members, never the module prologue — so this PR invented the semantics. And it stubs every export to an empty function, including values:

```js
'background only';
export const BRAND_COLOR = '#ff0000';   // main thread gets function(){}, not '#ff0000'
```

Wrong values rather than a build failure. Worse, `should_stub_module_level_directive` asserted that output as correct, so the bug was pinned by its own baseline.

The deeper problem is layering. A `<Background>` boundary already carries the information the stub needs; deriving module membership from it — which is what #3302's fold does, by removing the reference and letting the bundler's reachability analysis run — needs no annotation, no export-shape predicate, and no new error surface. This PR pushed a graph-cut mechanism into directive syntax and then spent its complexity budget fighting the consequences: mixed modules, class exports, barrels, re-exports, concise-body arrows, keep-alive.

And a stubbed module cannot host a main-thread island at all, so the mechanism trades away exactly the per-component reuse that makes incremental adoption work.

## What holds up

- **Classification agreement** is structural: the predicate is a pure function of module source evaluated before any pass, and both react loaders run ahead of `swc-loader`, so both targets see identical text. That part needs no verification.
- **Definition content agreement** is stronger than the in-place path, because a stubbed module produces no main-thread definition at all — the background is both the producer of the definition and the sender of the ops referencing it, so there is nothing to reconcile. #3302 keeps this property.
- **The two-view collector** (full definitions vs. identities only) is the one piece worth salvaging: #3302 reuses `collectMTSDefines` on the LEPUS target for the identity half, which also flips `reject_shared_identifiers` and makes shared-runtime imports inside fallbacks fail the build. Offered there as review note (3).

## Where #3297 goes

#3302 handles `enableMTSRendering: false`. The fold itself does not depend on that flag — it is gated, not limited. So #3297 becomes: enable the fold, full background collection and the subtraction under `enableMTSRendering: true`, so an app whose main thread *does* render business code can still have `<Background>` subtrees leave its bundle.

Opening that against `huxcc/vibrant-cori-0g6cxi`. Fixtures and the parity methodology from here carry over; the transform-side stub, the module-level directive and the scoped collector do not.

---

PR #3283, opened 2026-07-29
feat: introduce @lynx-js/rsbuild-plugin with an optionless pluginLynx()

upupming:
Bootstrap the engine-layer package so a Lynx app can eventually be built with the Rsbuild CLI directly. The Rspeedy plugins will migrate into `pluginLynx()` one PR at a time; the composition is empty for now and `@lynx-js/rspeedy` applies it as a no-op.

No changeset on purpose: releases will be batched once the migration series completes.

## Summary by CodeRabbit

* **New Features**
  * Added the `@lynx-js/rsbuild-plugin` package for configuring Lynx applications with Rsbuild.
  * Integrated Lynx plugin support into Rspeedy’s default build setup (auto-registered).
  * Added a public `pluginLynx()` entry point for plugin initialization.

* **Documentation**
  * Added setup and configuration examples for using the plugin.

* **Tests**
  * Added a test validating Lynx environment initialization with the plugin.
  * Updated coverage settings to exclude plugin-lynx test configuration files.

PR #3313, opened 2026-07-30
refactor: move pluginOutput and pluginMinify into @lynx-js/rsbuild-plugin

upupming:
Next step of the plugin migration (#3283, #3292, #3301, #3307). Unlike the previous moves, these two plugins took Rspeedy config fragments as options; the user values now flow through the Rsbuild config instead, making both plugins optionless:

- `toRsbuildConfig` forwards `output.filename` (object form only) and `output.minify`; `output.distPath` was already forwarded.
- `pluginOutput()` applies the Lynx defaults (`distPath.css`, `filename.css`, lowering `const`/`let` to `var`) on top of `getRsbuildConfig('original')`, yielding to user-set values.
- `pluginMinify()` reads `output.minify` (including the `mainThreadOptions`/`backgroundOptions` extensions) from the config it already receives.

The Rspeedy-side minify/output test matrices are untouched and stay green, which verifies the behavior is preserved; the only test change is the `toRsbuildConfig` shape snapshot gaining the two forwarded keys, plus two new cases pinning the string-form `filename` handling.

## Summary by CodeRabbit

- **New Features**
  - Added configurable minification, including support for disabling it or customizing JavaScript optimization settings.
  - Added Lynx output defaults for CSS paths and filenames while preserving custom output settings.
  - Added optional debug logging for Lynx-related build activity.

- **Bug Fixes**
  - Improved handling of output filenames and minification settings in build configuration.
  - Ensured generated runtime code remains compatible with supported Lynx environments.


upupming:
@coderabbitai review

colinaaa (approved):

PR #3346, opened 2026-07-31
feat: add a thread-aware `@lynx-js/react-signals` entry

HuJean:
## Summary by CodeRabbit

* **New Features**
  * Added thread-aware Signals support through a dedicated package.
  * Signals remain reactive on background threads and provide stable, read-only values during main-thread rendering.
  * Added signal, computed value, effect, batching, model, and React hook APIs.
  * Updated applications and benchmarks to use the new Signals APIs.

* **Documentation**
  * Updated contributor guides and A2UI documentation with Signals usage guidance.

* **Tests**
  * Added coverage for thread-specific rendering, signal behavior, and package aliasing.

## Checklist


packages/react-signals/package.json
```diff
@@ -0,0 +1,56 @@
...
+    "dist",
+    "README.md"
+  ],
+  "scripts": {
+    "build": "rslib build"
+  },
+  "dependencies": {
+    "@preact/signals": "^2.9.4",
+    "@preact/signals-core": "^1.14.4",
+    "preact": "catalog:lynxjs"
```
upupming: Shall we use peer dep? Duplicate preact version may exist if dep

upupming (dismissed):

PR #3354, opened 2026-08-03
chore(deps): align Typia with workspace TypeScript

### LINKED ISSUE #3353: `plugin-react` generates an invalid Typia validator after adding `experimental_transformBuiltinAttributeNames`
### Description

After `fa2ef9b6f` introduced `experimental_transformBuiltinAttributeNames`, `pnpm turbo build` fails when downstream examples invoke `pluginReactLynx`.

The generated validator incorrectly treats the entire `PluginReactLynxOptions | undefined` input as a string.

### Reproduction

```bash
pnpm install --frozen-lockfile
pnpm turbo build
```

The first relevant failures are:

```text
Invalid config on pluginReactLynx: `$input`.
  - Expect to be string
  - Got: object
```

and:

```text
Invalid config on pluginReactLynx: `$input`.
  - Expect to be string
  - Got: undefined
```

This can also be reproduced with the plugin test suite:

```bash
pnpm --filter @lynx-js/react-rsbuild-plugin test
```

### Actual behavior

`packages/rspeedy/plugin-react/src/validate.ts` declares:

```ts
typia.createAssertEquals<PluginReactLynxOptions | undefined>()
```

However, both `lib/validate.js` and the Rslib output contain a validator equivalent to:

```js
const __is = input => typeof input === 'string'
```

Consequently:

- `pluginReactLynx()` rejects `undefined`.
- `pluginReactLynx({ ... })` rejects objects.
- Downstream Rspeedy builds fail.
- Other concurrent Turbo tasks terminate with `SIGINT` or exit code 130 after the primary failures.

Rebuilding `@lynx-js/react-rsbuild-plugin` does not correct the generated validator.

### Expected behavior

The generated validator should accept `undefined` and valid `PluginReactLynxOptions` objects while rejecting invalid or unknown properties.

### Suspected cause

The regression was introduced when `PluginReactLynxOptions` began referencing the imported `TransformBuiltinAttributeNamesOptions` type in `fa2ef9b6f`.

This appears to trigger incorrect code generation in the current combination:

```text
typia@12.1.1
typia-rspack-plugin@3.0.1
typescript@6.0.3
```

### Possible fixes

- Introduce a local validation-safe representation of `TransformBuiltinAttributeNamesOptions`.
- Investigate whether Typia fails on the imported type or its `Readonly<Record<string, string>>` member.
- Add a build-output assertion ensuring the generated validator accepts both `undefined` and a valid options object.
- Avoid upgrading to Typia 13 without also resolving its TypeScript 7 toolchain requirements.

### Environment

- OS: macOS 14.7.8 (23H730), arm64
- Node.js: v24.18.1

Dugyu:
## Summary

Make Typia validator generation deterministic for
`@lynx-js/react-rsbuild-plugin` across different pnpm installation layouts.

## Reproducibility Note

The invalid validator was observed in an existing workspace after incremental
dependency updates. In that workspace, `pnpm turbo build` generated a validator
that expected the complete plugin options object to be a string.

A fresh clone followed by the commands below did not reproduce the failure:

```bash
pnpm install --frozen-lockfile
pnpm turbo build
```

This indicates that the failure depended on the existing pnpm virtual-store
layout rather than being consistently reproducible from the current lockfile.

## Root Cause

Typia creates TypeScript type objects in its transformer and inspects them using
`TypeFlags` imported by its internal packages.

In the affected workspace:

- the transformer used TypeScript 6.0.3;
- `@typia/core` and `@typia/transform` resolved TypeScript 5.9.3 from pnpm's
  virtual store.

Because `TypeFlags` values are not compatible across these TypeScript versions,
Typia misclassified `PluginReactLynxOptions | undefined` and generated a
validator equivalent to:

```js
typeof input === 'string'
```

The resolution was fragile because `@typia/core` and `@typia/transform` import
TypeScript without declaring it themselves. A clean installation happened to
resolve a compatible TypeScript instance, while the existing installation did
not.

The Rslib build also loaded the package tsconfig containing the ts-patch Typia
transformer while registering `TypiaRspackPlugin`. This allowed Typia
transformation to occur through two build paths and made the published output
dependent on the same incidental dependency resolution.

## Changes

- Define the exact TypeScript 6.0.3 version in the default pnpm catalog.
- Reference the catalog from the root package.
- Use a YAML anchor to assign the same version to `@typia/core` and
  `@typia/transform`, because `packageExtensions` does not support `catalog:`.

These scoped package extensions do not affect packages that require TypeScript
5.9, such as `ts-blank-space`.

## Verification

- `pnpm --filter @lynx-js/react-rsbuild-plugin test`
  - 190 passed, 1 skipped
- Targeted benchmark and React Compiler builds
- Full `pnpm turbo build`
  - 66/66 tasks passed
- Verified the published output:
  - accepts `undefined`;
  - accepts valid option objects;
  - rejects unknown properties.

Closes: #3353

## Summary by CodeRabbit

* **Chores**
  * Standardized development tooling configuration across the workspace.
  * Improved consistency and reliability for TypeScript-based build and validation processes.
  * Added guidance for maintaining compatible tooling configurations in future updates.
  * No user-facing functionality or interface changes were included in this release.

## Checklist

### DIFF
--- a/pnpm-workspace.yaml
+++ b/pnpm-workspace.yaml
@@ -99,6 +99,7 @@ peerDependencyRules:
 # Default catalogs
 catalog:
   "@microsoft/api-extractor": "7.58.12"
+  typescript: &typescript-version "6.0.3"
 
 catalogs:
   # Packages vendored from the Lynx toolchain.
@@ -150,6 +151,14 @@ overrides:
   "@rsdoctor/types>@rspack/core": "$@rspack/core"
 
 packageExtensions:
+  # Typia's internal packages import TypeScript without declaring it. Keep
+  # their TypeFlags aligned with the TypeScript instance used by the transformer.
+  "@typia/core@*":
+    dependencies:
+      typescript: *typescript-version
+  "@typia/transform@*":
+    dependencies:
+      typescript: *typescript-version
   "@rspack/test-tools@*":
     dependencies:
       wast-loader: ^1.14.1
--- a/.github/pnpm-11.instructions.md
+++ b/.github/pnpm-11.instructions.md
@@ -13,3 +13,5 @@ When bumping the `packageManager` pnpm version in package.json files, keep the `
 When adding Renovate npm minimum-release-age presets, mirror the delay in pnpm with `minimumReleaseAge` in `pnpm-workspace.yaml`; Renovate delegates lockfile maintenance to the package manager and does not enforce its own minimum release age for those updates.
 
 Do not add broad React or React DOM overrides in `pnpm-workspace.yaml`. Keep workspace React consumers on exact `react` and `react-dom` patch versions, and use scoped package metadata fixes when a tool package such as Rspress needs its internal React dependencies pinned to the same patch; Rspress SSG fails when pnpm resolves `react` and `react-dom` to different patch versions.
+
+When a tool's internal package imports TypeScript without declaring it, add an exact TypeScript dependency through a version-scoped `packageExtensions` entry. Do not override TypeScript globally, because tools such as `ts-blank-space` intentionally require an older TypeScript version.
--- a/package.json
+++ b/package.json
@@ -59,7 +59,7 @@
     "sort-package-json": "^3.7.1",
     "ts-patch": "^4.0.1",
     "turbo": "^2.10.5",
-    "typescript": "^6.0.3",
+    "typescript": "catalog:",
     "typescript-eslint": "^8.64.0",
     "vitest": "^3.2.4",
     "yaml": "^2.9.0"
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -16,6 +16,9 @@ catalogs:
     '@microsoft/api-extractor':
       specifier: 7.58.12
       version: 7.58.12
+    typescript:
+      specifier: 6.0.3
+      version: 6.0.3
   lynxjs:
     preact:
       specifier: npm:@lynx-js/internal-preact@10.29.1-20260519060144-e6da44c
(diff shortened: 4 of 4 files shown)


pnpm-workspace.yaml
```diff
@@ -150,6 +151,14 @@ overrides:
   "@rsdoctor/types>@rspack/core": "$@rspack/core"
 
 packageExtensions:
+  # Typia's internal packages import TypeScript without declaring it. Keep
+  # their TypeFlags aligned with the TypeScript instance used by the transformer.
+  "@typia/core@12.1.1":
```
upupming: We can just use * here to welcome future bump from renovate:

```suggestion
  "@typia/core@*":
```

pnpm-workspace.yaml
```diff
@@ -150,6 +151,14 @@ overrides:
...
+    dependencies:
+      typescript: *typescript-version
+  "@typia/transform@12.1.1":
```
upupming:
```suggestion
  "@typia/transform@*":
```

packages/rspeedy/plugin-react/rslib.config.ts
```diff
@@ -14,7 +14,7 @@ export default defineConfig({
         './src/loaders/invalid-import-error-loader.ts',
       index: './src/index.ts',
     },
-    tsconfigPath: './tsconfig.build.json',
+    tsconfigPath: './tsconfig.rslib.json',
```
upupming: I think we can safely keep original logic, because tsc used by rslib is not actually generate typia js code. We can refator this later since there are many other sub packages using TypiaRspackPlugin

upupming (dismissed):

Dugyu:
Update: I retested this from a fresh clone with a clean installation:

```bash
pnpm install --frozen-lockfile
pnpm turbo build
```

The build completed successfully.

This means the failure is not consistently reproducible from the current lockfile. The affected workspace had an existing pnpm virtual-store layout where Typia's undeclared internal TypeScript imports resolved to TypeScript 5.9.3, while the transformer used TypeScript 6.0.3. A fresh installation happened to resolve both sides to the same version.

The underlying dependency metadata remains fragile because `@typia/core` and `@typia/transform` import TypeScript without declaring it. Their resolution can therefore depend on the existing virtual-store layout after incremental dependency updates.

This PR should be considered a build reproducibility hardening change rather than a fix for a consistently failing clean build. It makes the TypeScript relationship explicit and prevents the validator output from depending on pnpm's incidental module layout.

packages/rspeedy/plugin-react/rslib.config.ts
Dugyu: Sounds good. I reverted the Rslib-specific tsconfig changes to keep this PR
focused on aligning Typia's internal TypeScript dependency.

The broader Typia/Rslib configuration can be investigated and refactored
consistently across packages in a follow-up.

upupming (approved):
LGTM, thanks! If you have time, can you also report an issue to typia, since this may be a bug of typia.

PR #3356, opened 2026-08-03
fix(web-core): guard event errors at wasm boundary

bytedragondance:
## Summary

- Catch cross-thread event dispatch and main-thread worklet exceptions at the Rust-to-JavaScript WebAssembly boundary.
- Consume the binding results so JavaScript exceptions cannot escape an active `MainThreadWasmContext` frame.
- Keep element `addEventListener` registration and callback handling unchanged; those are covered by the XML template / element-listener implementation.

## Testing

- [x] `pnpm install --frozen-lockfile`
- [x] `pnpm turbo build` (70/70 tasks)
- [x] `pnpm --filter @lynx-js/web-core test` (356/356 tests)
- [x] `pnpm turbo api-extractor -- --local` (25/25 tasks)
- [x] `pnpm changeset status --since=upstream/main`
- [x] `pnpm dprint check ...`
- [x] `pnpm eslint --no-warn-ignored packages/web-platform/web-core/tests/element-apis.spec.ts`

## Checklist

- [x] Tests updated.
- [x] Documentation updated.
- [x] Changeset added.

### DIFF
--- a/packages/web-platform/web-core/src/main_thread/client/element_apis/event_apis.rs
+++ b/packages/web-platform/web-core/src/main_thread/client/element_apis/event_apis.rs
@@ -331,7 +331,7 @@ impl MainThreadWasmContext {
           };
           is_caught = catch_handler.is_some();
           for handler in [bind_handler, catch_handler].iter().flatten() {
-            self.mts_binding.publish_event(
+            let _ = self.mts_binding.publish_event(
               handler,
               current_target_parent_component_id.as_deref(),
               serialized_event,
@@ -352,7 +352,7 @@ impl MainThreadWasmContext {
         if bind_handler.is_some() || catch_handler.is_some() {
           is_caught = catch_handler.is_some();
           if let Some(handler) = bind_handler {
-            self.mts_binding.publish_mts_event(
+            let _ = self.mts_binding.publish_mts_event(
               &handler,
               serialized_event,
               target_unique_id,
@@ -362,7 +362,7 @@ impl MainThreadWasmContext {
             );
           }
           if let Some(handler) = catch_handler {
-            self.mts_binding.publish_mts_event(
+            let _ = self.mts_binding.publish_mts_event(
               &handler,
               serialized_event,
               target_unique_id,
@@ -475,7 +475,7 @@ impl MainThreadWasmContext {
               .and_then(|binding| binding.borrow().component_id.clone())
           }
         };
-        self.mts_binding.publish_event(
+        let _ = self.mts_binding.publish_event(
           &handler,
           current_target_parent_component_id.as_deref(),
           serialized_event,
@@ -489,7 +489,7 @@ impl MainThreadWasmContext {
       let run_worklet_handler = current_target_element_data
         .get_framework_run_worklet_event_handler(&event_name_lowercase, "global-bindevent");
       if let Some(handler) = run_worklet_handler {
-        self.mts_binding.publish_mts_event(
+        let _ = self.mts_binding.publish_mts_event(
           &handler,
           serialized_event,
           target_unique_id,
--- a/.changeset/tough-wasps-guard.md
+++ b/.changeset/tough-wasps-guard.md
@@ -1,0 +1,5 @@
+---
+"@lynx-js/web-core": patch
+---
+
+Prevent event dispatch failures from escaping through the WebAssembly boundary.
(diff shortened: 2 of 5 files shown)


packages/web-platform/web-core/src/main_thread/client/element_apis/event_apis.rs
```diff
@@ -263,15 +263,17 @@ impl MainThreadWasmContext {
...
               handler,
               current_target_parent_component_id.as_deref(),
               serialized_event,
               target_unique_id,
               &target_element_dataset.clone().into(),
               *unique_id,
               &current_target_element_data.dataset.clone().into(),
-            );
+            ) {
+              self.defer_binding_error(error);
```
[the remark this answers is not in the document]
bytedragondance: Fixed in 7b617c21c. The Rust dispatcher now resolves and passes the target/currentTarget DOM references directly to `publishEvent` and `runWorklet`; `WASMJSBinding` no longer calls `get_dom_by_unique_id` during dispatch. The global-event fallback to currentTarget is preserved, and the unit test now fails if dispatch re-enters the WASM context for a DOM lookup.

packages/web-platform/web-elements/tests/web-elements.spec.ts
```diff
@@ -208,6 +208,33 @@ test.describe('web-elements test suite', () => {
...
+        lineCount: expect.any(Number),
+        lines: expect.arrayContaining([
+          {
+            start: expect.any(Number),
+            end: expect.any(Number),
+            ellipsisCount: expect.any(Number),
+          },
+        ]),
+      });
+    });
```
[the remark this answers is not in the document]
bytedragondance: Fixed in 7b617c21c. Truncated layout events now preserve line metadata from the pre-mutation measurement, and both the web-elements and ReactLynx tests wait for an actual truncation result and assert `ellipsisCount > 0`. The focused E2Es pass on Chromium, Firefox, and WebKit.

packages/web-platform/web-elements/src/elements/XText/XTextTruncation.ts
```diff
@@ -371,61 +374,53 @@ export class XTextTruncation
...
+  ) {
+    if (!this.#enableLayoutEvent) return;
+    let lines = measuredLines;
+    // Truncated layouts preserve their pre-mutation measurement snapshot.
+    // Otherwise, detail getters materialize and cache plain cloneable values.
+    const getTextMeasure = () =>
+      this.#textMeasure ??= new TextRenderingMeasureTool(
+        this.#dom,
+        this.#dom.getBoundingClientRect(),
+      );
```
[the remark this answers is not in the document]
bytedragondance: Fixed in b4e76dd48. Lazy layout-event measurement now uses the same `#inner-box` rectangle as the normal XText layout path, so padding and borders cannot shift line boundaries between the two paths. `pnpm turbo build` passed (66/66), and the cloneable layout-detail E2E passed on Chromium, Firefox, and WebKit (3/3).

packages/web-platform/web-core/ts/client/mainthread/elementAPIs/WASMJSBinding.ts
```diff
@@ -96,80 +125,88 @@ export class WASMJSBinding implements RustMainthreadContextBinding {
...
+        }
+      }
+    });
+  }
+
   runWorklet(
     handler: { value: unknown },
     eventObject: LynxCrossThreadEvent,
-    targetUniqueId: number,
+    target: DecoratedHTMLElement,
```
PupilTong: why do we need this?

packages/web-platform/web-elements/src/elements/XText/XTextTruncation.ts
```diff
@@ -371,61 +374,53 @@ export class XTextTruncation
...
-                      }
-                    },
-                  });
-                }
-              }
-            },
-          });
-        }
-      },
+  #materializeLayoutLines(
```
PupilTong: currently not a plan

bytedragondance: Agreed. I removed the Text layout changes and preserved the existing MTS-only lazy-read behavior.

packages/web-platform/web-core/ts/client/mainthread/elementAPIs/WASMJSBinding.ts
bytedragondance: Agreed. I restored the ID-based binding and kept this PR focused on the WASM exception boundary.

PupilTong (approved):

PR #3364, opened 2026-08-04
refactor: move pluginDev into @lynx-js/rsbuild-plugin

upupming:
Continues the migration series (#3283, #3292, #3301, #3307, #3313). Moves the last engine plugin, `pluginDev`, into `pluginLynx()`.

- `pluginDev()` is now optionless: user values are read from the Rsbuild config (`dev.assetPrefix`, `dev.client`, `server.host/port/base`, `output.filename.bundle/template`) via `getRsbuildConfig('original')`.
- New always-on `pluginServer` (`lynx:rsbuild:server`) defaults `server.host` to `0.0.0.0` when unset, replacing the default previously applied in `toRsbuildConfig`. This keeps the dev/preview bind host unchanged while preserving the raw user-host signal for IP discovery.
- The HMR runtime file `client/hmr/WebSocketClient.js` moves to the plugin package. `@lynx-js/webpack-dev-transport`, `@lynx-js/websocket`, `@lynx-js/web-rsbuild-server-middleware`, and `ipaddr.js` move with it.
- Engine-side tests port the full core dev matrix; the two deferred dev x sourcemap linkage tests are restored. Core keeps its integration tests except those requiring cross-package module mocks, which are covered by the engine ports.

## Deferred cleanups (for a follow-up sweep once the migration series lands)

Lynx-specific keys currently tunneled through the Rsbuild config namespace. The goal is for the Rspeedy-to-Rsbuild config to stay aligned with Rsbuild, so each of these should move to a Lynx-owned option surface (for example typed options on the DSL plugin) instead of widening or bending the Rsbuild types.

Marked with a `TODO` in code:

| Location | Item |
| --- | --- |
| `packages/rspeedy/plugin-lynx/src/plugins/minify.plugin.ts` | `output.minify.mainThreadOptions` / `backgroundOptions`. Also forces a duplicated local `Minify` interface, because Rsbuild's `Minify` is a type alias and cannot be augmented. |
| `packages/rspeedy/core/src/config/rsbuild/index.ts` | `dev.client.websocketTransport`. Rsbuild has no transport option at all: its web client always uses the global `WebSocket`, so the type will never widen upstream. |

Same class of problem, not marked in code yet:

| Location | Item |
| --- | --- |
| `packages/rspeedy/plugin-lynx/src/plugins/dev.plugin.ts` | Local `Filename` / `BundleFilenameContext` / `Client` interfaces duplicating Rspeedy public types, needed to read the `output.filename.bundle` / `template` tunnel keys. |
| `packages/rspeedy/plugin-lynx/src/plugins/output.plugin.ts` | The `.rspeedy` intermediate dist path is duplicated as a constant in the engine, previously imported from the Rspeedy config. |

Pre-existing on `main`, same family, listed so the sweep is complete:

| Location | Item |
| --- | --- |
| `packages/rspeedy/core/src/config/rsbuild/index.ts` | `output.sourceMap.js` is cast to `SourceMap` so it can carry `*-debugids`. Added in #342, unrelated to this series. |

Alternative worth evaluating for `websocketTransport`: the value is a module path fed to `ProvidePlugin`. If the engine passed the bare specifier `@lynx-js/websocket` instead of a pre-resolved absolute path, users could redirect it with Rsbuild's native `resolve.alias` and the tunnel key would disappear. The tradeoff is that resolution would then start from the user project instead of the plugin package.

Not included above: unrelated `TODO(doc)`, `TODO(guide)`, `TODO(OSS)`, and "remove in Rspeedy v1" markers that predate this series.

## Summary by CodeRabbit

* **New Features**
  * Development tooling is now integrated directly into the Lynx plugin.
  * Added improved development server handling for hosts, ports, asset prefixes, previews, and routes.
  * Added WebSocket-based hot module replacement support.
  * Development sourcemaps now better reflect server ports and asset paths.

* **Improvements**
  * Custom server hosts remain honored, while unspecified hosts use a suitable default.
  * Configuration values are forwarded more consistently to development tools.
  * Development configuration is now managed through `pluginLynx()`.

* **Documentation**
  * Updated release notes to reflect the development tooling migration.


upupming:
@coderabbitai review

upupming:
@coderabbitai review

upupming:
@coderabbitai review

colinaaa (approved):

PR #3382, opened 2026-08-04
feat(react): fold `&lt;Background&gt;` while the main thread keeps rendering

Huxpro:
Implements #3297, stacked on #3302 (based on `huxcc/vibrant-cori-0g6cxi`). Replaces #3305, which took a different route and was closed — reasoning [there](https://github.com/lynx-family/lynx-stack/pull/3305#issuecomment-5184530364).

```js
pluginReactLynx({ experimental_backgroundIslands: true })
```

#3302's fold never depended on the main thread being empty; it was only *gated* that way. This decouples them: every `<Background>` is still rewritten to its `fallback` on the main-thread target — so the deferred subtree's module closure leaves that bundle and its definitions are assembled there from the background compilation — while everything **around** the boundaries stays compiled and rendered on the main thread.

| | main thread compiles | first frame |
|---|---|---|
| `experimental_enableMTSRendering: false` (#3302) | the entry and its fallbacks | the fallbacks |
| this option | everything except deferred subtrees | the app, with fallbacks in the holes |
| neither | everything | the app; `<Background>` renders its fallback at runtime |

**No transform changes.** The fold, the collector and the subtraction all come from #3284/#3302 — this is the wiring that lets them run under the classic build.

## What this adds over #3302

For a single-entry app with a boundary, #3302's `'auto'` already folds and assembles. What is left:

1. **Per boundary, not per build.** #3302 flips the whole build to `experimental_enableMTSRendering: false`, so in a multi-entry app the entries *without* a boundary compile nothing for the main thread and render an empty first frame — #3302 warns about exactly this. Here they are untouched. Test: `an entry that defers nothing is untouched`.
2. **The flag keeps meaning what it says.** You get the fold while `experimental_enableMTSRendering` stays `true`, instead of opting into a switch documented as "business code is not compiled for the main thread" when it now is.

Two things this PR originally also carried have since been taken up in #3302 itself, and the merge of `40182fa5` adopts their versions:

- the **subtraction scope fix** (my review note (2)) — theirs makes `mainThreadEntry` required rather than optional, which is cleaner;
- the **module-concatenation regression** from prepending a second entry root, which they fixed with a generated single root (`loaders/mts-defines-entry-loader`) rather than the runtime self-registration used here.

The measurement that found the second one still stands and is worth recording — same fixture, same build, one variable, whether the main-thread entry gets the shim prepended:

| main-thread entry | bundle | unconcatenated module fns |
|---|---|---|
| `[...imports]` | 345,018 | 0 |
| `[mtsDefinesEntry, ...imports]` | **384,322** | **47** |

**+39,304 B from the extra entry root alone.** Islands keeps a single root because the runtime self-registers the assembled definitions (a `__BACKGROUND_ISLANDS__`/`__LEPUS__`-fenced block in `internal.ts`, dropped from every other build), so the `keeps the main-thread bundle concatenated` test holds independently of how #3302 solves it on its side.

## Definition conservation, checked against the classic build

```
owned(main thread) ⊎ assembled  ==  definitions(classic main-thread bundle)
```

On the three-island fixture: **4 owned + 3 assembled = 7**, and the classic build has exactly those 7 — nothing missing, nothing described twice. Both halves are asserted non-empty, so it cannot pass by doing nothing. This is the check I could not make hold in #3305, where the collection was scoped to a subset instead of subtracted from the total.

## Where the fold stops

`test/backgroundIslandsBoundary.test.ts` maps the boundary with 13 fixture modules across 4 build configurations — the full write-up is [in this comment](https://github.com/lynx-family/lynx-stack/pull/3382#issuecomment-5185176752). In short:

> The fold removes a JSX **reference**. A deferred module's top level stops running on the main thread **iff every reference to it lies inside a folded boundary**.

Eliminated with no annotation at all — including the transitive imports, a namespace member expression, a child returned from a call, a boundary nested in a first-screen component, a module that also defines a worklet, and render bodies carrying no directive. `'background only'` on a function never removed module top-level side effects; the fold does.

Not eliminated, and not fixable by any module-local transform: a module the shell also imports, a child slotted through a wrapper that owns the boundary, a component hoisted into a module-scope binding. `import 'background-only'` decides exactly this case — build OK when the module leaves the graph, build error in all three of the above.

## Tests

- `test/backgroundIslands.test.ts` — 7 cases: subtree cut with surroundings intact · assembled section carries real element-building code · an entry that defers nothing is untouched · conservation vs. the classic build · concatenation preserved · dev keeps the classic path · rejected combinations.
- `test/backgroundIslandsBoundary.test.ts` — 5 cases pinning both sides of the boundary above.

Suites after merging `40182fa5`: islands + boundary **12/12** · `react-webpack-plugin` **118/118** · `plugin-react` **241 passing**.

⚠️ `plugin-react` also has **6 failures that are pre-existing on this PR's base** — 3 in `mtsRendering.test.ts`, 3 in `rootBackgroundBuild.test.ts`. Verified by checking out `origin/huxcc/vibrant-cori-0g6cxi` (`40182fa5`) directly and re-running: identical 6. They belong to #3302.

## Open question

Same one I left on #3302: the deferred subtrees' definitions still ship in the first-frame chunk (2,095 B on this fixture). The boundary already says the subtree is not in the first frame, so those definitions could in principle be installed when the background is ready instead — turning the assembled section from a runtime module into a lazily-installed chunk, and the first frame into shell-only. That is the difference between trimming a couple of kB and not carrying the deferred vocabulary at all. Not attempted here; it needs the answer about whether deferring registration costs anything on the hydration path.

## Checklist

- [x] Documentation updated (or not required) — option TSDoc with the comparison table, changeset.
- [x] Changeset added, and when a BREAKING CHANGE occurs, it needs to be clearly marked (or not required) — new opt-in option, no behavior change without it.

https://claude.ai/code/session_01MvN9GY55qsuzCbahM2H1Hb


Huxpro:
## Where the fold stops, and why that is the right place

`01ca7813` adds `test/backgroundIslandsBoundary.test.ts` — 13 fixture modules through 4 build configurations, covering every shape I could construct where a deferred subtree has side effects. Recording the result here because the boundary it maps is the part worth reviewing, not the wiring.

### The headline

`'background only'` on a function **never** removed module top-level side effects — #3297 says so itself ("the module's top-level statements still execute on the main thread"). The fold does, with no annotation at all. For the enclosed shapes it is strictly stronger than the directive it replaces.

| shape | classic | islands |
|---|:---:|:---:|
| deferred module's own top level | present | **gone** |
| a module only the deferred one imports | present | **gone** |
| its render body, carrying no directive | present | **gone** |
| reached via `<UI.Card/>` (namespace member) | present | **gone** |
| child returned from a call, not a JSX identifier | present | **gone** |
| boundary nested inside a first-screen component | present | **gone** |
| deferred module that also defines a worklet | present | **gone** (worklet definition still reaches the main thread, via the assembly) |
| module the shell also imports | present | **present** |
| child slotted through a wrapper that owns the boundary | present | **present** |
| component hoisted into a module-scope binding | present | **present** |

Controls — shell, fallbacks, nested shell, worklet body — stay in every configuration. `enableMTSRendering: false` matches islands cell for cell. Development keeps everything, as designed.

### The rule

> The fold removes a JSX **reference**. Whether the module then leaves the bundle is decided by the bundler's reachability analysis over the whole graph.
>
> **A deferred module's top level stops running on the main thread iff every reference to it lies inside a folded boundary.**

The last three rows are one cause in three costumes — a reference authored *outside* the boundary:

- **shared with the shell** — it has to be in the bundle;
- **slotted** (`<Wrapper><Slotted/></Wrapper>`, boundary inside `Wrapper`) — folding it needs interprocedural knowledge of where a `children` prop lands, undecidable in general, and the parent evaluates `jsx(Slotted, …)` on the main thread regardless, so the binding must exist;
- **hoisted binding** (`const Picked = …`) — evaluated at module scope.

Not implementation gaps. No module-local syntactic transform can see those references.

### `import 'background-only'` decides this exactly

The pill is already the decision procedure for the rule above — it errors iff the module stayed in the main-thread graph, which is precisely when the fold failed to remove it:

| pill placed in | result |
|---|---|
| the deferred module (leaves the graph) | **build OK** |
| the module shared with the shell | **build error** |
| the slotted child | **build error** |
| the hoisted-binding component | **build error** |

Not a heuristic — it is checked by the same reachability analysis that decides the outcome.

### What is left uncovered, and by what

| | top level | render body |
|---|:---:|:---:|
| fold, reference enclosed | ✓ | ✓ |
| `'background only'` on the function | ✗ | ✓ (verified in every mode, including the classic build) |
| neither | — | — |

So exactly one thing has no mechanism: **the top level of a module that stays referenced.** Two sub-cases:

- **genuinely shared with the shell** — irreducible; the module must be split, and the pill is what tells the author so. (RSC is weaker here: a module imported from both graphs lands in both and its top level runs twice — no elimination is attempted.)
- **slotted / hoisted** — reducible *by declaration*: the reference must survive, the body need not. That is a checked module-level `'background only'` (exports kept as inert bindings, top level dropped, every non-inert export a build error). I built the unchecked version in #3305 and closed it — wrong as a primary mechanism, right as the escape hatch for exactly these two shapes.

Which leaves the API stack at three-of-four already in place:

```
<Background fallback>            use-site, per-instance   the model
'background only' (function)     def-site, per-function   bodies where the reference must live
'background only' (module)       def-site, per-module     ← not built; spec'd, for slotted/hoisted
import 'background-only'         def-site, per-module     decides + enforces the boundary
```

Roughly `<Suspense>`-shaped hole, `'use client'`-shaped module boundary, `server-only`-shaped pill.

**For this PR the practical takeaway is the first table**: the common shapes — including the transitive imports and the worklet case — need no annotation, and that is what the new test pins, on both sides.

---

Huxpro:
## The 6 failures are a stale-build artifact, not a regression on the base

> ⚠️ `plugin-react` also has **6 failures that are pre-existing on this PR's base** … They belong to #3302.

They don't — but the observation was worth chasing, because the trap is real and I walked into it myself during the base merge.

From a clean build at `ba604006` (`40182fa5` + a docs-only commit):

```
pnpm turbo run build --filter=@lynx-js/react-rsbuild-plugin
npx rstest run test/mtsRendering.test.ts test/rootBackgroundBuild.test.ts
→ 17 passed, 0 failed
```

### What produces them

`40182fa5` adopted the base's rename to `experimental_enableMTSRendering`, and that name is a **loader option key** — `plugin-react` writes it, `react-webpack-plugin`'s `getMainThreadTransformOptions` reads it to decide `foldBackgroundToFallback` + `collectMTSDefines`. `plugin-react`'s tests import from `../src/`, but resolve `@lynx-js/react-webpack-plugin` as a *package*, i.e. from its built `lib/`. If that `lib/` predates the rename, the loader reads the old key, gets `undefined`, and the fold silently never runs — every assertion of the form "the deferred logic left the main-thread chunk" then fails, and nothing else does.

Reproduced deliberately, by staling just that one key in the built lib:

```sh
sed -i 's/experimental_enableMTSRendering/enableMTSRendering/g' \
  packages/webpack/react-webpack-plugin/lib/loaders/options.js
```

```
→ 9 failed, 8 passed — 3 in mtsRendering.test.ts, 6 in rootBackgroundBuild.test.ts
```

Same two suites, same signature. The exact count varies with how stale the `lib/` is (a pre-merge build differs in more than the one key), which is why you saw 3 + 3 and this narrower simulation gives 3 + 6.

### Why it bites this stack specifically

The rename crosses a package boundary in a **silent** direction: an unknown option key is `undefined`, not an error, and `undefined !== false` just means "don't fold". No diagnostic anywhere. Anyone checking out either branch without rebuilding `@lynx-js/react-webpack-plugin` — or with a turbo cache entry keyed to pre-merge sources — sees it.

So: `pnpm turbo run build --filter=@lynx-js/react-rsbuild-plugin` before running `plugin-react` tests on this stack. Worth dropping the ⚠️ from the PR body once you can confirm on your side.

Separately, on the two items you list as taken up in #3302 — agreed on both, including that the required-`mainThreadEntry` shape is the cleaner one. Your +39,304 B measurement and my `p10-heavy` numbers are the same bug found from two directions; nice independent confirmation that it was the extra entry root and not something about the assembly itself.

---

PR #3388, opened 2026-08-05
feat(web-core): add element event listener PAPIs

jkzing:
> [!NOTE]
> **1 of 3.** This PR is the base of a stack that splits #3366 into reviewable
> pieces. It stands alone: nothing here depends on the other two.
>
> 1. **element event listener PAPIs** ← you are here
> 2. engine context proxy (`lynx.getEngine()`)
> 3. Lynx XML markup template support

## Summary by CodeRabbit

* **New Features**
  * Added element-level event listener APIs for registering and removing callbacks.
  * Supports capture, once, passive, closure, and binding options, including multiple listeners per event.
  * Event callbacks receive translated event details, target metadata, and dataset information.
  * Added support for client-side handler strings and listener cleanup during disposal.
  * Added compatible no-op APIs for server-side rendering.

* **Bug Fixes**
  * Prevented duplicate registrations and ensured targeted or complete listener removal.
  * Improved listener isolation, propagation handling, callback validation, and interoperability.

## What this adds

`__AddEventListener(element, name, callback, options)` and
`__RemoveEventListener(...)`: the **function callback** form of element event
binding.

The existing `__AddEvent` binds either a handler *name* (dispatched across
threads) or a worklet object, so it cannot carry a live main-thread closure. A
card that builds its UI directly from the Element PAPIs — rather than through
ReactLynx — needs the callback form, which is what these two provide.

## Behaviour

- `capture`, `once` and `passive` are honored.
- `closure_type` and `bind_type` select the binding semantics. A `kClient`
  binding with a **string** handler is routed to `__AddEvent`, since that is
  exactly the cross-thread path it already provides.
- `signal` is accepted for signature parity with the engine PAPI but is **not**
  implemented; use `__RemoveEventListener`.
- Registration identity is `(element, name, callback, capture)`, matching the
  engine's `Matches()` contract, so `once` / `passive` are not part of it and
  re-registering the same triple is a no-op, as with `EventTarget`.
- Listeners are detached on teardown, so a card that does not clean up after
  itself cannot leave closures attached to elements that outlive the view.
- SSR stubs both as no-ops, as it already does for `__AddEvent`.

Two cases worth calling out, both covered by tests:

- Elements not built through the Element PAPIs carry no unique id. They are
  ignored rather than all keyed under the string `"undefined"`, which would let a
  listener added on one be removed through another.
- A `kClient` removal clears only the cross-thread slot. Routing it through
  `__AddEvent` with no identifier takes that function's "clear everything"
  branch, which would also detach a worklet handler bound to the same element,
  type and name.

## Testing

24 unit tests in `tests/element-event-listener.spec.ts`, covering real DOM
dispatch, the option matrix, both delegation paths, dedup, disposal and the two
cases above. `web-core` is at 197 passing with no existing test changed.

## Checklist

- [x] Changeset added, and when a BREAKING CHANGE occurs, it needs to be clearly marked (or not required). — `minor`, additive only.


packages/web-platform/web-core/ts/client/mainthread/elementAPIs/createElementEventListenerAPIs.ts
```diff
@@ -0,0 +1,334 @@
...
+
+/**
+ * The element's Lynx unique id, or `undefined` when it has none.
+ *
+ * Every element built through the Element PAPIs carries one, but a card can
+ * hand these APIs an arbitrary object. Without this check such elements would
+ * all key on the string `"undefined"`, so listeners on unrelated elements would
+ * collide and removal would detach the wrong one.
+ */
+function lynxUniqueIdOf(element: HTMLElement): number | undefined {
```
PupilTong: use __GetElementUniqueID

jkzing: Done in 2bca483df — switched both the registry key and the dataset lookup to `__GetElementUniqueID`. Its `-1` sentinel is cleaner than the numeric check I had hand-rolled, and it keeps the id lookup going through one place.

Separately: thank you for the note about event bubbling being owned by the Rust delegation layer — that one is much more consequential and I have marked this PR as draft while I rework it. Reading `dispatch_event_by_path` and `#commonEventHandler` confirmed your point: JS attaches exactly one delegated listener per event name and hands Rust a `bubblePath`, and Rust owns the capture pass, the catch short-circuit, the bubble pass and global-bind. Binding per element with `addEventListener`, as this PR does, creates a parallel dispatch system that cannot interact with any of that.

Plan is to add a third handler slot alongside `framework_cross_thread_identifier` and `framework_run_worklet_identifier`, so the callback form rides the same path walk and gets capture/catch/global-bind semantics for free. Details in the thread with @jkzing; happy to move the Rust slot into its own PR first if you would prefer to review it separately.

PupilTong:
once breaks the documented registration identity — the callback fires twice. Every {once:true} add mints a fresh wrapper and files it unconditionally; the Rust-side dedup compares the wrapper, never the caller's callback, so it can't match. Measured, one click each:

sequence	callback invocations
add(h,{once:true}) × 2	2 (expected 1)
add(h,{}) then add(h,{once:true})	2 (expected 1)
onceWrappers is a single Map<callback, wrapper>, so the earlier wrapper is also orphaned — __RemoveEventListener resolves to the newest one and the old wrapper stays in the element's handler list, unreachable. The PR body states identity is (element, name, callback, capture) and that re-registering is a no-op "as with EventTarget"; that holds only on the non-once path, which is the only one tested. Keying onceWrappers by (eventName, eventType, callback) and early-returning on an existing wrapper would fix both halves — plus clearing any live wrapper in the non-once branch instead of just deleting the map entry.

packages/web-platform/web-core/ts/client/mainthread/elementAPIs/createElementEventListenerAPIs.ts
jkzing: Confirmed and fixed in b77f2d202 — thank you, this was a real bug and your diagnosis was exactly right.

I reproduced both rows before touching anything, and got the same numbers you measured:

| sequence | before | after |
|---|---|---|
| `add(h,{once:true})` × 2 | 2 | 1 |
| `add(h,{})` then `add(h,{once:true})` | 2 | 1 |

Root cause as you described: every `{once:true}` add minted a fresh wrapper and filed it unconditionally, and the handler table dedups against what it is given — the wrapper — so it could never match the caller's callback. The `Map<callback, wrapper>` then orphaned the earlier wrapper, leaving it in the element's handler list with nothing able to remove it.

Fix follows your suggestion:

- `onceWrappers` is keyed by `(eventName, eventType, callback)` rather than callback alone, so registrations for different events or passes no longer conflate
- the `once` branch returns early when that identity is already registered, so no second wrapper is minted
- the non-`once` branch now *clears the live wrapper* through the handler table rather than only deleting the map entry, and symmetrically the `once` branch clears a bare callback an earlier plain add left behind — otherwise the second row still ran twice

You were also right that the identity claim only held on the path that was tested. Four tests added to cover the untested half: the two sequences above, the reverse order (`once` then plain, where the plain registration correctly wins and keeps firing), and a cross-event case that would break under callback-only keying.

Both halves are mutation checked rather than just observed passing:

- disabling the dedup early-return → the "re-adding the same once listener" test fails
- collapsing the slot key to a constant → the cross-event test fails (it initially passed under that mutation, so I strengthened it to dispatch both events instead of only asserting the first)

PupilTong (approved):

PupilTong (approved):

PR #3393, opened 2026-08-06
fix(react): register background-collected definitions the main thread lacks

upupming:
# Summary

A production main-thread bundle can lose snapshot and worklet definitions whose only render path sits behind a background-only branch (for example `__MAIN_THREAD__ ? <A /> : <B />`): define folding and shaking remove them from the main thread, while the background still creates instances of them at runtime. The patch it then sends fails on the main thread with `Error: Snapshot not found`.

This change merges the definitions of both threads by id:

- The transform collects each module's snapshot/worklet definitions on both layers (`collectMTSDefines`), lowered to the main-thread syntax baseline.
- A runtime module in the main-thread chunk registers the background-collected definitions whose ids the main-thread bundle does not define; it renders empty when nothing is missing.
- The runtime registers them through the existing `snapshotCreatorMap` lazy path, so a definition is only instantiated when a patch first references it.

A worklet that closes over a `runtime: 'shared'` import is skipped during collection: its binding cannot be re-created outside its module.

# Verification

- `packages/react/transform/__test__/mts-defines.spec.js`: collection is identical across targets, lowered to `var`, and leaves the emitted code unchanged.
- `packages/webpack/react-webpack-plugin/test/cases/main-thread/merge-defines`: an integration build asserting the background-only snapshot and worklet land in the main-thread chunk and an already-defined snapshot is not duplicated.
- On-device (Android, LynxExplorer) with `examples/react-merge-defines`: the same bundle fails with `Snapshot not found: __snapshot_b0176_41bd1_1` without the injection, and renders the background branch correctly with it.

## Summary by CodeRabbit

- **New Features**
  - Added support for merging background-build snapshot and worklet definitions into main-thread builds.
  - Prevented missing snapshot errors while avoiding duplicate shared definitions.
  - Added a React merge-defines example (background-only snapshots and a main-thread worklet) with loading states and background-rendered content.
  - Exposed collected main-thread definitions through transform results.

- **Bug Fixes**
  - Improved reliability of main-thread rendering when features are defined only in background builds.

- **Tests**
  - Added coverage for definition collection, deduplication, rendering, and entry isolation.


packages/react/transform/src/lib.rs
```diff
@@ -845,17 +955,55 @@ fn transform_react_lynx_inner(
...
+              })
+              .collect::<Vec<_>>()
+          })
+        });
+
         TransformNodiffOutput {
           code: result.code,
           map: result.map,
-          errors: vec![],
+          errors: mts_define_errors,
```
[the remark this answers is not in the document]
upupming: Fixed in 1d6056750: the printing errors are now merged into the final `errors` field instead of being replaced by the handler diagnostics.

packages/react/transform/crates/swc_plugin_worklet/lib.rs
```diff
@@ -710,9 +761,31 @@ impl WorkletVisitor {
...
+      "const $loaded = loadWorkletRuntime(typeof globDynamicComponentEntry === 'undefined' ? undefined : globDynamicComponentEntry)" as Stmt,
+      loaded = self.worklet_runtime_loaded_ident.clone(),
+    );
+    collect_mts_define(
+      &self.mts_defs_collector,
+      MtsDefineKind::Worklet,
+      hash,
+      vec![ModuleItem::Stmt(guard), ModuleItem::Stmt(stmt)],
+    );
+  }
```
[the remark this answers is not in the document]
upupming: Each definition is rendered inside its own block, and the main-thread lowering turns the guard into a `var`, so there is no redeclaration error. Running `loadWorkletRuntime` once per definition matches the per-module emission in a normal bundle.

packages/webpack/react-webpack-plugin/src/loaders/background.ts
```diff
@@ -87,6 +88,13 @@ const backgroundLoader: LoaderDefinitionFunction<ReactLoaderOptions> = function(
       this.emitWarning(new Error(warning.text));
     }
   }
+  const buildInfo = (this as typeof this & {
+    _module?: { buildInfo?: Record<string, unknown> };
+  })._module?.buildInfo;
+  if (buildInfo && result.mtsDefines) {
+    buildInfo[MTS_DEFINES_BUILD_INFO] = result.mtsDefines;
+  }
```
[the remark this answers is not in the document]
upupming: Collection is always on for both layers, so `result.mtsDefines` is always an array and the key is rewritten on every rebuild; it cannot go stale.

packages/webpack/react-webpack-plugin/src/MTSDefines.ts
```diff
@@ -0,0 +1,168 @@
...
+export function selectMissingMTSDefines(
+  backgroundDefines: readonly MTSDefine[],
+  mainThreadDefines: readonly MTSDefine[],
+): MTSDefine[] {
+  const defined = new Set(
+    mainThreadDefines.map(({ kind, id }) => `${kind}:${id}`),
+  );
+  return backgroundDefines.filter(({ kind, id }) =>
+    !defined.has(`${kind}:${id}`)
+  );
```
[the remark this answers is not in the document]
upupming: Definitions sharing an id are interchangeable by contract — the id embeds the filename and content hashes, and the transform spec asserts both targets collect identical definitions for the same module.

packages/react/transform/crates/swc_plugin_snapshot/lib.rs
```diff
@@ -1781,8 +1869,109 @@ mod tests {
...
   use crate::JSXTransformer;
-  use swc_plugins_shared::{target::TransformTarget, transform_mode::TransformMode};
+  use swc_plugins_shared::{
+    mts_defines::{MtsDefineKind, MtsDefinesCollector},
+    target::TransformTarget,
+    transform_mode::TransformMode,
+  };
+
+  #[test]
+  fn should_collect_mts_defines_while_targeting_js() {
```
upupming: Since any target will generate snapshot & mts defines, shall we rename mtsDefines to other?

packages/react/transform/src/lib.rs
```diff
@@ -301,6 +319,77 @@ pub struct TransformNodiffOutput {
   /// @internal
   #[napi(js_name = "elementTemplates")]
   pub element_templates: Option<Vec<ElementTemplateAsset>>,
+  /// @internal
+  #[napi(js_name = "mtsDefines")]
+  pub mts_defines: Option<Vec<MTSDefine>>,
+}
+
+fn lower_to_main_thread_syntax(
```
upupming: I think we should remove this since we are using entry to generate code, so the code will be proccessed by swc loader of rspack, so we do not need to lower the syntax by ourselves.

packages/rspeedy/plugin-react/src/entry.ts
```diff
@@ -159,6 +160,8 @@ export function applyEntry(
 
         mainThreadChunks.push(mainThreadName)
 
+        mainThreadEntries[mainThreadEntry] = backgroundEntry
```
upupming: It is really strange here, mainThreadEntries[mainThreadEntry] = backgroundEntry

packages/webpack/react-webpack-plugin/src/loaders/options.ts
```diff
@@ -258,6 +258,7 @@ export function getMainThreadTransformOptions(
 
   return {
     ...commonOptions,
+    collectMTSDefines: true,
```
upupming: Do we really need this option? We can just make it by default.

packages/webpack/react-webpack-plugin/src/MTSDefines.ts
```diff
@@ -0,0 +1,138 @@
...
+/**
+ * The source of the module injected into a main-thread entry to register the
+ * definitions its bundle dropped. It is a regular module (added through
+ * `compilation.addEntry` as a \`data:\` URI), so a bundle with nothing missing
+ * gets no module and pays nothing. The definitions reference the runtime
+ * through direct member accesses on the namespace import, which the bundler
+ * narrows to the members they use. The \`require\` fallback of a development
+ * creator would escape the namespace and defeat that narrowing, so it is only
+ * emitted for definitions that call it.
+ */
```
upupming: Remove the comments

packages/webpack/react-webpack-plugin/src/MTSDefines.ts
```diff
@@ -0,0 +1,138 @@
...
+${renderMTSDefines(defines)}
+`;
+}
+
+export function renderMTSDefinesModuleURI(
+  defines: readonly MTSDefine[],
+): string {
+  return `data:text/javascript,${
+    encodeURIComponent(renderMTSDefinesModule(defines))
+  }`;
```
upupming: This will make the module id too long, do we have any better choice?

Image

packages/webpack/react-webpack-plugin/src/ReactWebpackPlugin.ts
```diff
@@ -404,6 +415,45 @@ class ReactWebpackPlugin {
...
+          const missing = selectMissingMTSDefines(
+            backgroundDefines,
+            collectLayerMTSDefines(compilation.modules, LAYERS.MAIN_THREAD),
+          );
+          if (missing.length === 0) {
+            return;
+          }
+          const request = renderMTSDefinesModuleURI(missing);
+          await Promise.all(
+            mainThreadEntries.map(([mainThreadEntry]) =>
```
upupming: Only key of mainThreadEntries is used by mainThreadEntries is a object, it is really strange.

packages/webpack/react-webpack-plugin/src/ReactWebpackPlugin.ts
```diff
@@ -404,6 +415,45 @@ class ReactWebpackPlugin {
...
+    const mainThreadEntries = Object.entries(options.mainThreadEntries ?? {});
+    if (mainThreadEntries.length > 0) {
+      const { EntryPlugin } = compiler.webpack;
+      compiler.hooks.finishMake.tapPromise(
+        this.constructor.name,
+        async (compilation) => {
+          const backgroundDefines = collectLayerMTSDefines(
+            compilation.modules,
+            LAYERS.BACKGROUND,
+          );
```
upupming: How do we assure only the defines of current entry is collected when multi entries exist.

packages/webpack/react-webpack-plugin/src/ReactWebpackPlugin.ts
```diff
@@ -404,6 +417,88 @@ class ReactWebpackPlugin {
...
+                reachableDefines(background),
+                reachableDefines(mainThread),
+              );
+              if (missing.length === 0) {
+                return;
+              }
+              const request = path.join(
+                compiler.context,
+                `__lynx-react-defines-for-main-thread.${mainThread}.js`,
+              );
```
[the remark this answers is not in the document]
upupming: Verified with a probe: a virtual module registered at a nested virtual path (e.g. `__lynx-react-defines-for-main-thread.pages/home__main-thread.js`) resolves its bare-specifier imports and bundles correctly — `VirtualModulesPlugin` never touches the real filesystem, so the path is only an identity for resolution, loader matching and the module id. A `..` can only escape `compiler.context` when the entry name itself contains a `/../` segment, and such a name already breaks the bundler's own `[name]`-derived output paths (`${entryName}/main-thread.js`) long before this virtual path is built. Keeping the entry name verbatim keeps the module id readable and collision-free, so we intentionally dropped the sanitization.

packages/react/transform/crates/swc_plugin_snapshot/lib.rs
upupming: Renamed to `definesForMainThread` across the transform output, the hand-maintained d.ts and the plugin, so the field reads as the definitions destined for the main thread on either target.

packages/react/transform/src/lib.rs
upupming: Removed the lowering passes; only hygiene and fixer remain before printing, and the injected virtual entry goes through the bundler's swc loader for syntax lowering.

packages/rspeedy/plugin-react/src/entry.ts
upupming: Replaced the record with `entryPairs: Array<{ mainThread, background }>` so the pairing is explicit and both members are load-bearing.

packages/webpack/react-webpack-plugin/src/loaders/options.ts
upupming: The option is removed entirely: the transform always collects, and the disable knob no longer exists.

packages/webpack/react-webpack-plugin/src/MTSDefines.ts
upupming: Removed.

packages/webpack/react-webpack-plugin/src/MTSDefines.ts
upupming: Replaced the data URI with a `VirtualModulesPlugin` module named `__lynx-react-defines-for-main-thread.<entry>.js`, so the module id is one short path.

packages/webpack/react-webpack-plugin/src/ReactWebpackPlugin.ts
upupming: Superseded by `entryPairs`: each pair's mainThread and background names are both used for the per-pair reachability diff.

packages/webpack/react-webpack-plugin/src/ReactWebpackPlugin.ts
upupming: Defines are now collected per entry pair via a module-graph BFS from each entry's own dependencies, and a regression test asserts two pairs stay isolated.

packages/webpack/react-webpack-plugin/src/loaders/background.ts
```diff
@@ -87,6 +88,13 @@ const backgroundLoader: LoaderDefinitionFunction<ReactLoaderOptions> = function(
       this.emitWarning(new Error(warning.text));
     }
   }
+  const buildInfo = (this as typeof this & {
+    _module?: { buildInfo?: Record<string, unknown> };
+  })._module?.buildInfo;
```
HuJean, line 96: Not need definesForBackground to merge definition?

upupming: No — the merge is one-directional because instances only flow one way. The background renders and serializes patches by snapshot id, and the main thread materializes them, so a definition the main thread lacks fails at patch time (`Snapshot not found`). The background never receives patches: it builds its own tree from its own bundle, a main-thread-only branch simply does not exist in the background VDOM, and no cross-thread id lookup ever happens there. Worklets likewise only execute on the main thread. So both compiles emit `definesForMainThread` (the main-thread side is used as the already-present baseline for the diff), and nothing needs to travel toward the background.

HuJean (dismissed):

packages/react/transform/crates/swc_plugin_worklet/lib.rs
```diff
@@ -344,6 +368,10 @@ impl VisitMut for WorkletVisitor {
       }],
     }
     .into();
+    self.collect_worklet_define(
+      collected_hash.filter(|_| !collector.saw_shared_identifiers()),
```
Yradex: The new merge path silently excludes a background-only Worklet when it closes over a `runtime: 'shared'` import. Unlike a Snapshot definition, which is compiler-generated and self-contained apart from the ReactLynx prelude and runtime globals, a Worklet definition contains the user's function body and still references bindings such as `Foo`. Filtering it here avoids emitting an unresolved binding, but when the owning module is absent from the main-thread graph it leaves the background output with an `_wkltId` and no matching `registerWorkletInternal`. The build succeeds, then the first invocation can fail the runtime lookup. Could we either materialize the shared dependency closure, or represent this as unmergeable and report a build error only when the main-thread graph does not already provide this id?

packages/webpack/react-webpack-plugin/src/ReactWebpackPlugin.ts
```diff
@@ -404,6 +419,115 @@ class ReactWebpackPlugin {
...
+                    { name: mainThread, layer: LAYERS.MAIN_THREAD },
+                    (err) => err ? reject(err) : resolve(),
+                  );
+                });
+              // The last entry dependency's exports become the entry's
+              // exports, so re-append the original entry request after the
+              // definitions: the startup sequence deduplicates it into the
+              // last slot and the entry keeps its own exports.
+              const originalRequest = compilation.entries.get(mainThread)!
+                .dependencies.at(-1)?.request;
```
Yradex: For a multi-import entry, re-appending only `dependencies.at(-1)` does not put the definitions before the entire startup sequence. With `[bootstrap, app]`, adding the virtual entry and then re-adding only `app` produces `bootstrap -> definitions -> app`; `bootstrap` can therefore use a background-only Snapshot or Worklet before it is registered. Could we preserve all original entry dependencies and re-append them after the definitions in their original order (while retaining the intended exports), and add a multi-import entry test?

Yradex (commented):
Two cases where the definition merge path remains incomplete:

packages/webpack/react-webpack-plugin/src/ReactWebpackPlugin.ts
```diff
@@ -404,6 +419,115 @@ class ReactWebpackPlugin {
...
+              return module ? [module] : [];
+            });
+            while (queue.length > 0) {
+              const module = queue.pop()!;
+              if (visited.has(module)) {
+                continue;
+              }
+              visited.add(module);
+              for (
+                const connection of moduleGraph.getOutgoingConnections(module)
```
Yradex, line 457: `getOutgoingConnections()` includes inactive/tree-shaken connections and async `import()` edges, so this traversal treats graph reachability as if every module were present during entry startup. For example, a matching definition in a main-thread async child is counted as present and suppresses the background definition, even though that child has not executed yet. In the other direction, a background lazy definition can be injected into the root entry with the wrong `globDynamicComponentEntry`. Could this use active connections plus ChunkGraph/initial-chunk ownership, and keep async definitions in their corresponding async chunk, rather than traversing the whole ModuleGraph?

packages/webpack/react-webpack-plugin/src/ReactWebpackPlugin.ts
```diff
@@ -404,6 +419,115 @@ class ReactWebpackPlugin {
...
+              virtualModules.writeModule(
+                request,
+                renderDefinesModule(missingSnapshot, missingWorklet),
+              );
+              const addEntry = (entryRequest: string) =>
+                new Promise<void>((resolve, reject) => {
+                  compilation.addEntry(
+                    compiler.context,
+                    EntryPlugin.createDependency(entryRequest),
+                    { name: mainThread, layer: LAYERS.MAIN_THREAD },
```
Yradex: Setting the entry's `layer` does not make this synthetic entry match Rspeedy's `issuerLayer: MAIN_THREAD` rule: an `EntryDependency` has no issuer. As a result, the virtual definitions module bypasses the main-thread SWC/ES2019 lowering. Collected Worklet definitions contain cloned user function bodies and are printed with the latest syntax, so optional chaining, nullish coalescing, class fields, etc. can reach the main-thread QuickJS bundle unlowered. Could we apply the dedicated main-thread lowering to this virtual request (without running the Worklet collector again), or lower each collected definition before injection, with a syntax regression test?

packages/webpack/react-webpack-plugin/src/loaders/background.ts
```diff
@@ -87,6 +91,18 @@ const backgroundLoader: LoaderDefinitionFunction<ReactLoaderOptions> = function(
...
+  if (buildInfo) {
+    if (result.definesForSnapshot) {
```
Yradex, line 98: An absent field is currently indistinguishable from a valid empty definition list. The Rsbuild plugin still accepts `@lynx-js/react: ^0.123.0`, including older transforms that do not emit `definesForSnapshot` or `definesForWorklet`; upgrading only the plugins therefore makes both sides appear empty, and the merge silently becomes a no-op. Could we raise the React compatibility floor for packages consuming this metadata and add a capability/schema handshake when `entryPairs` is enabled, so missing fields fail explicitly while `[]` remains a valid result?

Yradex (commented):
Three additional cases from the same review:

packages/webpack/react-webpack-plugin/src/ReactWebpackPlugin.ts
upupming: Fixed in 9fad46c79. The walk now stops at `import()` boundaries and pairs them across threads by the imported module's resource: a definition missing inside a lazy chunk is injected into that chunk's own main-thread boundary module (the loader appends an import for the rendered definitions and the module is rebuilt), so it executes when the chunk loads and keeps its own `globDynamicComponentEntry`; nothing from an async subgraph leaks into the root entry in either direction. Verified end to end in `examples/react-lazy-bundle` (a background-only component behind `__MAIN_THREAD__` inside the lazy component: previously the lazy main-thread section missed its snapshot, now it contains it while the root bundle stays clean) plus an integration case (`merge-defines-lazy`) and a unit test asserting async-only definitions stay out of the entry. An `import()` that exists only on the background side is intentionally left alone — injecting it elsewhere would run it under the wrong dynamic-component entry.

packages/webpack/react-webpack-plugin/src/ReactWebpackPlugin.ts
upupming: Fixed in 9fad46c79: every original entry dependency is re-appended after the definitions in its original order, so the startup dedup keeps the whole sequence intact behind the definitions and the exports stay on the last dependency. Added the multi-import test: with `[bootstrap, app]`, both modules observe the definitions as already registered and the entry exports come from `app`.

packages/react/transform/crates/swc_plugin_worklet/lib.rs
upupming: Implemented the second option in 480ad89c0. The transform now records such a worklet as an unmergeable definition (id only) instead of dropping it, and the plugin fails the build when the main thread does not already provide that id — with an error naming the worklet and the two ways out (make the owning module reachable from the main thread, or avoid closing over a shared import inside the worklet). When the id is present on the main thread everything behaves as before. Materializing the shared dependency closure stays as a follow-up: the injected module is a real module going through resolution and loaders, so emitting the shared imports (with per-define local renames and module-relative sources resolved against the owning module) is an additive extension, not a redesign.

packages/webpack/react-webpack-plugin/src/loaders/background.ts
upupming: Implemented the runtime handshake in 6ba841084: the current transform always emits both fields (an empty list stays `[]` in the buildInfo), so "no module in the background closure carries either key" is a reliable signal of a pre-merge transform — the plugin now fails the build for that entry with an explicit upgrade message instead of silently merging nothing. I left the peer range alone on purpose: the merge ships in a react patch release whose exact version is assigned by changesets at publish time, so a floor written today would be a guess, and tightening `^0.123.0` would cascade peer majors. The handshake catches the mismatch on any install topology, including ones a peer range cannot see.

packages/webpack/react-webpack-plugin/src/ReactWebpackPlugin.ts
upupming: Verified this does not reproduce with Rspeedy's actual configuration. Probe: a worklet body with `?.` and `??` behind a background-only branch, dev build of `examples/react-merge-defines` — the injected definitions module in `main-thread.js` contains the fully lowered form (`(_ref = e === null || e === void 0 ? void 0 : e.currentTarget) !== null && _ref !== void 0 ? _ref : null`, `const` → `var`), byte-consistent with the loader-processed business code around it. The main-thread lowering in Rspeedy is not keyed on `issuerLayer`, so the synthetic entry does not bypass it. Happy to add a syntax regression test if you want it pinned.

packages/webpack/react-webpack-plugin/src/loaders/background.ts
upupming: Settled on expressing the requirement through the version line instead of a runtime check (the handshake from 6ba841084 is reverted): `@lynx-js/react` now ships this as a minor, and `@lynx-js/react-rsbuild-plugin` — the package that turns the merge on via `entryPairs` — raises its peer floor to `^0.124.0`. Upgrading only the plugins now surfaces as an unmet peer instead of a silent no-op merge.

HuJean (approved):

gaoachao (approved):

colinaaa (approved):

PR #3396, opened 2026-08-06
fix(web-core): let a main-thread event handler call the Element PAPIs

### LINKED ISSUE #3395: [Bug][Web] __FlushElementTree() throws inside a main-thread event handler — the wasm element context is still borrowed during dispatch
### System Info

- `@lynx-js/web-core` **0.22.2** and **0.23.1** (current at time of filing), `dist/client_prod` in both cases
- Headless Chromium (Playwright), macOS
- Framework: reproduced with framework worklet code entirely bypassed, see below

### Details

Any `main-thread:bind*` handler that calls `__FlushElementTree()` throws on every event:

```
Error: recursive use of an object detected which would lead to unsafe aliasing in rust
    at __wbg___wbindgen_throw_… (static/js/client.js)
    at wasm-function[347]
    at s.take_timing_flags (static/js/client.js)
    at __FlushElementTree (static/js/async/web-core-main-chunk.js)
    …
    at o.runWorklet (static/js/async/web-core-main-chunk.js)
    at __wbg_runWorklet_… (static/js/client.js)
    at s.common_event_handler (static/js/client.js)
    at #commonEventHandler (static/js/async/web-core-main-chunk.js)
```

Mutate-then-flush is the documented main-thread scripting pattern, so this affects ordinary application code. The element writes themselves land — the host applies each eagerly — so the page still renders correctly. But the throw propagates out through web-core's own frames rather than the handler's, so it surfaces as an **uncaught page error** that the application cannot catch, once per event. A handler bound to a per-frame event produces hundreds of them in seconds.

### Cause

`WASMJSBinding.#commonEventHandler` calls `wasmContext.common_event_handler(...)`. That Rust method is still holding a borrow of the context when it re-enters JS through the `runWorklet` import to run the handler. If the handler then calls any element PAPI that takes `&mut self` on the same `wasmContext`, wasm-bindgen's borrow guard rejects it.

`__FlushElementTree` is the one an ordinary handler hits, via `wasmContext.take_timing_flags()` in `createElementAPI.js`:

```js
let timingFlagsAll = timingFlags.concat(wasmContext.take_timing_flags());
```

This also explains why the rest of a handler works fine. `__AddInlineStyle` calls the free wasm functions `set_inline_styles_number_key` / `add_inline_style_raw_string_key`, which take no borrow on the context object. And `WASMJSBinding.runWorklet` itself calls `get_dom_by_unique_id` without trouble, which suggests `common_event_handler` holds a shared borrow and only `&mut self` methods conflict.

### Reproduce Steps

Framework-independent — no ReactLynx, no third-party framework involved. Mount any bundle, then replace `runWorklet` on the main-thread realm with a bare flush:

```js
// The mts realm is an iframe inside <lynx-view>'s shadow root.
const frames = [];
const walk = (n) => {
  if (!n) return;
  if (n.nodeType === 1) { if (n.tagName === 'IFRAME') frames.push(n); if (n.shadowRoot) walk(n.shadowRoot); }
  for (const c of n.childNodes ?? []) walk(c);
};
walk(document.body);

for (const frame of frames) {
  const realm = frame.contentWindow;
  if (realm && typeof realm.__FlushElementTree === 'function') {
    realm.runWorklet = function () { realm.__FlushElementTree(); };
    break;
  }
}
```

Then let any main-thread-bound event fire. With a `<list>` driven by `invoke('autoScroll', { rate: '60', start: true })` and `main-thread:bindscroll`, this produced **286 uncaught errors in 5 seconds** with every line of framework worklet code out of the loop.

The equivalent application-level repro is the pattern from the Product Gallery tutorial:

```jsx
const onScrollMTS = (event) => {
  'main thread';
  __AddInlineStyle(scrollbar, 'height', h + 'px');
  __AddInlineStyle(scrollbar, 'top', t + 'px');
  __FlushElementTree();
};
```

### Impact

Any main-thread handler following the documented mutate-then-flush pattern. The visible symptom is console flooding rather than broken rendering, because element writes are applied eagerly on Web, but the errors are uncatchable by the application and drown out real diagnostics.

### Possible directions

Two that seem plausible from the outside — you'll know which fits the design:

1. Release the borrow before `common_event_handler` re-enters JS, so a handler sees a clean context.
2. Make `take_timing_flags` (and any other `&mut self` PAPI reachable from a handler) tolerate being called during dispatch — for example by deferring the drain to after the dispatch unwinds.

For what it's worth, downstream we currently work around this by attempting the flush inline and, only on a host that rejects it, republishing it after the dispatch unwinds. That keeps applications working but it is a mitigation in the framework layer, not a fix.

Huxpro:
Fixes #3395.

`__FlushElementTree()` from inside a `main-thread:bind*` handler — the documented mutate-then-flush pattern — threw `recursive use of an object detected which would lead to unsafe aliasing in rust` on every event. The element writes still landed, because the host applies each eagerly, but the throw unwound through web-core's own dispatcher frames rather than the handler's: it surfaced as an uncaught page error the application could not catch, and every handler ordered after the flush was skipped.

## Cause

`wasm-bindgen` wraps an exported struct in a `WasmRefCell` and holds a borrow for the whole exported call — *shared* for `&self`, *exclusive* for `&mut self`. Dispatch re-enters JS while `common_event_handler` is on the stack (through the `runWorklet` / `runElementClosure` bindings), so **any** `&mut self` export the handler reached failed that guard — not just `take_timing_flags` behind `__FlushElementTree`, but also `__AddEvent`, `__AddEventListener`, `__SetDataset`, `__SetCSSId` and `__CreateElement`.

There was a second, narrower instance of the same rule one level down: `dispatch_event_by_path` held the current target's `LynxElementData` across the call into JS, so a handler writing to the element it was dispatching on hit the inner `RefCell` instead.

## Changes

- Every export on `MainThreadWasmContext` now takes `&self` and reaches its mutable state through `RefCell`/`Cell` fields, so a nested call from a handler is a second shared borrow, which the guard allows.
- The dispatcher reads an element's handlers, dataset and parent component id out of its `LynxElementData` and drops the borrow before running any of them. This also subsumes the narrower `drop()` the `__AddEventListener` path already did, and applies it to the worklet and cross-thread paths and to `dispatch_global_bind_event`.
- `enable_event` and `update_css_og_style` no longer hold a field borrow across a call into JS.
- `create_element_common` reserves its slot before the attribute writes that enter JS, so a nested `__CreateElement` cannot be handed the same unique id now that the exclusive borrow is gone.
- `once` deferred its own unregistration to a microtask specifically to work around this; it now unregisters inline, before the callback runs.

## Verification

Reproduced and verified in headless Chromium, and in the jsdom + real-wasm unit harness.

**New e2e case** `basic-mts-flush-element-tree` — a `main-thread:bindtap` handler that flushes inline, with a bubble handler on the parent ordered after it. Before the fix it fails exactly as reported: the child turns green (the eager write lands) but the parent stays pink, and the page records `recursive use of an object detected which would lead to unsafe aliasing in rust` per tap. After the fix both handlers run and no page error is recorded.

Full `web-core-e2e` Playwright run (chromium), before vs after the Rust change — the only difference in the failure set is the new case flipping to green:

```
=== only BEFORE (fixed by the change) ===
basic-mts-flush-element-tree
=== only AFTER (regressions) ===
(none — two entries appeared, both verified flakes: basic-element-text-bindlayout
 is test.skip(true, …) and fails in the coverage fixture, and
 config-css-selector-false-inline-css-change-same-time passes 3/3 on both builds)
```

The ~106 failures common to both runs are pre-existing screenshot/font drift in this container, unchanged by the diff.

**New unit spec** `tests/event-dispatch-reentrancy.spec.ts` — 9 cases covering each PAPI reachable from a handler (`__FlushElementTree`, `__AddEvent`, `__AddEventListener`, `__SetDataset`/`__AddDataset`, `__CreateView`, `__SetCSSId`, writing to the dispatching element, the worklet binding, and dispatch continuing past a flush). All 9 fail on `main` and pass with the fix.

Also run: `web-core` rstest 285/285, `cargo test --all-features` 120/120, `cargo clippy` clean.

## Checklist

- [x] Documentation updated (or not required). — `web-core/AGENTS.md` gains the `&self`-only invariant and the no-borrow-across-JS rule under *Technical Learnings*.
- [x] Changeset added, and when a BREAKING CHANGE occurs, it needs to be clearly marked (or not required). — patch for `@lynx-js/web-core`; no API change, `&self` vs `&mut self` is invisible in the generated typings.

---
_Generated by [Claude Code](https://claude.ai/code/session_01WWbZAQ349VmkwnKVph9LXj)_

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed failures when Element API calls run during main-thread event handlers.
  * Handlers can safely update styles, datasets, listeners, and element trees during event dispatch.
  * Ensured subsequent and bubbling handlers continue running correctly.
  * Improved one-time event listener cleanup to occur immediately.

* **Tests**
  * Added coverage for reentrant event handling, element-tree flushing, and worklet-triggered updates.

* **Documentation**
  * Documented safe reentrant event-handling behavior.


PupilTong:
## On scope: a one-field fix, and an architectural alternative

Two observations — one about how much of this diff the reported bug actually needs, one about why the rest of it is needed at all.

### 1. The minimal fix for #3395 is a single field

`__FlushElementTree` fails only because `take_timing_flags` is `&mut self`. On its own:

```rust
timing_flags: RefCell<Vec<String>>,

pub fn take_timing_flags(&self) -> Vec<String> {
  std::mem::take(&mut *self.timing_flags.borrow_mut())
}
```

That closes the reported issue with no other signature touched.

I'm **not** proposing it as the fix. I built the wasm on `main` and called 24 PAPIs from inside a `tap` handler — 11 of them throw, not just the one behind `__FlushElementTree`:

```
__FlushElementTree  __CreateView  __CreateText  __AddEvent  __AddEventListener
__RemoveEventListener  __AddDataset  __SetDataset  __SetCSSId  __AppendElement
__RemoveElement   (+ gc(), and any catch/global-bind handler that flushes)
```

So the one-field version would leave "create an element and bind an event from a handler" broken. It's useful only as a scope reference point: the 14 signature changes are buying those other 10, not the one in the issue title.

### 2. Those 10 need it for an architectural reason that's worth fixing instead

The invariant this PR documents — *no `&mut self` export, never hold a borrow across a call into JS* — is a permanent tax on every future contributor, enforced only by review. It exists because the dispatcher runs user JS **while a Rust frame is on the stack**.

The alternative is to not do that. Have Rust compute the dispatch plan and return it; let JS drive the invocation loop:

```js
// WASMJSBinding.#commonEventHandler
const plan = this.wasmContext.plan_event(bubblePath, type, bubbles); // &self — borrow released on return
for (const step of plan) {
  try { this.#invoke(step, eventObject); } catch (e) { console.error(e); }  // no Rust frame live here
}
```

Three things make this cheaper than it looks:

- **The whole walk is statically computable.** `is_caught` is `catch_handler.is_some()` / `!catch_handlers.is_empty()` — it depends on the handler table, not on what a handler does at runtime. So capture → catch short-circuit → bubble → global-bind can all be resolved before anything runs.
- **The invocation functions are already on the JS side.** `publishEvent`, `runWorklet` and `runElementClosure` are methods on `WASMJSBinding`; Rust only calls back out through the binding. Moving the loop into JS *removes* a round-trip rather than adding one.
- **The snapshot semantics are identical** to what this PR already introduces, so there's no behavior delta to reason about.

With no Rust frame live during a handler, every export can stay `&mut self`, the `RefCell`/`Cell` fields go away, and the two new `AGENTS.md` rules stop needing to be rules.

It would also close two things this PR leaves open, both measured on this branch:

- **A handler that throws still skips every handler after it.** A child listener doing `throw new Error(...)` prevents both its same-element sibling and the parent's bubble handler from running (0 and 0), and the error never reaches the caller of `dispatchEvent`. That is the same symptom #3395 describes, just triggered by a handler bug instead of the borrow guard. A per-step `try/catch` in the JS loop closes it.
- **`__RemoveEventListener` mid-dispatch is now silently wrong.** Removing a not-yet-run sibling listener still lets it fire this event, because the dispatcher snapshots the list up front — a divergence from `EventTarget`. Before this PR the removal *threw*, so nobody could reach it; it becomes live here. A JS-side loop can re-check membership before each invocation and restore the spec behavior.

### Suggestion

Land this PR as-is. It's low-risk, mechanically verifiable, clippy/fmt/test clean, and the generated `.d.ts` are byte-identical — it unblocks #3395 now.

Then track the dispatch-plan refactor separately. When it lands, the `RefCell` fields revert to `&mut self` painlessly, and the invariant this PR had to write down stops being something a reviewer has to police.

PupilTong:
in flavor of https://github.com/lynx-family/lynx-stack/pull/3438

PR #3401, opened 2026-08-07
fix(css-serializer): drop the node:path dependency

jkzing:
## Why

`@lynx-js/css-serializer` cannot be bundled for a browser. `parse` imports
`generateHref`, which imported `node:path`, and the package exposes no subpath
exports to import around it — so the whole package is unusable in a browser
bundle even for consumers that never touch `@import` resolution:

```
× Reading from "node:path" is not handled by plugins (Unhandled scheme).
  … → css-serializer/dist/generateHref.js → node:path
```

`generateHref` now uses [`pathe`](https://www.npmjs.com/package/pathe) (137M
downloads/week, dual ESM+CJS, ships types, zero dependencies, already in this
lockfile). It replaces ~160 lines that hand-reimplemented `path.posix`:
**−145 / +23** in `src/generateHref.ts`.

Split out of #3390 at review request: that PR does not need this change, and a
behaviour change in a published package deserves its own review.

## ⚠️ Risks for the reviewer to weigh

`pathe` is **not** `path.posix` and **not** `path.win32`. It normalises `\` to
`/` by design. For every POSIX path it is byte-identical to `path.posix` — and
that is every path this repo produces — but Windows-shaped inputs change.

All values below are measured, `generateHref(projectRoot, filename, origin)`:

| projectRoot / filename | `@import` | **pathe (new)** | path.win32 (old, on Windows) | path.posix (old, on Linux/mac) |
| --- | --- | --- | --- | --- |
| `C:\proj` / `pages\index.css` | `./a.css` | `/pages/a.css` | `/pages/a.css` | `/a.css` |
| `C:\proj` / `pages\index.css` | `../shared/b.css` | `/shared/b.css` | `/shared/b.css` | `../shared/b.css` |
| `/` / `./index.css` | `\a.css` | `/a.css` | `/a.css` | `//a.css` |
| `/` / `./index.css` | `\\srv\share\x.css` | `/srv/share/x.css` | `/srv/share/x.css` | `///srv/share/x.css` |
| `/` / `./index.css` | `C:\x.css` | `../C:/x.css` | varies by drive † | `/C:/x.css` |
| `C:\proj` / `\abs\index.css` | `./a.css` | `../..//abs/a.css` | varies by drive † | `/a.css` |

† Those two rows had **no single** old Windows value. `path.win32.resolve` takes
the drive it resolves against from `process.cwd()`, so whenever no argument names
a drive, the old output depended on the build machine's drive letter:

| projectRoot / filename | `@import` | on a C: host | on D: | on E: |
| --- | --- | --- | --- | --- |
| `/` / `./index.css` | `C:\x.css` | `/x.css` | `/C:/x.css` | `/C:/x.css` |
| `C:\proj` / `\abs\index.css` | `./a.css` | `../abs/a.css` | `/D:/abs/a.css` | `/E:/abs/a.css` |

Both cells previously read `/C:/x.css` and `/abs/a.css`, which is what
`path.win32` returns **on Linux** - where the cwd has no drive at all, so win32
silently takes a "no device" branch. That is not a valid way to observe old
Windows behaviour, and the CI Windows runner (a C: host) contradicted both: those
two assertions are what made `Vitest (Windows)` fail on this PR.

Any `path.win32`-derived figure is therefore only valid for inputs where some
argument names a drive. The aggregate under Risk 1 was originally measured that
way and has been re-measured under faked drive-lettered cwds instead. This is the
second instance of that class of mistake here - the first was the differential
oracle using `node:path`'s platform-dependent default export, which is why it is
now pinned to `path.posix`.

The `\\srv\share\x.css` row is stable on every drive letter; it changes only if
the cwd is itself a UNC path on the same share (giving `/x.css`), which is not a
shape a build machine or a caller uses.

**Risk 1 — Windows hrefs change, and `pathe` matches neither oracle exactly.**

Re-measured, because the figure that was here before (`120 of 150`) had been taken
with `path.win32` on Linux. The matrix is given in full so it can be rebuilt; each
`path.win32` value was taken with `process.cwd()` faked to a drive-lettered path,
on C:, D: and E: hosts, two cwds per drive:

```
projectRoot : C:\proj | C:\ | \proj | \\srv\share | D:\a\b | /
filename    : pages\index.css | index.css | \abs\index.css | C:\proj\pages\index.css | ./index.css
@import     : ./a.css | ../shared/b.css | a\b.css | /a.css | \a.css | C:\x.css | \\srv\share\x.css
```

6 x 5 x 7 = 210 cases. Depth never mattered, only the device: no case differed
between the two cwds on the same drive. `@pkg/...` imports are excluded because
every implementation passes them through unchanged (60/60 agree), so counting them
would only inflate the ratio.

Classified before counting - an input whose old value depended on the build
machine's drive has no single old behaviour to agree *with*:

| | count |
| --- | --- |
| total | 210 |
| drive-sensitive, no single old value → excluded | 34 |
| comparable (drive-stable) | 176 |
| …of those, **unchanged** by this PR | **120 (68.2%)** |
| …of those, changed | 56 |

The 56 changes are 38 where an `@import` or an absolute `filename` names a
different root than `projectRoot` (cross-drive, or drive vs UNC), 13 involving a
`\`-rooted path with no device, and 5 with a UNC `projectRoot`. **None of the 56
consists only of POSIX paths** — every one needs a backslash or a drive letter,
which is the measured form of the "POSIX callers are unaffected" claim above.

Read 68.2% as a description of this matrix rather than a probability of breakage:
it deliberately over-samples adversarial shapes (drive roots, UNC roots,
cross-drive combinations) that no caller in this repo produces. The actionable
content is *which* classes change, listed above.

For contrast, the same matrix scored the invalid way gives 111/210, against
122/210 on a C: host and 120/210 on D: and E: — measuring `path.win32` on Linux
**understates** agreement, which is the direction that made the two cells above
wrong.

This is a one-off measurement, not part of the test suite: patching
`process.cwd()` across a whole matrix is a side effect I did not want in the
shared suite. The definition above is sufficient to rebuild it. What *is*
committed is a test asserting that no input compared against the win32 oracle is
drive-sensitive, so this cannot silently regress.

Note the `C:\proj` / `\abs\index.css` row produces `../..//abs/a.css`, a doubled
slash mid-string that no oracle emits.

**Risk 2 — a few absolute `projectRoot` values leak `process.cwd()` into the
href.** `pathe` normalises them into a UNC shape that its own `isAbsolute` regex
`/^[/\\]{2}(?!\.)/` then rejects, so `resolve` falls through to `cwd()`:

```
projectRoot='//'  cwd=/tmp  →  /tmp/a.css
projectRoot='//'  cwd=/     →  /a.css
```

Affected roots: `'//'`, `'//.'`, `'//./x'` and `'\'`, and only when combined
with a `.`-prefixed filename — i.e. `parse`'s own `'./index.css'` default. These
are pinned by value in the test suite rather than asserted as "unequal", so they
name what leaks and will turn red if `pathe` ever fixes it. A browser bundle has
no `process.cwd()` at all.

**Neither risk is observable in this repo**: `cssToAst` is the only `parse` call
site and it uses the defaults (`filename` `'./index.css'`, `projectRoot` `'/'`).
Both are recorded in the changeset so they reach the CHANGELOG.

If either is unacceptable, `path-browserify` is a drop-in alternative that is
byte-identical to `path.posix` (0 divergences) with no cwd leak — at the cost of
being unmaintained since 2020, CJS-only, no bundled types, and *keeping* the
POSIX-vs-Windows href change instead.

## What is verified

- **119 tests green**; package `tsc --noEmit` clean; `dprint` + `biome` clean;
  changeset guard passes — all on the committed bytes.
- **The differential suite still carries its weight.** It compares against a
  `node:path` oracle over ~1100 generated cases and **excludes only** inputs that
  diverge by construction (backslash-containing, and the degenerate `'//'` root).
  Each exclusion is justified in a comment, and the excluded inputs are pinned by
  value separately rather than dropped.
- A third divergence class is pinned rather than hidden: `pathe.normalize` keeps a
  leading `//` where `path.posix` collapses it (20 of 960 cases). It never reaches
  an href — `relative()` collapses it, proven over 1596 POSIX-input hrefs, none of
  which contain `//`.
- **Browser bundling proven**: `esbuild --platform=browser` exits 0 with **0**
  `node:path` occurrences in the output, and the negative control (import reverted
  to `node:path`) fails with `Could not resolve "node:path"` — so the probe
  discriminates.
- Ships as `patch`: this package is a `peerDependency` of `@lynx-js/web-core`, so
  a `minor` cascades that package to `1.0.0`, which
  `.github/scripts/check-no-major-changeset.cjs` rejects.

## Known dead code, left alone

`normalizeSlashes(projectPath)` on the non-`@` branch is now unreachable —
removing it keeps 119/119 green and `pathe` never returns a backslash (0 of 180
probed calls). The `@`-passthrough `normalizeSlashes(fullPath)` **is**
load-bearing: removing it fails 4 tests. Left in place to keep this PR to one
concern; happy to drop it if a reviewer prefers.

## Summary by CodeRabbit

- **Improvements**
  - CSS import links are generated consistently across platforms with POSIX-style path semantics.
  - CSS serialization now supports browser environments without relying on Node.js path utilities.
  - Relative project roots retain predictable resolution behavior in browser environments.
  - Windows path normalization is handled consistently, with documented platform differences.
  - Unsupported `//` project roots are now clearly documented.

- **Tests**
  - Expanded coverage validates path resolution, traversal, normalization, aliases, and generated import links across configurations.

### DIFF
--- a/.changeset/css-serializer-drop-node-path.md
+++ b/.changeset/css-serializer-drop-node-path.md
@@ -8,17 +8,15 @@ be bundled for browsers.
 `parse` imports `generateHref`, which imported `node:path`, and the package
 exposes no subpath exports to import around it. That made the whole package
 unusable in a browser bundle even for consumers that never touch `@import`
-resolution. `generateHref` now uses a small internal POSIX path helper instead.
+resolution. `generateHref` now uses `path-browserify`, which is Node's own POSIX
+`path` implementation ported verbatim and works in both a browser bundle and
+Node.
 
 `node:path`'s default export is already `path.posix` on every non-Windows
 platform, so href resolution is unchanged for any absolute `projectRoot` -
-verified against `node:path` itself over a generated input matrix. Two
-behaviours intentionally differ:
+verified against `node:path` itself over a generated input matrix. One behaviour
+intentionally differs:
 
-- A relative `projectRoot` no longer resolves against `process.cwd()`. It is
-  treated as rooted, making the output a pure function of the arguments rather
-  than of the directory the build ran from. `parse` defaults `projectRoot` to
-  `'/'`.
 - On Windows, resolution now uses POSIX semantics rather than `path.win32`, so
   hrefs no longer vary by platform. **This changes emitted `@import` hrefs for
   consumers that pass native Windows paths.** Measured examples, with
@@ -37,3 +35,10 @@ behaviours intentionally differ:
   is a `peerDependency` of `@lynx-js/web-core`, so a minor here cascades that
   package to `1.0.0`, which `.github/scripts/check-no-major-changeset.cjs`
   rejects.
+
+A relative `projectRoot` keeps resolving against `process.cwd()` in Node, exactly
+as before, because `path-browserify` is a faithful port. Note that a browser
+bundle has no `process.cwd()` and falls back to treating the path as rooted, so
+only an **absolute** `projectRoot` resolves identically in both environments.
+`parse` defaults `projectRoot` to `'/'` and every in-repo caller passes an
+absolute one.
--- a/.changeset/css-serializer-drop-node-path.md
+++ b/.changeset/css-serializer-drop-node-path.md
@@ -1,0 +1,23 @@
+---
+"@lynx-js/css-serializer": patch
+---
+
+Drop the `node:path` dependency from `@lynx-js/css-serializer` so the package can
+be bundled for browsers.
+
+`parse` imports `generateHref`, which imported `node:path`, and the package
+exposes no subpath exports to import around it. That made the whole package
+unusable in a browser bundle even for consumers that never touch `@import`
+resolution. `generateHref` now uses a small internal POSIX path helper instead.
+
+`node:path`'s default export is already `path.posix` on every non-Windows
+platform, so href resolution is unchanged for any absolute `projectRoot` -
+verified against `node:path` itself over a generated input matrix. Two
+behaviours intentionally differ:
+
+- A relative `projectRoot` no longer resolves against `process.cwd()`. It is
+  treated as rooted, making the output a pure function of the arguments rather
+  than of the directory the build ran from. `parse` defaults `projectRoot` to
+  `'/'`.
+- On Windows, resolution now uses POSIX semantics rather than `path.win32`, so
+  hrefs no longer vary by platform.
--- a/.changeset/css-serializer-drop-node-path.md
+++ b/.changeset/css-serializer-drop-node-path.md
@@ -25,7 +25,8 @@ normalizes Windows paths: `\` is a separator, a drive letter is a root, and
 backslash `projectRoot` with a relative `filename` - this now agrees with what a
 Windows build host used to emit, where a pure-POSIX helper would not. It is not a
 complete match: UNC roots, a `\`-rooted `filename` and a drive-letter `@import`
-still differ from the old Windows output. Measured examples:
+still differ, and for a drive-letter `@import` the old Windows output was not a
+single value at all. Measured examples:
 
 | `projectRoot` / `filename`    | `@import`           | now                | old, on Windows    | old, on Linux/macOS  |
 | ----------------------------- | ------------------- | ------------------ | ------------------ | -------------------- |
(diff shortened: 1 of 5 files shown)


colinaaa (dismissed):

colinaaa (approved):
I'm not sure if such a huge changeset is a good choice. But I'm fine with merging.

PR #3355, opened 2026-08-03
chore: upgrade Rslib to v1 beta

elecmonkey:
## Summary

- upgrade `@rslib/core` from 0.23.2 to 1.0.0-beta.1
- migrate deprecated `lib.autoExternal` settings to `output.autoExternal`
- explicitly configure `@typescript/native-preview` for declaration builds using `tsgo`

## Artifact compatibility

Compared all 670 Rslib outputs with 0.23.2. The differences are limited to internal chunk naming, compiler-generated identifiers, and bundle serialization. Declaration outputs and decoded Lynx executable sections are unchanged, with no behavior-affecting differences found.

## Suggestions

Some options explicitly configured in this repository are already defaults in Rslib v1, such as `format: 'esm'`, `bundle: true`, `dts: false`, and `output.target: 'node'`. We recommend removing these redundant explicit values in a follow-up cleanup to simplify the Rslib configurations.

## Summary by CodeRabbit

* **Chores**
  * Upgraded internal library build tooling to the Rslib 1.0 beta.
  * Improved declaration generation across multiple packages using the native TypeScript preview implementation.
  * Standardized build configuration while preserving existing output formats and bundling behavior.
  * No changes were made to published package behavior.
* **Documentation**
  * Added a changeset documenting the tooling upgrade.

PR #3437, opened 2026-08-11
fix(lynx-bundle-rslib-config): require @rslib/core >= 1.0.0-beta.1

upupming:
## Summary

Since #3355 the exported external-bundle config sets `output.autoExternal`, which only exists in Rslib v1 — Rslib 0.23.x does not read it, so `autoExternal` silently falls back to the CJS default `true`, workspace dependencies get auto-externalized, and the emitted bundle contains invalid code such as:

```js
module.exports = @scope/some-workspace-package;
```

The `peerDependencies` still declare `@rslib/core >= 0.22.0`, so 0.23.x consumers hit this at build time with no install-time warning (reproduced on `0.6.4-canary-20260810-948eece0`).

Bump the peer range to `^1.0.0-beta.1` so package managers surface the real requirement.

## Checklist

- [x] Changeset added

### DIFF
--- a/.changeset/lynx-bundle-rslib-config-rslib-peer.md
+++ b/.changeset/lynx-bundle-rslib-config-rslib-peer.md
@@ -1,0 +1,5 @@
+---
+"@lynx-js/lynx-bundle-rslib-config": minor
+---
+
+Require `@rslib/core ^1.0.0-beta.1` in `peerDependencies`, matching the `output.autoExternal` usage introduced by the Rslib v1 upgrade.
--- a/packages/rspeedy/lynx-bundle-rslib-config/package.json
+++ b/packages/rspeedy/lynx-bundle-rslib-config/package.json
@@ -50,7 +50,7 @@
     "@rstest/core": "catalog:rstest"
   },
   "peerDependencies": {
-    "@rslib/core": ">=0.22.0"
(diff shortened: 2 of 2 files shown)


.changeset/lynx-bundle-rslib-config-rslib-peer.md
```diff
@@ -0,0 +1,5 @@
+---
+"@lynx-js/lynx-bundle-rslib-config": patch
```
colinaaa: Update `peerDependencies` from `>=0.22.0` to `>=1.0.0-beta.1` should be considered as a breaking change. Maybe use a minor bump instead.

packages/rspeedy/lynx-bundle-rslib-config/package.json
```diff
@@ -50,7 +50,7 @@
     "@rstest/core": "catalog:rstest"
   },
   "peerDependencies": {
-    "@rslib/core": ">=0.22.0"
+    "@rslib/core": ">=1.0.0-beta.1"
```
colinaaa: Instead of `>=1.0.0-beta.1`, should we use `^1.0.0-beta.1` to avoid matching `2.x`?

colinaaa (approved):

PR #3442, opened 2026-08-11
feat(openui): add a Stack-only library condition

gaoachao:
## Problem

`createOpenUiLibrary()` statically reaches the complete default OpenUI catalog. OpenUI templates select components by runtime strings, so a bundler cannot infer that a particular application only emits `Stack`. The catalog CSS imports are also side effects, which keeps the other component styles and the embedded Material Icons font in the graph even when unused JavaScript can be eliminated.

## Change

- Move the existing 26-component default preset into an internal module without changing its default behavior.
- Add the build-wide `lynx-openui-stack-only` package condition. It resolves the same `createOpenUiLibrary` implementation to a preset that directly imports only `Stack`; caller-provided components and groups are still appended as before.
- Keep the existing public import and call site unchanged. No new public subpath is introduced.
- Add a public-package bundle-graph regression test that checks the default preset, the published package mapping, and the absence of every non-Stack catalog module, Material Icons, and Lynx UI under the condition.
- Document Lynx Speedy and native esbuild configuration in English and Chinese, and add a minor changeset for `@lynx-js/genui`.

The default resolver path still provides all 26 built-ins. The Stack-only condition is an explicit build-wide opt-in and must only be used when the Agent prompt, server schema, cached templates, and client library share the same Stack-plus-custom vocabulary.

## Usage

Application source remains unchanged:

```ts
import { createOpenUiLibrary } from '@lynx-js/genui/openui';

const library = createOpenUiLibrary({ components, componentGroups });
```

Lynx Speedy selects the preset in its resolver configuration:

```ts
resolve: {
  conditionNames: [
    'lynx-openui-stack-only',
    'import',
    'require',
    'node',
  ],
}
```

## Impact

Measured on the production Echo / Lynx Speedy build of `ai-flex-special`, with the same OpenUI import and `createOpenUiLibrary` call:

| Artifact | Before | After | Reduction |
| --- | ---: | ---: | ---: |
| `app.css` | 1,846,026 B | 14,169 B | 99.23% |
| `app.css`, gzip -9 | 832,896 B | 3,427 B | 99.59% |
| `template.js` | 3,478,886 B | 1,523,178 B | 56.22% |
| `template.js`, gzip -9 | 1,316,250 B | 448,572 B | 65.92% |
| `tasm.json` | 4,220,982 B | 2,110,061 B | 50.01% |
| `tasm.json`, gzip -9 | 1,233,664 B | 394,609 B | 68.01% |

Both app-service and Lepus source maps contain only `catalog/Stack/index.jsx` and `catalog/utils.js`. The other 25 built-ins, their CSS, `material-icons.css`, the embedded font, and `@lynx-js/lynx-ui` are absent.

## Validation

- `fnm exec --using=24.11.1 -- pnpm turbo build --summarize --filter '!benchx_cli' --filter '!@lynx-js/benchmark-*'` (65/65 tasks passed)
- `fnm exec --using=24.11.1 -- pnpm exec rstest run -c rstest.config.ts --project genui/openui` (2/2 tests passed)
- OpenUI typecheck, private package build, and public `@lynx-js/genui` build
- `pnpm pack --dry-run` verified both preset outputs and declarations are included while the nested private package manifest is not published
- Production Echo / Lynx Speedy build plus app-service, Lepus, and CSS source-map inspection
- Pre-commit ESLint, Biome, dprint, and package sorting checks

## Checklist

- [x] Tests updated.
- [x] Documentation updated.
- [x] Changeset added, and when a BREAKING CHANGE occurs, it needs to be clearly marked.


gaoachao:
Superseded by #3444.

PR #3450, opened 2026-08-11
feat(react): upgrade to Preact 11

upupming:
Migrates ReactLynx to Preact 11, consuming the published fork build: lynx-family/internal-preact#22.

Ships as a **minor** bump of `@lynx-js/react` with one breaking change: `useEffect` cleanups of unmounted components move from synchronous unmount to the after-paint flush, matching React. Page destroy still drains them synchronously, so cleanups that release native resources keep running before the runtime goes away.

## What changed

**Dependency** (`pnpm-workspace.yaml` catalog): `preact` resolves to `@lynx-js/internal-preact@11.0.0-rc.1-20260827072551-a63bfa4` (upstream 11.0.0-rc.1 plus all Lynx fork patches). `@lynx-js/preact-devtools` moves to `5.0.1-20260827072047-5f77e12`, which carries upstream preactjs/preact-devtools#522 (released-v11 adapter) alongside the fork's Lynx wiring. No overrides and no pkg.pr.new URLs remain, so the single-instance guard is back to its original form.

**Runtime adaptation** (`@lynx-js/react`):
- `Component._dirty`/`_force` booleans became the `_bits` bitfield (`__g`): `render-constants.ts` now exposes `BITS`/`COMPONENT_DIRTY`/`COMPONENT_FORCE`; renderToOpcodes, forceRootRender and the profiling hooks use bit ops.
- Hand-rolled jsx-runtime vnodes drop the removed `__d` slot.
- `SnapshotInstance`/`BackgroundSnapshotInstance`/`BackgroundElementTemplateInstance` implement `remove()` (Preact 11's `removeNode` calls `node.remove()`).
- New `withSyncEffectFlush()`: Preact 11 defers passive-effect cleanups of unmounted components to the after-paint flush (React-aligned, preactjs/preact#4299). Background destroy is the one path with no later turn to run them in — `callDestroyLifetimeFun` returns and native tears the runtime down — so both destroy paths wrap their `render(null)` with a synchronous scheduler, restoring v10's behavior. Ordinary unmounts keep the deferred semantics.
- v11 type adjustments: `useRef` requires an initial value; compat `Children.map` takes a context argument.

**Behavior changes surfaced by tests** (all judged equivalent-or-better):
- Context consumers no longer double-render on provider updates (preactjs/preact#4724): fewer `rLynxChange` flushes; patches merge into the first flush.
- List reorders use v11's LIS diff — equivalent final state via a different minimal move set.
- None outstanding for `<page />`: its attribute reset now runs as a layout effect (landed separately in #3622), so it stays inside the unmount patch as on v10.

## Test status (local)

- root `vitest`: 247 files, 2006 passed, 0 failed (includes react/runtime 816, element-template 692, testing-library, transform, signals, webpack, rspeedy projects)
- `rstest`: 202 files, 1847 passed, 0 failed
- `pnpm -r run test:type`: green
- `turbo api-extractor -- --local`: no API drift
- internal-preact fork suite (rc.1): 1152 passed / 186 skipped, lint + test:ts green

## Known open items

- `useLayoutEffect` remains aliased to `useEffect` in ReactLynx, so user effect cleanups are deferred on ordinary unmounts. Runtime code that needs a cleanup inside the unmount commit imports `useLayoutEffect` from `preact/hooks` directly (see `<page />`). Destroy is synchronous for everyone.
- Web-platform e2e suites left to CI.
- CodSpeed reports a real regression on the `*-destroyBackground` benchmarks (-5% to -17%, confirmed in Simulation mode, so not timing noise). This is the cost of correctness: destroy now actually runs every unmount cleanup. Before this PR's fix, Preact 11 deferred them to a flush that never happens during destroy, so those benchmarks were measuring a destroy that skipped the work entirely. Against the `main` (Preact 10) baseline — which also runs cleanups synchronously — the remaining gap is Preact 11's own unmount bookkeeping (the `errorParent` walk in `options.unmount`, plus the queue round-trip through `flushAfterPaintEffects`). Batching the flush to once per destroy instead of once per unmounted component would recover part of it, at the cost of no longer matching v10's ordering (cleanups would run after the tree is detached rather than during the unmount walk).

## Summary by CodeRabbit

- **New Features**
  - Upgraded the rendering engine to Preact 11.
  - Improved list reordering performance and consistency.
  - Added reliable node removal for background and snapshot rendering.

- **Bug Fixes**
  - Prevented duplicate context-consumer renders during provider updates.
  - Improved effect cleanup timing during unmount and page destruction.
  - Ensured incomplete renders reset correctly.
  - Updated developer tooling and compatibility across supported integrations.


upupming:
## CodSpeed analysis (net -2.82%, soft gate)

**Regressed** (5): concentrated on the unmount path — `008-many-use-state-destroyBackground` (-17.6% WallTime / -15.0% Simulation), `007-four-layer-views-destroyBackground` (-6.9% Simulation). This is the expected cost of Preact 11 deferring passive-effect cleanups of unmounted components to the after-paint flush (React-aligned ordering, preactjs/preact#4299): destroy now schedules cleanup work through the `unmountCleanups` queue plus an extra flush pass instead of running it inline. The `basic-performance-div-*` server-bench entries (-25%) are WallTime-on-hosted-runner measurements with a "different runtime environments" warning from CodSpeed, so treat them as noise until re-measured.

**Improved** (6): the render path got faster — `008-many-use-state-renderBackground` +11.8% (both modes), `012-attrs-compile-hydrate` +10.4%, plus smaller wins — consistent with v11's pruning (no context double-render, static-bail removal, LIS-based child diff).

If the destroy-path cost matters, follow-ups could batch the deferred cleanup flush or fast-path unmounts with no pending snapshot mutations; leaving as-is for this exploration.

packages/react/runtime/lepus/jsx-runtime/index.js
```diff
@@ -1,7 +1,7 @@
 // Copyright 2024 The Lynx Authors. All rights reserved.
 // Licensed under the Apache License Version 2.0 that can be found in the
 // LICENSE file in the root directory of this source tree.
-import { CHILDREN, COMPONENT, DIFF, DIRTY, DOM, FLAGS, INDEX, PARENT, SnapshotInstance } from '@lynx-js/react/internal';
+import { CHILDREN, COMPONENT, DIFF, DOM, FLAGS, INDEX, PARENT, SnapshotInstance } from '@lynx-js/react/internal';
```
Yradex, line 4: [should_fix] Keep function-component refs in props on the main thread

Preact 11’s `createElement` and JSX runtime only extract `ref` for non-function types, so function components receive it through `props` (and compat `forwardRef` now reads `props.ref`). This main-thread factory still deletes `ref` for every function component, while the background alias resolves to Preact 11’s JSX runtime; the ET factory and classic LEPUS `createElement`/`cloneElement` have the same v10 normalization. A component that branches on `props.ref` can therefore produce one first-screen tree on MTS and a different tree during BTS hydration. Please preserve `ref` for function-component props in all main-thread element factories, while retaining the required host/class ref handling, and add MTS/BTS parity coverage.

.changeset/preact-v11.md
```diff
@@ -0,0 +1,23 @@
...
+Upgrade the bundled Preact fork to Preact 11 (`@lynx-js/internal-preact`
+based on `11.0.0-rc.1`).
+
+**Breaking:** `useEffect` cleanups of unmounted components no longer run
+synchronously during unmount. They run in the after-paint flush instead,
+matching React. Page destroy is unaffected — it drains them synchronously,
+so cleanups that release native resources still run before the runtime
+goes away. Code that assumed a cleanup had already run right after a
+re-render removed the component needs to await a flush, or use
+`useLayoutEffect` from `preact/hooks` when the work must stay inside the
```
Yradex: [should_fix] Cover `useLayoutEffect` in the cleanup-timing migration

`@lynx-js/react` still exports `useLayoutEffect` as the same Preact passive effect hook, so the deferred unmount-cleanup change affects both public `useEffect` and public `useLayoutEffect`. The note only names `useEffect` and recommends importing `preact/hooks`, but generated consumers do not declare `preact` directly and that package is an implementation singleton, so this is not a reliable public migration path under isolated package managers and TypeScript resolution. Please document that ReactLynx’s exported `useLayoutEffect` is affected too and provide or document a supported API/import for cleanup that must stay inside the unmount commit.

upupming: Updated in changeset

packages/react/runtime/lepus/jsx-runtime/index.js
upupming: Good catch! Should be fixed in https://github.com/lynx-family/lynx-stack/pull/3688

packages/react/runtime/src/utils.ts
```diff
@@ -84,6 +85,31 @@ export function hook<T, K extends keyof T>(
...
+ *
+ * Preact 11 defers passive-effect cleanups of unmounted components to that
+ * flush. Page destroy is the one path with no later turn to run them in:
+ * `callDestroyLifetimeFun` returns and native tears the runtime down, so a
+ * deferred cleanup would never run at all.
+ */
+export function withSyncEffectFlush<T>(fn: () => T): T {
+  // eslint-disable-next-line @typescript-eslint/unbound-method
+  const previousRequestAnimationFrame = options.requestAnimationFrame;
+  options.requestAnimationFrame = cb => {
```
HuJean, line 99: no need to use lynx.requestAnimationFrame?

HuJean (approved):

colinaaa (approved):

luhc228 (approved):

PR #3453, opened 2026-08-11
feat(react): introduce MainThreadObject core

Huxpro:
## Summary

Implements the core `MainThreadObject` model from #3446 as an incremental evolution of the unchanged #2144 baseline commit `a2ab025746fa55d69971490bbe160a2b75157392`.

- adds `MainThreadObjectType<I, O>`, opaque `MainThreadObjectHandle<O>`, `defineMainThreadObjectType`, and `useMainThreadObject`
- uses a runtime-owned `WeakSet` brand instead of the forgeable `__MAIN_THREAD_VALUE__` marker
- version-checks the typed handle wire protocol and detects conflicting type registrations
- keeps one realized object identity across first-screen hydration and adds deterministic disposal
- preserves the existing two-entry `MainThreadRef` wire shape and behavior
- gives actionable background-access and lazy/runtime mismatch diagnostics
- retains `MainThreadValue` only as a deprecated compatibility adapter for the separately reviewable Motion migration

Closes #3447
Refs #3446

## Required baseline comparison

Review: [`a2ab025..agent/main-thread-object-core`](https://github.com/lynx-family/lynx-stack/compare/a2ab025746fa55d69971490bbe160a2b75157392...agent/main-thread-object-core)

```text
$ git diff --stat a2ab025746fa55d69971490bbe160a2b75157392..HEAD
310 files changed, 9555 insertions(+), 3516 deletions(-)
```

The large baseline stat includes the required merge of current `main`. The focused follow-up commit is:

```text
$ git diff --stat 8603345e1..HEAD
28 files changed, 858 insertions(+), 159 deletions(-)
```

### Mechanism mapping

| Original `MainThreadValue` mechanism | `MainThreadObject` replacement |
| --- | --- |
| `MainThreadValue<T>` conflates payload and target | `MainThreadObjectType<I, O>` plus opaque `MainThreadObjectHandle<O>` |
| `__MAIN_THREAD_VALUE__` duck-typed capture marker | module-owned `WeakSet` handle registry |
| `workletCapture` | `captureMainThreadObject` (deprecated alias retained for compatibility) |
| string + factory registration | immutable typed definition with create/dispose and collision detection |
| typed three-entry ref patch | protocol-versioned four-entry object patch; ordinary ref tuple unchanged |
| first-screen realized-value `WeakSet` | identity-preserving hydration plus redundant-target disposal |
| fields attached to third-party targets | runtime-owned WeakMaps/Sets only |

### Current-main merge deviation

[...]

Issue #3448, opened 2026-08-11
Migrate #3405 MotionValue from MainThreadValue to MainThreadObject


Huxpro:
Parent RFC: https://github.com/lynx-family/lynx-stack/issues/3446

Depends on #3447 and its core `MainThreadObject` implementation PR.

## Outcome

Migrate the MotionValue portion of #3405 from `MainThreadValue` to the accepted `MainThreadObject` primitive while preserving the public Motion API and behavior.

The purpose is not to add more declarative Motion features. The purpose is to make the library-internal cost and benefit of the new primitive directly observable.

## Scope

- migrate `useMotionValue` and its internal handle/type registration;
- migrate declarative style recognition and live MotionValue bindings where they depend on the old handle protocol;
- preserve real main-thread `MotionValue` identity and upstream-compatible method calls;
- preserve initial-value rendering on the background runtime;
- remove old `MainThreadValue`-specific duck typing, casts, and registration calls where the new primitive supersedes them;
- keep `motion.view`, `motion.text`, `motion.image`, `motion.create`, and existing tests behaviorally unchanged.

## Required review shape

The PR must be stacked on the core `MainThreadObject` PR and include a focused comparison of Motion library internals:

```text
MainThreadValue implementation
        ↓
MainThreadObject implementation
```

Include:

- before/after implementation snippets for `useMotionValue`;
- diffstat restricted to `packages/motion`;
- a list of adapter/forwarding code deleted, retained, or newly required;
- an explicit statement of how much upstream `MotionValue` code remains unchanged.

## Acceptance criteria

- [ ] Public `useMotionValue(initial)` API is unchanged.
- [ ] Main-thread functions receive the real MotionValue, not a forwarding facade.
- [ ] `style={{ x }}` retains background initial rendering and live main-thread updates.
- [ ] Multiple handlers and animations observe the same MotionValue identity.
- [ ] First-screen hydration does not recreate or reset a used MotionValue.
- [ ] No old `__MAIN_THREAD_VALUE__` duck typing remains in Motion.
- [ ] No behavior expansion is mixed into this migration PR.
- [ ] Unit tests and declarative Motion integration tests pass.
- [ ] PR description contains the requested before/after evidence.

## Decision evidence

The final PR summary should answer: did `MainThreadObject` remove a meaningful adapter layer, or did it mostly rename `MainThreadValue` and remove `.current`? This evidence determines whether the core primitive is worth standardizing.

Huxpro:
[...]

Issue #3446, opened 2026-08-11
RFC: MainThreadObject — target-runtime objects, handles, and bindings


Huxpro:
# RFC: MainThreadObject — target-runtime objects, handles, and bindings for ReactLynx

## Summary

Introduce a precise model for identity-bearing JavaScript objects that exist in the ReactLynx main-thread runtime but are referenced from background-rendered React code and captured by main-thread functions.

This RFC refines the `MainThreadValue` mechanism proposed by #2144 and used by #3405. The proposed name is `MainThreadObject`: the target-runtime semantic object. Its background representation is a `MainThreadObjectHandle`; rebinding is represented separately by `MainThreadBinding` / `MainThreadRef`.

Related work:

- #2144 — `MainThreadValue` core, commit `a2ab025746fa55d69971490bbe160a2b75157392`
- #3405 — declarative Motion components and MotionValue integration
- #1935 — Motion support tracking

## Motivation

`MainThreadValue` proves that ReactLynx needs more than JSON closure snapshots and `{ current }` cells. Libraries such as Motion need a stable, stateful object on the main thread with its real object model (`get`, `set`, subscriptions, animation state, identity), while background React can only retain a serializable descriptor.

The current name and type shape obscure the actual semantics:

- it is not a value shared between runtimes;
- the background object is an opaque handle, not the target object;
- `MainThreadValue<T>` uses `T` for the initialization payload but does not type the realized object;
- factory registration, ownership, capture, hydration, and rebinding are conflated;
- `MainThread.Element`, `MainThreadRef`, and MotionValue share handle infrastructure but have different creation and ownership semantics.

### 1. MainThreadObject

A JavaScript object whose object model is available in the main-thread runtime. It may be pure JavaScript or native-backed.

```text
MainThreadObject
├── MutableCell<T>               // target object of MainThreadRef
├── MotionValue<T>               // pure-JS, factory-realized
└── MainThread.Element           // native-backed JS object model
```

`MainThreadObject` describes the target-runtime semantic category. It does not imply native ownership or cross-runtime shared state.

### 2. MainThreadObjectType<I, O>

A static definition connecting a serializable initialization payload `I` to a realized main-thread object `O`.

```ts
interface MainThreadObjectType<I, O extends object> {
  readonly type: string
  create(initialValue: I): O
  dispose?(object: O): void
}
```

The exact public shape is open for discussion. The important part is that both initialization and output types are represented.

### 3. MainThreadObjectHandle<O>

[...]

PR #3455, opened 2026-08-11
refactor(motion): migrate Motion to MainThreadObject

Huxpro:
## Summary

Implements the Motion portion of the MainThreadObject RFC.

- migrates `useMotionValue` from the legacy `MainThreadValue` subclass/registry path to a module-stable typed `MainThreadObjectType`
- preserves the public `MotionValue<T>` return type and runtime behavior
- removes React-internal marker/serialization duck typing from declarative styles; Motion now owns background-only initial-value metadata in a `WeakMap`
- lands the declarative Motion component surface used to validate MainThreadObject-backed style bindings, variants, tap/hover gestures, and animation lifecycle callbacks
- adds disposal through `MotionValue.stop()`

Part of #3446.
Implements #3448.
Stacked on #3453; please review this PR with base `agent/main-thread-object-core`.

## Migration

Before:

```ts
class MotionValueHandle<T> extends MainThreadValue<T> {}
MainThreadValue.register(TYPE, motionValue)
return new MotionValueHandle(initialValue)
```

After:

```ts
const MotionValueType = defineMainThreadObjectType({
  type: '@lynx-js/motion/MotionValue',
  create: motionValue,
  dispose: value => value.stop(),
})

return useMainThreadObject(MotionValueType, initialValue)
```

The old `MainThreadValue` subclass, manual registration, casts around its serialized shape, and `__MAIN_THREAD_VALUE__` marker checks are gone from Motion. Motion keeps only a private weak association from its own returned handle to the initial style value; this does not inspect ReactLynx internals or broaden handle access in the background runtime.

The upstream `motion-dom` factory and public `useMotionValue<T>(): MotionValue<T>` contract are unchanged.

## Review map

- `packages/motion/src/hooks/useMotionValue.ts`: typed object definition, hook migration, lifecycle disposal, Motion-owned metadata
- `packages/motion/src/declarative/style.ts`: initial style and binding collection without React handle duck typing
- `packages/motion/src/declarative/motion.tsx`: declarative binding integration
- `packages/motion/__tests__`: handle shape, main-thread realization identity/state, initial style, binding, gestures, variants, lifecycle
- `packages/motion/README.md` and `examples/motion`: public declarative API and examples

## Diff scope

Against `agent/main-thread-object-core`:

- 22 files
- 1,607 insertions
- 40 deletions

This includes the focused declarative Motion implementation plus the MainThreadObject migration commit. No ReactLynx Core files are modified in this stacked PR.

## Validation

Passed:

- `pnpm --filter @lynx-js/motion test` — 15 files, 117 tests
- focused declarative/hook runtime integration tests
- ReactLynx MainThreadObject runtime tests — 27 tests
- React transform suite — 80 passed, 10 repository-configured skips (including all 52 fixture tests)
- `pnpm --filter @lynx-js/motion build` — JS build, declaration generation, publint
- targeted ESLint
- full `pnpm dprint check`
- full `pnpm biome check`
- commit-time lint-staged ESLint/Biome/dprint checks
- `git diff --check`
- search confirms no legacy `MainThreadValue` or `__MAIN_THREAD_VALUE__` use remains in Motion/examples

Environment notes:

- the initial sandboxed transitive Turbo build reached 10/11 tasks and stopped in the unrelated `@lynx-js/rspeedy` publint temporary-tarball step; the directly relevant Motion build subsequently passed outside that sandbox
- a standalone `@lynx-js/reactlynx-testing-library` rebuild could not resolve its already-built workspace dependency `@lynx-js/testing-environment` in this worktree; the matching tested Core worktree artifact was used to bootstrap the test runner, after which the complete Motion suite passed

## Compatibility

- `useMotionValue` API and MotionValue behavior remain unchanged
- first-frame style resolution remains on the background side using Motion-owned metadata
- worklet hydration yields one real MotionValue per object ID and preserves that identity across handlers
- declarative binding is exercised through the public Motion component surface

### DIFF
--- a/.changeset/declarative-motion-components.md
+++ b/.changeset/declarative-motion-components.md
@@ -1,0 +1,5 @@
+---
+"@lynx-js/motion": minor
+---
+
+Add declarative `motion.view`, `motion.text`, `motion.image`, and `motion.create()` components with initial styles, reactive animation targets, transitions, and live MotionValue style bindings.
--- a/.changeset/motion-main-thread-object.md
+++ b/.changeset/motion-main-thread-object.md
@@ -1,0 +1,5 @@
+---
+"@lynx-js/motion": patch
+---
+
+Migrate Motion values to the typed `MainThreadObject` primitive while preserving the public `useMotionValue` API and declarative style behavior.
--- a/.github/motion.instructions.md
+++ b/.github/motion.instructions.md
@@ -6,6 +6,6 @@ Lynx for Web wraps each MTS chunk with a mutable lexical `window` binding so dep
 
 The Lynx for Web MTS realm also exposes browser DOM constructors. Motion must replace the constructors it already shims, such as `Element` and `HTMLElement`, because Lynx main-thread elements are not browser DOM nodes.
 
-Keep `useMotionValue` typed as Motion's real `MotionValue`, but represent it as an opaque `MainThreadValue` handle on the background thread. Register the Motion factory from the hook execution path so it also runs during the main-thread render; module initialization alone may be shared or cached before that render.
+Keep `useMotionValue` typed as Motion's real `MotionValue`, but realize it through a module-stable `MainThreadObjectType` and represent it as an opaque handle on the background thread. Call `useMainThreadObject` from the hook execution path so the definition is registered during main-thread render; module initialization alone may be shared or cached before that render. Track background-only initial-style metadata in Motion-owned weak state instead of duck-typing ReactLynx handle fields.
 
 Keep declarative Motion semantics in `packages/motion`: render `initial` as ordinary Lynx style for the first frame, use `main-thread:ref` plus `runOnMainThread` for prop-driven animation updates, and bind MotionValue styles with the MTS `styleEffect`. Do not add Motion-specific lifecycle or prop handling to ReactLynx.
--- a/.github/motion.instructions.md
+++ b/.github/motion.instructions.md
@@ -7,3 +7,5 @@ Lynx for Web wraps each MTS chunk with a mutable lexical `window` binding so dep
 The Lynx for Web MTS realm also exposes browser DOM constructors. Motion must replace the constructors it already shims, such as `Element` and `HTMLElement`, because Lynx main-thread elements are not browser DOM nodes.
 
 Keep `useMotionValue` typed as Motion's real `MotionValue`, but represent it as an opaque `MainThreadValue` handle on the background thread. Register the Motion factory from the hook execution path so it also runs during the main-thread render; module initialization alone may be shared or cached before that render.
+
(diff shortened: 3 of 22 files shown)


Huxpro:
Documentation companion: lynx-family/lynx-website#1294 (https://github.com/lynx-family/lynx-website/pull/1294).

The Motion worked example explains `useMotionValue()` as a factory-realized pure-JS MainThreadObject, its disposal behavior, and why main-thread mutation does not rerender or synchronize back. The sample strictly typechecked against this branch's built artifacts. Published route after merge: https://lynxjs.org/react/main-thread-object.

f0rdream:
Design follow-up: #3467 is stacked directly on this PR.

The Motion-owned `motionValueHandleInitialValues` WeakMap in this branch is needed because declarative Motion must resolve the initial style in the background runtime, while the real MotionValue exists on the main thread. It avoids ReactLynx wire-field duck typing, so this PR remains correct and independently reviewable.

The follow-up asks whether that metadata should instead remain with the generic handle runtime. It adds type-scoped `isHandle` and `getInitialPayload` operations to the MainThreadObject type token, then removes the Motion WeakMap.

Besides reducing adapter bookkeeping, this preserves the correct lifecycle invariant: `useMainThreadObject` retains the original handle and realized MotionValue across rerenders, so initial-style resolution should continue using the original creation payload even if a later render passes another hook argument.

No Motion syntax or animation behavior changes in #3467.

f0rdream:
Scale testing for this Motion migration is tracked in draft PR #3477. Isolated native cases create 500 and 1,000 real `useMotionValue` handles and cover factory realization, capture, hydration, and disposal without handle-related exceptions. The earlier declarative `motion.view` variant exposed a more fundamental ordering race that also reproduces with ordinary `MainThreadRef`; it is now tracked separately in #3476 rather than changed in the object-hardening PR. The Motion adapter remains userspace-owned.

f0rdream:
Two focused follow-ups now update this design:

- #3502 fixes the core lifecycle boundary: `create`/`dispose` are capture-free Main Thread Functions and their dependencies stay out of BTS.
- #3503 applies that model to Motion and folds in the useful part of #3467: type-scoped `isHandle` / `getInitialPayload` replaces the Motion-owned WeakMap.

The public `useMotionValue<T>(): MotionValue<T>` contract is unchanged. In the current full example, the 411,904-byte `motion-dom` module appears only in the three `*__main-thread` chunks and in none of the BTS chunks. #3503 targets this PR's head branch because the active fork cannot push to it directly.

Review order: #3453 → #3502 → #3503 → #3477.

f0rdream:
Superseded by #3509, which contains the current Motion userspace and declarative implementation and depends on the standalone core PR #3477.

PR #3459, opened 2026-08-12
fix(motion): route value-specific transitions

Huxpro:
## Summary

- route each declarative MotionValue through upstream `motion-dom`'s named `animateMotionValue(key, ...)`
- preserve upstream `getValueTransition` semantics for `transition={{ x: {...}, opacity: {...} }}`
- apply the same property-aware path to Lynx-for-Web tap/hover fallbacks
- add a regression proving opacity can finish while x remains delayed

## Stack

Stacked on #3458 (`agent/motion-style-motion-value-types`).

## Validation

- `pnpm --filter @lynx-js/motion test -- declarative.test.tsx`: 15 files, 120 tests passed
- ESLint on changed files: passed
- package build generated JS and declarations; the final local publint packaging hook failed because its temporary tarball was missing (exit 0 from the underlying pack command)

Consumer Web/Lynx evidence is intentionally waiting for this commit's immutable `pkg.pr.new` preview before any conformance metric is promoted.

PR #3460, opened 2026-08-12
fix(motion): restore removed animate values

Huxpro:
## Summary

- track the previous base animate target keys on the main thread
- when a key disappears, animate its owned MotionValue back to the resolved static style/initial value
- keep interaction targets separate so tap/hover ownership is unchanged
- reuse upstream `animateMotionValue` for restoration

## Upstream contract

`packages/framer-motion/src/motion/__tests__/animate-prop.test.tsx` — **when value is removed from animate, animate back to value read from DOM**.

The Lynx adapter already has the resolved static style value, so it restores that value without introducing DOM reads.

## Stack

Stacked on #3459.

## Validation

- package suite: 15 files / 121 tests passed
- local same-stack consumer: removed value + `transition.default` + tap + hover, 20/20
- evidence build: passed
- no native API boundary introduced

Consumer conformance metrics remain unchanged until an immutable preview is available.

### DIFF
--- a/.changeset/motion-removed-animate-values.md
+++ b/.changeset/motion-removed-animate-values.md
@@ -1,0 +1,5 @@
+---
+"@lynx-js/motion": patch
+---
+
+Restore static style values when their declarative animate targets are removed.
--- a/packages/motion/__tests__/declarative.test.tsx
+++ b/packages/motion/__tests__/declarative.test.tsx
@@ -219,6 +219,44 @@ describe('declarative Motion', () => {
     );
   });
 
+  test('restores a static style when its animate value is removed', async () => {
+    function App() {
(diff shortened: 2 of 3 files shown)


Huxpro:
Superseded by the more complete ownership semantics in #3489. #3460 restored removed keys from an earlier `startingValues` snapshot; #3489 additionally matches upstream by using the current `initial` value and retaining the displayed value when a key leaves both `animate` and `initial`. I re-ported the combined atomic layer onto local `main → #3477 → #3509 → conformance` and verified 124/124 Motion tests plus TypeScript. Closing the weaker intermediate implementation avoids reviewing two competing restoration models.

PR #3462, opened 2026-08-12
feat(motion): support declarative transitionEnd

Huxpro:
## Summary
- expose upstream-compatible `transitionEnd` on declarative Motion targets
- keep `transitionEnd` out of animated key/value routing
- apply the latest completion values through existing MotionValues/styleEffect after the active animation generation completes
- prevent superseded animations from applying stale completion values

## Upstream contract
`packages/framer-motion/src/motion/__tests__/animate-prop.test.tsx` — `uses transitionEnd on subsequent renders`

## Evidence
- immutable bd151a1 baseline: Web settles at transitionEnd 100/200; Lynx stops at animate target 20 (3/3 reproduced)
- package suite: 15 files / 123 tests passed
- local stacked consumer: transitionEnd + tap + hover stress 15/15
- local consumer evidence build passed
- package JS and declarations generated; known final publint pack hook still fails because its temp tarball is missing after pack exits 0

This is stacked on #3460. Consumer metrics remain unchanged until an immutable preview package is available.

### DIFF
--- a/.changeset/motion-transition-end.md
+++ b/.changeset/motion-transition-end.md
@@ -1,0 +1,5 @@
+---
+"@lynx-js/motion": patch
+---
+
+Support declarative `transitionEnd` values and apply the latest target after animation completion.
--- a/packages/motion/__tests__/declarative.test.tsx
+++ b/packages/motion/__tests__/declarative.test.tsx
@@ -247,7 +247,7 @@ describe('declarative Motion', () => {
             initial={{ x: 0 }}
             animate={{
               x: step * 20,
-              transition: { duration: 0.01 },
+              transition: { type: false },
               transitionEnd: { x: step * 100 },
             }}
           />
--- a/packages/motion/__tests__/declarative.test.tsx
+++ b/packages/motion/__tests__/declarative.test.tsx
@@ -99,6 +99,24 @@ describe('declarative Motion', () => {
     expect(splitMotionTarget(definition, { duration: 1 })).toEqual({
       target: { opacity: 0.5, x: 80 },
       transition: { duration: 0.2 },
+      transitionEnd: undefined,
+    });
+  });
+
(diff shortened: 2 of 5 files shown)


Huxpro:
Follow-up aligned the regression with upstream `type: false` plus rapid rerenders. This exposed that instant MotionValue starts return a completion promise without an active animation handle. The completion phase now follows the promises returned by `value.start(...)`, while lifecycle start callbacks still require real active animations. Focused declarative test 15/15 and full package suite 123/123 pass.

Huxpro:
Final local validation at `20a4f03d6a`: package 15 files / 123 tests; generated JS/declarations (final publint temp-tarball hook remains the known separate failure); consumer evidence build; transitionEnd + tap + hover stress 15/15. The consumer hypothesis covers two successive completion values (100 then 200), while the package regression covers upstream `type: false` and latest rerender semantics.

Huxpro:
Exact consumer evidence is merged in Huxpro/motion#87 against immutable `2c805a2`: Web and Lynx remain `display:block` at the 120ms intermediate blue→red frame, then settle red and apply `display:none` only after named-variant completion. Focused 1/1; full suite 67/67; evidence build and package regression pass. Priority I4/F5/M1/R1/C1; no native boundary or Full Demo claim; weighted loss remains 7 across 68 tracked contracts.

Huxpro:
Confidence: **VERY CONFIDENT — implementation ready to review.** Re-ported onto the canonical local stack after the stronger removed-target ownership layer. Motion tests: 126/126; TypeScript build: pass. This covers normal and instant declarative `transitionEnd` with generation-based stale-write suppression entirely in Motion userspace. `transitionEnd`-only animate/gesture shapes remain deliberately separate in #3487/#3488.

Huxpro:
Superseded by #3519, restacked on the canonical conformance stack.

PR #3463, opened 2026-08-12
fix(motion): suppress completion after unmount

Huxpro:
## Summary

- suppress declarative `onAnimationComplete` callbacks after the component unmounts or its animation effect is replaced
- invalidate background-thread completion with an effect generation before scheduling main-thread cleanup
- add a dual-thread regression test for active-animation unmount

## Root cause

Animation completion originates on the main thread and is forwarded back to the background thread. Unmount cleanup also crosses the thread boundary, so a completed main-thread promise could enqueue a stale callback after ReactLynx had already removed the component.

## Validation

- upstream `animate-prop.test.tsx` contract reproduced in the consumer harness: baseline Lynx completed after unmount in 3/3 runs while Web suppressed completion
- `pnpm --filter @lynx-js/motion test`: 15 files, 124 tests passed
- focused lint, Biome, dprint, and diff checks passed
- local Web/Lynx consumer hypothesis plus hover regression: 10/10 across 5 repeats
- consumer evidence build passed

Native device verification was attempted because this crosses lifecycle/thread boundaries, but the available Sandbox Explorer runtime fails to register handlers before application code with `ReferenceError: queueMicrotask is not defined`. That environment mismatch is recorded as a native-evidence blocker, not as a Motion pass or platform-conformance blocker.

Consumer metrics remain unchanged until an immutable preview package is available.

### DIFF
--- a/.changeset/motion-unmount-cancellation.md
+++ b/.changeset/motion-unmount-cancellation.md
@@ -1,0 +1,5 @@
+---
+"@lynx-js/motion": patch
+---
+
+Suppress stale declarative animation completion callbacks after a component unmounts.
--- a/packages/motion/__tests__/declarative.test.tsx
+++ b/packages/motion/__tests__/declarative.test.tsx
@@ -276,6 +276,39 @@ describe('declarative Motion', () => {
     );
   });
 
+  test('does not complete an animation after unmount', async () => {
+    const onAnimationComplete = vi.fn();
+    function App() {
+      const [visible, setVisible] = useState(true);
+      return (
+        <view>
+          {visible
+            ? (
+              <motion.view
+                data-testid='box'
+                initial={{ x: 0 }}
+                animate={{ x: 40 }}
+                transition={{ duration: 0.08 }}
+                onAnimationComplete={onAnimationComplete}
+              />
+            )
+            : null}
+          <view data-testid='toggle' bindtap={() => setVisible(false)} />
+        </view>
+      );
+    }
+
+    const { getByTestId } = render(<App />, {
+      enableMainThread: true,
+      enableBackgroundThread: true,
+    });
+    fireEvent.tap(getByTestId('toggle'));
+    await act(async () => {
+      await new Promise(resolve => setTimeout(resolve, 120));
+    });
+    expect(onAnimationComplete).not.toHaveBeenCalled();
+  });
+
   test('restores a static style when its animate value is removed', async () => {
     function App() {
       const [ownsOpacity, setOwnsOpacity] = useState(true);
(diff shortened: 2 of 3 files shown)


Huxpro:
### Convergence record

| Field | Result |
| --- | --- |
| Gap | active animation unmount must suppress `onAnimationComplete` |
| Upstream source | `packages/framer-motion/src/motion/__tests__/animate-prop.test.tsx` — `unmount cancels active animations` |
| Priority | API importance: high; platform fit: high; MTS: low; ReactLynx: medium; CSS: none |
| Compatibility layer | declarative Motion lifecycle across ReactLynx background/main threads |
| Before | Web suppresses completion; Lynx fired completion after unmount in 3/3 baseline runs |
| After | local Web/Lynx hypothesis passes 5/5; package regression passes |
| Package evidence | 15 files / 124 tests |
| Dual-renderer evidence | unmount + hover regression 10/10 over 5 repeats |
| Gallery | atomic conformance card only; no Full Demo because no complete new usage pattern was unlocked |
| Native evidence | attempted; blocked before handler registration by Sandbox Explorer runtime missing global `queueMicrotask` |
| Weighted loss | unchanged at 20 until immutable preview verification allows consumer manifest promotion |

The native environment mismatch is not classified as a Web/Lynx Motion platform gap: the page event handler cannot be installed before the lifecycle probe executes.

Huxpro:
Confidence: **VERY CONFIDENT — implementation ready to review.** Re-ported directly after declarative `transitionEnd` on local `main → #3477 → #3509 → conformance`. Motion tests: 127/127; TypeScript build: pass. The generation invalidation is the same ownership mechanism already used for rerenders and prevents `onAnimationComplete` from escaping the ReactLynx unmount lifecycle; no platform-specific workaround.

Huxpro:
Canonical restack follow-up (not yet on this remote branch): exercising the current #3477 + #3509 foundation exposed a callback-thread binding bug in the declarative completion path. `runOnBackground(completeAnimationOnBackground)` was being created inside the async animation Promise continuation, after the Web/test runtime had restored background context, producing `runOnBackground can only be used on the main thread`.

The stacked fix binds the dispatcher synchronously while `updateMotionStyles` is still executing on MTS, then invokes that pre-bound dispatcher from the Promise continuation. After moving this fix down into the unmount-cancellation layer and cascading the stack rebase: 15 files / 132 tests pass, TypeScript passes, and there are no unhandled completion errors. This strengthens the implementation confidence; it should be included when the canonical stack is submitted.

Huxpro:
Superseded by #3520, which also includes the synchronous MTS binding fix before the async completion continuation.

PR #3465, opened 2026-08-12
test(motion): cover discrete visibility exits

Huxpro:
## Summary

- lock upstream discrete `display` and `visibility` exit contracts into the declarative Motion package suite
- verify each property remains active during an opacity exit and switches to its terminal hidden value only when the animation completes

## Why this is test-only

Both behaviors are already fixed by the property-aware MotionValue animation path introduced in stacked PR #3459. Passing the property name into upstream `motion-dom` selects its existing discrete visibility mixer, so no Lynx-specific animation engine is needed.

The immutable `bd151a1` package still uses the older property-agnostic path: both consumer reproductions fail 5/5 by hiding immediately. The current stack passes both cases.

## Validation

- `pnpm --filter @lynx-js/motion test`: 15 files, 126 tests passed
- local current-stack Web/Lynx visibility case plus hover regression: 10/10 across 5 repeats
- local current-stack Web/Lynx display case plus hover regression: 10/10 across 5 repeats
- consumer evidence builds passed
- no native run: these are CSS value-mixing contracts with no gesture, layout, measurement, lifecycle, or platform thread boundary

Consumer metrics remain unchanged until an immutable preview for the stack verifies these exact package paths.

### DIFF
--- a/.changeset/motion-discrete-display-test.md
+++ b/.changeset/motion-discrete-display-test.md
@@ -1,0 +1,5 @@
+---
+"@lynx-js/motion": patch
+---
+
+Preserve discrete display visibility until declarative exit animations complete.
--- a/packages/motion/__tests__/declarative.test.tsx
+++ b/packages/motion/__tests__/declarative.test.tsx
@@ -231,6 +231,44 @@ describe('declarative Motion', () => {
     expect(getByTestId('box').getAttribute('style')).toContain('display: none');
   });
 
+  test('switches visibility to hidden only after an opacity exit completes', async () => {
+    function App() {
+      const [hidden, setHidden] = useState(false);
+      return (
+        <view>
+          <motion.view
+            data-testid='box'
+            initial={{ visibility: 'visible', opacity: 1 }}
+            animate={hidden
+              ? { visibility: 'hidden', opacity: 0 }
+              : { visibility: 'visible', opacity: 1 }}
+            transition={{ duration: 0.08, ease: 'linear' }}
+          />
+          <view data-testid='toggle' bindtap={() => setHidden(true)} />
+        </view>
+      );
+    }
+
+    const { getByTestId } = render(<App />, {
+      enableMainThread: true,
+      enableBackgroundThread: true,
+    });
+
+    fireEvent.tap(getByTestId('toggle'));
+    await act(async () => {
+      await new Promise(resolve => setTimeout(resolve, 30));
+    });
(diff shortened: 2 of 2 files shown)


Huxpro:
### Convergence record

| Field | Result |
| --- | --- |
| Gap | `display: block → none` must switch only after the opacity exit |
| Upstream source | `packages/framer-motion/src/motion/__tests__/animate-prop.test.tsx` — `animate display block => none switches to none on animation end` |
| Priority | importance 4; platform fit 5; MTS 1; ReactLynx 0; CSS 1 |
| Compatibility layer | upstream `motion-dom` property-aware discrete value mixer |
| Before | immutable `bd151a1` Lynx switches to `none` during the exit; stable 5/5 mismatch |
| After | current stack holds `block` through the exit; focused display + hover 10/10 over 5 repeats |
| Package evidence | 15 files / 125 tests |
| Gallery | atomic conformance card; no Full Demo because it does not independently unlock a complete presence pattern |
| Native evidence | not required; no new gesture, layout, measurement, lifecycle, or platform thread boundary |
| Weighted loss | current 20; projected 19 after immutable preview promotion |

Implementation attribution: stacked PR #3459 already routes the property name into upstream `animateMotionValue`, which selects Motion's existing `display`/`visibility` mixer. This PR adds the permanent regression contract and does not introduce a Lynx-specific mixer.

Huxpro:
### Added visibility-exit contract

- upstream: `animate-prop.test.tsx` — `animate visibility visible => hidden switches to hidden on animation end`
- priority: I4 / F5 / M1 / R0 / C1
- immutable `bd151a1`: stable 5/5 mismatch; Lynx switches to `hidden` during the opacity exit
- current stack: package regression passes; package suite is now 15 files / 126 tests

This uses the same implementation attribution as display exit: #3459 passes the property name into upstream `motion-dom`, selecting its existing invisible-value mixer. No Lynx-specific mixer or platform Issue is added. Consumer promotion remains pending an immutable preview.

Huxpro:
Confidence: **VERY CONFIDENT — ready to review as conformance coverage.** Re-ported onto the canonical local stack; Motion tests rise from 127 to 129 and remain fully green, with TypeScript passing. This PR intentionally adds no new runtime capability: it proves the existing keyframe/`transitionEnd` path already handles discrete `display` and `visibility` exits. Metrics should move only for covered cases, not count this as an additional implementation feature.

Huxpro:
Superseded by #3521, restacked as a coverage-only draft on the canonical conformance stack.

PR #3466, opened 2026-08-12
fix(motion): support CSS custom properties

Huxpro:
## Summary

- preserve `CSSStyleDeclaration.setProperty` on the Lynx element style proxy
- let upstream `motion-dom` animate CSS custom properties without an adapter-side animation path
- accept typed ``--${string}`` keys in declarative `MotionStyle`, matching upstream target types
- add direct runtime and type-contract regressions

## Stack

Temporarily targets `main` only to run the full stack HEAD through CI and pkg.pr.new. After immutable preview validation it will return to #3465 (`agent/motion-discrete-display-exit`) as its review base.

## Validation

- focused `ElementCompt` suite: 22/22 passed
- explicit ESLint: passed
- isolated CSS custom-property typecheck: passed
- consumer baseline: Web reaches `#000`; old Lynx package stays at `#fff` in 5/5 runs
- immutable preview validation pending for `9aff526bec`

## Upstream contract

`motion-dom` routes keys beginning with `--` through `element.style.setProperty(name, value)`. This patch restores that Web API contract in `ElementCompt`; it does not add a second animation engine.

## Summary by CodeRabbit

- **New Features**
  - Added declarative Motion components with animated styles, variants, transitions, gestures, lifecycle callbacks, and MotionValue bindings.
  - Added `motion`, `motion.view`, and `useMotionValue` exports.
  - Added typed main-thread object APIs for stable, validated object handles and lifecycle management.
  - Added a selectable declarative Motion example.

- **Bug Fixes**
  - Improved CSS property handling, visibility during exit animations, animation cleanup, transition completion, and restoration of removed animation targets.

- **Documentation**
  - Expanded Motion usage guidance, supported features, limitations, and main-thread behavior.

### DIFF
--- a/.changeset/motion-css-custom-properties.md
+++ b/.changeset/motion-css-custom-properties.md
@@ -1,0 +1,5 @@
+---
+"@lynx-js/motion": patch
+---
+
+Preserve the `CSSStyleDeclaration.setProperty` contract so declarative Motion can animate CSS custom properties.
--- a/packages/motion/__tests__/declarative.test.tsx
+++ b/packages/motion/__tests__/declarative.test.tsx
@@ -9,6 +9,7 @@ import type { IntrinsicElements } from '@lynx-js/types';
 import { act, fireEvent, render } from '@lynx-js/react/testing-library';
 
 import { motion, useMotionValue } from '../src/index.js';
+import type { MotionStyle, MotionTarget } from '../src/declarative/types.js';
 import { ElementCompt } from '../src/polyfill/element.js';
 import {
   collectMotionValues,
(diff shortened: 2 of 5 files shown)


Huxpro:
Confidence: **VERY CONFIDENT — ready to review.** Re-ported onto the canonical local gh-stack. Motion tests: 131/131; TypeScript build: pass. Lynx API docs explicitly describe CSS variables as fully supported with Web-compatible syntax, cascade, scoping, and fallback. This patch only bridges the standards-shaped `style.setProperty(name, value)` call used by upstream `motion-dom` to Lynx `setStyleProperty`, preserving arbitrary `--*` names; it adds no property allowlist, CSS emulation, or platform-specific fallback.

Huxpro:
Superseded by #3522, restacked on the canonical conformance stack.

PR #3405, opened 2026-08-07
feat(motion): add declarative motion components

f0rdream:
Adds declarative Motion components for Lynx:

- `motion.view`, `motion.text`, `motion.image`, and `motion.create()`
- `initial`, reactive `animate`, and `transition` props
- live `MotionValue` style bindings composed with declarative targets
- declarative examples and documentation

## Summary by CodeRabbit

* **New Features**
  * Added declarative motion components for views, text, images, and custom components.
  * Added spring animations, transitions, initial styles, reactive targets, and live motion-value style bindings.
  * Added the `useMotionValue` hook for creating and updating animated values.
  * Added typed main-thread values with automatic hydration and serialization support.
  * Added a declarative motion demo and expanded usage documentation.

* **Bug Fixes**
  * Improved worklet handling for captured object and member-expression values.

* **Tests**
  * Added comprehensive coverage for declarative animations, motion values, hydration, and worklet transformations.

PR #3467, opened 2026-08-12
feat(react): expose MainThreadObject handle inspection

f0rdream:
This is a focused follow-up to #3455 and the MainThreadObject model in #3446 and #3453.

## Motivation

Declarative Motion needs the original MotionValue payload while rendering ordinary Lynx style in the background runtime. This produces the correct first frame before main-thread animation takes over.

#3455 preserves that value in a Motion-owned WeakMap:

```text
MotionValue handle -> initial style value
```

That works, but it duplicates metadata already owned by the ReactLynx handle runtime. It also creates a lifecycle mismatch: `useMainThreadObject` retains one handle and one realized main-thread object, while a later React render may pass a different hook argument. Updating a library-owned WeakMap on every render can then report a different initial value from the payload that created the stable main-thread object.

## Design

`defineMainThreadObjectType` now returns a type token with two source-runtime inspection operations:

```ts
type.isHandle(value)
type.getInitialPayload(handle)
```

Inspection is scoped to the exact object type. A MotionValue definition can recognize only MotionValue handles; another MainThreadObject type cannot read them.

ReactLynx owns the handle metadata and preserves the creation-time payload. Motion only asks its type token for that payload when resolving the initial declarative style:

```ts
if (motionValueType.isHandle(value)) {
  return motionValueType.getInitialPayload(value)
}
```

## Why this boundary

- ReactLynx owns generic handle identity and creation payload metadata.
- Motion owns style interpretation and declarative animation behavior.
- Library adapters do not depend on `_wvid`, `_initValue`, `_type`, protocol versions, or other wire fields.
- There is no global public inspector that lets one library inspect every MainThreadObject handle.
- Wrong-type and non-handle values are rejected.
- Rerenders cannot overwrite the payload associated with the stable handle.

This removes the Motion-specific `motionValueHandleInitialValues` bookkeeping without adding Motion-specific behavior to ReactLynx.

## API shape

The input definition and returned token are intentionally distinct:

```ts
interface MainThreadObjectTypeDefinition<I, O extends object> {
  type: string
  create(initialValue: I): O
  dispose?(object: O): void
}

interface MainThreadObjectType<I, O extends object>
  extends MainThreadObjectTypeDefinition<I, O> {
  isHandle(value: unknown): value is O
  getInitialPayload(value: O): I
}
```

Library authors provide only creation and disposal behavior. ReactLynx adds the type-scoped inspection methods when the definition is registered.

## Scope

This does not make the realized object available in the background runtime, synchronize main-thread mutations back to React, add RPC behavior, or add new Motion declarative syntax. It only exposes the source handle creation payload needed by adapters such as Motion initial-style resolution.

## Relationship to the existing work

- #2144 introduced the original MainThreadValue mechanism.
- #3405 introduced declarative Motion and demonstrated the initial-style requirement.
- #3446 defines the refined MainThreadObject model.
- #3453 implements the generic MainThreadObject core.
- #3455 migrates Motion to that core while retaining Motion-owned initial metadata.
- This PR removes that final duplicate metadata layer.

This PR is stacked directly on #3455 and contains one additional commit.


f0rdream:
Follow-up audit relative to this PR:

- The realized MotionValue remains the genuine factory-created `motion-dom` object; inspection is source-side and type-scoped.
- #3469 covers compiler capture safety.
- #3471 covers atomic transformation, typed hydration, divergence, and lazy registration.
- #3468 measures both ordinary-ref regression risk and the opt-in MotionValue lifecycle at scale.

The branch remains stacked on #3455. Any future shared-ref lifecycle work should be reconciled separately with these type-aware guarantees.

f0rdream:
Draft PR #3477 is stacked after this PR and preserves commit `0696dd0d` unchanged below the hardening commits. It builds on handle inspection to preserve class captures, validate hydration/divergence, support BTS-only lazy registration, and keep `MainThreadRef` on its fast path. Generic React patch versus direct-worklet ordering is tracked separately in #3476.

f0rdream:
Superseded by the consolidated lifecycle stack:

- #3502 now owns the generic type-scoped handle inspection API and removes lifecycle implementations from the returned type token.
- #3503 preserves this PR's Motion initial-payload migration while also moving the MotionValue factory and disposer to MTS-only Main Thread Functions.

No useful work is being dropped: the inspection commit is preserved in the new stack, and Motion no longer needs its private initial-value WeakMap. Closing this duplicate leaves one review path: #3453 → #3502 → #3503 → #3477.

PR #3474, opened 2026-08-12
fix(motion): restore declarative completion callbacks

Huxpro:
## Summary

- restore direct `onAnimationComplete` dispatch from the main-thread animation Promise
- remove the nested background closure whose captured generation ref is stale on Lynx for Web
- retain the main-thread animation generation gate, so interrupted or unmounted animations remain suppressed
- add a positive mounted-completion regression test alongside the existing unmount test

## Stack

Stacked on #3466 (`agent/motion-css-custom-property`).

## Evidence

Local consumer simulation against the Huxpro/motion dual renderer:

- Gallery + lifecycle + equal-target no-op + equal-keyframe no-op + active-unmount regression: **5/5 passed**
- active unmount reloads each renderer immediately before cancellation, proving the callback is not emitted after cleanup
- package source formatting and lint pre-commit hooks passed
- immutable preview: `@lynx-js/motion@48fc271` (with aligned `@lynx-js/react` and `@lynx-js/react-umd`)

Full immutable-preview parity and native evidence will be added after consumer validation.

### DIFF
--- a/.changeset/motion-declarative-completion-lifecycle.md
+++ b/.changeset/motion-declarative-completion-lifecycle.md
@@ -1,0 +1,6 @@
+---
+"@lynx-js/motion": patch
+---
+
+Restore declarative animation completion callbacks while retaining stale
+callback suppression after an effect cleanup.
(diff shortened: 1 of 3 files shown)


Huxpro:
Superseded by #3509. The canonical Motion userspace implementation already uses the direct background completion-callback path from this PR; the rebuilt local `main → #3477 → #3509` baseline passes the completion coverage within the 118/118 Motion suite. Closing this old stacked regression PR avoids duplicating a fix already carried by the new foundation.

PR #3477, opened 2026-08-12
feat(react): introduce MainThreadObject

f0rdream:
This PR is superseded by the upstream stacked implementation:

- #3788 — MainThreadObject foundation
- #3789 — nested and object-method capture/compiler support
- #3790 — transport and payload hardening

The stack is based on the latest `main` and removes the capability-specific core/full worklet-runtime split. The final stack keeps one normal worklet runtime and preserves MainThreadRef’s specialized path.

Motion integration and declarative syntax remain separate userspace work.


Huxpro:
Follow-up API refinement: #3513 replaces `isHandle + getInitialPayload` with `downcast(value) -> MainThreadObjectHandle | undefined` and exposes `creationPayload` on the successful typed handle. The branch is based directly on this PR's current head (`3181442c7`).

Huxpro:
Confidence update: **very confident / ready to review once the author is comfortable removing draft status**.

The full canonical dependent stack now runs on this exact head. All substantive #3477 CI checks are green; the remaining CodSpeed failure/Relative CI pending status is external benchmark reporting rather than a functional test failure. Local validation of the integrated stack passes the Motion package (15 files, 132 tests) and TypeScript. #3509 now targets an upstream mirror of this exact head, so the runtime layer is independently reviewable.

Huxpro:
Conformance-stack assessment: **very confident — recommended ready for review.**

The generic `MainThreadObject` runtime/compiler primitive is cleanly separated from Motion, and the full repository validation is green across Linux, Windows, Rust, Vitest, type checks, publishing, and Android emulator coverage. The downstream Motion stack is based on this exact head.

I am leaving the PR's author-owned draft state unchanged; this comment records the readiness recommendation rather than overriding the author's review intent.

f0rdream:
The review concern about `defineMainThreadObjectType` not executing on MTS is now isolated as [#3526](https://github.com/lynx-family/lynx-stack/issues/3526), with a stacked implementation in [#3527](https://github.com/lynx-family/lynx-stack/pull/3527).

The follow-up changes the registration guarantee from module/hook execution to type-token liveness across the paired BTS and MTS graphs:

- a type used by either graph retains and registers its `create` and optional `dispose` definitions;
- a type unused by both graphs, including an unused sibling export, remains tree-shakeable;
- empty MTS first screens and BTS-only lazy renders no longer require the defining module to execute on MTS;
- the public type token, handle wire format, hydration/lifecycle model, and `MainThreadRef` path remain unchanged.

#3477 remains the core primitive. #3527 is intentionally separate so this build-time registration decision can be reviewed after the runtime/compiler model.

packages/react/runtime/lazy/internal.js
```diff
@@ -53,6 +53,18 @@ export const {
   wrapWithLynxComponent,
 } = target[sExportsReactInternal];
 
+const captureMainThreadObjectImpl = target[sExportsReactInternal]
+  .captureMainThreadObject;
```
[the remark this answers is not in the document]
f0rdream: Checked against the current module initialization. `target[sExportsReactInternal]` is already destructured earlier in this file, so making only this later read optional cannot recover from an absent symbol. The supported compatibility case is an existing internal export object whose newer `captureMainThreadObject` property is missing; the current function guard diagnoses that case. Resolving without a code change.

packages/react/transform/crates/swc_plugin_worklet/extract_ident.rs
```diff
@@ -404,6 +404,14 @@ impl VisitMut for ExtractingIdentsCollector {
     self.pop_scope();
   }
 
+  fn visit_mut_method_prop(&mut self, n: &mut MethodProp) {
+    self.push_fn_scope(collect_inner_scope_decls(n));
+    self.next_block_decls_collected = true;
+    n.visit_mut_children_with(self);
+    self.next_block_decls_collected = false;
+    self.pop_scope();
```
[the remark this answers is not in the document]
f0rdream: Verified with the current Rust transform tests and generated JS/LEPUS snapshots. Non-computed object-method keys such as `create` and `dispose` are not emitted into `_c`; both methods compile directly to `_wkltId` descriptors. The proposed key insertion is therefore unnecessary for this AST path.

packages/react/runtime/src/snapshot/worklet/ref/mainThreadObject.ts
```diff
@@ -0,0 +1,373 @@
...
+
+      const id = this._wvid;
+      this._lifecycleObserver = lynx.getNativeApp().createJSObjectDestructionObserver?.(() => {
+        lynx.getCoreContext?.().dispatchEvent({
+          type: WorkletEvents.releaseWorkletRef,
+          data: { id },
+        });
+      });
+    }
+  }
```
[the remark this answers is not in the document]
f0rdream: Checked against the Lynx API contract and the existing `MainThreadRef` lifecycle path. `lynx.getNativeApp()` is required in the background runtime, and `MainThreadRef` uses the same unconditional access; only `createJSObjectDestructionObserver` is optional. Making `getNativeApp` optional here would silently disable release tracking, so the established contract is retained.

f0rdream:
Added the focused performance comparison in `30d37dd33`.

The benchmark contains two independent comparisons so ordinary-ref cost is not mixed with MotionValue construction:

1. **Specialized `MainThreadRef` vs a mutable cell implemented through `MainThreadObject`.** Both cases render 1,000 identical components with a captured `current` increment handler. Only the handle transport/lifecycle path changes.
2. **Existing `useMotionValueRef` workaround vs direct `MainThreadObject` MotionValue.** Both cases create 1,000 genuine `motion-dom` MotionValues and capture an increment handler. The workaround creates an empty `MainThreadRef`, then runs a Main Thread Function that constructs the MotionValue and assigns `ref.current`; the direct case realizes the MotionValue from the typed object's initial payload.

Separate one-million-operation cases measure only steady-state `current++` and MotionValue `.set()` access. They contain no subscribers or declarative style effects, so they do not claim to measure native drawing or animation frame performance.

Median of 10 interleaved production headless-Lynx runs:

| Comparison | Specialized / workaround | Generic / direct MTO | Difference |
| --- | ---: | ---: | ---: |
| Ref BTS render, 1,000 components | 32.090 ms | 33.064 ms | +3.0% |
| Ref MTS render | 26.554 ms | 27.000 ms | +1.7% |
| Ref hydration | 14.821 ms | 16.555 ms | +11.7% |
| Ref serialization | 3.197 ms | 4.612 ms | +1.415 ms |
| Ref 1M writes | 38.795 ms | 38.829 ms | +0.09% |
| Motion BTS render, 1,000 values | 70.833 ms | 33.489 ms | -52.7% |
| Motion MTS render | 38.495 ms | 26.983 ms | -29.9% |
| Motion hydration | 15.415 ms | 16.574 ms | +7.5% |
| Motion serialization | 3.314 ms | 3.885 ms | +0.571 ms |
| Motion 1M `.set()` calls | 481.698 ms | 468.274 ms | -2.8% |

Interpretation: generic typed-object metadata has a measurable creation/transport cost, but no meaningful mutable-cell access penalty after realization. The Motion improvement comes mainly from replacing per-hook Main Thread Function initialization with direct typed realization; both paths operate on the same real MotionValue afterward. This supports keeping the current specialized `MainThreadRef` fast path while using `MainThreadObject` for library-owned objects.

packages/react/transform/crates/swc_plugin_worklet/gen_stmt.rs
```diff
@@ -193,6 +227,121 @@ impl StmtGen {
...
+        continue;
+      };
+      let Prop::KeyValue(key_value) = prop.as_mut() else {
+        continue;
+      };
+      if !key_value.value.is_object() {
+        continue;
+      }
+      let source = match capture_root {
+        CaptureRoot::Ident => StmtGen::find_root_ident(&key_value.value).map(Expr::Ident),
```
Yradex, line 250: [P1] Preserve nested MainThreadObject handles during capture narrowing

The transform only probes the outer capture root. For `props.value.get()`, it emits `captureMainThreadObject(props) ?? { value: { get: props.value.get } }`. If `props.value` is the handle, `props` does not match the runtime brand, so the fallback reads `get` from the background handle: development throws while building the descriptor, and production captures `undefined`. Recurse through the narrowed capture tree and probe the exact member prefix that may contain a handle; cover nested lexical and `this.props.value` captures.

Yradex (commented):
Four blocking findings on the current head. These are contract and lifecycle correctness issues that should be addressed before merge.

packages/react/runtime/src/snapshot/worklet/ref/mainThreadObject.ts
```diff
@@ -0,0 +1,373 @@
...
+  protected _lifecycleObserver?: unknown;
+
+  /** @internal */
+  protected constructor(initialValue: I, type: string) {
+    if (__DEV__ && __JS__) {
+      assertSerializableMainThreadObjectPayload(initialValue, type);
+    }
+
+    this._wvid = allocateWorkletValueId();
+    this._initValue = initialValue;
```
Yradex, line 99: [P1] Snapshot the creation payload at handle construction

This stores the caller-owned `initialValue` by reference; `workletRefPool` serializes it later, and the `payload` getter returns the same object. Mutating the input before commit changes the factory input, while mutating `downcast(...).payload` after commit makes it diverge from the payload that created the target. This contradicts the PR's immutable creation-payload contract. Store a framework-owned JSON snapshot and expose a readonly immutable payload view.

packages/react/runtime/src/snapshot/worklet/ref/mainThreadObject.ts
```diff
@@ -0,0 +1,373 @@
...
+    || (typeof value === 'object' && value !== null
+      && typeof (value as Partial<Worklet>)._wkltId === 'string');
+}
+
+function assertCaptureFreeLifecycleFunction(
+  type: string,
+  name: 'create' | 'dispose',
+  value: ((...args: never[]) => unknown) | Worklet,
+): void {
+  if (typeof value === 'function' || value._c === undefined) {
```
Yradex: [P1] Validate every lifecycle capture channel

This returns early whenever `_c` is absent, but a Worklet descriptor may also carry `_jsFn` and arbitrary top-level fields produced by `this` captures. A lifecycle that captures a local `runOnBackground` callback is emitted with `_jsFn`, and `this.helper` is emitted as a top-level `helper` field; both pass this capture-free check. Reject capture-bearing own properties beyond the allowed protocol fields and add transform coverage for JS-function and `this` captures.

packages/react/runtime/src/worklet-runtime/workletRef.ts
```diff
@@ -75,9 +224,82 @@ const getFromWorkletRefMap = <T>(
 };
 
 function removeValueFromWorkletRefMap(id: WorkletRefId): void {
+  disposeMainThreadObject(impl!._workletRefMap[id]);
```
Yradex: [P1] Always remove released entries when a disposer throws

`disposeMainThreadObject` deletes the target metadata before invoking the user disposer. If that disposer throws, the following map deletion never runs, leaving a page-lifetime strong reference with no metadata and no automatic retry. Delete the entry in `finally` while preserving the disposer error. Apply the same exception-safe ordering to hydration completion so one disposer failure cannot skip later framework cleanup.

Yradex:
### Suggestion: keep deterministic disposal out of the v1 public contract

Would it be safer to remove `dispose` from the public `MainThreadObjectTypeDefinition` for v1, rather than expose it with a stronger-looking lifecycle contract than the runtime can currently provide?

Today, normal target release starts only when the destruction observer attached to the BTS handle fires. React unmount does not directly release the target. On Web this observer is backed by `FinalizationRegistry`, and the React runtime API also treats the observer capability as optional, so there is no prompt—or even universal eventual—cleanup guarantee.

The same callback is also used for three different events:

1. the BTS handle is finalized;
2. a redundant target is discarded during hydration;
3. an unadopted first-screen target is cleaned up.

The latter two are deterministic physical-instance discard points. The first is GC-driven and is not a reliable signal that the logical owner has ended. Calling all three `dispose` makes it easy for library authors to depend on it for correctness-critical cleanup such as stopping animations, removing subscriptions/listeners, or releasing native resources, even though the timing cannot be guaranteed.

A narrower v1 contract could be:

- expose only `type` and `create` publicly;
- document that v1 targets must be GC-safe JavaScript objects and may live until page teardown;
- keep deterministic hydration/discard cleanup internal, with a name that reflects that narrower purpose;
- let adapters such as Motion own explicit stop/cancel behavior through a deterministic scope or unmount protocol;
- reconsider public disposal only after owner, release timing, handle escape, exactly-once, and use-after-release semantics are defined.

This would not solve retention by itself, but it would keep the public API honest and reversible while preserving the identity, capture, and hydration parts of MainThreadObject. It also resolves the open disposal question in #3446 conservatively. What do you think?

packages/react/transform/crates/swc_plugin_worklet/gen_stmt.rs
f0rdream: Fixed in 0430e2b1d. Capture narrowing now recursively probes every object-valued prefix with its exact source expression, including nested lexical captures and `this.props.value`. Rust JS/LEPUS snapshots cover the emitted descriptors, and the testing-library integration executes both functional and class-component nested-prop handlers against a guarded handle.

packages/react/runtime/src/snapshot/worklet/ref/mainThreadObject.ts
f0rdream: Fixed in 0430e2b1d. Lifecycle validation now inspects every own descriptor field, permits only the capture-free protocol fields, and rejects `_jsFn`, non-empty or invalid `_c`, and flattened `this` capture fields. Dedicated JS/LEPUS transform fixtures expose both capture channels, with runtime rejection coverage.

packages/react/runtime/src/worklet-runtime/workletRef.ts
f0rdream: Fixed in 0430e2b1d. Release deletion now runs in `finally`, and hydration always installs the compatible first-screen value and completes first-screen bookkeeping even if disposal of the replaced value throws; the original disposer error is preserved. Tests verify no stale strong reference, no duplicate disposal, and usable hydration state after the exception.

packages/react/runtime/src/snapshot/worklet/ref/mainThreadObject.ts
f0rdream: Fixed in 0430e2b1d. Handle construction now creates a framework-owned deep-frozen JSON payload snapshot; the same snapshot is queued for realization, and `payload` is exposed as `Readonly<I>`. The regression test mutates both the caller-owned input and the returned nested payload and verifies that serialization and factory input remain stable.

packages/react/runtime/src/worklet-runtime/hydrate.ts
```diff
@@ -60,22 +61,22 @@ function hydrateCtxImpl(
...
+  handle: WorkletRefImpl<unknown>,
+  value: object,
 ) {
-  if ('_initValue' in value) {
+  if (!isHydratedWorkletValue(value)) {
     // The ref has not been accessed yet.
     return;
   }
-  lynxWorkletImpl!._refImpl._workletRefMap[refId] = value;
+  hydrateWorkletValue(handle, value as WorkletRef<unknown>);
```
Huxpro, line 110: The hydration walk can interpret a MainThreadObject descriptor's `_initValue` user payload as worklet metadata and clobber `_workletRefMap` entries. `transformWorkletInner` got protection for this in this PR (the payload "remains untouched"), but `hydrateCtxImpl` recurses into every non-`_wvid`/`_jsFn` key — including `_initValue` — on both the ctx and first-screen sides.

Concretely: `useMainThreadObject(type, { _wvid: 3, current: 0 })` is a valid JSON payload that passes `snapshotMainThreadObjectPayload`. If its handle is captured by a worklet that isn't invoked on first screen, hydration reaches the descriptor's `_initValue` key, recurses into the payload, `isMutableCell(payload)` is true (own `current` + numeric `_wvid`), `assertCompatibleWorkletValue` passes (no `_type` → mutable-cell expected), and `hydrateWorkletValue` assigns the payload into `impl._workletRefMap[3]`, silently corrupting whatever unrelated `MainThreadRef` has id 3.

Suggestion: skip recursion into `_initValue` when the enclosing object is a MainThreadObject descriptor (mirroring the `transformWorkletInner` guard), so payload data can never be treated as ref handles during hydration.

---

packages/react/runtime/src/snapshot/worklet/ref/mainThreadObject.ts
```diff
@@ -0,0 +1,407 @@
...
+  const guardedHandle = new Proxy(handle, {
+    get(target, property, receiver): unknown {
+      if (property in target) {
+        return Reflect.get(target, property, receiver) as unknown;
+      }
+      throw new Error(
+        `MainThreadObject handle for "${type}" cannot access "${
+          String(property)
+        }" in the background runtime. Use the object only inside a main-thread function.`,
+      );
```
Huxpro, line 315: This dev-only `get` trap throws for *every* property not present on the handle, which turns legitimate engine/framework probes into false-positive crashes in development:

- `await handle` or `Promise.resolve(handle)` — the spec reads `.then` on any object being resolved, so returning a handle from an async function crashes.
- React dev-mode reads `$$typeof` when a handle is held in props/state.
- `console.log(handle)` under Node/Vitest reads `Symbol.toStringTag` / `nodejs.util.inspect.custom` via `util.inspect`.

None of these actually misuse the object, and production (plain object) behaves fine — so this only breaks dev/test runs, where the guard is active.

Suggestion: return `undefined` (instead of throwing) for well-known probe keys — at minimum `then`, `$$typeof`, and all `symbol`-typed properties — and keep the throw for everything else, which preserves the diagnostic for real "called a method on the background handle" mistakes.

---

packages/react/runtime/lazy/react.js
```diff
@@ -50,4 +50,31 @@ export const {
...
+
+function assertMainThreadObjectRuntimeExport(value, name) {
+  if (typeof value !== 'function') {
+    throw new Error(
+      `This lazy bundle requires ReactLynx runtime export ${name} for MainThreadObject. Upgrade the main template runtime or rebuild the lazy bundle with a compatible @lynx-js/react version.`,
+    );
+  }
+}
+
 export default target[sExportsReact]['default'];
```
Huxpro: Nit: these `defineMainThreadObjectType`/`useMainThreadObject` wrappers plus `assertMainThreadObjectRuntimeExport` (~25 lines) are copy-pasted verbatim between `lazy/react.js` and `lazy/compat.js`, with a third near-identical inline variant in `lazy/internal.js`. A future change to the guard or its error message (e.g. a new export or version hint) that touches only one file will leave the others stale — a small shared helper module imported by all three lazy entry files would do the same job once.

---

Huxpro (commented):
Code review of this PR — three inline comments below, plus one transform bug with a ready-made fix.

**Transform bug (fix available):** object-method worklets that reference `this.x` emit the extracted capture spread `...{ x: this.x }` inside the enclosing object literal, where `this` is the outer lexical scope — `undefined` at ESM module top level — not the object being defined. For the exact shape this PR promotes (`defineMainThreadObjectType({ helper: 1, create() { "main thread"; return this.helper; } })`), importing the module throws `TypeError: Cannot read properties of undefined` before the friendly capture-free validation in `defineMainThreadObjectType` can run; in non-module scopes it silently captures the wrong object. The new snapshot `should_expose_object_method_capture_channels_js.js` shows the bad output.

Fix: emit the transformed prop as a getter when `this`-props were extracted, so `this` stays bound to the receiver on property read — matching both the original method-call semantics (`obj.create()` has `this === obj`) and the existing class-member handling. I couldn't push to this PR's fork branch directly, so the commit is on a branch in this repo, ready to cherry-pick or merge:

- Commit: lynx-family/lynx-stack@252a9f8f (`fix(react): bind object-method worklet this-captures to the receiver`, based on this PR's head `0430e2b`)
- Branch: [`claude/review-pr-3477-t9bpl3`](https://github.com/lynx-family/lynx-stack/tree/claude/review-pr-3477-t9bpl3)

`cargo test -p swc_plugin_worklet` passes (64/64) with the updated snapshots; fmt and clippy are clean.

---

packages/react/runtime/src/worklet-runtime/hydrate.ts
f0rdream: Valid and fixed in 2cc595ca5. Hydration now treats every worklet-value descriptor with a numeric `_wvid` as atomic and returns immediately after handling it, so `_initValue` is never traversed as metadata. The regression uses a valid `{ _wvid, current }` payload and verifies that an unrelated `MainThreadRef` map entry cannot be overwritten.

packages/react/runtime/src/snapshot/worklet/ref/mainThreadObject.ts
f0rdream: Valid and fixed in 2cc595ca5. The development guard now returns `undefined` for `then`, `$$typeof`, and symbol keys while retaining errors for missing application string properties. Coverage includes direct probes and `Promise.resolve(handle)`.

packages/react/runtime/lazy/react.js
f0rdream: The duplication is real, but this is a maintainability refactor rather than a correctness issue. I am leaving the three lazy entry wrappers explicit in this PR because their export surfaces and compatibility loading paths differ, and sharing them would expand the lazy-bundle change boundary without changing behavior.

f0rdream:
Follow-up review changes are now included:

- Huxpro’s object-method receiver fix from `252a9f8f` is included as `f10dfdc8d`, preserving `xuan.huang` as the author.
- `2cc595ca5` makes worklet-value descriptors atomic during hydration, permits harmless development-proxy probes, and uses realized-object metadata before structural mutable-cell checks.
- The disposal suggestion is accepted: v1 no longer exposes `dispose` on `MainThreadObjectTypeDefinition`. Handle release can be GC-driven, so the public contract now exposes only `type` and `create`; targets must remain safe until runtime release or page teardown. The existing runtime-only callback has been renamed internally as physical-instance discard behavior and is not reachable from the public definition.

packages/react/transform/crates/swc_plugin_worklet/tests/__swc_snapshots__/lib.rs/should_extract_member_expr_2_lepus.js
```diff
@@ -10,23 +10,23 @@ let onTapLepus = {
...
-        mmmm: {
-            nnnn: {
-                'oooo': this.mmmm.nnnn['oooo']
-            }
-        }
+        mmmm: ((__mainThreadObjectSource)=>captureMainThreadObject(__mainThreadObjectSource) ?? {
+                nnnn: ((__mainThreadObjectSource)=>captureMainThreadObject(__mainThreadObjectSource) ?? {
+                        'oooo': __mainThreadObjectSource['oooo']
+                    })(__mainThreadObjectSource.nnnn)
+            })(this.mmmm)
```
Yradex, line 29: Will this have bundle size & performance issues?

f0rdream: Yes, there is a measurable cost, and the earlier MainThreadRef/MainThreadObject comparison did not include this recursive-prefix change. I ran a focused offline comparison between `b453f4670` and current `2cc595ca5`.

Brief methodology: generate JS-target worklets with depths 1/3/5/10 and 1/10/100 handlers; compare raw/minified/gzip output, warmed transform time, and execution time of the compiled capture-descriptor initializer. The full run was repeated three times. I also tested handles at root, middle, and deep prefixes.

The worst synthetic case (depth 10, 100 handlers, all ordinary objects) was:

- minified output: 18,271 -> 32,301 B (+14,030 B); minified+gzip: 2,004 -> 2,167 B (+163 B);
- transform: 5.565 -> 16.905 ms;
- descriptor construction: 3.136 -> 27.421 us total for 100 descriptors;
- prefix lookups: 100 -> 1,100.

A handle at the root short-circuits: 4.482 -> 4.521 us for 100 descriptors, effectively the previous cost. Costs scale with object-valued prefixes and occur during descriptor construction. The generated registered handler body is byte-identical, so this does not add steady-state handler execution overhead.

Full methodology, matrix, limitations, and handle-placement results: https://github.com/lynx-family/lynx-stack/issues/3446#issuecomment-5354425789

f0rdream: @Yradex There is measurable cost on bundle size and rendering perfoamcne. 

For rendering performance, there are some test done here: https://github.com/lynx-family/lynx-stack/issues/3446#issuecomment-5329420215. For rendering performance does not meaningfully slow the rendering despite a `captureMainThreadObject ` wrapper.

For bundle size, the output is larger for about 14B for each handler in the worst case, which after gzip is acceptable. And this problem is addressable, I will later add a new PR to minimize the string footprint, like what we did before, to change `__workletClosure` to `_c`, that should be a separate PR

f0rdream:
## MainThreadObject base-to-head performance comparison (refreshed 2026-09-10)

This supersedes the earlier report, which used pre-stack revisions. The current comparison uses the merge-base of the complete stack and current `main`:

- Base: [`2150f254528a5a575a6e3148b47ef422386f444e`](https://github.com/lynx-family/lynx-stack/commit/2150f254528a5a575a6e3148b47ef422386f444e)
- Head: [`a3b40c183c78aeeb30e1783bb1977dada639ae18`](https://github.com/lynx-family/lynx-stack/commit/a3b40c183c78aeeb30e1783bb1977dada639ae18)

Both revisions were built in production mode through the repository's Turbo build. Raw sizes are filesystem bytes; gzip values use deterministic `gzip -n -9`; delta is `head - base`.

### `benchmark/react` production bundle size

Case `020` was applied to the base fixture as well, so it has a true base-to-head comparison.

| Case | Base raw | Head raw | Raw delta | Base gzip | Head gzip | Gzip delta |
| --- | ---: | ---: | ---: | ---: | ---: | ---: |
| `001-fib` | 340,389 | 340,930 | +541 | 99,411 | 99,496 | +85 |
| `002-hello-reactLynx` | 264,505 | 265,046 | +541 | 71,178 | 71,257 | +79 |
| `003-hello-list` | 403,223 | 407,315 | +4,092 | 96,253 | 97,002 | +749 |
| `004-various-update` | 298,350 | 305,648 | +7,298 | 81,694 | 84,835 | +3,141 |
| `005-load-script` | 466,727 | 479,044 | +12,317 | 113,803 | 116,783 | +2,980 |
| `006-static-raw-text` | 263,538 | 264,079 | +541 | 70,918 | 71,010 | +92 |
| `007-four-layer-views` | 269,181 | 269,722 | +541 | 71,762 | 71,849 | +87 |
| `008-many-use-state` | 269,788 | 270,329 | +541 | 72,334 | 72,404 | +70 |
| `009-eval-bench` | 260,689 | 261,230 | +541 | 68,347 | 68,424 | +77 |
| `010-transform-exponentiation-operator` | 260,718 | 261,259 | +541 | 68,347 | 68,424 | +77 |
| `011-transform-async-to-generator` | 260,782 | 261,323 | +541 | 68,365 | 68,443 | +78 |
| `012-attrs-compile` | 265,344 | 265,885 | +541 | 71,329 | 71,412 | +83 |
| `013-attrs-runtime` | 265,744 | 266,285 | +541 | 71,481 | 71,559 | +78 |
| `014-attrs-mixed` | 265,901 | 266,442 | +541 | 71,520 | 71,597 | +77 |
| `015-attrs-component` | 267,713 | 268,254 | +541 | 71,720 | 71,795 | +75 |
| `016-use-state-local-attribute-update` | 273,788 | 274,329 | +541 | 71,074 | 71,157 | +83 |
| `017-use-signal-local-attribute-update` | 298,675 | 299,216 | +541 | 75,967 | 76,073 | +106 |
| `018-use-state-full-attribute-update` | 274,311 | 274,852 | +541 | 71,128 | 71,211 | +83 |
| `019-use-signal-full-attribute-update` | 299,198 | 299,739 | +541 | 76,020 | 76,128 | +108 |
| `020-many-main-thread-refs` | 289,642 | 296,934 | +7,292 | 81,897 | 85,026 | +3,129 |
| **Total** | **5,858,206** | **5,897,861** | **+39,655** | **1,544,548** | **1,555,885** | **+11,337** |

The ordinary entries add 70–108 B gzip. The larger deltas are `003` (+749 B), `004` (+3,141 B), `005` (+2,980 B), and the new MainThreadRef workload `020` (+3,129 B).

### Real lynx-ui production bundle size

Fixture: [`lynx-family/lynx-ui@65373a4b`](https://github.com/lynx-family/lynx-ui/commit/65373a4b45e8902cc3fa76f6552ef12d55127f06). Plain compares base and head runtimes with identical source. API adoption compares the head MotionValue workaround with the head MainThreadObject implementation.

| Comparison | Base/from raw | Head/to raw | Raw delta | Base/from gzip | Head/to gzip | Gzip delta |
| --- | ---: | ---: | ---: | ---: | ---: | ---: |
| TabGroup plain | 212,651 | 217,155 | +4,504 | 86,682 | 88,855 | +2,173 |
| TabGroup head workaround -> head object | 303,913 | 308,227 | +4,314 | 127,214 | 128,905 | +1,691 |
| Swiper plain | 184,551 | 188,962 | +4,411 | 73,981 | 76,137 | +2,156 |
| Swiper head workaround -> head object | 278,780 | 283,083 | +4,303 | 117,483 | 119,195 | +1,712 |

### CI repetition model

The checked workflow runs the complete benchmark suite sequentially once in CodSpeed simulation, once in CodSpeed walltime, and once for the separate Perfetto collection. The repository has 20 bench:* page cases, so one CI job launches each page case once in each CodSpeed pass (20 simulation launches + 20 walltime launches) and once for Perfetto (20 ptraces). The successful CI log shows each benchx_cli command once in each pass; there is no fixed page-launch round count configured.

For CodSpeed-instrumented Node benchmarks, walltime can add warmup and dynamically chosen measurement rounds based on timing bounds. That is separate from the benchx_cli page launches above. The workflow pins CodSpeed runner v4.17.5 and sets no min/max round options. The current PR benchmark check fails in simulation before producing a walltime result.

This is consistent with CodSpeed documentation: CPU simulation runs once, while walltime warms up and chooses measurement rounds from timing bounds ([instrument overview](https://codspeed.io/docs/instruments), [CLI/instrument options](https://codspeed.io/docs/cli)).

### Rendering performance on a physical Android device

I rebuilt the base and head production bundles and ran representative cases 004, 008, and 020 on the same physical Android device (PJD110, serial 3635650d). The corrected run uses 48 paired cycles per case, base then head, with a fresh Perfetto trace for every launch: 288 bundle launches total.

Values are milliseconds. Base/head are sample medians. The delta is the median of the 48 paired head-minus-base differences; its interval is a deterministic bootstrap 95% interval over those paired differences. Because this is paired, the delta need not equal the subtraction of the two displayed medians.

| Case | Metric | Base median | Head median | Paired delta (95% CI) |
| --- | --- | ---: | ---: | ---: |
| 004-various-update | ReactLynx::renderBackground | 0.207768 | 0.209203 | +0.000779 ms (+0.37%; -0.003403, +0.007524) |
|  | ReactLynx::renderMainThread | 0.031714 | 0.030668 | -0.002029 ms (-6.40%; -0.003731, -0.000779) |
|  | ReactLynx::hydrate | 0.174229 | 0.176341 | +0.000328 ms (+0.19%; -0.002419, +0.002911) |
|  | ReactLynx::BSI::hydrate | 0.142168 | 0.143111 | +0.000574 ms (+0.40%; -0.000738, +0.003280) |
| 008-many-use-state | ReactLynx::renderBackground | 20.999380 | 20.794934 | +0.008631 ms (+0.04%; -0.187247, +0.257111) |
|  | ReactLynx::renderMainThread | 12.422262 | 12.194815 | -0.179765 ms (-1.45%; -0.279046, -0.008118) |
|  | ReactLynx::hydrate | 0.105452 | 0.105021 | +0.000964 ms (+0.91%; -0.001702, +0.004305) |
|  | ReactLynx::BSI::hydrate | 0.053854 | 0.054366 | +0.000389 ms (+0.72%; -0.000779, +0.001968) |
| 020-many-main-thread-refs | ReactLynx::renderBackground | 35.213301 | 35.196163 | -0.065703 ms (-0.19%; -0.483657, +0.400734) |
|  | ReactLynx::renderMainThread | 27.029455 | 26.781754 | -0.200736 ms (-0.74%; -0.743330, +0.233208) |
|  | ReactLynx::hydrate | 23.402636 | 23.441217 | -0.205041 ms (-0.88%; -0.468159, -0.032718) |
|  | ReactLynx::BSI::hydrate | 21.197451 | 21.253006 | -0.162668 ms (-0.77%; -0.298111, +0.014801) |

The larger differences from the earlier short run do not survive as a general regression signal: most intervals include zero, while the non-zero paired results are small and mixed (renderMainThread improves in 004/008; hydration improves in paired 020). The base/head median subtraction can point the other way for 020 hydration because the comparison is intentionally paired.

### Lynx-ui real-case rendering

The lynx-ui production bundles built at both revisions, so the bundle-size table above remains valid. I am not presenting a lynx-ui rendering time: the available benchx_cli runtime throws `TypeError: cannot read property current of undefined` on both base and head for this fixture, and the physical com.lynx.explorer installation is currently unusable (the installed APK had an activity/dex mismatch; the clean reinstall is blocked while Android reports the user as RUNNING_LOCKED). Those are harness/device setup failures, not performance measurements. A valid real-case result requires an unlocked device with the official LynxExplorer app installed, then the same paired-cycle protocol.

### Conclusion

The stack still has the measured bundle-size cost documented above, but the stronger 48-cycle physical-device benchmark does not establish a broad rendering slowdown. The lynx-ui real-device comparison remains intentionally unclaimed until the LynxExplorer path is repaired; the previous CLI exceptions must not be turned into timings.

packages/react/transform/crates/swc_plugin_worklet/tests/__swc_snapshots__/lib.rs/should_extract_member_expr_2_lepus.js
HuJean: Can we perf in a real case?

packages/react/runtime/__test__/snapshot/lazy.test.js
```diff
@@ -136,6 +141,129 @@ describe('Lazy Exports', () => {
...
+    const definition = {
+      type: '@test/lazy-compatible',
+      create: value => ({ value }),
+    };
+
+    const objectType = ReactMainThreadObjectExports.defineMainThreadObjectType(definition);
+    expect(objectType).toMatchObject({ type: definition.type });
+    expect(objectType).not.toHaveProperty('create');
+    expect(objectType.isHandle).toBeTypeOf('function');
+    expect(objectType.getInitialPayload).toBeTypeOf('function');
```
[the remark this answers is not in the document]
f0rdream: Valid with respect to the intended v1 contract. The old assertions happened to match the stale implementation at the prior head, but both were wrong relative to the agreed API. The test now asserts downcast; I also fixed the separate CI failure here by removing the stale fourth undefined registration argument in c38a6175e.

packages/react/runtime/src/worklet-runtime/workletRef.ts
```diff
@@ -94,19 +337,58 @@ function updateWorkletRef(
...
+            existing,
+            'initialization patch',
+          );
+        }
+      } else {
+        impl!._workletRefMap[id] = createWorkletValue(handle);
       }
     });
   });
 }
```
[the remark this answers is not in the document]
f0rdream: Valid. A failing entry could prevent later unrelated refs in the already-consumed patch from being initialized. The full MTO runtime now processes every entry, records the first error, and rethrows after the batch; a regression test covers a bad MTO followed by a normal MainThreadRef. The core build retains its small no-MTO fast path. Fixed in c38a6175e.

packages/webpack/react-webpack-plugin/test/worklet-runtime.test.ts
```diff
@@ -173,9 +195,14 @@ describe('worklet-runtime bundler guardrails', () => {
         expect(
           countOccurrences(
             lepusChunk['worklet-runtime'],
-            'globalThis.lynxWorkletImpl = {',
+            'globalThis.lynxWorkletImpl',
+          ),
+        ).toBeGreaterThanOrEqual(expectedInitSignatureCount);
```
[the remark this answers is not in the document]
f0rdream: Valid. The guardrail now counts only assignment-shaped lynxWorkletImpl initializations and asserts the exact expected count. Fixed in c38a6175e.

packages/react/etc/react.api.md
```diff
@@ -137,6 +140,20 @@ export interface Lynx {
...
+    readonly isHandle: (value: unknown) => value is O;
+    readonly type: string;
+}
+
+// @public
+export interface MainThreadObjectTypeDefinition<I, O extends object> {
+    readonly create: (initialValue: I) => O;
+    readonly dispose?: (object: O) => void;
+    readonly type: string;
+}
```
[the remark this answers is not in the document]
f0rdream: Valid. The runtime-split rebase lost the intended v1 downcast contract and left the old inspection API in the generated report. Restored MainThreadObjectHandle, exact-type downcast, readonly creationPayload, exports, tests, changeset text, and API report in c38a6175e. Public disposal remains absent.

packages/react/transform/crates/swc_plugin_worklet/lib.rs
```diff
@@ -75,6 +75,54 @@ impl Default for WorkletVisitor {
...
+        .clone()
+        .ident()
+        .unwrap_or(Ident::dummy().into())
+        .into(),
+      method.function.clone(),
+      &mut collector,
+      false,
+      &mut self.named_imports,
+      self.worklet_runtime_loaded_ident.clone(),
+    );
```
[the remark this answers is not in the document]
f0rdream: Valid. Object-method keys are not lexical bindings. The transform now passes a dummy identifier, with a regression fixture where method create calls an imported outer create; both JS and Lepus snapshots were updated. Fixed in c38a6175e.

packages/react/runtime/src/snapshot/worklet/ref/mainThreadObject.ts
```diff
@@ -0,0 +1,370 @@
...
+ * @public
+ */
+export interface MainThreadObjectType<I, O extends object> {
+  /** A globally unique and stable type key. */
+  readonly type: string;
+  /** Return whether a value is a source-runtime handle of this exact type. */
+  readonly isHandle: (value: unknown) => value is O;
+  /** Read this type's initialization payload from one of its source handles. */
+  readonly getInitialPayload: (value: O) => I;
+}
```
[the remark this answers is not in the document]
f0rdream: Valid. The downcast rewrite existed before the runtime-split reconstruction but was accidentally omitted from the rebuilt PR head. It is restored in c38a6175e with an exported public handle type and readonly creationPayload; isHandle and getInitialPayload are removed, exact type identity is enforced, and no public disposer is exposed.

packages/react/runtime/src/snapshot/worklet/ref/mainThreadObject.ts
```diff
@@ -0,0 +1,387 @@
...
+    this._wvid = allocateMainThreadRefId();
+    this._initValue = initialValue;
+    this._type = type;
+    this._mtoVersion = MAIN_THREAD_OBJECT_PROTOCOL_VERSION;
+    registerMainThreadObjectHandle(this, type, initialValue);
+
+    if (__JS__) {
+      addMainThreadRefInitValue(
+        this._wvid,
+        initialValue,
```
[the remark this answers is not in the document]
f0rdream: There's a similar problem for MainThreadRef, maybe we could leave it now. Because fixing this would requiring freeze the object recursively. For now, we have added a ReadOnlyType at types to prevent the problem.

```
const options = { nested: { value: 1 } }
useMainThreadRef(options)

options.nested.value = 2  // If changed before JSON.stringify(patch), MTS receives 2.
```

packages/react/runtime/src/snapshot/worklet/ref/mainThreadObject.ts
```diff
@@ -0,0 +1,387 @@
...
+
+  const prototype = Object.getPrototypeOf(value) as object | null;
+  if (!Array.isArray(value) && prototype !== Object.prototype && prototype !== null) {
+    return path;
+  }
+
+  ancestors.add(value);
+  const entries = Array.isArray(value)
+    ? value.map((item, index) => [String(index), item] as const)
+    : Object.entries(value);
```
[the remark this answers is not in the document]
f0rdream: Fixed in 56c7b71bb. Array validation now compacts actual entries while retaining their original indices, so sparse holes are skipped without accepting an explicitly stored `undefined`. Added coverage for both cases.

Yradex (commented):
Follow-up review of 56c7b71b: seven findings from the full implementation review, with reproducers and suggested fixes inline. The evidence includes current-source compiler, resolver/module-graph and runtime probes; this does not claim a complete native end-to-end validation. Previously discussed payload semantics and explicitly deferred static-registration/lifecycle work are not repeated here.

packages/react/transform/crates/swc_plugin_worklet/extract_ident.rs
```diff
@@ -404,6 +404,14 @@ impl VisitMut for ExtractingIdentsCollector {
     self.pop_scope();
   }
 
+  fn visit_mut_method_prop(&mut self, n: &mut MethodProp) {
+    self.push_fn_scope(collect_inner_scope_decls(n));
```
Yradex, line 408: [P1] Collect object-method locals before extracting captures

`collect_inner_scope_decls(n)` now receives a `MethodProp`, but `BindingCollector` has no corresponding method visitor to enter its function body. Its first block is therefore treated as a nested block and only `var` declarations are collected; the subsequent `next_block_decls_collected = true` prevents another collection pass.

This affects an ordinary capture-free factory:

```js
create(initial) {
  'main thread';
  const state = { value: initial };
  return state;
}
```

A collector/generator probe using this commit's Rust source emits `_c: { state }` outside the method and injects `let { state } = this['_c'];` before the original `const state`. The descriptor references an out-of-scope variable and the registered function has duplicate lexical declarations. `let` and local function variables also fail; nested object methods inside an existing worklet are affected too.

Could we complete the `MethodProp` scope handling in `BindingCollector` and add regressions for top-level `const`/`let` and nested object-method locals, checking that the generated code parses and executes? The current direct-return method fixtures do not exercise this.

packages/react/package.json
```diff
@@ -37,6 +37,11 @@
       "types": "./runtime/lib/document.d.ts",
       "default": "./runtime/lib/document.js"
     },
+    "./main-thread-object": {
+      "types": "./runtime/lib/main-thread-object.d.ts",
+      "lazy": "./runtime/lazy/main-thread-object.js",
```
Yradex, line 42: [P1] Route the public subpath through the lazy-aware alias

Adding a `lazy` export here does not make standalone lazy imports select it. `plugin-react-alias/src/index.ts` applies the `lazy` condition through its private resolver and installs aliases for a fixed list of requests; `@lynx-js/react/main-thread-object` is missing from that list, and the bundler's default resolver does not receive that condition.

Resolving this public subpath in a standalone lazy therefore selects `runtime/lib/main-thread-object.js`, not the host-forwarding bridge. That implementation brings its own type/handle registries, ID allocator and initialization queue, while imports from the root entry forward to the host.

Resolver and current-source dual-instance probes show that a root-entry type token passed to the subpath hook is rejected as an invalid token; host and lazy can also both allocate ID 1, and the host neither flushes the lazy's private queue nor recognizes its handles.

Could we add the subpath to the lazy-aware aliases and cover a real standalone-lazy public import, including root/subpath interoperability? The current lazy unit test imports `../../lazy/main-thread-object.js` directly, so it bypasses the failing resolution boundary.

packages/webpack/react-webpack-plugin/src/ReactWebpackPlugin.ts
```diff
@@ -489,10 +542,30 @@ class ReactWebpackPlugin {
...
+                      ),
+                  );
+                })
+              )
+            );
             lepusCode.chunks.push({
               name: 'worklet-runtime',
               source: new RawSource(fs.readFileSync(
-                options.workletRuntimePath,
+                usesMainThreadObject
```
Yradex, line 565: [P1] Prevent a first lazy bundle from locking the page into the core runtime

Per-bundle export usage does not account for the page-wide lifetime of `globalThis.lynxWorkletImpl`. A same-version loading sequence can be:

1. The host has no MTF, so the outer `registerWorkletInternal` gate injects no runtime, even if the host supports lazy bundles.
2. Ordinary standalone lazy A has an MTF but no MTO usage and selects core. It initializes the shared global runtime first.
3. MTO-using lazy B contains full, but `loadWorkletRuntime(B)` returns early because the global already exists. B's full chunk is never loaded, and type registration fails because core has no `registerMainThreadObjectType`.

A production Webpack module-graph probe using the current lazy bridge sources produces `usedExports = []` for the MTO entry when only ordinary hooks are used: it is not necessarily indeterminate/full. A separate current-source loader/bindings probe loads only A and reproduces B's registration error. These are module-graph and source-level probes, not a full Rspeedy/native end-to-end run.

Could we conservatively use full for standalone lazy bundles, or otherwise guarantee that a lazy-capable host initializes full before any lazy? Reinitializing full over core would need to preserve existing worklet/ref state. Please cover the no-MTF host -> ordinary lazy A -> MTO lazy B sequence; this is separate from deferred static type-registration analysis.

packages/rspeedy/plugin-react/src/entry.ts
```diff
@@ -321,6 +321,11 @@ export function applyEntry(
         workletRuntimePath: await resolve(
           `@lynx-js/react/${isDev ? 'worklet-dev-runtime' : 'worklet-runtime'}`,
         ),
+        workletCoreRuntimePath: await resolve(
```
Yradex, line 324: [P1] Preserve production builds with the supported older React peers

This unconditionally resolves the newly added `@lynx-js/react/worklet-core-runtime` export during configuration, even when the application uses no worklets or MainThreadObject. The plugin still declares `@lynx-js/react` peers `^0.124.0 || ^0.125.0 || ^0.126.0`, and the existing 0.125.0 package does not export this path.

Updating only the plugin can therefore break an otherwise supported application's production build with an export-not-found error. A resolver probe with the base package manifest confirms this. The `workletCoreRuntimePath || workletRuntimePath` fallback in `ReactWebpackPlugin` cannot help because this awaited resolution fails before the plugin is instantiated.

Could we detect support for the optional core export at this package-version boundary and keep the existing full runtime for older packages? Alternatively, the minimum peer version and release compatibility contract would need to change explicitly. A production-config regression using a supported pre-MTO React package would cover the gap left by workspace-only tests.

packages/react/transform/crates/swc_plugin_worklet/lib.rs
```diff
@@ -80,6 +80,81 @@ impl Default for WorkletVisitor {
...
+    let Some(worklet_type) = self.check_is_worklet_block(body) else {
+      n.visit_mut_children_with(self);
+      return;
+    };
+
+    let mut collector = ExtractingIdentsCollector::new(ExtractingIdentsCollectorConfig {
+      custom_global_ident_names: self.cfg.custom_global_ident_names.clone(),
+      shared_identifiers: Some(self.shared_identifiers.clone()),
+    });
+    method.visit_mut_with(&mut collector);
```
Yradex, line 101: [P2] Exclude computed method keys from the factory's captures

This is specifically the computed-key case, unlike the earlier discussion about non-computed `create`/`dispose` keys. Visiting the entire `MethodProp` also visits the expression used to name the property:

```js
const name = 'create';
defineMainThreadObjectType({
  type: 'example',
  [name](initial) {
    'main thread';
    return { value: initial };
  },
});
```

The current collector/generator emits `_c: { name }`, although `name` is evaluated when constructing the definition object and is not a dependency of the factory body. The capture-free validation then rejects this factory.

Could we analyze only the method's parameters/body for its captures, while leaving the computed key evaluated at the original object-definition site? This needs a separate regression from local `const`/`let` collection: merely adding the missing `BindingCollector` scope visitor does not necessarily exclude the key expression.

packages/react/runtime/src/worklet-runtime/workletRef.ts
```diff
@@ -39,12 +68,118 @@ const createWorkletRef = <T>(
...
+    throw new Error(
+      `MainThreadObject protocol mismatch for type "${type}": runtime supports version ${MAIN_THREAD_OBJECT_PROTOCOL_VERSION}, but the handle or bundle uses ${
+        String(protocolVersion)
+      }. Rebuild the main template and lazy bundle with compatible @lynx-js/react versions.`,
+    );
+  }
+}
+
+function isHydratedWorkletValue(value: unknown): value is object {
+  return isMutableCell(value)
```
Yradex, line 172: [P2] Check registered target identity before probing mutable-cell fields

`isMutableCell(value)` reads `value._wvid` before checking the authoritative MainThreadObject WeakMap. Here `value` can be the arbitrary object returned by the user's factory, not a framework descriptor.

For example, a factory can return:

```js
new Proxy({ get: () => initial }, {
  get(target, key) {
    if (!(key in target)) throw new Error(`Unknown property ${String(key)}`);
    return Reflect.get(target, key);
  },
});
```

With the current `workletRef.ts` and `hydrate.ts` source, first-screen `target.get()` succeeds, but `hydrateCtx()` fails with `Unknown property _wvid`. The public target contract does not prohibit proxies/accessors or require third-party objects to tolerate reads of framework-private fields.

Could we recognize registered MTO targets through the WeakMap first, and only use the existing mutable-cell shape check for other values? This avoids invoking user getters/proxy traps on an already-known target and needs no additional validation mechanism. A target-side hydration regression would cover it.

packages/react/runtime/src/snapshot/worklet/ref/mainThreadObject.ts
```diff
@@ -0,0 +1,387 @@
...
+ * Define a main-thread object type for use by a library-provided hook.
+ *
+ * @param definition - Stable type key and target-object factory function.
+ * @returns An immutable object type definition.
+ * @public
+ */
+export function defineMainThreadObjectType<I, O extends object>(
+  definition: MainThreadObjectTypeDefinition<I, O>,
+): MainThreadObjectType<I, O> {
+  if (typeof definition.type !== 'string' || definition.type.length === 0) {
```
Yradex, line 160: [P2] Reject the reserved `main-thread` type key at definition

The public contract currently permits any non-empty, globally unique stable string, but `workletRef.ts` treats `type === 'main-thread'` as the legacy mutable-cell protocol in both realization and hydration classification.

Consequently, `defineMainThreadObjectType({ type: 'main-thread', create(...) { ... } })` is accepted but its factory is silently bypassed. A current-source initialization probe with `[1, 42, 'main-thread', 1]` produces `{ current: 42, _wvid: 1 }` with zero factory calls; invoking a target method such as `get()` then fails.

Could we explicitly reject and document this reserved key here, with a regression test? Distinguishing typed objects using the protocol marker is another option, but rejecting the legacy sentinel at the public boundary is the smaller change.

f0rdream:
Closing this monolithic PR in favor of the upstream stacked series:

- #3788 — foundation
- #3789 — capture/compiler support
- #3790 — hardening

The new series is based on the latest `main`, preserves the contributor-authored object-method capture commit, and removes the unused core/full worklet-runtime split.

PR #3478, opened 2026-08-12
fix(motion): bridge background MotionValue set

Huxpro:
## Summary

- let MainThreadObject types expose a narrow, explicit set of background-callable methods
- bridge only `MotionValue.set()` from ReactLynx's background runtime to the authoritative main-thread upstream MotionValue
- keep synchronous reads/subscriptions main-thread-only instead of pretending Web-equivalent RPC
- add runtime and declarative regression coverage for `style={{ x }}` updated from a background tap without a React rerender

## Stack

Stacked on #3474 (`agent/motion-completion-control-regression`).

## Compatibility boundary

The bridge is opt-in per object type and method. Motion exposes only the one-way `set()` mutation; synchronous `get()`, `getVelocity()`, `isAnimating()`, and subscription return values remain a documented platform gap (Huxpro/motion#62).

## Validation

- React runtime TypeScript build
- React MainThreadObject/worklet tests: 28/28
- API Extractor
- immutable preview dual-renderer verification will be attached after pkg.pr.new publishes `dd1b4f6`

## Summary by CodeRabbit

* **New Features**
  * Added declarative Motion components for views, text, images, and custom components.
  * Added animated styles, variants, transitions, tap and hover interactions, MotionValue bindings, and typed main-thread object APIs.
  * Added stable MotionValue hydration with background updates.

* **Bug Fixes**
  * Improved animation cleanup, unmount cancellation, style restoration, visibility handling, and CSS custom-property updates.
  * Added support for `initial={false}`, `transitionEnd`, property-specific transitions, and removed animation targets.

* **Documentation**
  * Expanded Motion usage, compatibility, MotionValue, and declarative animation documentation.

### DIFF
--- a/.changeset/motion-value-hydration.md
+++ b/.changeset/motion-value-hydration.md
@@ -1,0 +1,6 @@
+---
+"@lynx-js/react": patch
+"@lynx-js/motion": patch
+---
+
+Preserve live MotionValue subscriptions when first-screen main-thread objects hydrate into their canonical background identities.
--- a/.changeset/motion-value-hydration.md
+++ b/.changeset/motion-value-hydration.md
@@ -3,4 +3,4 @@
 "@lynx-js/motion": patch
 ---
 
-Preserve live MotionValue subscriptions when first-screen main-thread objects hydrate into their canonical background identities.
+Allow explicitly selected main-thread object methods to bridge background calls, and support background `MotionValue.set()` without exposing synchronous methods across runtimes.
--- a/packages/motion/__tests__/declarative.test.tsx
+++ b/packages/motion/__tests__/declarative.test.tsx
@@ -508,6 +508,39 @@ describe('declarative Motion', () => {
     expect(getByTestId('box').getAttribute('style')).toContain('scale(1.5)');
   });
 
+  test('updates a MotionValue-backed style from a main-thread event', async () => {
+    function App() {
+      const x = useMotionValue(-36);
+      function move() {
+        'main thread';
+        x.set(36);
+      }
+      return (
+        <motion.view
+          data-testid='box'
+          main-thread:bindtap={move}
+          style={{ x }}
+        />
+      );
+    }
+
+    const { getByTestId } = render(<App />, {
+      enableMainThread: true,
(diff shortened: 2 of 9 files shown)


Huxpro:
Consumer verification is merged in Huxpro/motion#63: the immutable preview was exercised with a real `style={{ x: MotionValue }}` binding and background `MotionValue.set`, focused 5/5 and full dual-renderer 40/40. The Android sandbox reached the exact bundle but the available Playground client rejected it at decode time, so native evidence remains explicitly false. Repository CI for f3beab5 is green across Ubuntu/Windows builds, Vitest, React coverage, Rust, Playwright, type and publish checks.

Huxpro:
Confidence: **DIRTY / HACKY — moved back to draft; do not review as part of the canonical stack yet.**\n\nReasons after re-auditing against the latest #3477 (`3181442c`):\n\n1. #3477 now explicitly defines a MainThreadObject source value as an opaque handle, **not an RPC stub**. This PR changes the generic ReactLynx API to `backgroundMethods: (handle) => Partial<O>` and installs a Proxy solely to make one Motion method callable. That reopens the cross-runtime object model #3477 just closed.\n2. The apparent API is only partially compatible: `set()` becomes an asynchronous one-way dispatch while `get()`, subscriptions, velocity, animation state, and return values remain unavailable. Typing it as `Partial<O>` makes a generic object look more RPC-like than the runtime contract actually is.\n3. The first commit added a subscription-preserving `hydrate` hook, but the bridge commit removes that hook and restores replacement hydration. Therefore the final net patch does not match the “preserve MotionValue subscriptions on hydration” portion of the PR history/title.\n4. This expands ReactLynx public architecture for a Motion-specific convenience. The cleaner direction is to keep #3477’s opaque handle invariant, treat background `MotionValue.set()` as the recorded platform gap (Huxpro/motion#62), and design any future command bridge as an explicit ReactLynx primitive with protocol/lifecycle semantics—not `Partial<O>` masquerading as the target object.\n\nI am restacking gesture conformance without #3478 to verify that gestures do not require this bridge.

PR #3479, opened 2026-08-12
fix(motion): report tap animation lifecycle

Huxpro:
## Summary

- report `onAnimationStart` / `onAnimationComplete` for `whileTap` activation and restoration
- preserve upstream Motion definition semantics for tap, hover/animate fallback, and base fallback targets
- suppress stale completion when a tap animation is interrupted

## Conformance gap

Priority: Importance 4 / Lynx fit 4 / MTS effort 2 / ReactLynx effort 1 / CSS effort 0.

Upstream contracts:
- `packages/framer-motion/src/gestures/__tests__/press.test.tsx` — press gesture variant applies and unapplies
- `packages/motion-dom/src/animation/interfaces/visual-element.ts` — lifecycle callbacks receive the active animation definition

## Validation

- TypeScript project build
- ESLint, Biome, dprint pre-commit checks
- focused declarative tests for apply/restore definitions and interrupted completion

Stacked on #3478 (`agent/motion-value-hydration`). The base is temporarily `main` to obtain the full CI and immutable pkg.pr.new preview; it will be restored to the stack branch after workflow dispatch.

## Summary by CodeRabbit

* **New Features**
  * Added declarative Motion components for views, text, images, and custom components.
  * Added `useMotionValue` support with main-thread retention and background updates.
  * Added transitions, variants, gestures, lifecycle callbacks, `transitionEnd`, and CSS custom-property animation support.
  * Added typed main-thread object APIs and a selectable declarative Motion demo.

* **Bug Fixes**
  * Improved animation cleanup, cancellation, visibility, style restoration, and unmount behavior.
  * Preserved CSS property naming and animation completion behavior.

* **Documentation**
  * Expanded Motion usage guidance, supported capabilities, and limitations.

PR #3481, opened 2026-08-12
fix(motion): report tap animation lifecycle

Huxpro:
## Summary

- report `onAnimationStart` / `onAnimationComplete` for `whileTap` activation and restoration
- preserve upstream Motion definition semantics for tap, hover/animate fallback, and base fallback targets
- aggregate multi-property completion and suppress stale completion after interruption

## Conformance gap

Priority: Importance 4 / Lynx fit 4 / MTS effort 2 / ReactLynx effort 1 / CSS effort 0.

Upstream contracts:
- `packages/framer-motion/src/gestures/__tests__/press.test.tsx` — press gesture variant applies and unapplies
- `packages/motion-dom/src/animation/interfaces/visual-element.ts` — lifecycle callbacks receive the active animation definition

## Validation

- Motion declarative regression 20/20
- TypeScript project check
- ESLint, Biome, dprint pre-commit checks
- immutable preview dual-renderer lifecycle case is the acceptance gate

Stacked on #3478 (`agent/motion-value-hydration`). The base is temporarily `main` to obtain full CI and the immutable pkg.pr.new preview; it will be restored to the stack branch after workflow dispatch.

Supersedes #3479 because GitHub stopped emitting Actions checks for its updated HEAD.


Huxpro:
Superseded by #3482 so there is one canonical replacement while Actions dispatch is investigated.

PR #3485, opened 2026-08-12
fix(motion): apply gesture transitionEnd values

Huxpro:
## Summary
- apply target-local `transitionEnd` after `whileTap` and `whileHover` animations finish
- wait for every animated property before committing final values
- suppress stale final writes with the existing interaction generation guard
- restore lower-priority hover/animate/style values when each gesture ends

## Why
Upstream Motion applies a gesture variant's `transitionEnd` only after that gesture animation completes, then unapplies it when the gesture exits. Lynx already parsed `transitionEnd` but only consumed it for base `animate`.

## Stack
- Base: #3484 (`agent/motion-hover-animation-lifecycle`)
- This PR contains only gesture `transitionEnd` behavior.

## Validation
- `@lynx-js/motion`: 131/131 tests
- TypeScript and dprint: pass
- package build + Publint: pass
- Huxpro/motion dual-renderer headless harness: 42/42
- real browser tap: opacity 0.5 → transitionEnd 0.75 → rest 1
- real browser hover: opacity 0.9 → transitionEnd 0.8 → rest 1

## Conformance priority
Importance 3 · Lynx fit 5 · MTS 1 · ReactLynx 0 · CSS 0. No platform blocker and no new Full Demo; this completes semantics in the existing gesture pattern.

PR #3486, opened 2026-08-12
fix(motion): restore gesture transition ownership

Huxpro:
## Summary
- restore tap and hover values with the active lower-priority target transition
- preserve hover transition ownership when tap releases over an active hover
- otherwise prefer the base animate transition, then the initial/component fallback transition

## Why
Upstream animation state assigns transitions to the target being restored. When gesture-owned keys fall back to the base style, it uses the initial transition if present. Lynx previously reused the exiting tap/hover transition, producing different release timing.

## Stack
- Base: #3485 (`agent/motion-gesture-transition-end`)
- This PR contains only gesture restoration transition ownership.

## Validation
- `@lynx-js/motion`: 132/132 tests
- TypeScript, dprint, package build, Publint: pass
- Huxpro/motion dual-renderer headless harness: 42/42
- tap→rest and hover→rest both reach an instant base transition inside a strict 30ms window on Web and Lynx
- slow restoration interruption regression remains green

## Conformance priority
Importance 4 · Lynx fit 5 · MTS 1 · ReactLynx 0 · CSS 0. No platform blocker and no new Full Demo.

### DIFF
--- a/.changeset/motion-gesture-rest-transition.md
+++ b/.changeset/motion-gesture-rest-transition.md
@@ -1,0 +1,5 @@
+---
+"@lynx-js/motion": patch
+---
+
+Restore tap and hover values with the active lower-priority target transition instead of reusing the exiting gesture transition.
--- a/packages/motion/__tests__/declarative.test.tsx
+++ b/packages/motion/__tests__/declarative.test.tsx
@@ -607,6 +607,31 @@ describe('declarative Motion', () => {
     expect(getByTestId('button').getAttribute('style')).toContain('opacity: 1');
   });
 
+  test('uses the initial transition when a tap restores its base value', async () => {
+    const { getByTestId } = render(
+      <motion.view
+        data-testid='button'
+        initial={{ opacity: 1, transition: { type: false } }}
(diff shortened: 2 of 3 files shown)


Huxpro:
Evidence for exact HEAD `667b3f1b9e03732674025c6740cbc25a6967e100`:

- `@lynx-js/motion`: 132/132 tests
- TypeScript, dprint, package build, Publint: pass
- Huxpro/motion dual-renderer headless harness: 42/42
- tap→rest and hover→rest use the lower-priority `transition: { type: false }` and settle inside the same strict 30ms window on Web and Lynx
- tap-over-hover transition ownership and slow-restoration interruption regressions remain green

Priority: Importance 4 · Lynx fit 5 · MTS 1 · ReactLynx 0 · CSS 0. No architecture blocker and no new Full Demo.

Immutable-preview gate remains closed: `https://pkg.pr.new/lynx-family/lynx-stack/@lynx-js/motion@667b3f1b9e` currently returns E404, so Hux pins/manifest/metrics/docs/loss remain unchanged.

PR #3487, opened 2026-08-12
fix(motion): apply transitionEnd-only gestures

Huxpro:
## What changed

- apply `whileTap` and `whileHover` definitions that only contain `transitionEnd`
- preserve Motion lifecycle callbacks even when the gesture has no animated target values
- defer the final write to the next main-thread frame and suppress stale writes after gesture exit
- add package regressions for application, lifecycle order, restoration, and interruption

## Why

Upstream `motion-dom` schedules `transitionEnd` even when `animateTarget` creates no animations. The Lynx declarative gesture path previously treated an empty target as a no-op, so a valid `transitionEnd`-only gesture never applied.

This PR is stacked on #3486.

## Validation

- `@lynx-js/motion`: 134/134 tests
- TypeScript: pass
- dprint: pass
- package build + publint: pass
- local hypothesis only: Web/Lynx dual-renderer focused gesture suite 5/5
- local hypothesis only: full dual-renderer suite 43/43

Immutable `pkg.pr.new` evidence will be added when the exact package is available. No consumer metrics or conformance status are moved before that gate.

## Priority

I2 / F5 / M1 / R0 / C0. This is an upstream-specified long-tail gesture target shape with low architecture conflict. It extends an existing usage pattern, so it does not add a separate Full Demo.

### DIFF
--- a/packages/motion/__tests__/declarative.test.tsx
+++ b/packages/motion/__tests__/declarative.test.tsx
@@ -607,6 +607,65 @@ describe('declarative Motion', () => {
     expect(getByTestId('button').getAttribute('style')).toContain('opacity: 1');
   });
 
+  test('applies a transitionEnd-only tap definition', async () => {
+    const lifecycle: string[] = [];
+    const onAnimationStart = vi.fn((value: MotionTarget) => {
+      lifecycle.push(`start:${String(value.transitionEnd !== undefined)}`);
+    });
+    const onAnimationComplete = vi.fn((value: MotionTarget) => {
+      lifecycle.push(`complete:${String(value.transitionEnd !== undefined)}`);
+    });
+    const definition = { transitionEnd: { opacity: 0.4 } };
+    const { getByTestId } = render(
+      <motion.view
+        data-testid='button'
+        style={{ opacity: 1 }}
+        whileTap={definition}
+        transition={{ type: false }}
+        onAnimationStart={onAnimationStart}
+        onAnimationComplete={onAnimationComplete}
+      />,
+      { enableMainThread: true, enableBackgroundThread: true },
+    );
+
+    fireEvent.touchstart(getByTestId('button'));
+    await act(async () => {
+      await Promise.resolve();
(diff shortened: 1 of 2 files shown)


Huxpro:
### Convergence record

- **Gap:** `whileTap` / `whileHover` with `transitionEnd` and no animated target keys.
- **Upstream contract:** `packages/motion-dom/src/animation/interfaces/visual-element-target.ts`; `animateTarget` schedules `transitionEnd` through `frame.update` even when `animations.length === 0`, while `animateVisualElement` still reports start/complete.
- **Priority:** I2 / F5 / M1 / R0 / C0.
- **Layers:** MTS scheduling only; no ReactLynx, CSS, or native rendering change.
- **Before:** gesture lifecycle ran no animation and the final value was never written.
- **After:** lifecycle reports the gesture definition, the final value is written on the next MTS frame, release restores rest, and generation invalidation suppresses a stale scheduled write.
- **Package:** 134/134 tests; TypeScript, dprint, package build, and publint pass.
- **Dual renderer:** local hypothesis focused gesture suite 5/5; full suite 43/43.
- **Gallery:** existing gesture card/test surface extended; no separate Full Demo because this does not unlock a new broad usage pattern.
- **Native:** no immutable native evidence yet. The implementation uses ReactLynx's main-thread `requestAnimationFrame`, which requires Lynx SDK 2.16+.
- **Loss:** unchanged until the exact immutable preview package proves the consumer case. Manifest/docs/metrics remain gated.

Exact preview target: `https://pkg.pr.new/lynx-family/lynx-stack/@lynx-js/motion@527a97ec28`.

Huxpro:
Native follow-up: Lynx Sandbox `aries_10` + UIApp/Playground successfully created a Lynx DevTool session and fetched `main.lynx.bundle` over ADB reverse, but the host rendered `Decode error: unknown Decode Error`. No native behavior claim is made; immutable native evidence remains unavailable because the Sandbox host cannot decode the current Rspeedy bundle.

PR #3488, opened 2026-08-12
fix(motion): apply transitionEnd-only targets

Huxpro:
## What changed

- apply `animate` definitions that contain `transitionEnd` but no animated target keys
- keep `transitionEnd` keys owned by the active definition so they do not trigger an incorrect removed-key restoration
- create and bind a MotionValue when the final property was not previously animated
- preserve lifecycle notification and generation-based stale-write suppression
- avoid capturing active animation handles in ReactLynx rerender snapshots

## Why

Upstream `motion-dom` schedules `transitionEnd` even when `animateTarget` creates no animations. Its animation state also treats `transitionEnd` keys as owned by the active definition. The Lynx declarative path previously skipped that final write and could misclassify the key as removed.

This PR is stacked on #3487.

## Validation

- `@lynx-js/motion`: 136/136 tests
- TypeScript and dprint: pass
- package build + publint: pass
- local hypothesis only: focused dual-renderer transitionEnd/lifecycle suite 4/4
- local hypothesis only: full Web/Lynx dual-renderer suite 44/44

Immutable `pkg.pr.new` evidence will be added when the exact package is available. Consumer manifest, metrics, docs, and loss stay unchanged before that gate.

## Priority

I3 / F5 / M1 / R0 / C0. This completes an upstream target-shape invariant using existing MotionValues and the existing MTS lifecycle bridge. It extends the existing declarative target pattern rather than unlocking a separate Full Demo.

### DIFF
--- a/packages/motion/__tests__/declarative.test.tsx
+++ b/packages/motion/__tests__/declarative.test.tsx
@@ -549,6 +549,41 @@ describe('declarative Motion', () => {
     expect(onUpdateAnimationStart).toHaveBeenCalledOnce();
   });
 
+  test('uses transitionEnd without lifecycle when initial is false', async () => {
+    const onAnimationStart = vi.fn();
+    const onAnimationComplete = vi.fn();
+    const { getByTestId } = render(
+      <motion.view
+        data-testid='box'
+        initial={false}
+        animate={{
+          x: 20,
+          y: 20,
+          transitionEnd: { x: 10, z: 20 },
+        }}
+        transition={{ type: false }}
+        onAnimationStart={onAnimationStart}
+        onAnimationComplete={onAnimationComplete}
+      />,
+      { enableMainThread: true, enableBackgroundThread: true },
+    );
+
+    await act(async () => {
+      await new Promise(resolve => setTimeout(resolve, 30));
+    });
+    expect(getByTestId('box').getAttribute('style')).toContain(
+      'translateX(10px)',
+    );
+    expect(getByTestId('box').getAttribute('style')).toContain(
+      'translateY(20px)',
+    );
+    expect(getByTestId('box').getAttribute('style')).toContain(
+      'translateZ(20px)',
+    );
+    expect(onAnimationStart).not.toHaveBeenCalled();
+    expect(onAnimationComplete).not.toHaveBeenCalled();
+  });
+
   test('animates a MotionValue-backed style', async () => {
     function App() {
       const value = useMotionValue(1);
--- a/packages/motion/__tests__/declarative.test.tsx
+++ b/packages/motion/__tests__/declarative.test.tsx
@@ -572,9 +572,6 @@ describe('declarative Motion', () => {
       { enableMainThread: true, enableBackgroundThread: true },
     );
 
-    await act(async () => {
-      await new Promise(resolve => setTimeout(resolve, 30));
-    });
     expect(getByTestId('box').getAttribute('style')).toContain(
       'translateX(10px)',
     );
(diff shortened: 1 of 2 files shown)


Huxpro:
### Convergence record

- **Gap:** `animate={{ transitionEnd: … }}` with no animated target keys.
- **Upstream contract:** `packages/motion-dom/src/animation/interfaces/visual-element-target.ts` applies `transitionEnd` with zero animations; `packages/framer-motion/src/render/utils/__tests__/animation-state.test.ts` verifies swapping a property from `target` to `transitionEnd` keeps it owned by the active definition.
- **Priority:** I3 / F5 / M1 / R0 / C0.
- **Layers:** MTS declarative target ownership and scheduling; no ReactLynx or CSS behavior change.
- **Before:** the empty target produced no MotionValue/final write and a previous target key could be treated as removed.
- **After:** transitionEnd-only updates create/bind the missing value, preserve key ownership, report lifecycle, and ignore stale writes after a newer update.
- **Package:** 136/136 tests; TypeScript, dprint, package build, and publint pass.
- **Dual renderer:** local hypothesis focused transitionEnd/lifecycle suite 4/4; full suite 44/44.
- **Gallery:** hypothesis uses a focused existing target card; no Full Demo because this is an edge target shape, not a new broad usage pattern.
- **Native:** no claim. The same Sandbox host used for #3487 fetched the bundle but failed with `Decode error: unknown Decode Error`.
- **Loss:** unchanged until exact immutable preview verification; manifest/docs/metrics remain gated.

Exact preview target: `https://pkg.pr.new/lynx-family/lynx-stack/@lynx-js/motion@692097415d`.

Huxpro:
Upstream completion audit added `initial={false}` coverage from `animate-prop.test.tsx`: the final `transitionEnd` values are applied, but mount lifecycle callbacks remain suppressed. Updated head `b6775eb91d`; package is now 137/137, focused dual-renderer transitionEnd/initial-false/lifecycle is 3/3, build + publint/typecheck/dprint pass. Exact preview target is now `https://pkg.pr.new/lynx-family/lynx-stack/@lynx-js/motion@b6775eb91d`; metrics remain gated.

Huxpro:
Final lifecycle audit: transitionEnd-only base updates now assert exact `start → complete` behavior in package tests and both headless renderers, while the ordinary animated lifecycle regression still proves no circular snapshot payload. Head `d48c4ef201`; package 137/137; focused dual renderer 3/3. Exact preview target updated to `https://pkg.pr.new/lynx-family/lynx-stack/@lynx-js/motion@d48c4ef201`.

Huxpro:
Initial snapshot audit tightened the upstream `initial={false}` contract: transitionEnd values are now merged into the first rendered animate state (x/y/z = 10/20/20 immediately), with no mount start/complete callbacks. Head `a7f6085842`; package 137/137; focused dual renderer 3/3; build/publint/typecheck/dprint pass. Exact preview target updated to `https://pkg.pr.new/lynx-family/lynx-stack/@lynx-js/motion@a7f6085842`.

Huxpro:
Exact named-variant race evidence is merged in Huxpro/motion#86 against immutable `2c805a2`. The test first proves the instant `on` variant can apply its `display:flex` transitionEnd, restores `off` to `display:none`, then switches `on→off` within one turn and confirms deferred completion cannot write stale flex. Focused Web-first/Lynx 1/1; full suite 66/66; evidence build passes. Priority I4/F5/M1/R1/C0; no native boundary or Full Demo claim; weighted loss remains 7 across 67 tracked contracts.

PR #3489, opened 2026-08-12
fix(motion): retain values removed from animate

Huxpro:
## Summary

- match upstream Motion ownership when a key is removed from `animate`
- restore the current `initial` value when present
- retain the current MotionValue snapshot when the key leaves both `animate` and `initial`
- keep the retained value serializable for later Lynx main-thread gesture snapshots

## Upstream conformance

Mirrors the three adjacent `animate` prop contracts in `packages/framer-motion/src/motion/__tests__/animate-prop.test.tsx`:

- restore the originally defined initial value
- restore the currently defined initial value
- perform no animation when removed from both animate and initial

## Verification

- `@lynx-js/motion`: 140/140 package tests
- focused Web/Lynx headless hypothesis: 1/1
- focused retained-value + transitionEnd-only gesture regression: 2/2
- full Web/Lynx suite: new case passes; 44/45 total with one pre-existing CDP touch retry double-dispatch flake, which passes 1/1 on isolated rerun
- package artifact publint: all good

Priority: I4 / F5 / M2 / R0 / C0. This closes a common conditional-target pattern but does not justify a separate Full Demo.

### DIFF
--- a/.changeset/motion-removed-target-retention.md
+++ b/.changeset/motion-removed-target-retention.md
@@ -1,0 +1,5 @@
+---
+"@lynx-js/motion": patch
+---
+
+Retain the current MotionValue when a property is removed from both `animate` and `initial`.
--- a/packages/motion/__tests__/declarative.test.tsx
+++ b/packages/motion/__tests__/declarative.test.tsx
@@ -430,6 +430,99 @@ describe('declarative Motion', () => {
     expect(getByTestId('box').getAttribute('style')).toContain('opacity: 0.8');
   });
 
+  test('restores a value removed from animate to its initial value', async () => {
+    function App() {
+      const [active, setActive] = useState(true);
+      return (
+        <view>
+          <motion.view
+            data-testid='box'
+            initial={{ opacity: 0 }}
+            animate={active ? { opacity: 1 } : {}}
+            transition={{ type: false }}
+          />
+          <view data-testid='toggle' bindtap={() => setActive(false)} />
+        </view>
+      );
+    }
+
+    const { getByTestId } = render(<App />, {
+      enableMainThread: true,
+      enableBackgroundThread: true,
+    });
+    await act(async () => {
+      await new Promise(resolve => setTimeout(resolve, 30));
+    });
+    expect(getByTestId('box').getAttribute('style')).toContain('opacity: 1');
+    fireEvent.tap(getByTestId('toggle'));
+    await act(async () => {
+      await new Promise(resolve => setTimeout(resolve, 30));
+    });
+    expect(getByTestId('box').getAttribute('style')).toContain('opacity: 0');
+  });
+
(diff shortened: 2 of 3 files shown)


Huxpro:
Conformance evidence (pre-immutable hypothesis)

- Upstream source: `animate-prop.test.tsx` lines covering removal from `animate` with original initial, current initial, and no initial fallback.
- Package suite: 140/140.
- Headless focused ownership comparison: Web + Lynx pass 1/1.
- Focused ownership + transitionEnd-only gesture regression: 2/2.
- Full headless suite: the new case and its gesture serialization regression pass; 44/45 overall due to the existing CDP touch retry occasionally dispatching two taps (`Tapped 2`), and that existing case passes 1/1 on isolated rerun.
- Package build emitted all 24 dist files. Direct packed-tarball publint reports `All good!`; the Rslib publint hook itself is affected by `@publint/pack` not matching pnpm absolute tarball filenames in this environment.
- Priority: I4/F5/M2/R0/C0.
- Native claim: none. The currently available Sandbox host is already recorded as unable to decode this Rspeedy bundle, so no repeated unverifiable native claim.
- Full Demo: no; this is a foundational conditional-target ownership contract, not a broad new API surface.

Metrics/docs remain unchanged until the exact pkg.pr.new artifact for this HEAD is available and passes the same consumer suite.

Huxpro:
Exact consumer evidence is merged in Huxpro/motion#84: immutable `2c805a2` restores static opacity/rotate after a named variant is removed, keeps later style updates reactive, lets variant re-entry reclaim ownership, and masks style changes while the variant remains active. Focused Web-first/Lynx test 1/1; complete dual-renderer suite 64/64; evidence build passes. This case is pure declarative ownership (no native host boundary), so no Sandbox or Full Demo claim. Priority I4/F5/M1/R1/C0; weighted loss remains 7 while tracked coverage grows to 65.

Huxpro:
Additional exact consumer evidence is merged in Huxpro/motion#85: immutable `2c805a2` switches from an opacity variant to an x-only variant, restores omitted opacity from static style, and still applies x. Focused Web-first/Lynx test 1/1; complete dual-renderer suite 65/65; evidence build passes. Pure declarative ownership: no Sandbox or Full Demo claim. Priority I4/F5/M1/R1/C0; tracked coverage is now 66 with weighted loss unchanged at 7.

Huxpro:
Confidence: **VERY CONFIDENT — ready to review.** Re-ported the stronger #3489 semantics together with #3460’s static-style case onto the canonical local stack. Motion tests: 124/124; TypeScript build: pass. The combined behavior matches upstream ownership: current `initial` wins when present, otherwise a value removed from both `animate` and `initial` retains its displayed snapshot. No MTS or CSS workaround is introduced.

Huxpro:
Superseded by #3518, restacked on the canonical conformance stack.

PR #3490, opened 2026-08-12
fix(motion): render initial transform origin

Huxpro:
## Summary

- render `originX`, `originY`, and `originZ` aliases into the first `transform-origin` style snapshot
- match upstream percentage, pixel, and missing-axis defaults
- keep subsequent updates on the existing upstream `motion-dom` styleEffect path

## Upstream conformance

Mirrors `build-styles.test.ts` (`Builds transformOrigin with correct default value types`): `{ originX: 0.2, originY: "60%", originZ: 10 }` renders `20% 60% 10px`.

## Verification

- `@lynx-js/motion`: 142/142 package tests
- focused Web/Lynx headless transform-origin hypothesis: 1/1
- focused transform-origin + removed-target + gesture regression: 3/3
- full Web/Lynx suite: new case passes through its final assertions on isolated rerun; the initial full run exposed a harness timing assertion and the existing CDP retry double-tap flake, both tightened in the consumer hypothesis

Priority: I3 / F5 / M1 / R0 / C1. No Full Demo: this improves a compositional transform detail but does not unlock a broad new usage pattern.

### DIFF
--- a/.changeset/motion-transform-origin-initial.md
+++ b/.changeset/motion-transform-origin-initial.md
@@ -1,0 +1,5 @@
+---
+"@lynx-js/motion": patch
+---
+
+Render declarative transform origin aliases in the initial style snapshot.
--- a/packages/motion/__tests__/declarative.test.tsx
+++ b/packages/motion/__tests__/declarative.test.tsx
@@ -47,6 +47,20 @@ describe('declarative Motion', () => {
     });
   });
 
+  test('resolves initial transform origin aliases with upstream defaults', () => {
+    expect(
+      resolveInitialStyle(
+        undefined,
+        { originX: 0.2, originY: '60%', originZ: 10 },
+      ),
+    ).toEqual({
+      transformOrigin: '20% 60% 10px',
+    });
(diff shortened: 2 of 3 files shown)


Huxpro:
Conformance evidence (pre-immutable hypothesis)

- Upstream source: `packages/framer-motion/src/render/html/utils/__tests__/build-styles.test.ts`, `Builds transformOrigin with correct default value types`.
- Package suite: 142/142.
- Focused Web/Lynx transform-origin case: 1/1; first snapshot resolves 0%/0%, later update resolves to the element width/height. A plain React/ReactLynx control updates in the same case, confirming no host-style blocker.
- Focused transform-origin + removed-target + transitionEnd-only gesture regression: 3/3.
- Priority: I3/F5/M1/R0/C1.
- No native claim: the available Sandbox host remains unable to decode the current Rspeedy bundle.
- No Full Demo: origin composition is useful but does not unlock a broad new Motion pattern.

The Hux harness remains a local hypothesis until the exact pkg.pr.new package for this HEAD exists; metrics/docs/loss stay unchanged until immutable replay.

PR #3493, opened 2026-08-12
feat(motion): propagate delayChildren

Huxpro:
## Summary

- propagate numeric `delayChildren` through inherited base variant labels
- accumulate delay through uncontrolled descendants
- preserve Motion ownership: an explicit child `animate` starts a new controlling root
- continue to use each child's upstream MotionValue animation rather than introducing a second scheduler

## Upstream contracts

- `packages/framer-motion/src/motion/__tests__/delay.test.tsx` — `in variant children via delayChildren`
- `packages/framer-motion/src/motion/__tests__/variant.test.tsx` — `delay propagates throughout children`

## Validation

- @lynx-js/motion: 147/147
- package JS + declarations: generated successfully
- direct pnpm pack + publint tarball: All good
- full repository build is unavailable in this environment because cargo/Rust is not installed; the package build then reaches the known @publint/pack temporary-tarball hook failure after successful JS/declaration generation

## Scope boundary

This PR intentionally covers numeric `delayChildren` only. Dynamic `delayChildren: stagger()`, `staggerChildren`, `when`, controls, gesture propagation, child ordering, and subtree completion aggregation remain tracked in Huxpro/motion#10.

Stacked on #3492.

PR #3491, opened 2026-08-12
test(motion): validate declarative conformance stack

Huxpro:
## Validation-only rollup — DO NOT MERGE

This draft exists only because feature-base PRs do not publish `pkg.pr.new` packages. It rolls the atomic Motion stack onto `main` so the exact motion/react/react-umd set can be consumed by the Hux conformance harness.

Capability ownership remains with #3483–#3490 and #3492–#3499. Review and merge those atomic PRs, not this rollup.

Current validation HEAD: `ed0c9f29bb`

This immutable preview includes #3499 (a memoized inherited child restores identity values when a later parent variant removes an animated transform key). Hux will pin the exact motion/react/react-umd set, rerun the focused upstream Web/Lynx contract, then the complete suite before any metrics move.

PR #3494, opened 2026-08-12
feat(motion): support variant inheritance opt-out

Huxpro:
## Summary

- add the upstream-compatible declarative `inherit` prop
- make `inherit={false}` publish an empty variant context so parent labels do not propagate through that component
- keep the prop out of host forwarding
- add a package test proving an inherited `initial` label stops at the boundary

## Stack

- base: #3493 (`agent/motion-delay-children`)
- this PR is one atomic Motion contract

## Validation

- `@lynx-js/motion`: 148/148 tests
- dprint + Biome: pass
- motion API extractor/type build: pass
- JS/declaration build completes; direct packed tarball passes publint
- required full Turbo build attempted; this local environment lacks `cargo`, so it stops in pre-existing Rust/wasm dependencies before the Motion package build

## Priority and scope

I4 / F5 / MTS1 / ReactLynx1 / CSS0. This is a context opt-out with no new animation engine, host API, MTS callable, CSS behavior, Full Demo, or native boundary. Parent-driven dynamic orchestration is intentionally still tracked in Hux issue #10.

### DIFF
--- a/.changeset/fair-birds-inherit.md
+++ b/.changeset/fair-birds-inherit.md
@@ -1,0 +1,5 @@
+---
+"@lynx-js/motion": minor
+---
+
+Support `inherit={false}` to stop parent variant labels propagating through a declarative Motion component.
--- a/packages/motion/README.md
+++ b/packages/motion/README.md
@@ -47,6 +47,7 @@ The declarative layer currently supports:
 - object targets and string/array/function `variants` for `initial` and
   `animate`, including `custom`, target-local `transition`, and parent label
   inheritance for base targets when a child does not define its own label;
+  `inherit={false}` prevents labels propagating through a variant boundary;
   numeric `delayChildren` also propagates through inherited labels
 - `initial={false}` to render the final `animate` keyframe without a mount
   animation while preserving later target updates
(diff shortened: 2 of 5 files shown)


Huxpro:
Immutable consumer proof is complete via validation-only #3491 at c394dd3:

- exact packages: https://pkg.pr.new/@lynx-js/motion@c394dd3 plus matching react/react-umd
- @lynx-js/motion package suite: 148/148
- Hux evidence build: pass
- complete headless upstream-Web / Lynx-for-Web conformance suite: 50/50
- manifest: 51 tracked, 46 conformant, 1 partial, 4 blocked; weighted loss remains 10 after denominator expansion
- priority: I4 / F5 / MTS1 / ReactLynx1 / CSS0
- harness/evidence merged: https://github.com/Huxpro/motion/pull/70

The contract is deliberately the upstream context boundary: inherit={false} prevents inherited initial/animate labels propagating through that component. It does not claim to remove descendants from parent-driven visual-element orchestration; that larger boundary remains in https://github.com/Huxpro/motion/issues/10. No Full Demo or native claim was added.

PR #3495, opened 2026-08-12
feat(motion): propagate initial false to variant children

Huxpro:
## Summary

- include `initial={false}` in the declarative variant context
- let inherited variant children render their final animate keyframe on the first frame without a mount animation
- add a focused package test for the parent/child contract

## Stack

- base: #3494 (`agent/motion-variant-inherit-opt-out`)
- this PR is one atomic Motion contract

## Validation

- `@lynx-js/motion`: 149/149 tests
- JS/declaration build + publint: pass
- pre-commit ESLint/Biome/dprint: pass

## Priority and scope

I5 / F5 / MTS1 / ReactLynx1 / CSS0. This extends the existing variant context and upstream MotionValue path; it adds no animation engine, host API, MTS callable, CSS behavior, Full Demo, or native boundary.

PR #3496, opened 2026-08-13
fix(motion): preserve explicit child mount animations

Huxpro:
## Summary
- inherit parent `initial={false}` only when the child also inherits the parent animate label
- preserve mount animation ownership for children with their own object or explicit animate prop
- add a package regression that observes the child static first frame, animated target, and lifecycle start

## Why
Upstream Motion context applies parent `initial={false}` to inherited variant descendants, but does not rewrite an independently controlled child. The previous #3495 condition applied `initial={false}` to every child with no explicit initial, causing an explicit object animate target to render immediately without lifecycle.

## Verification
- `@lynx-js/motion` declarative suite: 41/41
- `@lynx-js/motion` build and declaration generation
- Publint passed

## Stack
Atomic follow-up stacked on #3495. Exact pkg.pr.new validation will be performed through validation-only #3491 after this PR publishes its head.

### DIFF
--- a/packages/motion/__tests__/declarative.test.tsx
+++ b/packages/motion/__tests__/declarative.test.tsx
@@ -152,7 +152,7 @@ describe('declarative Motion', () => {
       enableBackgroundThread: true,
     });
     await act(async () => {
-      await new Promise(resolve => setTimeout(resolve, 30));
+      await new Promise(resolve => setTimeout(resolve, 100));
     });
     expect(getByTestId('child').getAttribute('style')).toContain('opacity: 0');
     expect(getByTestId('child').getAttribute('style')).toContain(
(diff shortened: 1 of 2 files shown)


Huxpro:
Exact cross-repo evidence is now merged in Huxpro/motion#79.

- immutable preview: `dfb913f` (`@lynx-js/motion`, `@lynx-js/react`, and `@lynx-js/react-umd`)
- package declarative suite: 41/41; build and Publint pass
- focused upstream-Web/Lynx-for-Web contract: 3/3
- complete dual-renderer suite: 59/59
- monitor: 60 tracked / 55 conformant / 1 partial / 4 blocked; API 39 supported / 3 partial / 8 blocked; weighted loss remains 8 after rounding
- priority: I5 / F5 / M1 / R1 / C0

This validates upstream `initial=false doesn't propagate to props`: inherited `initial={false}` applies only to children inheriting the parent animate label, while an explicit object child preserves its mount animation and `start|complete` lifecycle. No Full Demo or native claim is added because this is a narrow context-ownership contract.

PR #3497, opened 2026-08-13
fix(motion): apply initial variant transition end

Huxpro:
## Summary

Apply a resolved initial variant's `transitionEnd` values on the first rendered frame, matching upstream Motion's `applies applyOnEnd if set on initial` contract.

## Stack

- base: validation rollup `agent/motion-complex-value-audit` (`dfb913f`)
- one semantic change and one package regression test

## Verification

- before fix on exact `dfb913f`: upstream Web computes `display: none`; Lynx-for-Web remains `display: flex`
- package declarative suite after fix: 42/42
- exact pkg.pr.new Hux validation will be attached after this branch is rolled into the validation publisher

This only changes initial target composition: `transitionEnd` overrides the normal initial target as upstream does. No new MTS or CSS path is introduced.

### DIFF
--- a/packages/motion/__tests__/declarative.test.tsx
+++ b/packages/motion/__tests__/declarative.test.tsx
@@ -960,6 +960,27 @@ describe('declarative Motion', () => {
     expect(onAnimationComplete).not.toHaveBeenCalled();
   });
 
+  test('applies transitionEnd from an initial variant on the first frame', () => {
+    const { getByTestId } = render(
+      <motion.view
+        data-testid='box'
+        initial='visible'
+        variants={{
+          visible: {
+            opacity: 1,
+            transitionEnd: { display: 'none' },
+          },
+        }}
+      />,
+      { enableMainThread: true, enableBackgroundThread: true },
+    );
+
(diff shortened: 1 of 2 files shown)


Huxpro:
Exact cross-repo evidence is merged in Huxpro/motion#81.

- immutable preview: `2c805a2` (`@lynx-js/motion`, `@lynx-js/react`, and `@lynx-js/react-umd`; all report the same commit header)
- before fix on exact `dfb913f`: upstream Web computes `display: none`, Lynx-for-Web remains `display: flex`
- package declarative suite after fix: 42/42
- focused upstream-Web/Lynx-for-Web contract: 1/1
- complete dual-renderer suite: 61/61
- monitor: 62 tracked / 57 conformant / 1 partial / 4 blocked; API 39 supported / 3 partial / 8 blocked; weighted loss remains 8 after rounding
- priority: I3 / F5 / M1 / R1 / C0

This verifies upstream `applies applyOnEnd if set on initial`: the initial named variant composes `transitionEnd` over its ordinary target on the first rendered frame. No Full Demo or native claim is added because this is a narrow target-composition contract.

PR #3498, opened 2026-08-13
fix(motion): restore inherited variant style fallback

Huxpro:
## Summary

Restore a child's static style when a value disappears from an inherited parent variant.

## Root cause

The removed-key path used the child's complete initial snapshot as its fallback. For an inherited label, that snapshot contains the initial variant value, but upstream Motion intentionally falls back to the child `style` value when a later parent variant omits the key.

This change explicitly carries inherited-animation ownership into the host hook and separates:

- initial values used as animation starting values; and
- values used only when a previously animated key is removed.

Direct `animate` targets keep their existing initial-value fallback semantics.

## Stack

- Base: #3497 (`agent/motion-initial-transition-end`)
- This PR is one atomic ReactLynx ownership fix; no MTS, CSS, or native host change.

## Validation

- new package regression: `a(0.5) → b(1) → c({})` restores child `style.opacity=0`
- declarative suite: 43/43
- complete `@lynx-js/motion` suite: 152/152
- pre-commit ESLint/Biome/dprint: pass
- declaration build: pass
- `pnpm pack`: pass
- Rslib's Publint post-build step is locally blocked by its temp-tarball lookup even though independent pack succeeds; recorded rather than hidden

Consumer validation with an immutable preview package will be attached after the validation rollup publishes this commit.

### DIFF
--- a/.changeset/sharp-rabbits-inherit.md
+++ b/.changeset/sharp-rabbits-inherit.md
@@ -1,0 +1,5 @@
+---
+"@lynx-js/motion": patch
+---
+
+Restore child style values when an inherited variant removes an animated key.
--- a/packages/motion/__tests__/declarative.test.tsx
+++ b/packages/motion/__tests__/declarative.test.tsx
@@ -169,6 +169,53 @@ describe('declarative Motion', () => {
     );
   });
 
+  test('restores child style when an inherited variant removes a value', async () => {
+    function App() {
+      const [phase, setPhase] = useState(0);
+      const label = phase === 0 ? 'a' : phase === 1 ? 'b' : 'c';
+      return (
+        <view>
+          <motion.view initial='a' animate={label}>
+            <motion.view
+              data-testid='child'
+              style={{ opacity: 0 }}
+              variants={{
+                a: { opacity: 0.5 },
+                b: { opacity: 1 },
+                c: {},
+              }}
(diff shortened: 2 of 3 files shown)


Huxpro:
Exact consumer evidence is merged in Huxpro/motion#88 against immutable `cd567e7` (motion/react/react-umd all return matching x-commit-key). Before: inherited parent `a→b→c({})` restored Web child style opacity 0 but left Lynx at initial variant opacity 0.5. After #3498: both restore 0 while delayed inherited animation starts and direct animate initial fallback remain covered. Package declarative 43/43; package full 152/152; focused 1/1; evidence build pass; complete dual-renderer 68/68. Priority I4/F4/M1/R2/C0; no native/Sandbox or Full Demo claim; 69 tracked, weighted loss 7.

Huxpro:
Follow-up upstream evidence: Huxpro/motion#89 verifies `new child items animate from initial to animate` on the immutable `cd567e7` motion/react/react-umd set. A child mounted after its parent has settled still resolves inherited `hidden` and reaches `visible` through a neutral Motion wrapper. Focused Web/Lynx 1/1; complete suite 69/69. No additional lynx-stack source change was required. Native remains unclaimed because Sandbox Playground SDK 0.0.1 cannot decode the current Rspeedy bundle.

PR #3499, opened 2026-08-13
fix(motion): restore inherited removed values

Huxpro:
## Summary

- restore identity values when an inherited variant removes transform keys from a child
- keep static style fallback from #3498 higher priority than the identity fallback
- cover a memoized child that does not render in lockstep with its parent

## Upstream contract

`packages/framer-motion/src/motion/__tests__/variant.test.tsx` — `Children correctly animate to removed values even when not rendering along with parents`

Before this patch, `hidden → visible → hidden` left a memoized child at `x=100` on Lynx while Web restored `x=0`.

## Verification

- declarative package suite: 44/44
- complete `@lynx-js/motion` suite: 153/153
- declaration generation succeeds
- ESLint and staged Biome/dprint checks pass
- independent `npm pack --dry-run`: succeeds

The Rslib build emits JS and declarations, then the local Publint post-build step hits the existing missing temporary tarball issue; independent packing succeeds.

## Stack

Base: #3498 (`agent/motion-inherited-variant-style-fallback`)

### DIFF
--- a/.changeset/bright-moles-restore.md
+++ b/.changeset/bright-moles-restore.md
@@ -1,0 +1,5 @@
+---
+"@lynx-js/motion": patch
+---
+
+Restore transform identity values when inherited variants remove keys from memoized children.
--- a/packages/motion/__tests__/declarative.test.tsx
+++ b/packages/motion/__tests__/declarative.test.tsx
@@ -4,7 +4,7 @@
 
 import { beforeEach, describe, expect, test, vi } from 'vitest';
 
-import { useState } from '@lynx-js/react';
+import { memo, useState } from '@lynx-js/react';
 import type { IntrinsicElements } from '@lynx-js/types';
 import { act, fireEvent, render } from '@lynx-js/react/testing-library';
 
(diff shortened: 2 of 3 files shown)


Huxpro:
Exact consumer evidence is now published in Huxpro/motion#90.

- immutable package set: `ed0c9f2` for `@lynx-js/motion`, `@lynx-js/react`, and `@lynx-js/react-umd`
- all three responses: HTTP 200 and `x-commit-key: lynx-family:lynx-stack:ed0c9f2`
- before (`cd567e7`): Web restored the memoized inherited child to `x=0`; Lynx retained `x=100`
- after (`ed0c9f2`): hidden → visible → hidden settles at `{ opacity: 0, x: 0 }` in both renderers
- focused dual-renderer: 1/1; complete suite: 70/70
- package validation: declarative 44/44; full motion 153/153
- manifest: 71 tracked / 66 conformant / 1 partial / 4 blocked; weighted loss remains 7 after rounding

Native is not claimed: the current Rspeedy bundle remains undecodable by Sandbox Playground SDK 0.0.1. No Full Demo was added because this is an ownership-semantic correction, not a newly complete end-to-end pattern.

PR #3502, opened 2026-08-13
refactor(react): isolate MainThreadObject lifecycle code

f0rdream:
## Motivation

`MainThreadObject` represents an identity-bearing JavaScript object that exists only in the ReactLynx main-thread runtime. Its `create` and `dispose` implementations therefore should not be shipped in the background-thread bundle.

The initial API in #3453 accepts ordinary shared-runtime functions. With a real MotionValue factory, that pulls `motion-dom` into both BTS and MTS even though BTS can only hold an opaque handle.

This PR is a focused follow-up to #3453 and targets its head branch so the lifecycle boundary can be fixed before the Motion integration is finalized.

## Design

Library authors still define a type once at module scope:

```ts
import { createValue } from './value.js' with { runtime: 'shared' };

const valueType = defineMainThreadObjectType({
  type: '@library/Value',
  create(initialValue) {
    'main thread';
    return createValue(initialValue);
  },
  dispose(value) {
    'main thread';
    value.dispose();
  },
});
```

The compiler turns both lifecycle methods into capture-free Main Thread Function descriptors. The returned `MainThreadObjectType` is an immutable public token containing only:

- `type`
- `isHandle(value)`
- `getInitialPayload(handle)`

The token does not expose the lifecycle implementation.

On MTS, the type definition is registered during module evaluation. The registry keeps the descriptors and resolves each one lazily, once, after compiler-appended worklet registration has completed. The first MTS hook use also idempotently ensures registration for runtimes that retain a module across a per-page worklet-registry reset. BTS receives only descriptors and opaque handles.

Lifecycle functions must be capture-free. Dependencies must use shared-runtime imports, which lets the same source identifier resolve inside MTS without pulling its implementation into BTS.

This also teaches the worklet transform to recognize object method syntax, so the natural `create() { 'main thread' }` form has the same behavior as an arrow Main Thread Function.

## Why this should land first

The bundle boundary is part of the core `MainThreadObject` contract. Fixing it after library integrations would either preserve duplicated code for compatibility or require every library to adopt a temporary conditional workaround. Making lifecycle functions MTS-owned first gives Motion and future libraries one stable model.

This path is isolated from ordinary `MainThreadRef`: its two-field wire representation and recursive transform path are unchanged. Lifecycle resolution runs only for typed `MainThreadObject` definitions.

## Bundle evidence

An isolated MotionValue build using the original ordinary shared callback produced a 266,017-byte Lynx bundle and placed the 411,904-byte `motion-dom` module in both `mto__main-thread` and `mto` chunks.

With Main Thread Function lifecycle callbacks, the bundle is 191,980 bytes (74,037 bytes / 27.8% smaller). `motion-dom` appears only in `mtoFactoryWorklet__main-thread`; the BTS chunk has no `motion-dom` module.

The stacked Motion example confirms the same result across its real entries: `motion-dom` is present only in `main__main-thread`, `declarative__main-thread`, and `mini__main-thread`, and absent from the `main`, `declarative`, and `mini` BTS chunks.

## Validation

- `pnpm turbo build` (69/69)
- `pnpm turbo api-extractor -- --local` (25/25)
- `UPDATE=1 cargo test -p swc_plugin_worklet` (60/60)
- React transform fixture suite (52/52)
- React runtime snapshot/worklet/guardrail suite (96 files; 785 passed, 2 skipped, 4 todo)
- focused retained-module registration test (18/18 in `workletRef.test.jsx`)
- `pnpm dprint check`
- scoped ESLint and pre-commit checks

Related RFC: #3446

Stacked on: #3453


f0rdream:
Superseded by the upstream stacked MainThreadObject implementation in #3788 → #3789 → #3790. The lifecycle work is now part of the foundation, with the v1 API retaining no public dispose method.

PR #3504, opened 2026-08-13
feat(motion): support explicit beforeChildren timing

Huxpro:
## Summary

Add the explicit-duration, non-repeating slice of upstream Motion's `when: "beforeChildren"` contract.

A parent variant now contributes its explicit numeric delay and duration to the existing inherited variant context, so inherited descendants start after the parent timing window. Empty-target and repeating parents are deliberately excluded.

Stacked on #3499. Remaining automatic/repeating duration, `afterChildren`, dynamic delay/stagger, controls, and gesture propagation stay tracked in Huxpro/motion#10.

## Verification

- exact package tests: 4/4
- declarative package suite: 48/48
- full motion package suite: 157/157
- Rslib build, declarations, and Publint: pass
- immutable validation package: `51d11ee` for motion/react/react-umd, matching `x-commit-key`
- consumer old/new proof: old `ed0c9f2` Lynx jumped the inherited child to opacity 1 at 200ms while Web held 0.1; `51d11ee` holds both at 0.1 and settles only after the 600ms parent duration
- consumer focused headless: 1/1
- consumer complete headless: 74/74
- Android Explorer loads the exact bundle and exposes initial child opacity 0.1; its SDK 0.0.1 did not dispatch either DevTool tap or scaled ADB touch into the page, so native timing remains unclaimed

Consumer evidence: Huxpro/motion#95.

### DIFF
--- a/.changeset/motion-before-children.md
+++ b/.changeset/motion-before-children.md
@@ -1,0 +1,5 @@
+---
+"@lynx-js/motion": patch
+---
+
+Honor `when: "beforeChildren"` for declarative variants with a non-repeating, explicit parent duration.
--- a/packages/motion/README.md
+++ b/packages/motion/README.md
@@ -49,7 +49,10 @@ The declarative layer currently supports:
   inheritance for base targets when a child does not define its own label;
   `inherit={false}` prevents labels propagating through a variant boundary;
   parent `initial={false}` also suppresses inherited child mount animations;
-  numeric `delayChildren` also propagates through inherited labels
+  numeric `delayChildren` also propagates through inherited labels;
+  `when: "beforeChildren"` waits for a parent's non-repeating,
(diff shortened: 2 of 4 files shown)


Huxpro:
Consumer proof is now published in Huxpro/motion#95. The immutable `51d11ee` motion/react/react-umd set passes the exact dual-renderer case 1/1 and the complete headless suite 74/74; old `ed0c9f2` Lynx jumped the child to opacity 1 inside the parent window, while this head holds 0.1 at 200ms and settles after the 600ms parent duration. Package declarative is 48/48 and full motion is 157/157. Native Explorer loads the exact bundle and exposes the expected initial 0.1, but its SDK 0.0.1 did not dispatch the page tap, so native timing remains explicitly unclaimed.

PR #3506, opened 2026-08-13
feat(motion): honor value-specific beforeChildren timing

Huxpro:
## Summary

Extend #3504's explicit-duration `when: "beforeChildren"` slice to value-specific parent transitions.

Instead of reading only top-level `duration`/`delay`, this uses upstream Motion's `getValueTransition` routing for every animated key and delays inherited children until the longest explicit value window completes. If any parent value is automatic, instant, or repeating, the adapter still makes no orchestration claim.

Stacked on #3504.

## Verification

- new exact tests:
  - waits for the longest value-specific `delay + duration`
  - refuses the claim when one parent value uses automatic spring timing
- declarative: 50/50
- full Motion package: 159/159
- Motion Rslib build, declarations, Publint pass
- full Turbo build remains environment-blocked before Motion by missing `cargo` in React transform/Web Core WASM tasks

This narrows the remaining issue #10 boundary to automatic/repeating completion, `afterChildren`, dynamic stagger, controls, and gesture propagation. Consumer immutable-package proof will be added after pkg.pr.new publishes this head.

### DIFF
--- a/.changeset/motion-before-children.md
+++ b/.changeset/motion-before-children.md
@@ -2,4 +2,4 @@
 "@lynx-js/motion": patch
 ---
 
-Honor `when: "beforeChildren"` for declarative variants with a non-repeating, explicit parent duration.
+Honor `when: "beforeChildren"` for declarative variants whose non-repeating parent values have explicit durations.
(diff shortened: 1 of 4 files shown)


Huxpro:
Consumer evidence is merged as Huxpro/motion#96. On the identical upgraded scene, immutable `51d11ee` fails because Lynx child opacity is already 1 at 400ms; immutable `36e144e` holds 0.1 through the longest parent value window (opacity 100ms, x 200ms delay + 600ms duration) and later settles. Focused dual-renderer 1/1, complete consumer 74/74, package declarative 50/50, package full 159/159, commit hooks and Rslib/declarations/Publint pass. Native timing remains unclaimed under the existing SDK 0.0.1 event-dispatch boundary.

PR #3507, opened 2026-08-13
feat(motion): wait for finite beforeChildren repeats

Huxpro:
## Summary

Extend the explicit `when: "beforeChildren"` completion window to finite repeats.

For each parent value, the adapter now computes:

`delay + duration × (repeat + 1) + repeatDelay × repeat`

and still takes the longest value-specific window. Infinite repeat and automatic timing remain unclaimed.

Stacked on #3506.

## Verification

- finite repeat + repeatDelay holds the child until the parent completes
- Infinity remains excluded and starts no fake wait
- declarative: 51/51
- full Motion package: 160/160
- Rslib build, declarations, Publint pass

Consumer immutable-package proof will follow after the validation rollup publishes this head.

### DIFF
--- a/.changeset/motion-before-children.md
+++ b/.changeset/motion-before-children.md
@@ -2,4 +2,4 @@
 "@lynx-js/motion": patch
 ---
 
-Honor `when: "beforeChildren"` for declarative variants whose non-repeating parent values have explicit durations.
+Honor `when: "beforeChildren"` for declarative variants whose finite parent values have explicit durations.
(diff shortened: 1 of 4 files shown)


Huxpro:
Consumer evidence is merged as Huxpro/motion#97. The same finite-repeat scene fails on immutable `36e144e` because Lynx child opacity is already 1 at 400ms; immutable `013e20e` holds 0.1 through the sample and settles after the 700ms parent window (100ms delay + two 200ms iterations + 200ms repeatDelay). Focused dual-renderer 1/1, complete consumer 74/74, package declarative 51/51, package full 160/160, commit hooks and Rslib/declarations/Publint pass. Windows validation build succeeded and Ubuntu published the package. Native timing remains unclaimed under the existing SDK 0.0.1 event-dispatch boundary.

PR #3508, opened 2026-08-13
test(motion): cover initial false beforeChildren mount

Huxpro:
## Summary

Add a regression for the upstream composition of `initial={false}` with explicit `when: "beforeChildren"`.

On mount, the inherited child must render its visible variant immediately and must not report an animation start. The parent orchestration window must not manufacture a child delay when mount animation is disabled.

Stacked on #3507. This is test-only because the current implementation already conforms.

## Verification

- exact new regression passes
- declarative: 52/52
- full Motion package: 161/161

No capability or loss metric moves from this package-only regression; consumer dual-renderer evidence follows separately.

### DIFF
--- a/packages/motion/__tests__/declarative.test.tsx
+++ b/packages/motion/__tests__/declarative.test.tsx
@@ -344,6 +344,38 @@ describe('declarative Motion', () => {
     );
   });
 
+  test('initial false skips a beforeChildren mount delay', async () => {
+    const onAnimationStart = vi.fn();
+    const { getByTestId } = render(
+      <motion.view
+        initial={false}
+        animate='visible'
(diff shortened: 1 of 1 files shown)


Huxpro:
Consumer evidence is merged as Huxpro/motion#98. With a ten-second parent `beforeChildren` window, inherited `initial={false}` still renders the child immediately at opacity 1 / x 24 in both Web and ReactLynx; no artificial mount delay is introduced. Focused dual-renderer 1/1, complete consumer 74/74, package declarative 52/52 and full 161/161. This is regression depth only, so API/conformance/loss/native/Full Demo claims remain unchanged.

Issue #1935, opened 2025-11-13
[Feature] Support motion in Lynx


f0rdream:
- [x] Add initial support for Motion  #1062 
  - [x] Add polyfills for Element and Web APIs 
  - [x] Add `animate`, `stagger` with `MainThread.Element`
  - [x] Demos using `animate`, `stagger`
- [ ] Support for import third-party libraries in MTS
- [x] Support getComputedStyle
- [ ] Documentation
- [ ] Further usage exploration
  - [ ] Motion Value in MTS
  - [ ] Declartive sytax in ReactLynx like `motion.view`

PR #3509, opened 2026-08-13
feat(motion): add declarative MotionValue support

f0rdream:
Depends on #3790.

Adds the declarative Motion layer for ReactLynx, including initial and target styles, transitions, variants, tap and hover gestures, animation callbacks, and live MotionValue style bindings.

`useMotionValue` uses ReactLynx `MainThreadObject` handles so Motion values keep their identity and execute on the main thread while remaining usable from background-rendered React code. Examples and scale benchmarks cover the declarative API and MotionValue lifecycle.

This PR is userspace work on top of the ReactLynx MainThreadObject stack. It does not add another ReactLynx runtime path or public disposal API.

Related RFC: #3446. Motion support tracking: #1935.


Huxpro:
Downstream stack note: #3509 and #3477 are currently sibling fork branches rather than linear ancestry. That is why standalone #3509 CI reports missing `defineMainThreadObjectType` / `useMainThreadObject` exports, and GitHub cannot use the fork head as the base of our upstream conformance PRs. I rebuilt the canonical sequence locally as `main → #3477 → #3509`; the Motion suite is green (118/118) with the #3477 transform. Please rebase/update #3509 onto #3477 so the conformance layers can be represented as the real remote stack.

Huxpro:
Core API follow-up: #3513 replaces the two-step handle inspection consumed here with `const handle = motionValueType.downcast(value)` and `handle.creationPayload`. Once #3513 is accepted, this Motion PR should switch its two call sites from `isHandle/getInitialPayload` to the downcast result.

Huxpro:
Naming update for #3513: the final Motion migration is `const handle = motionValueType.downcast(value)`, then `handle.payload` on success (not `creationPayload`).

Huxpro:
Restacked this PR on #3477 as a true dependent layer.

- #3477 remains `main` → `feat/main-thread-object-hardening`.
- #3509 now targets the upstream mirror of that exact #3477 head.
- The fork head was fast-forwarded with a two-parent merge commit, preserving both original histories; only the three overlapping benchmark registration files required additive resolution (`020` ordinary refs plus `021`/`022` MotionValue cases).
- The resulting #3509 diff now contains only the Motion adoption/declarative layer and its small handle-inspection addition.

This is now the canonical foundation for the conformance stack.

Huxpro:
Confidence update: **very confident architecture and ready for foundation review once the author is comfortable removing draft status**.

This PR is now a real dependent diff on #3477 and its new CI run has Ubuntu/Windows builds, code style, type, React, RSTest, Android emulator, and Vitest checks green; the remaining jobs are still running. The canonical conformance stack begins at #3515, with each later semantic gap isolated above this foundation. The complete local stack passes 15 Motion test files / 132 tests and package TypeScript.

Huxpro:
Conformance-stack assessment: **OK — keep draft and iterate.**

The failed Web Elements shard was unrelated to the Motion diff: 437 tests passed, two viewpager cases were reported flaky, and `nyc report` failed in coverage processing. Re-running failed jobs produced a fully green repository CI set, including Linux/Windows builds, Vitest, types, Rust, publishing, and Android emulator coverage.

The remaining reason to stay draft is architectural/acceptance evidence, not CI: gesture `onAnimationComplete` ownership and interruption semantics still need stable package-level coverage, and native proof remains limited by the available host SDK. #3523 and #3524 now cover gesture start, settle, and restoration while retaining those completion/native gaps explicitly.

f0rdream:
## Restacked on the ReactLynx implementation

This PR now targets [#3790](https://github.com/lynx-family/lynx-stack/pull/3790), completing the stack after [#3788](https://github.com/lynx-family/lynx-stack/pull/3788) and [#3789](https://github.com/lynx-family/lynx-stack/pull/3789). Its head is `5ba8764d9e30275134edec131cb6dca937f04308`.

The ReactLynx stack is based on latest `main` (`1d81bd618`) and uses one normal `worklet-runtime`; the old capability-specific runtime split is removed. This PR contains only the userspace Motion layer. It uses `motionValueType.downcast(value)?.creationPayload` and keeps disposal internal rather than exposing a public `dispose` API.

Issue #3447, opened 2026-08-11
Implement MainThreadObject core as a reviewable evolution of #2144


Huxpro:
Parent RFC: https://github.com/lynx-family/lynx-stack/issues/3446

## Outcome

Implement the `MainThreadObject` core model as a reviewable evolution of the `MainThreadValue` mechanism from #2144 / the first commit of #3405.

## Required starting point

The preferred branch history keeps this commit unchanged as an ancestor:

`a2ab025746fa55d69971490bbe160a2b75157392` — `feat(motion): add main-thread MotionValue support`

Add the `MainThreadObject` refactor as follow-up commits rather than flattening or rewriting the original commit. If current `main` must be incorporated, merge it without rewriting `a2ab025` while the design is under review.

The PR description must include:

- `git diff --stat a2ab025..HEAD`;
- a link or command for reviewing `a2ab025..HEAD`;
- a short mapping from each `MainThreadValue` mechanism to its `MainThreadObject` replacement;
- any unavoidable deviations caused by conflicts with current `main`.

## Scope

- model `MainThreadObjectType<I, O>` so both initialization payload and realized object types exist;
- represent the source-runtime value as an opaque `MainThreadObjectHandle<O>`;
- provide a library-author API for defining and realizing main-thread object types;
- preserve stable object identity through first-screen hydration;
- retain the existing `MainThreadRef` wire shape and behavior during this PR;
- replace the forgeable `__MAIN_THREAD_VALUE__` convention with a runtime-owned brand/registry;
- detect conflicting registrations for the same type key;
- define serializable initialization constraints and development diagnostics;
- make transform/lazy-runtime compatibility failures explicit;
- avoid attaching React bookkeeping fields to third-party objects.

## Go/no-go gate

Stop and publish a written assessment instead of forcing an incremental PR if the work requires wholesale replacement of all four original pillars:

1. worklet capture strategy;
2. worklet-value ID/ref pool;
3. factory registry and realization;
4. first-screen hydration identity.

The assessment must explain why preserving `a2ab025` no longer provides useful reviewability and recommend either a fresh design or retaining `MainThreadValue`.

## Non-goals

- Motion declarative API changes;
- migration of Motion consumers;
- renderer-bound Element ref refactoring;
- generic native-resource ownership;
- RPC or background observation of main-thread mutations.

## Acceptance criteria

- [ ] Implementation PR has `a2ab025` as an unchanged ancestor.
[...]

PR #3513, opened 2026-08-13
refactor(react): downcast MainThreadObject handles

Huxpro:
## Summary

Replace the two-step MainThreadObject handle inspection API with a proof-carrying downcast:

```ts
const handle = objectType.downcast(value);
if (handle !== undefined) {
  use(handle.payload);
}
```

A successful downcast returns the original runtime-branded handle, not the realized main-thread object and not a detached metadata record. It performs no creation, realization, dereference, or disposal.

This also removes the duplicated initial-payload entry from the runtime WeakMap: Core retains only the nominal type brand there, while the canonical creation payload stays on the handle.

## Stack

- Depends on #3477
- Motion consumer: #3509
- Umbrella RFC: #3446
- Core tracking issue: #3447
- Motion migration issue: #3448

The branch is based directly on #3477 head `3181442c7`. GitHub will show the parent changes until #3477 lands.

## Validation

Passed:

- `pnpm dprint fmt`
- `pnpm biome check` (2253 files)
- `pnpm turbo build --filter @lynx-js/react-runtime`
- `pnpm turbo api-extractor --filter @lynx-js/react -- --local`
- targeted ESLint for the changed runtime sources
- commit-time lint-staged checks
- `git diff --check`

Coverage added for:

- identity-preserving successful downcast
- exact-type rejection
- primitive/null rejection
- rejection of forged wire-shaped objects
- `undefined` creation payload remaining distinguishable from a failed match
- lazy runtime export forwarding

Unable to run locally:

- full `pnpm turbo build`: the environment has no `cargo`; it stops in the Rust/WASM transform packages before this change compiles
- focused runtime Vitest: the test transform imports the missing Rust-built `@lynx-js/react-transform` artifact, so Vitest reports worker setup errors before collecting tests
- full ESLint after retrying with an 8 GiB heap: repository build outputs are absent because the full build is blocked, producing unrelated unresolved-workspace-package errors; changed runtime sources pass targeted ESLint

No transform code changes are included.


Huxpro:
API naming follow-up applied in `4bcfbee55`: the typed handle projection is now `handle.payload` instead of `handle.creationPayload`. Runtime build, API Extractor, Biome, targeted ESLint, commit hooks, and diff checks pass.

f0rdream:
Superseded by the upstream stacked MainThreadObject implementation in #3788 → #3789 → #3790. The downcast contract is included in #3788, including the contributor-authored capture work retained in #3789.

PR #3483, opened 2026-08-12
fix(motion): report tap animation lifecycle

Huxpro:
## Summary

- report `onAnimationStart` / `onAnimationComplete` for `whileTap` activation and restoration
- aggregate multi-property completion and suppress stale completion after interruption
- preserve active hover/animate/base fallback definition semantics

## Conformance priority

Importance 4 / Lynx fit 4 / MTS 2 / ReactLynx 1 / CSS 0.

## Validation

- Motion declarative regression 20/20
- TypeScript project check
- ESLint, Biome, dprint pre-commit checks
- immutable preview dual-renderer lifecycle case is the acceptance gate

Temporarily based on `main` for full CI/pkg.pr.new, then stacked on #3478. Canonical replacement for #3479/#3481/#3482 after their updated SHAs did not receive main Test workflow runs.

## Summary by CodeRabbit

* **New Features**
  * Added declarative Motion components for views, text, images, and custom components.
  * Added spring animations, variants, transitions, gesture interactions, lifecycle callbacks, and `transitionEnd` support.
  * Added `useMotionValue` for reactive animation values and main-thread updates.
  * Added typed main-thread object APIs and improved MotionValue hydration.
* **Bug Fixes**
  * Preserved styles, visibility, and completion behavior during animation updates, exits, cancellations, and unmounts.
  * Improved CSS property handling, including custom properties.
* **Documentation**
  * Expanded Motion usage guidance, supported features, and limitations.
* **Tests**
  * Added comprehensive coverage for declarative animations, gestures, MotionValues, and main-thread objects.

PR #3523, opened 2026-08-13
fix(motion): report tap animation lifecycle

Huxpro:
## What

Report `onAnimationStart`/`onAnimationComplete` for tap-driven animations, including interruption ownership and restoration ordering.

## Stack

Depends on #3522. Supersedes #3483.

## Validation

- `@lynx-js/motion`: 15 test files, 132 tests passed on the complete stack
- package TypeScript check passed
- headless Chromium probe verifies press/release state, callbacks, and final scale/color

## Confidence

**OK — stays draft.** The behavior is correct in headless validation, but lifecycle completion crosses the MTS/background boundary and gesture interruption ownership remains more complex than base animation callbacks. Native validation is additionally blocked by the available Lynx Sandbox client's obsolete SDK/runtime contract.

## Checklist

- [x] Tests updated.
- [x] Documentation updated.
- [x] Changeset added.

### DIFF
--- a/.changeset/motion-tap-animation-lifecycle.md
+++ b/.changeset/motion-tap-animation-lifecycle.md
@@ -1,0 +1,5 @@
+---
+"@lynx-js/motion": patch
+---
+
+Report declarative animation lifecycle callbacks when tap targets apply and restore.
--- a/.changeset/motion-tap-animation-lifecycle.md
+++ b/.changeset/motion-tap-animation-lifecycle.md
@@ -2,4 +2,4 @@
 "@lynx-js/motion": patch
 ---
 
-Report declarative animation lifecycle callbacks when tap targets apply and restore.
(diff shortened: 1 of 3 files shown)


Huxpro:
Conformance status: **OK — keeping this PR in draft.**

New package-level coverage now proves tap target starts, every animated value reaches its target, release starts restoration for every value, and every value returns to its base state. The full `@lynx-js/motion` suite passes (134/134; declarative 25/25), and the stacked branch is CLEAN.

Remaining readiness gap: `onAnimationComplete` is not yet reliably observable in the package test scheduler, and native Sandbox verification is still gated by the available client/SDK. The non-enumerable active-animation handle is also a fragile implementation detail. I would promote this to ready after completion/cancellation has stable acceptance coverage or the bookkeeping is refactored into an independently testable helper.

PR #3484, opened 2026-08-12
feat(motion): report hover animation lifecycle

Huxpro:
## Summary
- report `onAnimationStart` and `onAnimationComplete` for `whileHover` activation and restoration
- aggregate multi-property completion and suppress stale callbacks when another interaction interrupts hover
- share interaction animation cancellation tracking with the preceding tap lifecycle layer
- document the exact declarative lifecycle boundary

## Why
Upstream Motion reports the applied hover definition and then the restored base definition through the animation lifecycle callbacks. Lynx already applied the visual hover states, but omitted these callbacks.

## Stack
- Base: #3483 (`agent/motion-gesture-lifecycle-final`)
- This PR contains only the hover lifecycle layer.

## Validation
- `pnpm --filter @lynx-js/motion test` — 130 passed
- `pnpm exec tsc -p packages/motion/tsconfig.json --noEmit`
- `pnpm exec dprint check ...`
- `pnpm --filter @lynx-js/motion build` — Publint passed
- Huxpro/motion dual-renderer headless harness — 42 passed, including exact `start:hover → complete:hover → start:rest → complete:rest` ordering and interruption suppression

## Conformance priority
Importance 3 · Lynx fit 4 · MTS 2 · ReactLynx 1 · CSS 0

PR #3524, opened 2026-08-13
feat(motion): report hover animation lifecycle

Huxpro:
## What

Extend animation lifecycle reporting to hover-driven animations and add a readable interaction probe showing idle/hover/press state, gesture counts, and animation start/completion counts.

## Stack

Depends on #3523. Supersedes #3484.

## Validation

- `@lynx-js/motion`: 15 test files, 132 tests passed on the complete stack
- package TypeScript check passed
- headless Chromium verifies idle → hover → press → release → leave with 5/5 animation lifecycles and no console errors

## Confidence

**OK — stays draft.** Web/headless behavior is solid, but hover is meaningful only on pointer-capable Lynx clients and shares the cross-thread interruption/ownership complexity of tap lifecycle. Native proof awaits a current Sandbox host SDK.

## Checklist

- [x] Tests updated.
- [x] Documentation updated.
- [x] Changeset added.

### DIFF
--- a/.changeset/motion-hover-animation-lifecycle.md
+++ b/.changeset/motion-hover-animation-lifecycle.md
@@ -1,0 +1,5 @@
+---
+"@lynx-js/motion": patch
+---
+
+Report declarative animation lifecycle callbacks when hover targets apply and restore, including multi-property completion and interrupted animation suppression.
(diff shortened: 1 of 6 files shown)


Huxpro:
Conformance status: **OK — keeping this PR in draft.**

New package-level coverage now proves global hover entry starts every animated value, target styles settle, pointer exit starts restoration for every value, and base styles are restored. The full `@lynx-js/motion` suite passes (134/134; declarative 25/25), and the stacked branch is CLEAN.

Remaining readiness gap: the same completion/cancellation acceptance hole as #3523 remains, plus the global mouse-move bridge needs native client evidence. The non-enumerable active-animation handle is still fragile. I would promote this only after stable completion/interruption coverage and a Sandbox run on a compatible client.

PR #3527, opened 2026-08-13
fix(react): register MainThreadObject definitions statically

### LINKED ISSUE #3526: MainThreadObject registration across dual-thread tree shaking
# Static registration for `MainThreadObject` across dual-thread tree shaking

Related work: #3446 defines the `MainThreadObject` model, #3477 implements the
core primitive, and #3527 implements the registration design described here.

## Summary

`MainThreadObject` lets a library create a stateful JavaScript object on the main thread while background React code retains an opaque handle to it. Before a BTS handle can be realized, MTS must know the matching `create` and optional `dispose` functions.

The first design registered those functions when the type's module or `useMainThreadObject` hook executed on MTS. ReactLynx compiles and tree-shakes BTS and MTS as separate module graphs, so that execution is not guaranteed. BTS can retain a component and create a handle while MTS removes the same component, skips its first-screen render, or never evaluates its type module.

The proposed design makes the `MainThreadObjectType` token the tree-shaking root for its lifecycle definition:

- if the token is live in either BTS or MTS, register its lifecycle on MTS;
- if the token is unused in both graphs, remove the token and lifecycle code;
- keep the public API, handle payload, hydration model, and `MainThreadRef` path unchanged.

### BTS and MTS are independent build graphs

ReactLynx produces two views of an application:

- **BTS** runs background React rendering and later updates;
- **MTS** runs the main-thread first-screen path and Main Thread Functions.

The same source module can have different reachability in the two graphs. Optimizing one graph does not imply that the corresponding module remains in the other graph.

```mermaid
flowchart LR
  S["Application source"] --> C["ReactLynx Compiler"]
  C --> B["BTS module graph"]
  C --> M["MTS module graph"]
  B --> BO["BTS tree shaking"]
  M --> MO["MTS tree shaking"]
```

This is valid and useful. An application can intentionally render a feature only on BTS, and an MTS first-screen build should be able to remove code it never uses.

### What a `MainThreadObject` needs

A library defines how a serializable payload becomes a real MTS object:

```ts
interface MainThreadObjectTypeDefinition<I, O extends object> {
  type: string

  // Runs on MTS to create the real object.
  create(initialPayload: I): O

  // Runs on MTS when ReactLynx releases the object.
  dispose?(object: O): void
}
```

The hook returns one logical object for the component lifecycle:

```ts
const CounterType = defineMainThreadObjectType<number, Counter>({
  type: '@example/Counter',
  create(initialValue) {
    'main thread'
    return new Counter(initialValue)
  },
  dispose(counter) {
    'main thread'
    counter.close()
  },
})

function useCounter(initialValue: number): Counter {
  return useMainThreadObject(CounterType, initialValue)
}
```

On BTS, `useCounter` returns an opaque handle. When a Main Thread Function captures that value, MTS resolves the handle to the real `Counter`. The resolution requires `CounterType` to be registered on MTS first.

## Motivation

Consider a valid thread-specific render:

```tsx
import { CounterFeature } from './CounterFeature.js'

export function App() {
  return __MAIN_THREAD__ ? null : <CounterFeature />
}
```

`CounterFeature` uses `CounterType` and creates a handle on BTS. The MTS graph can remove `CounterFeature`, its hook, and the module that defines `CounterType`.

The old registration path then becomes:

```text
BTS                                      MTS

CounterFeature is live                   CounterFeature is removed
        │                                         │
        ▼                                         ▼
create Counter handle                    CounterType module never executes
        │                                         │
        └──────── handle patch ──────────────────►│
                                                  ▼
                                      "type is not registered"
```

Users can encounter the same shape through:

- an empty or smaller MTS first screen;
- `__MAIN_THREAD__` branches;
- a lazy component whose later React render happens only on BTS;
- different BTS and MTS conditions;
- a library type imported only by a BTS-retained feature.

### Why `MainThreadRef` does not need this registration design

`MainThreadRef` uses one runtime-owned target type: `MutableCell`.

```text
BTS MainThreadRef descriptor ──► MTS runtime creates built-in MutableCell
```

The MTS runtime already contains the constructor, so it does not wait for a library module to register a factory. `MainThreadRef` can still participate in general BTS/MTS render-order problems, but it does not have this extensible-type registration failure.

`MainThreadObject` allows every library to provide a different factory. The build must preserve the factory connected to each live type without retaining every library type.

## Goals

- Register a `MainThreadObject` type before any BTS handle of that type is realized on MTS.
- Treat use in either BTS or MTS as a reason to retain the lifecycle definition.
- Preserve tree shaking at type granularity.
- Support optional `dispose` and referenced `runtime: 'shared'` imports.
- Keep the public type token and serialized handle format unchanged.
- Leave ordinary `MainThreadRef` creation, capture, and cleanup on the existing path.
- Produce actionable diagnostics for source shapes the Compiler cannot register statically.

## Non-goals

- Dynamic construction of object types inside functions or React components.
- Serializing executable factories inside BTS handle payloads.
- Changing capture, hydration identity, reference counting, or disposal timing from #3477.
- Solving React patch and direct-worklet ordering tracked by #3476.
- Making later MTS mutations observable from BTS React rendering.
- Adding Motion-specific behavior to ReactLynx.

### 1. Use a statically recognizable type definition

A compiler-managed definition must:

- use a direct or aliased named import of `defineMainThreadObjectType` from `@lynx-js/react`;
- be assigned at module scope;
- use a static string for `type`;
- define `create` and optional `dispose` as inline Main Thread Functions;
- avoid ordinary closure captures. Explicit `with { runtime: 'shared' }` imports are supported.

```ts
import {
  defineMainThreadObjectType,
  useMainThreadObject,
} from '@lynx-js/react'
import { Counter } from './Counter.js' with { runtime: 'shared' }

export const CounterType = defineMainThreadObjectType<number, Counter>({
  type: '@example/Counter',
  create(initialValue) {
    'main thread'
    return new Counter(initialValue)
  },
  dispose(counter) {
    'main thread'
    counter.close()
  },
})
```

Separate factory references, dynamic type strings, object spreads, and definitions created during render are rejected because their lifecycle cannot be connected to one static type token reliably.

### 2. Emit an opaque marker in both graphs

For each valid definition, the Compiler already knows:

- the stable type key;
- the worklet ID of `create`;
- the worklet ID of `dispose`, when present.

It emits an internal marker import in both BTS and MTS output. The transformed shape is conceptually:

```js
import {
  mainThreadObjectDefinition as marker,
} from '@lynx-js/react/internal/main-thread-object-definition?type=...&create=...&dispose=...'

export const CounterType = /*#__PURE__*/ defineMainThreadObjectType(
  definition,
  marker,
)
```

The second argument and marker module are compiler-private. They do not appear in the public TypeScript signature, the returned token, or the BTS handle payload.

The call remains pure. If `CounterType` is unused, normal tree shaking can remove the call and marker import together.

### 3. Union marker liveness across BTS and MTS

After Rspack optimizes both graphs, the React plugin checks whether each marker import is active.

```mermaid
flowchart TD
  T["MainThreadObject type token"] --> BM["BTS marker"]
  T --> MM["MTS marker"]
  BM --> BL{"Live after BTS optimization?"}
  MM --> ML{"Live after MTS optimization?"}
  BL --> U["BTS live OR MTS live"]
  ML --> U
  U -->|"yes"| R["Retain lifecycle and register on MTS"]
  U -->|"no"| D["Remove type and lifecycle code"]
```

Liveness is the union of the paired graphs. It does not require both threads to render the same component.

The optimizer determines final import activity late in compilation. When the recorded marker state changes, the plugin requests an additional pass so definition injection uses stable liveness information.

### 4. Inject only the live lifecycle definitions

For every live marker, the build locates the collected worklet definitions for `create` and `dispose`, then generates an MTS registration fragment equivalent to:

```js
ReactLynx.registerMainThreadObjectDefinition({
  type: '@example/Counter',
  create: { _wkltId: 'counter-create-id' },
  dispose: { _wkltId: 'counter-dispose-id' },
})
```

This fragment runs with the MTS entry before a BTS handle patch can require the type.

Lifecycle code may reference `runtime: 'shared'` imports. The Compiler records only the shared import specifiers used by that lifecycle function. The plugin generates its lifecycle module beside the owning source module, so relative imports keep their original resolution base.

The same collection and injection model is applied to paired entry and lazy/async scopes.

### 5. Keep hook registration as a fallback

`useMainThreadObject` still performs an idempotent registration on MTS when the hook executes. This covers direct MTS rendering and registry resets.

The fallback is no longer the cross-thread guarantee. BTS-only use is handled by the compiler marker and build-time injection.

## End-to-end flow

```mermaid
sequenceDiagram
  participant C as Compiler
  participant P as Rspack plugin
  participant M as MTS runtime
  participant B as BTS React

  C->>C: Transform type and lifecycle functions
  C->>P: Emit marker in BTS and MTS graphs
  P->>P: Union optimized marker liveness
  P->>M: Inject registration for live types
  M->>M: Register create and dispose
  B->>B: Create opaque handle
  B->>M: Send handle patch or capture
  M->>M: Realize or reuse the typed object
```

Registration is determined by build reachability. Object identity and lifetime continue to follow #3477.

## Required behavior

| BTS token | MTS token | Build result | Runtime result |
| --- | --- | --- | --- |
| live | live | retain one definition | type is registered once; duplicate registration is idempotent |
| live | removed or not rendered | retain definition | BTS handle can be realized on MTS |
| removed | live | retain definition | first-screen/direct MTS use works |
| removed | removed | remove definition | no registration or lifecycle code in the bundle |
| Type A live | Type B live | retain both independently | type/protocol validation still prevents invalid hydration aliasing |

Using one exported type must not retain unused sibling types from the same module.

### Build time

- One small marker import is emitted per static type definition in each graph.
- The plugin scans optimized marker connections and may run an additional pass when liveness changes.
- Lifecycle retention is per type and per lifecycle function.

### Bundle size

- A live type adds its registration fragment and required `create`/`dispose` definitions.
- An unused type, including an unused sibling export, remains removable.
- Shared-runtime imports are retained only when referenced by a live lifecycle function.

### Runtime

- A compiler marker is associated with a live type token through one private `WeakMap` entry per defined type.
- Registration is idempotent.
- The handle wire format and realization protocol do not change.
- Ordinary `MainThreadRef` objects do not enter marker storage or typed-object realization, so this design adds no per-ref runtime wrapper or traversal.

## Diagnostics

The implementation should fail close to the source of an invalid definition:

- **Compiler error:** the type is not a static module-scope definition with inline lifecycle functions.
- **Build error:** a marker references a missing lifecycle definition or a lifecycle definition has no owning module.
- **Runtime error:** the same type key is registered with a different lifecycle definition.
- **Lazy compatibility error:** a newly compiled lazy bundle requires static registration support that the main runtime does not expose.

## Compatibility

- No public `MainThreadObjectType` API change.
- No serialized handle or protocol change.
- No change to hydration, ownership, reference counting, or `dispose` timing.
- No change to `MainThreadRef` or `MainThread.Element` semantics.
- Existing transformed code keeps hook-time registration. The BTS-only guarantee requires rebuilding with the new Compiler and React webpack plugin.
- New lazy bundles detect an incompatible older main runtime and report a targeted error.

### Register during MTS module evaluation

This is the previous design. Independent graph reachability means the module may never execute on MTS while BTS still produces a handle.

### Retain every `MainThreadObject` definition

This guarantees registration but disables type-level tree shaking. A library entry containing many controllers or value types would retain all lifecycle code.

### Send the factory with the BTS handle

Factories are executable MTS code and can depend on transformed worklet definitions and shared-runtime imports. Putting them in each handle would expand the wire contract, duplicate code, and reintroduce closure serialization problems.

### Add every library type to the ReactLynx runtime

This matches the built-in `MutableCell` path used by `MainThreadRef`, but it removes library extensibility and makes ReactLynx own library-specific object construction.

### Build library objects through `MainThreadRef`

Libraries would need to construct and assign `.current`, preserve first-screen identity, guard BTS access, and implement cleanup themselves. The cell-shaped API also leaks into the library's public object model.

## Validation plan

Production-mode bundle tests cover:

- BTS-only token use;
- MTS-only token use;
- use in both graphs;
- a token unused in both graphs;
- used and unused sibling types in one module;
- `create` and optional `dispose` retention;
- relative `runtime: 'shared'` imports;
- malformed markers and invalid static definitions;
- lazy runtime registration forwarding and compatibility diagnostics.

Runtime and compiler suites continue to cover handle identity, hydration validation, class and lexical capture, disposal, duplicate definitions, and ordinary `MainThreadRef` isolation in #3477.

## Rollout and review order

```text
#3477  MainThreadObject core semantics
   │
   ▼
#3527  static registration and tree-shaking guarantee
   │
   ▼
#3509  Motion userspace integration
```

#3527 is intentionally separate from #3477. Reviewers can evaluate the public/runtime object model first, then the build-time registration guarantee, then the Motion consumer.

Related infrastructure:

- #3393 transports definitions missing from a paired MTS graph;
- #3299 tracks the broader shared-runtime import and worklet-definition transport problem;
- #3476 remains deferred because its React patch/direct-worklet ordering issue also exists for `MainThreadRef`.

f0rdream:
Closes #3526.

Ensures a used `MainThreadObject` type is registered on MTS even when its defining React hook or module is not evaluated there.

- emits an opaque compiler marker for static type definitions in both thread graphs
- retains `create` and optional `dispose` definitions when either graph uses the type while removing unused sibling types
- preserves shared-runtime lifecycle imports relative to the owning module
- keeps the public type token, handle wire format, and `MainThreadRef` runtime path unchanged

Stacked on #3477. After #3477 lands, this PR reduces to the static registration change.


f0rdream:
Closing this stale implementation branch so the new stack is unambiguous. The static-registration design remains tracked by #3526 and can be rebased onto #3788 or #3790 when the core stack is ready.

Issue #3538, opened 2026-08-15
RFC: useMainThreadEvent — main-thread event functions over a lifecycle-safe callable transport


Huxpro:
> Implementation: #3539. The RFC text is also versioned in-repo at `docs/superpowers/specs/2026-08-15-main-thread-callable-transport-rfc.md`. **v2** renames the public API into the `useEvent` lineage and adds the "Naming", binding-decomposition, and "Convergence" sections (see the v2 comment below for the delta).

## Summary

Introduce a generic, lifecycle-safe way to hand an **arbitrary consumer-authored main-thread function** — including functions nested inside plain objects and arrays — to long-lived main-thread consumers, with:

- a stable, serializable background-thread handle per component instance;
- captured values that are **re-synchronized on every rerender**;
- deterministic **release on unmount** (plus a GC fallback);
- a **stable realized function identity** on the main thread, so main-thread code that retained the function keeps observing the latest implementation.

The proposal has two layers with deliberately different names:

- **Mechanism** (internal): the *callable transport* — a `_wcid`-branded handle, a main-thread registry of identity-stable wrapper functions, and an ordered per-commit ctx patch channel.
- **Public API** (for library authors): *main-thread event functions* — `useMainThreadEvent(fn)` for a single function and `useMainThreadEvents(value)` for values with nested functions (e.g. a Motion `transition` object containing an easing function array), in the direct lineage of React's `useEvent`/`useEffectEvent` (see "Naming" below).

This RFC follows the model laid out by the MainThreadObject RFC (#3446): it names the background representation an opaque handle, keeps wire fields internal, and separates identity, transport, and lifecycle. It does not depend on the MainThreadObject implementation (#3477) and composes with it.

### The concrete blockers

Two Motion conformance gaps are hard-blocked on this capability (both are adapter blockers, not Lynx engine gaps):

1. **Easing callbacks** (Huxpro/motion#37): upstream Motion accepts easing function arrays per keyframe segment:

   ```tsx
   <motion.view
     animate={{ x: [-42, 0, 42] }}
     transition={{
       duration: 0.8,
       ease: [
         (progress) => { 'main thread'; return progress },
         (progress) => { 'main thread'; return progress * progress },
       ],
     }}
   />
   ```

[...]

PR #3539, opened 2026-08-15
feat(react): add useMainThreadEvent — main-thread event functions over a callable transport

Huxpro:
Implements RFC #3538 (following the MainThreadObject RFC #3446 model): **main-thread event functions** — the main-thread counterpart of React's `useEvent`/`useEffectEvent` — built on a generic, lifecycle-safe **callable transport** that registers consumer-authored main-thread functions (including functions nested inside objects and arrays) as identity-stable main-thread callables. This is the upstream capability that unblocks the two Motion adapter conformance gaps Huxpro/motion#37 (easing function arrays) and Huxpro/motion#55 (`transformTemplate`).

## What this adds

Public API (`@lynx-js/react`):

- `useMainThreadEvent(fn)` — registers one main-thread function for the component lifetime as an event function: **stable identity** on the main thread, captured values **re-synchronized on every committed render**, **released on unmount** (with a GC-observer fallback). The `useEffectEvent` isomorphism is exact — the `useRef` cell becomes a main-thread registry slot, the `useInsertionEffect` assignment becomes a per-commit ctx patch, and the stable deref-and-call wrapper is the same wrapper — and every documented `useEffectEvent` constraint transfers, enforced structurally by the thread boundary (the handle is inert on the background thread; the main thread has no user-facing render phase). See the RFC's "Naming: the useEvent lineage" section for why the un-narrowed `useEvent` semantics are safe across the boundary.
- `useMainThreadEvents(value)` — deep transport: walks a value (e.g. a Motion `transition` object), replaces each nested main-thread function with a path-keyed handle, keeps slot identity across rerenders, and releases dropped slots and on unmount.
- `MainThreadEvent` — the opaque, serializable handle class.

Transport mechanics (mirrors the `MainThreadRef`/`_wvid` architecture; internal names keep the *callable transport* terminology and the wire format is `{ _wcid }`):

- New `_wcid` discriminant branch in `transformWorkletInner`, checked after the common discriminants so hot paths pay nothing; handles hydrate to callables at any nesting depth in worklet captures and parameters.
- A `_callableImpl` registry on `lynxWorkletImpl` (positive ids from the background thread, negative first-screen ids cleared at hydration finish, per-id stable wrapper functions).
- A new ordered ctx patch channel (`rLynxChangeCallableCtx`): **updates flush before the commit's patch update** (same-commit main-thread tasks observe fresh ctxs) and **releases flush after it** — the distributed form of `useEffectEvent`'s "callable from an unmounting Effect's cleanup" guarantee. An unmounting commit's tasks (e.g. Motion stopping an animation, which samples one final frame through the consumer easing) can still call the function; the declarative Motion conformance run below caught the release-before-stop ordering as a real bug.
- Staged ctxs go through `onPostWorkletCtx`, so nested `runOnBackground` handles keep their `_execId` lifecycle; installed ctxs are retained with `JsFunctionLifecycleManager` so replaced ctxs release their background exec contexts.

The RFC text is versioned at `docs/superpowers/specs/2026-08-15-main-thread-callable-transport-rfc.md`, including the binding decomposition of the #3446 algebra (binding writer / exposure / ownership) and a convergence section (compiler auto-workletization for end users; a type-directed `useMainThread(value)` transport for library authors; a shared slot substrate as a follow-up refactor).

## Verification that this unblocks Motion

1. **Runtime unit tests** (`packages/react/runtime/__test__`, 29 new tests): registry update/release, stable wrapper identity across ctx updates, first-screen negative ids, nested-handle hydration in arrays/captures/params, release event, `_jsFn` retention, release-after-patch ordering. 100% coverage maintained.
2. **Dual-runtime integration tests with the real compiler transform** (`packages/react/testing-library`, 4 new tests): an easing function array nested in a transition object arrives callable on the main thread with captured values updating across rerenders; `transformTemplate` produces `translateY(30px) translateX(30px)` (issue #55's acceptance string); a retained event function keeps a stable identity while observing fresh captures; event functions release on unmount.
3. **Motion engine tests** (`packages/motion`, 2 new tests): a consumer easing transported from the background thread drives `mini`'s real `animate` loop frame-by-frame (~84 main-thread invocations, every intermediate frame equal to the captured plateau value); a `transformTemplate` composes `translateY(30px) translateX(30px)` through a real animation loop.
4. **Declarative adapter proof on the #3509 stack** (not part of this PR's diff): on a branch merging this PR with #3509 plus an ~50-line adapter wiring — `useMainThreadEvents` around the three worklet transitions, `useMainThreadEvent` for a new public `transformTemplate` prop, and a template-composed `transform` derived value built with upstream `buildTransform`/`transformValue` — both conformance cases pass through the real `motion.view` path:
   - `transitions/easing-function-array` (Huxpro/motion#37): both segment callbacks execute; the element settles at `translateX(42px)`.
   - `targets/transform-template` (Huxpro/motion#55): the element renders `translateY(30px) translateX(30px)`.

   All 122 existing motion tests on that merged stack stay green; all 849 react runtime tests, 167 testing-library tests, and 107 motion tests on this branch pass.

## Known CodSpeed note

`004-various-update/main-thread.js_LoadScript` regresses ~-5% (~29µs) in Simulation mode: the one-time evaluation cost of the new runtime modules in the main-thread template (+~0.2KiB). The hot-path cost was removed by ordering the `_wcid` discriminant after the common ones; the remaining delta is the feature's irreducible footprint and needs a maintainer accept/acknowledge on the CodSpeed dashboard. The `web-platform` WallTime rows in the report are cross-environment measurement noise per CodSpeed's own warning (that package is untouched).

@coderabbitai summary

## Checklist

- [x] Documentation updated (or not required) — RFC in `docs/superpowers/specs/`, TSDoc on all public API, api-extractor report regenerated.

https://claude.ai/code/session_01DgHaTiHU1DWDAzsHdLd9GV


Huxpro:
On the CodSpeed check, for whoever reviews it:

- The one regression attributable to this PR was `004-various-update/main-thread.js_LoadScript` (-5% in both Simulation and WallTime on the first run): a combination of one-time script-eval growth from the new runtime modules (~+0.2KiB in the main-thread template per RelativeCI) and the new `_wcid` discriminant check sitting ahead of the much hotter `_wkltId` branch in `transformWorkletInner`. Commit 30c0114 reorders the callable check after the existing discriminants so worklet ctxs, refs, and background-function handles pay nothing for it — and that benchmark no longer regresses on the latest run.
- The remaining flagged entries (`basic-performance-text-200`, `basic-performance-small-css`) are WallTime-only results from `web-platform/web-core-e2e`, which this PR does not touch; CodSpeed's own report warns these were compared across different runtime environments on hosted runners, and the same suite flapped +33% in the other direction on the previous run. They look like measurement noise and probably just need acknowledging on the CodSpeed dashboard, which needs maintainer access.

---

Huxpro:
Correction to my previous comment — I read a truncated check summary: the `004-various-update/main-thread.js_LoadScript` Simulation regression **does persist** on 30c0114 (-5.14%, 536.4µs → 565.4µs). The discriminant reorder helps the event-dispatch hot path but cannot reduce this one, because LoadScript measures script evaluation: the ~29µs is the one-time cost of evaluating the new callable runtime modules in the main-thread template (+~0.2KiB per RelativeCI). That is the irreducible footprint of the feature short of lazy-loading the module, and reads to me like an accepted-cost acknowledgment call for a maintainer on the CodSpeed dashboard.

The `web-platform` WallTime rows remain cross-environment measurement noise per CodSpeed's own warning (this PR does not touch that package, and the same suite flapped +33%/+17% in the other direction across runs).

---

Huxpro:
## New commit: the primitive family over the same slot pool

`4d11046b4` adds two further write policies over the callable slot pool this PR introduces, each with its own RFC issue:

- **`useMainThreadInstance(create, dispose?)`** (#3540) — create-once main-thread objects with deterministic dispose, inheriting the release-after-patch rule. This is the transport-collapsed form of MainThreadObject's minimum contract (#3446): the `defineMainThreadObjectType` registry collapses into two workletized closures.
- **`useMainThreadEffect(fn, deps?)`** (#3541) — the main-thread counterpart of `useLayoutEffect`: ctx installed pre-patch, run flushed post-patch via one new message (`rLynxRunCallableCtx`), cleanup before next run and on release. Post-patch is exposed as a callback API deliberately; pre-patch is not (that restriction *is* `useMainThreadEvent`).
- **One new substrate rule** (#3542): releases tear down LIFO (reverse declaration order), destructor-style, so an effect cleanup may use an instance declared above it, whose dispose then runs after the cleanup.

The commit-cycle protocol is now: `ctx updates < patch < effect runs < releases`, asserted literally in `callableProperties.test.jsx`.

Verification: react/runtime 883 tests at 100% coverage, testing-library 171 (includes new real-transform e2e for instance realize/dispose and effect run/cleanup/LIFO), motion 107 — all green locally.

Scope note: this keeps the PR as one coherent "callable slot pool + its write policies" change, but the new commit is cleanly separable — happy to split `4d11046b4` into its own PR if reviewers prefer the event transport to land alone.

---

PR #3551, opened 2026-08-17
feat(react): add the background-only element

upupming:
## Summary

Add the `<background-only>` element, a subtree boundary that is resolved entirely at compile time:

```jsx
<view>
  <background-only fallback={<Skeleton />}>
    <Feed />
  </background-only>
</view>
```

The main thread renders `fallback` on the first screen, the background thread renders `children`, and the first-screen hydration replaces the fallback through the normal update patch. Because the boundary is folded at build time, the modules only `children` reach — including their top-level side effects — leave the main-thread bundle, while `fallback` stays as ordinary main-thread code and may itself contain user components, hooks and worklets.

The snapshot and worklet definitions the deferred subtree needs on the main thread are registered there by the definition merge from #3393, so folding the subtree away does not produce `Snapshot not found`.

## Context

Implements #3297 (component-level `'background only'` on the definition-merge infrastructure). Related RFCs: #3296 (root `<Background fallback>`), #3298 (the symmetric `<MainThread>` opt-in direction). Builds on #3393.

This supersedes @Huxpro's `<Background>` line of work — #3382 (and the closed #3305), which fold the same boundary on top of #3302. The semantics, the fold-to-`fallback` rewrite, the "no `fallback` renders nothing" degenerate case and the "a spread cannot be folded" diagnostic all come from that series; what changes here is the carrier and the direction of the fold.

### Why a built-in element rather than a component

The earlier design exposed the boundary as a component imported from `@lynx-js/react`, so the transform had to decide whether a given JSX element *is* the boundary by resolving its binding: named imports, aliases (`import { Background as B }`), namespace access (`<ReactLynx.Background>`). That decision cannot be made reliably in general — a re-export through a barrel file, a component passed as a value (`const B = cond ? Background : View`), or any wrapper defeats it, and each miss is silent: the subtree either stays in the main-thread bundle or disappears from it. A boundary that the compiler must fold has to be syntax, not a value. An intrinsic element name is matched lexically, cannot be aliased or passed around, and lets misuse be reported (`fallback` is the only attribute, spreads are a build error) instead of silently doing the wrong thing.

### Why the fold is target-independent

`<background-only>` is desugared into `__MAIN_THREAD__ ? fallback : <>children</>` **before** the snapshot backend runs, identically for both targets, and the branch is folded later by the existing `defineDCE` pass. Rewriting the element straight to `fallback` on the main-thread target only would make the two targets walk different trees, and the snapshot id counter — `__snapshot_<file>_<content>_<n>` — would drift apart, so the same id would name a different definition on each thread. That is safe when the main thread compiles no business code at all (`experimental_enableMTSRendering: false`), but not in the normal dual-thread mode this targets. Keeping the desugar target-independent keeps every id and definition identical across threads, which is what the definition merge relies on.

## Element templates

`<background-only>` was a build error under `experimental_useElementTemplate` while the deferred subtree's templates could not reach the main thread. #3563 fixed that by collecting element templates from the background build, so the restriction is gone and the pairing is covered by a test instead.

## Tests

- Rust unit tests for the desugar
- Transform integration tests: per-thread output, `fallback`-less and error cases
- Bundle-level test: the fallback stays in the main-thread bundle, the children's snapshot and worklet definitions are injected into it, the children's component body is gone
- Bundle-level test pinning what the boundary does to module side effects: the top-level side effects of a module only the children reach leave the main-thread bundle, while those of the module holding the boundary stay (that module renders the boundary, so it is evaluated on both threads)
- Bundle-level test under `experimental_useElementTemplate`: both the fallback and the deferred subtree's templates reach the encoded bundle. Negative-validated against #3563's collection.
- Runtime test with `@lynx-js/react/testing-library`: the fallback on the main-thread first screen, the children once the background thread takes over

## Summary by CodeRabbit

- **New Features**
  - Added the `<background-only>` element for rendering a fast main-thread fallback before displaying content in the background thread.
  - Supports fallback and child content, including nested elements and common JSX child patterns.
  - Added validation for unsupported or duplicate attributes.

- **Compatibility**
  - Updated React integrations to support `@lynx-js/react` version 0.125.

- **Documentation**
  - Added public JSX type definitions and usage guidance for `<background-only>`.


Huxpro:
Reviewing this as the author of the superseded `<Background>` line (#3302/#3382), so I checked the claim this PR makes about that work rather than take it on faith. It holds, and it is worse for my side than the PR states.

## The target-independence argument is a correctness fix, not a preference

The PR is charitable to #3302 — "that is safe when the main thread compiles no business code at all (`experimental_enableMTSRendering: false`)". It is not safe there either. It is safe only when the `fallback` and the deferred children never both contain host elements **in the same module**.

Reproduced through the transform, one file, both sides holding host elements:

```
LEPUS (folded): _1 = fallback   _2 = tail
JS    (full)  : _1 = fallback   _2 = real   _3 = tail
```

`__snapshot_<hash>_<file>_2` names `tail` on the main thread and `real` on the background thread. #3302's assembly then subtracts by `kind:id`, sees the main thread "owns" `_2`, and drops the background's `_2`. Built end to end on the same fixture:

| build | deferred element's definition in the main-thread chunk |
| --- | --- |
| `experimental_enableMTSRendering: true` | present |
| `experimental_enableMTSRendering: false` | **absent** |

At hydration the background asks for `_2` and the main thread builds `hdr`. Silent wrong tree, no diagnostic anywhere.

My matrix never caught it because the fixtures deliberately split `sk.jsx` / `real.jsx` into separate modules so the bundle question would be answerable — and snapshot ids are per-file, so the two counters never had to agree. The fixture-design decision I documented as rigor is exactly what hid it. Desugaring to `__MAIN_THREAD__ ? fallback : <>children</>` ahead of the snapshot backend is the right fix.

## Two corrections to the rationale

**"a re-export through a barrel file, a component passed as a value, or any wrapper defeats it" — the wrapper is not fixed here.**

Re-export and computed element type genuinely are: an intrinsic name cannot be re-exported or stored in a variable, so those shapes become unrepresentable. The wrapper is a different problem:

```jsx
function Deferred({ children }) {
  return <background-only fallback={<A/>}>{children}</background-only>;
}
<Deferred><Real/></Deferred>   // the caller still evaluates jsx(Real, {})
```

`children` is constructed at the call site, on the main thread, so `Real`'s module stays. Identical to `i04-wrapper` in the matrix, and #3382's own boundary analysis lists "child slotted through a wrapper that owns the boundary" as not eliminated. Changing the carrier fixes *identification*; the wrapper is about *where the reference lives*, which no carrier change reaches.

Worth correcting because `desugars_to_thread_conditional_in_expression_position` is precisely this shape — as written, the test set reads as endorsing a spelling that yields no bundle win.

Also: aliases and namespace access were not actually broken. `i01-aliased` and `i02-namespace` both drop the deferred code under import-binding tracking. The argument stands on re-export / wrapper / dynamic type; overstating it invites a reviewer to find the counterexample.

**The JSDoc contradicts itself and the tests.**

> Resolved at compile time, so the main-thread bundle keeps only `fallback`.

The main-thread bundle also keeps the children's element definitions — it has to, or hydration cannot build the real tree, which is what "injects the children snapshot into the main thread" asserts two files over. The next sentence ("Modules that nothing but `children` reference are dropped") is the accurate one. The boundary is **module-level, not markup-level**: JSX written inline in the boundary's own module still contributes its snapshot to the main-thread bundle.

## The one thing I would block on

Target-independence is the whole justification, and nothing pins it. Both transform fixtures use `fallback={<text>loading</text>}` with `<Feed/>` children — `Feed` is a component, so the deferred side contributes no snapshot in that file and the counters cannot diverge. Add a case with host elements on both sides asserting the id assignment agrees across targets; the repro above drops straight in.

## Questions

1. **`Optional::new(..., jsx_backend_enabled)`.** With `jsx_backend_enabled = use_snapshot_plugin || use_element_template_plugin`, a config with neither leaves `<background-only>` undesugared and it survives to runtime as `_jsx("background-only", …)`. Is that path reachable in a real build? If not, worth an assertion; if so, it should error rather than leak. (Line 667 nearby uses `jsx_backend_enabled && !preserve_jsx` — I read the omission of `preserve_jsx` here as deliberate and correct, but flagging it.)

2. **Which mechanism supplies the children's snapshot?** `define_dce` runs *after* `snapshot`, so the main thread sees both branches and generates the children's definitions itself, and `snapshotCreatorMap[id] = …` is a property assignment a tree-shaker will not remove. So: children in a separate module (your tests) — the fold makes the import unused, the module is shaken, and #3393's merge brings the definition back. Children inline in the boundary's own module — the definition simply stays, and the merge is not involved. The PR credits #3393 for both; worth separating, otherwise "what breaks if #3393 is off" has no answer.

3. **`fallback` coverage.** The PR states the fallback "may itself contain user components, hooks and worklets" — that was the headline capability of the superseded design — but every fixture uses a bare `<text>`. At minimum one case with a component + hook in the fallback.

4. **Multiple children.** `desugars_to_thread_conditional_in_child_position` has two, but only as a desugar snapshot. Main thread puts one node in the slot, background puts a Fragment; I expect the dynamic part treats the Fragment as a single vnode and this is fine, but there is no end-to-end case.

## Nits

- `children_to_expr` wraps even a single child in a Fragment — one extra vnode per boundary on the background thread.
- Whitespace-only children (`<background-only fallback={<A/>}>\n</background-only>`) make `children.is_empty()` false, producing `<></>` rather than the `null` a self-closing boundary gets.
- `take_fallback` reports the spread error and then folds anyway; the build fails regardless, but an early return would be tidier.
- `Ident::new_no_ctxt("__MAIN_THREAD__")` will capture a user's local binding of that name. Consistent with `defineDCE` elsewhere, but it is a hygiene hole.
- `fallback?: React.JSX.Element` is narrower than what the transform accepts — `fallback={cond ? <A/> : <B/>}` works but does not type.
- Three things now named background-only: the element (subtree boundary), the `'background only'` directive (function body), and `import 'background-only'` (module assertion). Defensible as a family, but the element and the package assert different properties.

## Verdict

Approve the direction. Syntax-not-value is the right call and the target-independent desugar is a genuine correctness improvement over what I built — on the merits I would take this over #3302's fold. It supersedes that line cleanly, and #3302 needs either the same desugar or a content-addressed subtraction key before it goes anywhere regardless.

---

packages/react/transform/src/lib.rs
```diff
@@ -612,6 +613,11 @@ fn transform_react_lynx_inner(
     let enable_ui_source_map =
       !use_element_template_plugin && snapshot_plugin_config.enable_ui_source_map.unwrap_or(false);
 
+    let background_only_plugin = Optional::new(
+      visit_mut_pass(BackgroundOnlyVisitor::new()),
+      jsx_backend_enabled,
```
Huxpro, line 618: `jsx_backend_enabled` is `use_snapshot_plugin || use_element_template_plugin`, so this also runs under `experimental_useElementTemplate` — but nothing here tests that combination.

The desugar puts a `CondExpr` in a JSX child position, and the element-template backend wants shapes it can serialize statically. #3382 rejected the pairing outright rather than guess. Right now this is the third option: it runs, and nobody knows what comes out.

Either add an element-template case to `background-only.spec.js`, or gate on `use_snapshot_plugin` and make `<background-only>` a build error under element templates. Silently producing something is the one outcome the rest of this design goes out of its way to avoid.

---

packages/react/transform/src/lib.rs
```diff
@@ -612,6 +613,11 @@ fn transform_react_lynx_inner(
     let enable_ui_source_map =
       !use_element_template_plugin && snapshot_plugin_config.enable_ui_source_map.unwrap_or(false);
 
+    let background_only_plugin = Optional::new(
```
Huxpro, line 616: The desugar is unconditional, and `__MAIN_THREAD__` is a define in development too, so the fold happens in dev as well — which means editing a module that only a `<background-only>` subtree reaches will not hot-update the main thread.

#3382 deliberately kept development on the runtime path (`if (this.hot) return false` in the loader's fold gate) so every module stays refreshable and only production pays the boundary. That is a real ergonomic difference for anyone iterating on deferred code.

If it is intentional — the fold is target-independent here, so dev and prod at least agree with each other — a line in the PR description or the TSDoc saying so is enough. If not, the loader gate is the place to put it.

---

packages/react/runtime/jsx-runtime/index.d.ts
```diff
@@ -22,5 +22,38 @@ export namespace JSX {
     key?: React.Key | null | undefined;
   }
   interface IntrinsicClassAttributes<T> extends React.JSX.IntrinsicClassAttributes<T> {}
-  interface IntrinsicElements extends Lynx.IntrinsicElements {}
+  interface IntrinsicElements extends Lynx.IntrinsicElements {
+    /**
+     * Renders `fallback` on the main thread and `children` on the background thread.
+     *
+     * Resolved at compile time, so the main-thread bundle keeps only `fallback`.
```
Huxpro: "the main-thread bundle keeps only `fallback`" is not quite what happens, and the PR's own snapshot shows it: in `renders nothing on the main thread without a fallback`, the LEPUS output still carries `__snapshot_a94a8_test_2` with the full `__CreateRawText("background")` creator — the deferred child's element definition, compiled into the main thread.

That is correct and necessary (hydration has to build those elements on the main thread), but it means what leaves is the **module closure**, not the element definitions. The next sentence gets this exactly right — "Modules that nothing but `children` reference are dropped from it, including their top-level side effects" — so the first sentence is the only one overclaiming.

Suggest something like: "Resolved at compile time, so the main thread renders only `fallback`. Modules that nothing but `children` reference are dropped from its bundle, including their top-level side effects; the element definitions `children` need are still registered there for hydration."

Worth fixing now — as written it reads like a size guarantee, and someone will file a bug when they find a deferred subtree's definitions in the first-frame chunk.

---

packages/react/runtime/jsx-runtime/index.d.ts
```diff
@@ -22,5 +22,38 @@ export namespace JSX {
...
+     * <view>
+     *   <background-only fallback={<Skeleton />}>
+     *     <Feed />
+     *   </background-only>
+     * </view>
+     * ```
+     *
+     * @public
+     */
+    'background-only': {
```
Huxpro, line 52: Naming, and it is the kind of thing that is free now and impossible later: `background-only` takes a bare name out of the same namespace as platform host elements. Lynx element names are open — if any platform ever ships an element called `background-only`, it is shadowed by a compiler construct, and by then there will be apps depending on both meanings.

Everything else about this element is deliberately unspoofable: matched lexically, cannot be aliased, cannot be passed as a value. A reserved prefix would extend that to the name itself — `lynx:background-only`, `lynx-background-only`, whatever fits the conventions here — and would also make it obvious at the call site that this is not a real element.

Not blocking, but it is much cheaper to decide before this ships than after.

---

packages/react/transform/crates/swc_plugin_background_only/lib.rs
```diff
@@ -0,0 +1,319 @@
...
+}
+
+fn is_background_only(el: &JSXElement) -> bool {
+  matches!(&el.opening.name, JSXElementName::Ident(ident) if ident.sym.as_ref() == "background-only")
+}
+
+fn take_fallback(el: &mut JSXElement) -> Option<Box<Expr>> {
+  let mut fallback = None;
+  let mut seen_fallback = false;
+  for attr in el.opening.attrs.drain(..) {
```
Huxpro, line 26: Minor / non-blocking: `drain(..)` empties the attribute list even on the paths that only emit a diagnostic, so after an error the element still desugars with the offending attributes discarded. Since any emitted error fails the build, nothing downstream sees the result — but it does mean a spread is silently dropped rather than preserved, so if diagnostics ever become recoverable (or a `--no-fail-fast`-style mode appears) this recovers into something the author did not write.

Retaining unconsumed attributes on the error path would be more obviously correct, at the cost of a slightly less tidy loop. Entirely your call.

---

Huxpro (commented):
I wrote the `<Background>` series this supersedes (#3382, #3305), so I went after the two load-bearing claims in the "why" section rather than take them on faith. **Both hold, and the id one holds harder than the PR states — it is not a hazard this design avoids, it is a live bug in the design it replaces.**

## The snapshot-id argument is correct, and #3302's fold trips it today

Ids are `__snapshot_{filename_hash}_{content_hash}_{counter}` with a monotonic per-file counter (`swc_plugin_snapshot/lib.rs:1393-1398`), so only position distinguishes them. One boundary does not drift — the fold deletes the id rather than renumbering its neighbours. **Two boundaries in one file do.** Same source, `foldBackgroundToFallback` the only variable:

```
MAIN THREAD, fold ON   :: _2 -> FALLBACK1 | _3 -> FALLBACK2
MAIN THREAD, fold OFF  :: _2 -> FALLBACK1 | _3 -> DEFERRED1 | _4 -> FALLBACK2 | _5 -> DEFERRED2
BACKGROUND             :: _1 .. _5
```

`_3` is `FALLBACK2` on the main thread and `DEFERRED1` on the background. Built end to end through the definition merge on my #3382 branch:

| id | classic | folded |
|---|---|---|
| `_2` | FALLBACK1 | FALLBACK1 |
| `_3` | **DEFERRED1** | **FALLBACK2** |
| `_4` | FALLBACK2 | FALLBACK2 |
| `_5` | DEFERRED2 | DEFERRED2 |

`DEFERRED1`'s definition is **absent from the main-thread bundle entirely** — the subtraction sees `_3` as already owned by the main thread and drops the background's `_3`, while `FALLBACK2` ends up stored twice under two ids. Hydration then builds the first deferred subtree from the second fallback's definition. Silently wrong, no diagnostic. Reproduces under `experimental_enableMTSRendering: false` too, so it belongs to #3302's fold, not to my extension of it.

Worth saying plainly because it is the strongest argument in this PR and the description undersells it: keeping the desugar target-independent is not a tidiness preference, it is what makes the ids mean the same thing on both threads. I will report this against #3302 separately.

It also exposes a hole in my own test: `owned ⊎ assembled == classic` compares id **strings**, so it passes with the wrong definitions behind them. Any port of that check should compare what an id denotes, not that it exists.

## The syntax-not-a-value argument also holds

This matches what #3302's own matrix measured: `i03` (re-export) and `i05` (computed element type) were real, silent detection misses. A lexically matched intrinsic element removes that class by construction, which binding resolution cannot.

One thing to state explicitly so readers do not over-read it: **this does not fix a wrapper that slots children in from its caller.** Your own `desugars_to_thread_conditional_in_expression_position` is exactly that shape, and the caller's `<App />` reference still keeps `App` in the main-thread graph. "cannot be aliased or passed around" is true of the boundary, not of the deferred subtree. The theoretical limit is unchanged — a fold can only delete a reference it can see, in the module where the boundary is written.

## Two commits pushed

Took you up on the offer to just fix the small ones directly:

- **`ab57235`** — a repeated `fallback` was silently taking the last one, i.e. compiling a first screen the author did not write. Now an error, consistent with how spreads and unknown attributes are already treated.
- **`92b408e`** — `children_to_expr` always built a fragment, so the common single-child boundary compiled to `_jsx(_Fragment, { children: _jsx(Feed, {}) })` and paid a vnode per background render for nothing. A lone child now goes straight into the conditional; text and spread children still get the fragment, as do multi-child boundaries. Whitespace spanning a line break and `{/* comments */}` do not count towards the child total, so ordinary formatting does not defeat it.

Rust tests 7/7, `background-only.spec.js` 6/6, full transform suite 94 passed. Snapshots regenerated; the background output loses its `_Fragment` import. `index.d.ts` was left alone — the napi build rewrites it wholesale, so it needs restoring after any local rebuild.

## Worth carrying over from the series this replaces

1. **The boundary map.** #3302's `bgmatrix` (22 cases × 3 settings) and #3382's `backgroundIslandsBoundary.test.ts` (13 fixture modules) pin *where the fold stops*: a module the shell also imports, a child slotted through a wrapper, a component hoisted into a module-scope binding. The limits here are identical, and two bundle fixtures do not record them. Undocumented limits get rediscovered as bug reports.
2. **`import 'background-only'`.** The repo already resolves it to a failing module on the main-thread layer. It is what turns the leaks above from silent dead weight into a named build error. Nothing in this PR uses it.

Neither is a reason to hold this — the design is better than what it replaces and I would land it and close #3382 in its favour. They are the parts of that work worth salvaging.

---

packages/react/transform/src/lib.rs
upupming: Gated, and thank you for pushing on it — it was not just untested, it was broken.

I built it end to end: an element-template case whose `<background-only>` children live in their own module, asserting on `encodeData.elementTemplate` in `beforeEncode`. The encoded template map came back with the fallback and without the deferred subtree:

```
keys:         _et_27e47289cc89, _et_d7c09654e316, _et_builtin_raw_text
hasFallback:  true
hasDeferred:  false
```

The desugar itself survives the element-template backend — ET ids are content hashes (`_et_<hash>`), not the positional counter, so nothing drifts and the conditional folds fine. What is missing is the other half: #3393 merges the definitions the main thread lacks for the snapshot backend only, and there is no equivalent for element templates. So the deferred subtree would have no template to build from when the background hands over — the ET analogue of `Snapshot not found`, at runtime, on device.

`<background-only>` is now a build error under `experimental_useElementTemplate`, with a test pinning the diagnostic. Merging element templates across threads is the feature that would lift it, and that is its own PR.

packages/react/transform/src/lib.rs
upupming: Intentional, and now written down — in the TSDoc rather than only the PR, since that is where someone hits it:

> The fold is the same in development, so what renders there is what renders in production, at the cost of the deferred modules not being refreshable on the main thread.

The reasoning is the one you named: the fold here is target-independent by construction, so gating it on `this.hot` would make development render a first screen that production never renders — the deferred subtree would be live on the main thread in dev and absent in prod. For a boundary whose whole purpose is to change what the first screen is, I would rather the two agree and pay the refresh cost than have dev quietly disagree with what ships.

Worth noting the cost is narrower than it looks: the main thread never renders those modules in either mode, so what is lost is main-thread refresh for code that only ever runs on the background thread. The background side stays refreshable.

packages/react/runtime/jsx-runtime/index.d.ts
upupming: Correct, and I had verified exactly that output earlier without noticing my own sentence contradicted it. Reworded:

> Resolved at compile time, so the main thread renders only `fallback`. Modules that nothing but `children` reference leave its bundle, including their top-level side effects, while the element definitions the deferred subtree needs stay behind so it can hydrate.

The second clause is now explicit rather than implied by the next sentence, since "the definitions stay, the modules leave" is the part that is easy to get backwards.

packages/react/transform/crates/swc_plugin_background_only/lib.rs
upupming: Taken — the loop now keeps every attribute it refuses. Spreads, unknown attributes, a non-JSX `fallback` value and (on top of your later commit) a repeated `fallback` all get pushed back onto the element instead of being dropped on the way out.

You are right that nothing observes it today because any emitted error fails the build. I applied it anyway for the reason you gave: the recovery path should reconstruct what the author wrote, not a tidied-up version of it, so that it stays correct if diagnostics ever become recoverable.

upupming: (also picked up your two commits on the branch — the repeated-`fallback` rejection and the lone-child unwrap — and rebased onto them; the repeated-`fallback` path now retains the attribute too, for the same reason as above.)

packages/react/transform/__test__/background-only.spec.js
```diff
@@ -0,0 +1,217 @@
...
+      `,
+      options('LEPUS'),
+    );
+
+    expect(result.errors.map((error) => error.text)).toMatchInlineSnapshot(`
+      [
+        "<background-only> accepts only one \`fallback\` attribute",
+      ]
+    `);
+  });
```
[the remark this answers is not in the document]
upupming: This finding does not match the code it points at: lines 180-199 are the unknown-attribute test, and there was no "first `fallback` is retained" behavior to protect — a repeated `fallback` used to resolve to the *last* one, matching how JSX handles repeated attributes.

That behavior is gone either way. A repeated `fallback` is now a build error, so nothing is silently chosen, and the test at line 180 asserts the diagnostic.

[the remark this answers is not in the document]
upupming: Fair on the substance, and done — the finding was right that only the diagnostic was asserted.

The recovery is observable, so it can be pinned: emitting a diagnostic does not stop the transform, so `result.code` is still produced (655 chars for this input). The test now asserts both halves — the error, and that the surviving branch is the first `fallback` and not the second.

upupming:
@Huxpro heads-up on a force-push: your two commits on this branch were blocking the CLA check.

`license/cla` was failing with *1 out of 2 committers have signed* — the unsigned one being `claude`, because both commits were authored by `Claude <noreply@anthropic.com>`. It is not specific to this branch: #3382 and #3302 fail the same check today, #3546 passes because its commits carry your own identity, and no commit authored by `Claude` has ever landed on main.

So I re-authored those two to `Huxpro <5563315+Huxpro@users.noreply.github.com>`. Only the author metadata changed — `git diff` between the old and new tips is empty, the messages including your trailers are untouched, and the commits stay yours rather than being folded into mine. CLA now reports all committers signed.

If you have either commit checked out locally, they are now `10c8ae2` (repeated `fallback`) and `85659e3` (lone-child unwrap); the pre-rewrite tip was `cc800c9` if you need it.

Both behaviours are kept as you wrote them. I only extended the repeated-`fallback` path to push the rejected attribute back onto the element, for the same recovery-fidelity reason as the other error paths.

packages/react/transform/src/lib.rs
upupming: Lifted — #3563 landed, so the restriction this thread produced is gone.

Verified before removing the gate rather than assuming the fix covered it: with `main` merged in, an element-template build of a `<background-only>` whose children live in their own module now encodes both templates.

```
keys:        _et_20d292f64bb2, _et_d7c09654e316, _et_builtin_raw_text, _et_c63f175491fe
hasFallback: true
hasDeferred: true      // was false before #3563
```

The build error and its test are gone, and `element-template/background-only` covers the pairing instead. It is negative-validated against the mechanism it depends on: removing the background collection #3563 added makes this case fail, and only it and `merge-templates`.

gaoachao (approved):

HuJean (approved):

colinaaa (approved):

PR #3553, opened 2026-08-17
refactor(react): reach lynx-core's app object through lynx.getApp()

upupming:
## Summary

The runtime reached lynx-core's app object through the `lynxCoreInject` global that the AMD wrapper injects. `lynx.getApp()` returns that same instance and has existed since the first open-source commit, so every reference now goes through it: the app-level callback assignments in `injectTt` and the element-template `init`, `setupLynxEnv`'s init-data params, the before-publish hook, the global-props emitter, the lazy-bundle exports and the ReactLynx 2 class-component compat.

Resolving through `lynx` also stays correct once several cards share a runtime chunk, where the injected global is whichever card evaluated it first.

The global's type declaration is replaced by a `LynxApp` interface returned from `lynx.getApp()`, so `lynxCoreInject` is now absent from the package including its tests.

No behavior change: the same object receives the same assignments under the same names.

### Notes for reviewers

- `lynx.getApp()` is not part of `@lynx-js/types`, so this PR declares it in the runtime's own `types/types.d.ts` — the same place the `lynxCoreInject` shape was declared. Both accesses are equally untyped upstream; this one at least goes through `lynx`.
- It has been a public constructor member of lynx-core's `Lynx` class since the first open-source commit (`@lynx-js/lynx-core@0.1.1`), and the class needs it to implement `setTimeout` and friends. I could not check pre-open-source 2.x cores from the repo, so if anyone knows of a supported engine without it, say so — a missing `getApp` would throw at `injectTt` rather than fail quietly.
- `@lynx-js/testing-environment` gains `lynx.getApp()`; anything that reaches into `lynxTestingEnv.backgroundThread.lynxCoreInject.tt` keeps working, since both point at the same object.

### Left alone deliberately

`lynxCoreInject` also names the parameter the AMD wrapper passes to a bundle, which is a different thing from the global. Those stay: `runtime-wrapper-webpack-plugin` (which defines the wrapper), `web-platform/web-core` (which implements it for the web), the `react/transform` extract-str output, and the REPL examples that demonstrate the protocol. `plugin-debug-metadata`'s release banner also stays, because its `lynxCoreInject` branch is the fallback for engines older than 2.13 — older than `getApp` itself.

## Verification

`packages/react/runtime` suites all green (770 + 49 + 692) with the 100% coverage gate satisfied.

Verified on a physical Android device (Pixel 5) with the `examples/react` app: the eight callbacks land on `lynx.getApp()` under the engine's own names, and first screen, the post-hydration `publishEvent` swap, a tap reaching the background thread and card destruction all behave as before.

### DIFF
--- a/.changeset/app-object-via-get-app.md
+++ b/.changeset/app-object-via-get-app.md
@@ -1,8 +1,7 @@
 ---
-"@lynx-js/react-runtime": patch
+"@lynx-js/react": patch
 "@lynx-js/cache-events-webpack-plugin": patch
 "@lynx-js/testing-environment": patch
-"@lynx-js/reactlynx-testing-library": patch
 ---
 
 Reach lynx-core's app object through `lynx.getApp()` instead of the
--- a/.changeset/app-object-via-get-app.md
+++ b/.changeset/app-object-via-get-app.md
@@ -1,0 +1,12 @@
+---
+"@lynx-js/react-runtime": patch
+"@lynx-js/cache-events-webpack-plugin": patch
+"@lynx-js/testing-environment": patch
+"@lynx-js/reactlynx-testing-library": patch
+---
+
+Reach lynx-core's app object through `lynx.getApp()` instead of the
+`lynxCoreInject` global the AMD wrapper injects. It is the same instance, so
+behavior is unchanged, and resolving it through `lynx` also stays correct once
+several cards share a runtime chunk. `@lynx-js/testing-environment` now exposes
+`lynx.getApp()` alongside the object it already provided.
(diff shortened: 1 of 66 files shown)


upupming:
Marked `callBeforePublishEvent` optional in a68a1a9 — only `ReactApp` defines it, so the optional chaining in `tt.ts` was right and the type was not. Left `_params` required as suggested.

@coderabbitai review

Yradex (approved):

HuJean (approved):

colinaaa (approved):

PR #3558, opened 2026-08-17
perf(react): omit profile hooks from default Web bundles

Huxpro:
## Stack dependency

This PR is stacked on #3546. Review the top commit as the reachability-only delta. Once #3546 lands, this branch can be rebased onto main without changing the implementation contract.

## Summary

- Make exhaustive ReactLynx component profiling hooks unreachable from default production Web bundles.
- Keep full hooks for explicit Web profiling and all Lynx artifacts, preserving native host-driven production recording.
- Scope module replacement to the linked React runtime root so similarly named user modules are untouched.
- Treat REACT_PROFILE=false as an explicit false value.

## Performance evidence

In the same callback-fixed production app:

| Artifact | Before | After | Delta |
| --- | ---: | ---: | ---: |
| Web raw | 97,862 B | 94,532 B | -3.40% |
| Web gzip | 33,030 B | 31,903 B | -3.41% |
| Lynx raw | 99,367 B | 99,367 B | exact control |

Default Web contains none of the tested profiling markers. Explicit Web profile and both default/profile Lynx artifacts retain them.

## Compatibility

- Explicit performance.profile keeps Web profiling.
- Lynx keeps hooks for native recording.
- Direct-webpack fallback keeps hooks when the plugin cannot prove the linked worklet runtime root.
- Module replacement is restricted to React runtime paths, not lookalike user modules.

## Validation

- React webpack plugin artifact tests: 18/18.
- Runtime focused tests: 12/12.
- Production TypeScript config, dprint, and diff checks passed.
- Real default/profile Web/Lynx four-state artifact probe passed.
- A full local Turbo retry is environmentally blocked before code execution by pnpm 11's read-only SQLite store; CI should provide the authoritative full-repository result.

## Checklist

## Summary by CodeRabbit

* **New Features**
  * Added configurable profiling for ReactLynx component hooks.
  * Profiling can be enabled through `performance.profile` or native host recording.
  * Web profiling now produces User Timing traces when explicitly enabled.
  * Production Web builds exclude component-hook profiling by default to reduce overhead.

* **Documentation**
  * Updated profiling configuration guidance for Web and native environments.

* **Bug Fixes**
  * Prevented browser performance API availability from incorrectly activating profiling.


Huxpro:
Stacked on #3546. Please review commit b3877351630a055db9532f0c05f94199bde56da4 as the reachability-only delta; I will rebase after #3546 lands.

PR #3562, opened 2026-08-17
refactor: move the CSS minimizer into @lynx-js/rsbuild-plugin

upupming:
Moves the CSS minimizer registration from `applyDefaultPlugins` into `pluginLynx()`, together with its dependency.

Rspeedy replaces Rsbuild v2's built-in LightningCSS minimizer with the cssnano-based one from `@rsbuild/plugin-css-minimizer`, because several LightningCSS transforms are equivalent on the Web but produce CSS that Lynx cannot handle. That is a Lynx constraint rather than a Rspeedy product decision, so it belongs to the engine. Today a raw `rsbuild` + `pluginLynx` user keeps the LightningCSS minimizer and hits those transforms.

The registration itself is unchanged, including the `isPluginExists` guard that lets a user supply their own configured `pluginCssMinimizer`. The only difference is how it is expressed: the plugin API has `isPluginExists` but no `addPlugins`, so the guarded registration runs the upstream plugin's own `setup` instead of adding it to the instance. No behaviour or option handling is reimplemented here.

`@rsbuild/plugin-css-minimizer` moves from `@lynx-js/rspeedy` dependencies to `@lynx-js/rsbuild-plugin`, and stays as a devDependency of `@lynx-js/rspeedy` because its tests import it directly.

Verified: both test projects pass unchanged, including the existing `custom css minimizer` test in `@lynx-js/rspeedy` that pins the user-supplied-plugin behaviour, and `@examples/react` `main.lynx.bundle` is byte-identical to `main`.

## Summary by CodeRabbit

* **New Features**
  * CSS minimization is now included automatically through the Lynx plugin.
  * Existing CSS minimizer configurations are preserved to avoid duplicate processing.

* **Bug Fixes**
  * Improved CSS optimization setup for more consistent production builds.

* **Chores**
  * Updated release metadata for patch versions of the affected packages.

### DIFF
--- a/.changeset/move-css-minimizer.md
+++ b/.changeset/move-css-minimizer.md
@@ -1,0 +1,6 @@
+---
+"@lynx-js/rsbuild-plugin": patch
+"@lynx-js/rspeedy": patch
+---
+
+Move the cssnano-based CSS minimizer into `pluginLynx()`.
--- a/packages/rspeedy/core/package.json
+++ b/packages/rspeedy/core/package.json
@@ -56,12 +56,12 @@
     "@lynx-js/debug-metadata-rsbuild-plugin": "workspace:^",
     "@lynx-js/rsbuild-plugin": "workspace:*",
     "@rsbuild/core": "catalog:rsbuild",
-    "@rsbuild/plugin-css-minimizer": "2.0.1",
     "@rsdoctor/rspack-plugin": "~1.6.1"
   },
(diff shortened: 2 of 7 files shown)


upupming:
@coderabbitai review

upupming:
@coderabbitai review

colinaaa (approved):

PR #3563, opened 2026-08-17
fix(webpack): register element templates collected from the background build

upupming:
## Summary

Element templates were only ever collected from the main-thread build: `lynx:element-templates` is written by the main-thread loader alone, so `beforeEncode` could only see templates belonging to modules the main-thread module graph reaches. Whatever the background renders but the main thread never imports had no template in the encoded bundle, and building those elements when the background hands over its first patch had nothing to build from — the element-template counterpart of the `Snapshot not found` that #3393 fixed for the snapshot backend.

The background loader now stores its templates the same way, so the existing per-entrypoint collection picks them up. Merging stays id-keyed and is safe by construction: element-template ids are content hashes, so the same id carries the same template on both threads, and the collision check in `mergeElementTemplate` keeps that assumption honest.

Reaching this needs no new API — it is the same shape as the definition merge in #3393, one layer lower.

## Reproducing it without the fix

The gap is not tied to any one feature. A plain thread conditional is enough:

```jsx
<view>
  {__MAIN_THREAD__ ? <text text='main-thread-shell' /> : <Deferred />}
</view>
```

`<Deferred />`'s module leaves the main-thread bundle, and with it its template. The same happens for a subtree dropped by `sideEffects` shaking. `merge-templates` pins this: it fails on `main` and passes here.

## Tests

- `element-template/merge-templates` — a deferred subtree's template reaches the encoded bundle. Negative-validated: reverting the loader change fails this case and only this case.
- `element-template/merge-templates-lazy` — a lazy component's template still does not leak into the main bundle. The collection is scoped per entrypoint chunk group, and this pins that the extra background modules do not widen it.
- The existing `element-template/basic` case asserts an exact template count and is unchanged, which is the evidence that the two threads produce byte-identical templates for a shared id rather than colliding.

## Not covered

A lazy bundle's own encode currently receives an empty template map — its chunk groups are unnamed, and the collector keys on entrypoint names. That predates this change and is untouched by it; element templates inside lazy bundles need their own fix.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved element-template generation so content from deferred and lazily loaded components is correctly collected and merged into the appropriate output.
  * Prevented lazy-only content from appearing in the main template when it should remain separate.

* **Tests**
  * Added coverage for merged templates, lazy loading, deferred content, and background builds.

* **Chores**
  * Documented the change in the upcoming patch release.

### DIFF
--- a/.changeset/element-template-merge.md
+++ b/.changeset/element-template-merge.md
@@ -1,0 +1,5 @@
+---
+"@lynx-js/react-webpack-plugin": patch
+---
+
+Collect element templates from the background build as well, so a template the main-thread bundle does not reach is still registered for it.
--- a/packages/webpack/react-webpack-plugin/src/loaders/background.ts
+++ b/packages/webpack/react-webpack-plugin/src/loaders/background.ts
@@ -9,6 +9,7 @@ import {
   DEFINES_FOR_SNAPSHOT_BUILD_INFO,
   DEFINES_FOR_WORKLET_BUILD_INFO,
 } from '../Defines.js';
+import { ELEMENT_TEMPLATE_BUILD_INFO } from './main-thread.js';
 import { getBackgroundTransformOptions } from './options.js';
 import type { ReactLoaderOptions } from './options.js';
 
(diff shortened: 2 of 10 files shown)


upupming:
Related: #3393 did this for the snapshot backend (`definesForSnapshot` / `definesForWorklet` collected from the background build and registered on the main thread). This is the same gap one layer lower, for element templates.

It also unblocks a limitation in #3551: `<background-only>` is a build error there under `experimental_useElementTemplate` precisely because the deferred subtree lost its templates. Once this lands that restriction can be lifted, which I would rather do as a follow-up than fold into either PR.

Yradex (dismissed):

upupming:
Both nitpicks applied — the second one was a real hole, not a style point.

**Exact merged set.** `merge-templates` now asserts three user templates rather than only that two markers appear. That is the assertion that actually pins the content-hash contract: the `<view>` wrapper is compiled by both threads, and it stays one entry instead of two because the id is `sha256` of the canonical compiled template (`template_identity.rs:27`). A duplicate registration would produce a different id and push the count to four.

**Vacuous conditional.** Correct, and worth more than a nitpick: `if (isMainTemplate)` meant the case passed if `main__main-thread` never reached `beforeEncode` at all. It now records that it was observed and asserts it in `afterEmit`. Negative-validated by pointing the predicate at a chunk-group name that never matches — the case fails, where before it passed silently.

Yradex (approved):

colinaaa (approved):

PR #3569, opened 2026-08-18
refactor: move the htmlPlugin default into @lynx-js/rsbuild-plugin

upupming:
Lynx does not use HTML, so `tools.htmlPlugin` is disabled. That default lived in `toRsbuildConfig`, which means a raw `rsbuild` + `pluginLynx` build emits a `main.html` next to its assets. It is a Lynx platform fact rather than a Rspeedy option, so it moves into `pluginLynx()`.

Measured before and after on a raw `rsbuild` + `pluginLynx` build with a plain JS entry:

```
- dist/.rspeedy/main/main.css
- dist/static/js/main.<hash>.js
- dist/main.html          <- gone after this change
```

`tools.htmlPlugin` is not part of the Rspeedy config schema, so no Rspeedy user could set it and the behaviour there is unchanged. In the engine it sits next to the existing `output.environment.const` default, in the merge position that lets a raw Rsbuild user opt back in; verified that setting `tools.htmlPlugin: true` still resolves to `true`.

Verified: both test projects pass, both packages type-check, and `@examples/react` `main.lynx.bundle` is byte-identical to `main`.

## Summary by CodeRabbit

* **Changes**
  * The HTML plugin default is now managed by the Lynx plugin.
  * Generated Lynx builds continue to avoid incompatible modern JavaScript syntax in output.
  * Existing behavior is preserved while configuration is simplified.

* **Bug Fixes**
  * Prevented unintended HTML processing during Lynx builds.

* **Tests**
  * Added coverage confirming the HTML plugin is disabled by default in normalized configuration.

### DIFF
--- a/.changeset/move-html-plugin.md
+++ b/.changeset/move-html-plugin.md
@@ -1,0 +1,6 @@
+---
+"@lynx-js/rsbuild-plugin": patch
+"@lynx-js/rspeedy": patch
+---
+
+Move the `tools.htmlPlugin` default into `pluginLynx()`.
--- a/packages/rspeedy/core/src/config/rsbuild/index.ts
+++ b/packages/rspeedy/core/src/config/rsbuild/index.ts
@@ -134,8 +134,6 @@ export function toRsbuildConfig(
 
       cssLoader: config.tools?.cssLoader,
 
-      htmlPlugin: false,
-
       rspack: config.tools?.rspack,
 
       swc: config.tools?.swc,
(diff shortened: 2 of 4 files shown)


upupming:
@coderabbitai review

upupming:
@coderabbitai review

upupming:
@coderabbitai review

upupming:
@coderabbitai review

colinaaa (approved):

PR #3570, opened 2026-08-18
refactor: read the bundle filename from the Rsbuild config

upupming:
Stacked on #3569.

`applyEntry` resolves the Lynx template filename from the Rspeedy config, reached through `api.useExposed(Symbol.for('rspeedy.api'))`. That is one of the reasons the template pipeline only runs under Rspeedy. The same value already travels through the Rsbuild config as `output.filename.bundle` / `template`, so it can be read from there instead.

One gap had to be closed first: `toRsbuildConfig` dropped the string form of `output.filename`, because only the object form was forwarded. The string form carries the same `bundle`/`template` meaning, so it is now resolved with the existing `getFilename` helper rather than dropped. Only the string case changes; `undefined` stays `undefined` and the object form is still forwarded as-is.

No behaviour change under Rspeedy: for every input shape the resolved template filename is the same as before.

| `output.filename` | resolved bundle name |
| --- | --- |
| unset | `[name].[platform].bundle` |
| `'x.bundle'` | `x.bundle` |
| `{ bundle: 'y.bundle' }` | `y.bundle` |

Verified: `@lynx-js/rspeedy`, `@lynx-js/rsbuild-plugin` and `@lynx-js/react-rsbuild-plugin` test suites pass, both packages type-check, and `@examples/react` `main.lynx.bundle` is byte-identical to `main`.

## Summary by CodeRabbit

* **Bug Fixes**
  * Bundle and template filenames are now correctly read from the Rsbuild output configuration.
  * String-based filename settings are applied consistently to generated bundle and template files.
  * Existing object-based filename configurations continue to work unchanged.

* **Documentation**
  * Added release notes describing the updated filename configuration behavior.

### DIFF
--- a/.changeset/read-bundle-filename-from-rsbuild-config.md
+++ b/.changeset/read-bundle-filename-from-rsbuild-config.md
@@ -1,0 +1,6 @@
+---
+"@lynx-js/react-rsbuild-plugin": patch
+"@lynx-js/rspeedy": patch
+---
+
+Read the bundle filename from the Rsbuild config instead of the Rspeedy API.
--- a/packages/rspeedy/core/src/config/defaults.ts
+++ b/packages/rspeedy/core/src/config/defaults.ts
@@ -65,7 +65,7 @@ function getEnableChunkSplitting(config: Config): boolean {
 
 const DEFAULT_FILENAME = '[name].[platform].bundle'
 
-function getFilename(filename: string | Filename | undefined): Filename {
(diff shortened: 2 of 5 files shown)


upupming:
@coderabbitai review

upupming:
@coderabbitai review

upupming:
@coderabbitai review

colinaaa (approved):

colinaaa (approved):

luhc228 (approved):

PR #3572, opened 2026-08-18
refactor: emit the Lynx template for non-Rspeedy callers

upupming:
Stacked on #3570.

`applyEntry` sets up the whole Lynx template pipeline (`LynxTemplatePlugin`, the runtime wrapper, `LynxEncodePlugin`, the web plugin) inside a block gated on `callerName === 'rspeedy'`. That is why building with plain Rsbuild and `pluginReactLynx` produces JavaScript chunks but no `.bundle`.

Nothing inside the block needs Rspeedy any more: its last Rspeedy read, the bundle filename, moved to the Rsbuild config in #3570. Only the gate was left, so it is inverted from "opt in for Rspeedy" to "opt out for the callers that must not emit a template". `rslib` builds libraries and `rstest` runs tests; both already have dedicated branches in `pluginReactLynx` and neither should produce a Lynx template. The only caller newly included is plain `rsbuild`.

Measured on `@examples/react`, built once through Rspeedy and once through plain Rsbuild with `pluginLynx()` + `pluginReactLynx()` over the same source:

| | `main.lynx.bundle` |
| --- | --- |
| Rspeedy | 94824 bytes |
| Rsbuild, before this change | not emitted |
| Rsbuild, after this change | 94875 bytes |

The remaining 51 bytes are a `/*! LICENSE: ... */` banner, because `output.legalComments: 'none'` is still applied by `toRsbuildConfig` rather than by the engine. That is the next step.

Verified: `@lynx-js/react-rsbuild-plugin` tests pass, `@lynx-js/react-umd` (the `rslib` caller) still builds, and `@examples/react` `main.lynx.bundle` under Rspeedy is byte-identical to `main`.

## Summary by CodeRabbit

- **New Features**
  - Lynx templates are now generated for supported callers beyond Rspeedy.
  - Template generation remains disabled for Rslib and Rstest workflows.
  - This provides more consistent template availability across supported build workflows.

- **Documentation**
  - Added release notes describing the expanded template generation behavior.

PR #3574, opened 2026-08-18
refactor: move the legalComments default into @lynx-js/rsbuild-plugin

upupming:
Stacked on #3572.

Lynx encodes everything into a single template, so a linked license file has nowhere to point at. `output.legalComments` is therefore forced to `none`, but that default lives in `toRsbuildConfig`, which means a raw Rsbuild build keeps Rsbuild's own `linked` default and emits a `/*! LICENSE: ... */` banner. It is a Lynx platform fact, so it moves into `pluginLynx()`.

The user value still flows through, because `output.legalComments` is a native Rsbuild option: the engine reads `getRsbuildConfig('original')` and only supplies the default when the user did not set one, the same pattern the plugin already uses for `distPath.css` and `filename.css`.

Measured on `@examples/react`, same source built through Rspeedy and through plain Rsbuild with `pluginLynx()` + `pluginReactLynx()`:

| | `main.lynx.bundle` |
| --- | --- |
| Rspeedy | 94824 bytes |
| Rsbuild, before this change | 94875 bytes |
| Rsbuild, after this change | 94824 bytes |

The two files are now the same size, and the only remaining difference is 64 bytes of `background.<contenthash>.js` naming. The embedded JavaScript is identical, so the hash is fed by a config value that has not moved yet; the normalized configs still differ on `output.sourceMap.css`, `output.inlineScripts`, `output.dataUriLimit`, `output.cssModules.localIdentName` and the `output.filename.bundle` default. Those are the next steps.

Verified: `@lynx-js/rspeedy` and `@lynx-js/rsbuild-plugin` tests pass, and `@examples/react` `main.lynx.bundle` under Rspeedy is byte-identical to `main`.

## Summary by CodeRabbit

* **Bug Fixes**
  * Standardized the default handling of legal comments during builds.
  * Legal comments are now omitted by default while preserving any explicitly configured setting.
* **Chores**
  * Updated release metadata for the affected build packages.

### DIFF
--- a/.changeset/move-legal-comments.md
+++ b/.changeset/move-legal-comments.md
@@ -1,0 +1,6 @@
+---
+"@lynx-js/rsbuild-plugin": patch
+"@lynx-js/rspeedy": patch
+---
+
+Move the `output.legalComments` default into `pluginLynx()`.
--- a/packages/rspeedy/core/src/config/rsbuild/index.ts
+++ b/packages/rspeedy/core/src/config/rsbuild/index.ts
@@ -60,9 +60,7 @@ export function toRsbuildConfig(
 
       inlineScripts: config.output?.inlineScripts,
 
-      // TODO(OSS): change the default value to `linked`(or `undefined`) when OSS.
-      // We expect to use different default legalComments with Rsbuild
-      legalComments: config.output?.legalComments ?? 'none',
+      legalComments: config.output?.legalComments,
 
       minify: config.output?.minify,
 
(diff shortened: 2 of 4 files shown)


upupming:
@coderabbitai review

upupming:
@coderabbitai review

colinaaa (approved):

luhc228 (dismissed):

colinaaa (dismissed):

luhc228 (approved):

PR #3577, opened 2026-08-18
fix(rsbuild-plugin): alias the @rspack/core/hot subpaths the dev client imports

upupming:
## The bug

`@lynx-js/webpack-dev-transport` imports `@rspack/core/hot/log.js` and
`@rspack/core/hot/log-apply-result.js` (added in #3434), but `@rspack/core` is
only a `devDependency` of that package. `pluginDev` compensates by aliasing the
`@rspack/core/hot/*` subpaths it knows about, and those two were never added, so
a **development-mode** build fails with `Module not found`. Production builds
are unaffected.

It passed on pnpm only because pnpm's hidden hoisted store
(`node_modules/.pnpm/node_modules`) is on the resolution path of anything inside
`.pnpm/`, so the phantom dependency resolved by accident. `test-publish` only
ever installed with pnpm, so CI never saw it.

Verified with the same generated project on each package manager:

| install | dev build |
| --- | --- |
| npm | fails |
| yarn 4, `nodeLinker: node-modules` | fails |
| pnpm | passes |

## Changes

- **`@lynx-js/rsbuild-plugin`** -- alias both missing subpaths in `pluginDev`.
- **`@lynx-js/rsbuild-plugin`** tests -- the alias assertion no longer hardcodes
  a list. It reads the dev-transport client sources and asserts every
  `@rspack/core/hot/*` specifier they import is aliased, so the next added
  import cannot regress silently. Confirmed it fails without the fix.
- **`test-publish`** -- the repeated smoke-test blocks are now one function with
  a package-manager dimension, adding npm and yarn 4 next to the existing pnpm
  runs. The npm variant reproduces this bug and passes with the fix.

Two Yarn settings are needed for the mock registry: `enableImmutableInstalls`
(Yarn enables it whenever `CI` is set, but the projects are generated with no
lockfile) and `npmMinimalAgeGate` (Yarn quarantines any version published less
than `1d` ago, and the canary versions reach the registry seconds earlier).

## Not included

Running the same smoke test under pnpm with `hoist: false` surfaces several
pre-existing packaging problems unrelated to this fix. That is kept separate in
#3581 rather than bundled here.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed development-build resolution issues for Rspack hot-log modules when dependencies are not hoisted.
  * Added aliases to ensure required hot-module imports resolve correctly.

* **Tests**
  * Expanded publish smoke tests to cover npm, pnpm, and Yarn workflows.
  * Added validation that Rspack hot-module imports have corresponding aliases.

* **Chores**
  * Simplified dependency installation and script argument handling in publish workflows.

### DIFF
--- a/.changeset/rspack-hot-subpath-alias.md
+++ b/.changeset/rspack-hot-subpath-alias.md
@@ -1,0 +1,5 @@
+---
+"@lynx-js/rsbuild-plugin": patch
+---
+
+Alias `@rspack/core/hot/log.js` and `@rspack/core/hot/log-apply-result.js`, which `@lynx-js/webpack-dev-transport` imports without declaring `@rspack/core`. Development builds failed to resolve them on any package manager that does not hoist `@rspack/core`, such as npm and Yarn.
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -220,32 +220,70 @@ jobs:
         fi
         node packages/tools/canary-release/snapshot.js
         pnpm --recursive publish --no-git-checks --access public --registry=http://localhost:4873
-        cd `mktemp -d`
-        npx --registry http://localhost:4873 create-rspeedy-canary@latest --template react --dir create-rspeedy-regression --tools eslint --skill lynx-devtool
-        cd create-rspeedy-regression
-        npx --registry http://localhost:4873 upgrade-rspeedy-canary@latest
-        pnpm install --registry=http://localhost:4873
-        pnpm run build
-        pnpm run build --mode development
-        pnpm run lint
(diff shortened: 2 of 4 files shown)


upupming:
@coderabbitai review

upupming:
Broadened the alias guard as suggested: it now walks `clientDir` recursively, accepts `.ts/.tsx/.mts/.cts`, and matches `from`, bare side-effect `import`, and dynamic `import()` under either quote style. `declare module` is deliberately still excluded -- a type shim needs no alias.

Verified the guard still fails when the aliases are removed.

colinaaa (approved):

PR #3580, opened 2026-08-18
refactor(react): unify typed Element Template attributes

Yradex:
## Summary by CodeRabbit

* **New Features**
  * Typed list elements now preserve and apply classes, event handlers, and refs during rendering.
  * Generic typed elements now retain list metadata and serialize attributes correctly.
  * Attribute updates remain consistent when hydration handles are rebound.

* **Bug Fixes**
  * Improved ref callback behavior and event-handler mapping after hydration.
  * Ensured typed-list attributes are available during initial creation and subsequent updates.

## Overview

Typed Element Template attributes currently take different preparation paths depending on where an instance is materialized. In particular, background typed instances prepare slot-0 attributes before insertion, while the main-thread typed list path can forward raw attributes. This makes attribute, event, and ref behavior depend on the rendering layer.

This PR establishes one shared typed-attribute path before adding authored `<page />` support in later stacked changes:

- defines the fixed slot-0 spread plan and attribute preparation in `runtime/template/typed-attributes.ts`;
- makes `BackgroundTypedElementTemplateInstance` own the slot-0 invariant, raw/prepared attributes, and their event/ref lifecycle;
- reserves the main-thread handle before preparing attributes, then composes list-owned callbacks after preparation so internal callbacks are not filtered as user spread values;
- makes main-thread insertion and patching address typed attributes through the same shared slot constant instead of local literals;
- covers empty and populated slots, initial materialization, updates, hydration handle rebinding, event replacement, and ref attachment.

This layer does not add `<page />` behavior, change the public API or native schema, or introduce compatibility fallbacks. The subsequent PRs in the stack build page support on this common typed-instance contract.

## Review order

1. `runtime/template/typed-attributes.ts`
2. `background/attr-slots.ts` and `background/instance.ts`
3. `runtime/render/render-opcodes.ts` and `runtime/render/render-main-thread.ts`
4. `runtime/patch.ts`
5. Focused background and render tests

## Checklist

Documentation and a changeset are not required for this internal, unreleased Element Template refactor.

### DIFF
--- a/packages/react/runtime/__test__/element-template/runtime/render/render-opcodes-into-element-template.et.test.tsx
+++ b/packages/react/runtime/__test__/element-template/runtime/render/render-opcodes-into-element-template.et.test.tsx
@@ -91,7 +91,12 @@ describe('renderOpcodesIntoElementTemplate', () => {
   it('creates exact list through typed native create with slot-0 refs as listChildren', () => {
     const itemRef = { kind: 'item-ref' };
     const listRef = { kind: 'list-ref' };
-    const attributes = { id: 'typed-list' };
+    const handler = vi.fn();
+    const attributes = {
+      bindtap: handler,
+      className: 'feed',
+      id: 'typed-list',
+    };
     createElementTemplate.mockReturnValueOnce(itemRef);
     createTypedElementTemplate.mockReturnValueOnce(listRef);
 
@@ -123,6 +128,8 @@ describe('renderOpcodesIntoElementTemplate', () => {
     const typedCreateCall = createTypedElementTemplate.mock.calls[0]!;
     expect(typedCreateCall[0]).toBe('list');
     expect(typedCreateCall[1]).toEqual({
+      bindtap: '-2:0:bindtap',
+      class: 'feed',
       id: 'typed-list',
       'component-at-index': expect.any(Function),
       'component-at-indexes': expect.any(Function),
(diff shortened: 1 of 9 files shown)


packages/react/runtime/__test__/element-template/runtime/render/render-opcodes-into-element-template.et.test.tsx
```diff
@@ -91,7 +91,12 @@ describe('renderOpcodesIntoElementTemplate', () => {
   it('creates exact list through typed native create with slot-0 refs as listChildren', () => {
     const itemRef = { kind: 'item-ref' };
     const listRef = { kind: 'list-ref' };
-    const attributes = { id: 'typed-list' };
+    const handler = vi.fn();
+    const attributes = {
+      bindtap: handler,
+      className: 'feed',
+      id: 'typed-list',
```
upupming, line 98: How about ref

upupming (approved):

packages/react/runtime/__test__/element-template/runtime/render/render-opcodes-into-element-template.et.test.tsx
Yradex: Ref is already covered by the shared spread adapter's native-create test, the typed-instance ref attach test, and the typed-list hydration rebinding test. This case uses `bindtap` and `className` to verify that typed-list materialization goes through the same `adaptSpreadAttrSlot` path, so adding `ref` here would duplicate the adapter coverage.

packages/react/runtime/src/element-template/background/instance.ts
```diff
@@ -492,16 +518,10 @@ export class BackgroundElementTemplateInstance {
...
-  if (value == null || typeof value !== 'object' || Array.isArray(value)) {
-    return null;
-  }
-  return value as TypedElementAttributesCommand;
-}
-
 export class BackgroundTypedElementTemplateInstance extends BackgroundElementTemplateInstance {
   constructor(type: string) {
     super(type);
+    this.attributeSlots = [null];
```
HuJean, line 524: why need null for index 0

HuJean (approved):

PR #3583, opened 2026-08-18
feat(react): add ET authored page background runtime

Yradex:
## Overview

This PR adds the background-runtime layer for authored Element Template `<page />` support on top of the shared typed-attribute foundation.

- registers the existing logical page root in the background instance manager at reserved handle `0`;
- routes page attributes, events, and refs through the ordinary typed attribute-slot pipeline;
- adds the transparent internal Page helper that writes `attributes`, returns `$0`, and clears attributes on unmount;
- retains one narrow lifetime token so a keyed replacement stale passive cleanup cannot clear its successor;
- preserves page-root ownership across init, destroy, reload, and hydration setup.

This layer does not add main-thread page instructions, the serialized page-root hydrate envelope, or authored JSX lowering. Those are separate later PRs in the stack.

## Review order

1. `background/manager.ts` and the reserved page-root registration
2. `background/instance.ts` and the thin `BackgroundPageRootInstance`
3. `runtime/page/authored-page.ts`
4. init/reload integration
5. focused manager, instance, root-render, and authored-page tests

## Checklist

Documentation is maintained in the parent project Track. A changeset is not required for this internal, unreleased Element Template layer.

## Summary by CodeRabbit

* **New Features**
  * Added authored page support with configurable attributes, children, and keys.
  * Added consistent page-root handling across rendering, hydration, reloads, and native integration.
  * Preserved page attributes across updates and hydration while safely removing outdated attributes.

* **Bug Fixes**
  * Improved page-root protection during cleanup and invalid-handle handling.
  * Fixed page-root registration and replacement during reloads and hydration.

* **Tests**
  * Expanded coverage for page rendering, attribute updates, hydration, reloads, teardown, and authored-page behavior.

### DIFF
--- a/packages/react/runtime/src/element-template/background/manager.ts
+++ b/packages/react/runtime/src/element-template/background/manager.ts
@@ -3,17 +3,19 @@
 // LICENSE file in the root directory of this source tree.
 
 import type { BackgroundElementTemplateInstance } from './instance.js';
+import { ELEMENT_TEMPLATE_PAGE_HANDLE_ID } from '../protocol/page.js';
 
 export const backgroundElementTemplateInstanceManager: {
   nextId: number;
   values: Map<number, BackgroundElementTemplateInstance>;
   register(instance: BackgroundElementTemplateInstance): void;
+  registerPageRoot(instance: BackgroundElementTemplateInstance): void;
   updateId(oldId: number, newId: number): void;
   get(id: number): BackgroundElementTemplateInstance | undefined;
   getRawAttributeValueByEventValue(eventValue: string): unknown;
   clear(): void;
 } = {
-  nextId: 0,
+  nextId: ELEMENT_TEMPLATE_PAGE_HANDLE_ID,
   values: new Map<number, BackgroundElementTemplateInstance>(),
 
   register(instance: BackgroundElementTemplateInstance): void {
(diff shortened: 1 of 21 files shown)


packages/react/runtime/src/element-template/background/manager.ts
```diff
@@ -3,26 +3,34 @@
...
 
   register(instance: BackgroundElementTemplateInstance): void {
     instance.instanceId = this.nextId += 1;
     this.values.set(instance.instanceId, instance);
   },
 
+  registerPageRoot(instance: BackgroundElementTemplateInstance): void {
+    this.values.delete(instance.instanceId);
+    instance.instanceId = ELEMENT_TEMPLATE_PAGE_HANDLE_ID;
+    this.values.set(ELEMENT_TEMPLATE_PAGE_HANDLE_ID, instance);
```
upupming, line 29: What if multiple <page /> elements are used, shall we throw?

Yradex: If you mean using multiple `<page />`s, I think yes. But it's a little hard to check that. I could try to implement that in the next PR.

packages/react/runtime/src/element-template/runtime/page/authored-page.ts
```diff
@@ -0,0 +1,37 @@
...
+  props,
+): ComponentChild {
+  const lifetime = useRef<object>({});
+  const root = __root as unknown as BackgroundAuthoredPageRoot;
+
+  if (__BACKGROUND__) {
+    root.setAuthoredPageAttributes(lifetime.current, props.attributes ?? null);
+  }
+
+  useEffect(() => () => root.clearAuthoredPageAttributes(lifetime.current), []);
```
HuJean: Only need in background？

Yradex: Yes. setAuthoredPageAttributes is only needed on background, where the helper updates the logical BackgroundPageRootInstance. On the main thread, renderToOpcodes handles __ElementTemplatePage directly and consumes props.attributes to emit __OpPageStart. The cleanup is background-only as well because the main-thread useEffect implementation is a no-op.

HuJean (approved):

PR #3584, opened 2026-08-18
feat(react): render and hydrate through the typed page root

Yradex:
## Overview

This PR adds the main-thread render, physical page commit, serialization, and background hydration layer for authored Element Template `<page />`.

- emits dedicated PageStart/PageEnd render instructions carrying prepared page attributes;
- materializes all opcodes before mutating the physical page, then applies page attributes and slot-0 roots;
- serializes the typed physical page as the single hydrate envelope;
- hydrates the reserved background page root and reconciles attributes only after child matching succeeds;
- keeps target-`0` patching and reload cleanup at the existing page/native boundary.

This layer builds on the background page-root owner. It does not yet lower authored `<page />` JSX; the transform and compiled fixtures are the next PR.

## Review order

1. `runtime/render/render-to-opcodes.ts`
2. `runtime/render/render-opcodes.ts`
3. `runtime/render/render-main-thread.ts`
4. `background/page-root-hydrate.ts`
5. reload/patch and focused protocol tests

## Checklist

Documentation is maintained in the parent project Track. A changeset is not required for this internal, unreleased Element Template layer.

## Summary by CodeRabbit

* **New Features**
  * Added page-level serialization for rendering and hydration.
  * Added support for authored page attributes and page boundaries.
  * Improved handling of multiple roots and typed page references.
  * Added support for string-based element template identifiers.

* **Bug Fixes**
  * Improved hydration of page roots, including missing slots and attributes.
  * Prevented page updates when child hydration or rendering fails.
  * Improved reload cleanup and attribute reconciliation.

* **Tests**
  * Expanded coverage for page rendering, hydration, reloads, attributes, and opcode validation.

### DIFF
--- a/packages/react/runtime/src/element-template/native/reload-main-thread.ts
+++ b/packages/react/runtime/src/element-template/native/reload-main-thread.ts
@@ -5,10 +5,12 @@
 import { applyUpdatePageData } from '../../core/lynx-page-data.js';
 import { increaseReloadVersion } from '../../core/reload-version.js';
 import { profileEnd, profileStart } from '../debug/profile.js';
+import { ELEMENT_TEMPLATE_PAGE_ROOT_SLOT_INDEX } from '../protocol/page.js';
+import type { SerializedPageRoot } from '../protocol/types.js';
 import { destroyAllElementTemplateListStates } from '../runtime/list/list.js';
 import { __page } from '../runtime/page/page.js';
 import { __root, setRoot } from '../runtime/page/root-instance.js';
-import { removeMainThreadRootRefs, renderMainThread } from '../runtime/render/render-main-thread.js';
+import { renderMainThread } from '../runtime/render/render-main-thread.js';
 import { resetTemplateId } from '../runtime/template/handle.js';
 import { resetElementTemplateMainThreadBackgroundFunctionRuntime } from '../runtime/template/main-thread-background-function.js';
 import { clearMainThreadDynamicAttrState } from '../runtime/template/main-thread-dynamic-attr-state.js';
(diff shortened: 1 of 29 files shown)


packages/react/runtime/src/element-template/native/reload-main-thread.ts
```diff
@@ -24,13 +26,19 @@ export function reloadMainThread(data: unknown, options: UpdatePageOption): void
     applyUpdatePageData(data, options);
 
     destroyAllElementTemplateListStates();
+    // TODO: Replace this cleanup-only serialization with a direct page-children
+    // or clear-slot PAPI once native exposes one.
+    const page = __SerializeElementTemplate(__page) as SerializedPageRoot;
+    for (const root of page.elementSlots?.[ELEMENT_TEMPLATE_PAGE_ROOT_SLOT_INDEX] ?? []) {
+      const rootRef = elementTemplateRegistry.get(root.uid as number)!;
```
upupming, line 33: can uid be string?

packages/react/runtime/src/element-template/runtime/render/render-opcodes.ts
```diff
@@ -204,6 +214,28 @@ export function renderOpcodesIntoElementTemplate(
         i += 3;
         break;
       }
+      case __OpPageStart: {
+        if (__DEV__ && stackTop !== 0) {
+          throw new Error('Element Template authored <page /> must be the outermost element.');
```
upupming, line 219: Is this limitation too strict?

packages/react/runtime/src/element-template/native/reload-main-thread.ts
Yradex: No. Later We will change its type to `number` as well.

packages/react/runtime/src/element-template/runtime/render/render-opcodes.ts
Yradex: Why, is using `<page />` inside other elements supported now?

upupming: yes

Yradex: why? I thought it would be illegal

upupming: Agree, page should be the top element

HuJean (approved):

packages/react/runtime/src/element-template/runtime/render/render-opcodes.ts
```diff
@@ -204,6 +214,28 @@ export function renderOpcodesIntoElementTemplate(
...
+        isInsideAuthoredPage = true;
+        i += 2;
+        break;
+      }
+      case __OpPageEnd: {
+        if (__DEV__) {
+          isInsideAuthoredPage = false;
+        }
+        i += 1;
+        break;
```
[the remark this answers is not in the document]
Yradex: Thanks for the suggestion. I don't think we should add these checks here.

`renderOpcodesIntoElementTemplate` only consumes opcodes produced by `renderToString` in the runtime path. In development, the producer emits `__OpPageEnd` only after the page subtree has finished rendering, and `renderMainThread` discards the entire opcode array if opcode production throws. The existing page checks cover user-authored semantic errors (nested/multiple pages and roots outside the page), while unmatched, premature, or missing page-end instructions would require the internal producer/consumer contract itself to be broken.

The producer test already verifies that a development page sequence ends with `__OpPageEnd`. Adding parser branches and tests for manually constructed impossible sequences would be redundant defensive validation, so I prefer to keep the current implementation.

PR #3596, opened 2026-08-19
refactor: move the debug metadata plugin into @lynx-js/rsbuild-plugin

upupming:
Part of the series that moves the Lynx build engine out of `@lynx-js/rspeedy` and into `pluginLynx`.

`pluginLynxDebugMetadata` was the last engine plugin still applied from Rspeedy's own default plugin list. Applying it from `pluginLynx` gives an Rsbuild build the same debug metadata as an Rspeedy build.

Stacked on #3578.

## Summary by CodeRabbit

- **Bug Fixes**
  - Debug metadata support is now included through the Lynx plugin.
  - Existing Rspeedy configurations continue to receive debug metadata without requiring separate setup.

- **Chores**
  - Updated package release metadata to reflect the change.

### DIFF
--- a/.changeset/engine-debug-metadata.md
+++ b/.changeset/engine-debug-metadata.md
@@ -1,0 +1,6 @@
+---
+"@lynx-js/rsbuild-plugin": patch
+"@lynx-js/rspeedy": patch
+---
+
+Move the debug metadata plugin from `@lynx-js/rspeedy` into `pluginLynx`.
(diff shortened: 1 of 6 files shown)


upupming:
The CodSpeed check is measurement noise, not a regression from this PR. It flags `basic-performance-div-100` in `packages/web-platform/web-core-e2e` at 1.4 ms -> 1.5 ms, measured with the WallTime instrument on a standard hosted runner, which CodSpeed's own report warns gives inconsistent data. This PR only moves where a plugin is applied and does not touch the web platform. The PR stacked on top of this one flags a different benchmark in the same suite while only adding a test file. `Done` passes on both.

upupming:
@coderabbitai review

luhc228 (dismissed):

luhc228 (approved):

PR #3597, opened 2026-08-19
test: cover Rsbuild and Rspeedy bundle parity

upupming:
Part of the series that moves the Lynx build engine out of `@lynx-js/rspeedy` and into `pluginLynx`.

The point of the series is that an Rsbuild build with only `pluginReactLynx()` produces the same bundle as an Rspeedy build. This locks that in: the same fixture is built both ways, each in its own process, in production and in development, and the two bundles are compared.

The comparison ignores identifiers minted per build rather than derived from the source. `chunk.hash` is not reproducible across builds, and both the content hash in a filename and the debug metadata release are keyed on it, so the filename hash is turned off and the release is stripped.

That reproducibility problem is not ours: swc's `inline_globals` keeps the span of the value it substitutes for a define, and those values come from a process-wide cache, so their spans belong to a `SourceMap` from an earlier call. The module's source map therefore moves between builds, which moves `build_info.hash`, `chunk.hash` and the content hash. Fix proposed upstream in swc-project/swc#12129; with it, 40 Rspeedy builds of the same source produce one bundle instead of ten. Once that lands the two allowances in this test can go away.

Stacked on #3596.

## Summary by CodeRabbit

* **Tests**
  * Added coverage to verify that development and production bundles generated by Rsbuild and Rspeedy remain equivalent.
  * Normalized build-specific metadata during comparisons for reliable results.
  * Isolated builds and cleaned up temporary output after testing.

* **Chores**
  * Added a parity build utility using consistent React, environment, output, and deterministic build settings.
  * Documented the parity coverage in the upcoming patch release notes.

### DIFF
--- a/.changeset/rsbuild-rspeedy-parity.md
+++ b/.changeset/rsbuild-rspeedy-parity.md
@@ -1,0 +1,5 @@
+---
+"@lynx-js/react-rsbuild-plugin": patch
+---
+
+Cover that an Rsbuild build and an Rspeedy build produce the same bundle.
--- a/.changeset/rsbuild-rspeedy-parity.md
+++ b/.changeset/rsbuild-rspeedy-parity.md
@@ -1,5 +1,0 @@
----
-"@lynx-js/react-rsbuild-plugin": patch
----
-
-Cover that an Rsbuild build and an Rspeedy build produce the same bundle.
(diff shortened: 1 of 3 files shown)


upupming:
The CodSpeed check is measurement noise, not a regression from this PR. It flags `basic-performance-div-1000` in `packages/web-platform/web-core-e2e` at 7.4 ms -> 8.8 ms, measured with the WallTime instrument on a standard hosted runner, which CodSpeed's own report warns gives inconsistent data. This PR adds a test file and a changeset and cannot affect a browser benchmark; the PR below it in the stack flags a different benchmark in the same suite. `Done` passes on both.

upupming:
@coderabbitai review

upupming:
Good catch — applied, and to the Rsbuild branch too, since `build()` returns `{ close }` on both. Each fixture build now closes before the process exits.

luhc228 (approved):

PR #3598, opened 2026-08-19
feat(genui): support A2UI v1.0 core protocol

PupilTong:
## Summary

- add a lean A2UI v1.0 core path across the ReactLynx client and GenUI server
- support `createSurface`, `updateComponents`, `updateDataModel`, and `deleteSurface`, including inline surface initialization and v1 renderer actions
- share the server protocol schema across parsing and validation, normalize v1 function calls, and keep streaming catalog-aware
- preserve v0.9 client read compatibility while updating the playground, CLI template, package exports, documentation, and changeset

## Scope

This intentionally implements one active catalog and the core rendering/action flow only. Renderer RPC, capability negotiation, per-component catalog routing, and v1 themes remain out of scope.

## Validation

- `pnpm turbo build` — 71/71 tasks
- A2UI client tests — 144/144
- GenUI server tests — 152/152
- Playground tests — 148/148
- `pnpm turbo api-extractor -- --local` — 25/25 tasks
- dprint, scoped Biome (390 files), scoped ESLint, changeset status, and `git diff --check`


PupilTong:
in flavor of 3933

PR #3601, opened 2026-08-20
fix(react/transform): stop sharing parsed `defineDCE` values across transforms

upupming:
Production builds of unchanged sources are not reproducible: the same module can come out with a different chunk hash between two builds, and with minification on, every mangled name in the chunk moves with it.

## Cause

`defineDCE.define` values go to swc's `GlobalPassOption::build`, which parses each value into an anonymous file of the `SourceMap` it is handed and memoizes the parsed expression in a **process-wide** cache. A span only means anything to the `SourceMap` that produced it, so every later transform in the process substitutes values carrying positions of a `SourceMap` that is already gone.

What comes out depends on where that stale position lands in the file currently being compiled:

- the transform that parses the value emits mappings into the anonymous file (the define text itself is reported as a source);
- a transform that hits the cache emits nothing for it, or a mapping to an unrelated line if the file it is compiling is long enough.

The generated code is identical either way — only the mappings move. But `SourceMapDevToolPlugin` puts the map on the module, the map is part of the module's build hash in Rspack, so the chunk hash moves, and the minifier orders its short names by character frequency over the whole chunk, so one changed string permutes every mangled name.

Which of those happens is not stable across builds: the cache key is order sensitive and `define` is a `HashMap`, and the cached position is derived from the size of whichever module the process transformed first.

Full write-up and the measurements behind it: https://github.com/swc-project/swc/pull/12129

## Fix

Build the `globals` / `global_exprs` maps in `transform_react_lynx_inner` and drop the spans of the parsed values, instead of going through `GlobalPassOption::build`. Nothing is shared between transforms, and a substituted value can no longer carry a position into the source map of the file being compiled. Values are still parsed against the caller's `SourceMap`, so a syntax error in a define value keeps its position.

Parsing per transform instead of once per process costs a handful of tiny parses per file, next to nothing against a full transform.

The upstream fix (swc-project/swc#12129) does the same thing inside `GlobalPassOption::build`, but this does not need to wait for it, and stays correct after it lands.

## Tests

The first commit adds `__test__/sourcemap.spec.js`, which transforms the same input twice and compares the maps. It fails there — same code, different mappings:

```
- "mappings":"AAAA,OAAO,MAAM,OAAb,KAA2C;AAC3C,QAAQ,GAAG,CAAC"
+ "mappings":"AAAA,OAAO,MAAM,YAA8B;AAC3C,QAAQ,GAAG,CAAC"
```

The define name in that test is deliberately not shared with any other test: the cache is per process, so another test using the same define set would warm it up first and the test would pass for the wrong reason.

Verified locally on both backends — `pnpm vitest run` (wasm) and `USE_NAPI=1 pnpm vitest run`: 89 passed, plus `cargo test` for the crates.

## Checklist

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved source map stability when transforming identical React Lynx input repeatedly.
  * Production builds now produce more reproducible output by avoiding cross-build cache interference.
  * Invalid define values are now reported consistently during transformation.

* **Tests**
  * Added regression coverage to verify identical generated code and source maps across repeated transformations.

### DIFF
--- a/.changeset/stable-define-source-maps.md
+++ b/.changeset/stable-define-source-maps.md
@@ -1,0 +1,5 @@
+---
+"@lynx-js/react": patch
+---
+
+Build the `defineDCE` value map without swc's process-wide cache, so the source map of a module no longer depends on which module was transformed before it. It made production builds non-reproducible: the source map is part of the module hash in Rspack, so an unchanged source could produce a different chunk hash, and with minification on, different mangled names.
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2715,7 +2715,6 @@ dependencies = [
  "convert_case 0.11.0",
  "getrandom 0.4.3",
  "hex",
- "indexmap",
  "napi",
  "napi-build",
  "napi-derive",
--- a/packages/react/transform/Cargo.toml
+++ b/packages/react/transform/Cargo.toml
@@ -10,14 +10,13 @@ crate-type = ["cdylib"]
 # Default enable napi4 feature, see https://nodejs.org/api/n-api.html#node-api-version-matrix
 convert_case = { workspace = true }
 hex = { workspace = true }
-indexmap = { workspace = true }
 napi = { workspace = true, features = ["serde-json"] }
 napi-derive = { workspace = true }
 rustc-hash = { workspace = true }
 serde = { workspace = true, features = ["derive"] }
 serde_json = { workspace = true, features = ["preserve_order"] }
 sha-1 = { workspace = true }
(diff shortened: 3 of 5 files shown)


upupming:
Parked as a draft: the same bug is being fixed upstream in swc-project/swc#12129, and duplicating `GlobalPassOption::build` here without its cache is not worth carrying.

Keeping the branch around because the first commit is still the proof: once `swc_core` is bumped to a version containing the upstream fix, `__test__/sourcemap.spec.js` passes without the second commit, and it stays as a regression test.

PR #3610, opened 2026-08-20
feat(react): transport element template ref init values

Yradex:
This is 3/6 of the Element Template MainThreadRef stack and follows the merged #3609.

## Why this change

#3609 preserves `main-thread:ref` values through the Element Template transform and prepares them as ET-owned MainThreadRef wrappers. A MainThreadRef still has two pieces of state that cross threads separately:

```ts
const ref = new MainThreadRef('idle');

// Serialized with a callback or attribute slot.
ref.toJSON(); // { _wvid: 1 }

// Required to create the corresponding main-thread ref entry.
[1, 'idle'];
```

The `_wvid` identifies the ref, but it intentionally does not carry `current`. The background runtime records the initial value in the shared MainThreadRef init-value queue. Snapshot already drains that queue and sends it to the main-thread ref runtime; Element Template previously did not. As a result, ET could transport a ref handle while the main-thread `_workletRefMap` still had no entry for that id.

## Protocol and data flow

This PR adds the optional field below to the existing JSON-encoded ET update payload:

```ts
interface ElementTemplateUpdateCommitContext {
  ops: ElementTemplateUpdateCommandStream;
  flushOptions: ElementTemplateFlushOptions;
  mainThreadRefInitValuePatch?: [id: number, value: unknown][];
}
```

For example, creating two refs on the background thread now contributes this data to the next ET update envelope:

```ts
new MainThreadRef('idle');
new MainThreadRef({ count: 1 });

// JSON payload sent by createElementTemplateUpdateEvent(...)
{
  "ops": [],
  "flushOptions": { "emptyPatch": true },
  "mainThreadRefInitValuePatch": [
    [1, "idle"],
    [2, { "count": 1 }]
  ]
}
```

There are two background send boundaries:

- During initial hydration, ET first matches the serialized typed page to the background tree. A successful match drains the init-value queue and includes it in the hydration update envelope.
- After hydration, each Preact commit drains the queue together with native ops, flush options, refs, and delayed `runOnMainThread` calls. A ref-only commit emits an `emptyPatch` update so the values are not left waiting for an unrelated native operation.

The queue uses take semantics: once an envelope owns a patch, the global queue is empty. This matches Snapshot and prevents the same id/value pair from being replayed by a later commit.

On the main thread, the patch listener now processes an envelope in this order:

```ts
const payload = JSON.parse(event.data.payload);

updateWorkletRefInitValueChanges(payload.mainThreadRefInitValuePatch);
checkReloadVersion();
applyElementTemplateUpdateCommands(payload.ops);
runDelayedBackgroundFunctions();
clearFirstScreenMainThreadRefs();
runDelayedMainThreadFunctions(payload.delayedRunOnMainThreadData);
__FlushElementTree(...);
```

Applying init values before delayed main-thread functions is required because a function in the same envelope may read the newly created ref. Init values are also applied before the stale-reload early return. This deliberately follows Snapshot: native ET operations and delayed function calls from an old reload are discarded, while MainThreadRefs created before `reloadTemplate` may still be referenced afterward and therefore keep their initial-value update.

At a successful hydration boundary, ET also follows Snapshot's ref ownership order. Worklet-context hydration first copies any accessed negative-ID first-screen ref value into its positive background ID. ET then replays delayed `runOnBackground` calls and clears the temporary negative-ID map before delayed `runOnMainThread` tasks and native flush. The positive hydrated entries remain available; only the first-screen aliases are released.

## Ownership and failure behavior

MainThreadRef init values are one-shot synchronization data, not a transactional native patch:

- A failed hydration match discards the queued values because the incomplete tree must not publish externally observable work.
- Serialization failure does not put an already-taken patch back into the queue. The commit path cleans up delayed return ids, pending refs, and detached subtree state before rethrowing the original serialization error. RuntimeProxy `getXxxContext()` and `dispatchEvent()` are internal native contracts and are called directly rather than modeled as recoverable failures.
- Full-lifetime destroy drains any values that were never assigned to an update envelope, preventing an old runtime from leaking them into a later lifetime.
- Existing delayed-function return-id cleanup and pending event/ref cleanup remain responsible for their own queues; this PR does not combine those ownership models with MainThreadRef storage.

The payload keeps the existing JSON boundary rather than adding a second bridge method. Initial values are required to satisfy the same JSON-serializable value contract as other ET protocol data. This PR intentionally adds no ET-only codec or runtime validation; unsupported values retain the existing Snapshot/native transport behavior instead of defining a second value model.

## Scope and follow-ups

This PR establishes initial-value transport and ownership only. It does not yet:

- resolve a prepared `main-thread:ref` to a native `FiberElement`;
- attach or clean up direct MainThreadRefs with element lifecycle;
- align attachment and cleanup with typed-list holder materialization and recycling; or
- add the final smoke example that exercises the complete feature.

Those behaviors remain in the following PRs in the stack. A changeset is not required because Element Template remains unreleased.

## Checklist

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved hydration reliability by correctly applying and cleaning up reference initialization updates.
  * Prevented stale, failed, or partially applied updates from leaving temporary references, delayed callbacks, or pending events.
  * Ensured references are initialized before related delayed tasks and cleaned up when elements are destroyed or hydration fails.
  * Improved commit recovery and update timing/profiling behavior for stale reloads.
* **New Features**
  * Added automatic cleanup of temporary first-screen references after hydration completes.

### DIFF
--- a/packages/react/runtime/__test__/element-template/test-utils/mock/mockNativePapi.ts
+++ b/packages/react/runtime/__test__/element-template/test-utils/mock/mockNativePapi.ts
@@ -174,6 +174,7 @@ export function installMockNativePapi(
     g.__LYNX_REPORT_ERROR_CALLS.push(error);
     nativeLog.push(['lynx.reportError', error]);
   });
+  const mockCreateJSObjectDestructionObserver = vi.fn().mockImplementation(() => ({}));
 
   const mockSetAttribute = vi.fn().mockImplementation((element: unknown, name: string, value: unknown) => {
     nativeLog.push(['__SetAttribute', formatNode(element), name, value]);
@@ -354,8 +355,15 @@ export function installMockNativePapi(
   vi.stubGlobal('__FlushElementTree', mockFlushElementTree);
   const currentLynx = (globalThis as unknown as { lynx?: any }).lynx;
   const baseLynx = (currentLynx && typeof currentLynx === 'object') ? currentLynx : {};
+  const baseGetNativeApp = typeof baseLynx.getNativeApp === 'function'
+    ? baseLynx.getNativeApp
+    : undefined;
   vi.stubGlobal('lynx', {
     ...baseLynx,
+    getNativeApp: () => ({
+      ...baseGetNativeApp?.call(baseLynx),
+      createJSObjectDestructionObserver: mockCreateJSObjectDestructionObserver,
+    }),
     reportError: mockReportError,
   });
 
--- a/packages/react/runtime/src/core/main-thread-ref.ts
+++ b/packages/react/runtime/src/core/main-thread-ref.ts
@@ -55,7 +55,7 @@ export class MainThreadRef<T> {
       addMainThreadRefInitValue(this._wvid, initValue);
       const id = this._wvid;
       this._lifecycleObserver = lynx.getNativeApp().createJSObjectDestructionObserver?.(() => {
-        lynx.getCoreContext?.().dispatchEvent({
+        lynx.getCoreContext().dispatchEvent({
           type: WorkletEvents.releaseWorkletRef,
           data: {
             id,
--- a/packages/react/runtime/src/element-template/native/patch-listener.ts
+++ b/packages/react/runtime/src/element-template/native/patch-listener.ts
@@ -4,9 +4,11 @@
 
 import type { ClosureValueType } from '@lynx-js/react/worklet-runtime/bindings';
 import {
+  clearFirstScreenMainThreadRefs,
   flushDelayedRunOnBackgroundFunctions,
   runRunOnMainThreadTask,
   setEomShouldFlushElementTree,
+  updateWorkletRefInitValueChanges,
 } from '@lynx-js/react/worklet-runtime/bindings';
 
 import { markTiming, setPipeline } from '../../core/performance.js';
@@ -27,13 +29,6 @@ export function installElementTemplatePatchListener(): void {
 
   listener = (event: Pick<ElementTemplateUpdateEvent, 'data'>) => {
     const { patchOptions } = event.data;
-    if (
-      typeof patchOptions.reloadVersion === 'number'
-      && patchOptions.reloadVersion < getReloadVersion()
-    ) {
-      return;
-    }
-
     const { flowIds, pipelineOptions } = patchOptions;
     const shouldProfilePatch = !!flowIds
       && typeof lynx.performance?.profileStart === 'function'
@@ -50,6 +45,21 @@ export function installElementTemplatePatchListener(): void {
 
     const payload = JSON.parse(event.data.payload) as ElementTemplateUpdateCommitContext;
     markTiming('parseChangesEnd');
+
+    if (payload.mainThreadRefInitValuePatch?.length) {
+      updateWorkletRefInitValueChanges(payload.mainThreadRefInitValuePatch);
+    }
+
+    if (
+      typeof patchOptions.reloadVersion === 'number'
+      && patchOptions.reloadVersion < getReloadVersion()
+    ) {
+      markTiming('mtsRenderEnd');
+      if (shouldProfilePatch) {
+        lynx.performance.profileEnd();
+      }
+      return;
+    }
 
     const hasOps = payload.ops.length > 0;
     const flushOptions = payload.flushOptions;
@@ -78,12 +88,14 @@ export function installElementTemplatePatchListener(): void {
         markTiming('mtsRenderEnd');
         if (isHydration) {
           flushDelayedRunOnBackgroundFunctions();
+          clearFirstScreenMainThreadRefs();
         }
       }
     } else {
       markTiming('mtsRenderEnd');
       if (isHydration) {
         flushDelayedRunOnBackgroundFunctions();
+        clearFirstScreenMainThreadRefs();
       }
     }
     if (delayedRunOnMainThreadData?.length) {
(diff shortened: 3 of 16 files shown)


Yradex:
@coderabbitai review

packages/react/runtime/__test__/element-template/test-utils/mock/mockNativePapi.ts
```diff
@@ -354,8 +355,15 @@ export function installMockNativePapi(
...
   const currentLynx = (globalThis as unknown as { lynx?: any }).lynx;
   const baseLynx = (currentLynx && typeof currentLynx === 'object') ? currentLynx : {};
+  const baseGetNativeApp = typeof baseLynx.getNativeApp === 'function'
+    ? baseLynx.getNativeApp
+    : undefined;
   vi.stubGlobal('lynx', {
     ...baseLynx,
+    getNativeApp: () => ({
+      ...baseGetNativeApp?.(),
+      createJSObjectDestructionObserver: mockCreateJSObjectDestructionObserver,
```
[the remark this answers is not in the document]
Yradex: Fixed in 0ec4e40ec. The mock now invokes the preserved method as `baseGetNativeApp?.call(baseLynx)`, matching the receiver used by the original `lynx.getNativeApp()` call.

packages/react/runtime/src/element-template/protocol/types.ts
```diff
@@ -200,4 +201,5 @@ export interface ElementTemplateUpdateCommitContext {
   isHydration?: boolean | undefined;
   reloadVersion?: number | undefined;
   delayedRunOnMainThreadData?: RunWorkletCtxData[] | undefined;
+  mainThreadRefInitValuePatch?: MainThreadRefInitValuePatch | undefined;
```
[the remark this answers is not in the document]
Yradex: This transport intentionally matches the existing Snapshot contract rather than introducing an ET-only codec. Snapshot already serializes the same shared `MainThreadRefInitValuePatch` with `JSON.stringify` in `snapshot/worklet/ref/updateInitValue.ts`, and values crossing the native/MTS boundary are required to be JSON-serializable protocol values. Functions, symbols, cycles, and BigInt are outside that contract; adding runtime validation or a second value model here would diverge ET from Snapshot. The PR description now states this boundary explicitly, so no code change is needed for this finding.

Yradex:
@coderabbitai review

packages/react/runtime/__test__/element-template/test-utils/mock/mockNativePapi.ts
```diff
@@ -174,6 +174,7 @@ export function installMockNativePapi(
     g.__LYNX_REPORT_ERROR_CALLS.push(error);
     nativeLog.push(['lynx.reportError', error]);
   });
+  const mockCreateJSObjectDestructionObserver = vi.fn().mockImplementation(() => ({}));
```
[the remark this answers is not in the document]
Yradex: This mock is intentionally private. It only supplies the native `createJSObjectDestructionObserver` contract so MainThreadRef construction can run in ET tests; no test in this change needs to assert observer creation or reset it independently. Exposing it through `MockNativePapi` would enlarge the shared test-utility API without a consumer, so I am keeping the result contract minimal.

packages/react/runtime/src/core/main-thread-ref.ts
```diff
@@ -55,7 +55,7 @@ export class MainThreadRef<T> {
       addMainThreadRefInitValue(this._wvid, initValue);
       const id = this._wvid;
       this._lifecycleObserver = lynx.getNativeApp().createJSObjectDestructionObserver?.(() => {
-        lynx.getCoreContext?.().dispatchEvent({
+        lynx.getCoreContext().dispatchEvent({
```
HuJean, line 58: No need to protect?

packages/react/runtime/src/element-template/native/patch-listener.ts
```diff
@@ -27,13 +29,6 @@ export function installElementTemplatePatchListener(): void {
 
   listener = (event: Pick<ElementTemplateUpdateEvent, 'data'>) => {
     const { patchOptions } = event.data;
-    if (
-      typeof patchOptions.reloadVersion === 'number'
-      && patchOptions.reloadVersion < getReloadVersion()
```
HuJean, line 32: Change the behavior when patchOptions.reloadVersion < getReloadVersion() ?

packages/react/runtime/src/core/main-thread-ref.ts
Yradex: No. `createJSObjectDestructionObserver?.()` remains the SDK capability gate. Once native exposes that API and invokes the callback, `getCoreContext()` and `dispatchEvent()` are required RuntimeProxy APIs. Guarding them would silently skip the ref release and hide an internal contract violation, so this path intentionally calls them directly.

packages/react/runtime/src/element-template/native/patch-listener.ts
Yradex: Yes, intentionally. The MainThreadRef init-value patch is still applied when the containing update envelope is stale; ET ops, delayed main-thread functions, and native flush remain skipped. Snapshot sends the same patch through `updateMTRefInitValue`, which explicitly ignores `reloadVersion` because refs created before `reloadTemplate` may still be referenced afterward. ET co-locates the patch with the regular update payload, so applying it before the stale check preserves Snapshot semantics. This is covered by the stale-update test in `element-template-patch.test.tsx` and the hydration race regression test in `hydration-listener.test.ts`.

HuJean (approved):

PR #1161, opened 2025-06-25
feat(tailwind-preset): Add createLynxPreset, expand Lynx plugin coverage, and enforce strict config types

Dugyu:
## Summary

This PR introduces the following updates to `@lynx-js/tailwind-preset`:

### TypeScript Compatibility Fixes
– Fixed type errors when using the Lynx Tailwind Preset in `tailwind.config.ts`.
– Enforced strict typing on the Tailwind config object.
– Refined types for `createPlugin` and `createUtilityPlugin`.

### Replaced & Updated Core Plugins
– Removed the unofficial plugin key `outline`.
– Removed deprecated `variants()` syntax (used only in Tailwind v1/v2).
– **inset**: Replaced shorthand with explicit `top`, `right`, `bottom`, and `left` utilities.
– **rotate**: Added support for `rotateX`, `rotateY`, and `rotateZ` (aligns with Tailwind v4).
– **translate**: Added support for `translateZ`.
– **skew & scale**: Updated their implementation to use the same `CSSTransformValue`.
– **transform**: Now supports `translateZ`, `rotateX`, `rotateY`, and `rotateZ` utilities.
– **boxShadow**: Uses Lynx-compatible color and shadow format; CSS variables not supported due to a boxShadow rendering bug.
– **gridColumn & gridRow**: Now implemented using longhand properties.
– **display**: Added `relative` and `linear` values.
– **alignContent**: Removed unsupported values.
– **textAlign**: Removed `justify`.
– **overflow**: Removed unsupported values.
– **backgroundClip**: Removed `text`.
– **justifyContent**: Removed `normal`.
– **visibility**: Removed `collapse`.
– **whitespace**: Retains only `normal` and `nowrap`.
– **wordBreak**: Removed `overflow-wrap` property and `keep-all` value.
– **perspective**: (Added `perspective-*` utlities (aligns with Tailwind v4).

### Enabled Utilities via Tailwind Core Plugins:

– **verticalAlign**
– **justifyItems**
– **justifySelf**
– **size**
– **flexBasis**
– **textIndent**
– **aspectRatio**
– **animation**

### Added Lynx Specific Plugins

– **direction**: add Lynx specific plugin to handle `ltr`, `rtr` and `lynx-ltr`.

### Theme Object Refinements

– **boxShadow**: Uses Lynx-compatible color formatting.
– **transitionProperty**: Removed unsupported values: `filter`, `stroke`, `backdrop-filter`.
– **zIndex**: Removed `auto`.
[...]

Issue #69, opened 2025-03-05
[Feature]: Support tailwindcss v4 and v3


mthezi:
### What problem does this feature solve?

如果能兼容的话应该是挺好的

### What does the proposed API of configuration look like?

直接使用

gaoachao:
Hi @mthezi ,

Actually, Tailwind CSS v3 is already supported. Please check our documentation: https://lynxjs.org/rspeedy/plugin.html#rsbuild-plugins. However, support for Tailwind CSS v4 is still in progress.

For updates on v4 support, please track this issue: https://github.com/rspack-contrib/rsbuild-plugin-tailwindcss/issues/17

Thank you for your interest!

gaoachao:
@mthezi This example might be helpful. https://github.com/lynx-family/lynx-examples/pull/20/file

bravo680git:
> [@mthezi](https://github.com/mthezi) This example might be helpful. [lynx-family/lynx-examples#20](https://github.com/lynx-family/lynx-examples/pull/20)

Thank you very much, it works great,  but hot reloading doesn't work. After changing a tailwind class, I must stop and re run dev again to make it running. So, are there any solutions for it?

inyono:
> > [@mthezi](https://github.com/mthezi) This example might be helpful. [lynx-family/lynx-examples#20](https://github.com/lynx-family/lynx-examples/pull/20)
> 
> Thank you very much, it works great, but hot reloading doesn't work. After changing a tailwind class, I must stop and re run dev again to make it running. So, are there any solutions for it?

I have the same behavior. Interestingly enough, the styles reload correctly when saving `App.css` instead of `App.tsx`.

devlopersabbir:
Still pending thiss features ??
We can we use hot reload?
@inyono  @bravo680git

Willaiem:
@inyono @bravo680git 

I managed to temporary fix the issue with hot reload. It isn't a perfect solution, but it works.

- install "Run on Save" extension 
https://marketplace.cursorapi.com/items?itemName=emeraldwalk.RunOnSave

- create ``.vscode/settings.json`` file in your Lynx project and paste this code:
```json
{
    "emeraldwalk.runonsave": {
      "commands": [
        {
          // TEMP: needed to run hot reload when styles are changed
          "match": "\\.(tsx)$",
          "cmd": "type src/index.css > src/index.css"
        }
      ]
    }
  }

```

[...]

PR #3627, opened 2026-08-24
test(tailwind-preset): compile the preset instead of reading a stale fixture

KealanAU:
## Summary

`@lynx-js/tailwind-preset` has no test that compiles Tailwind, so nothing checks what the preset actually emits. #3386 asked for one as a follow-up. This adds it, and regenerates `src/__tests__/output.css`, which stopped matching the preset in July 2025.

Follow-up to #3386. This does not touch v4, which is tracked in #69.

## Changes Made

- `config.test.ts` runs the Tailwind v3 CLI over `tailwind.config.ts` at test time. It no longer reads a committed fixture.
- A file snapshot compares the compiled CSS against `output.css`, so the fixture cannot drift again.
- `output.css` is regenerated. It still carried `--tw-translate-x`, `--tw-ring-*`, `--tw-shadow*` and `::backdrop`, which the preset stopped emitting after #1161 and #1773.
- `.dprint.jsonc` excludes `output.css`. The formatter adds semicolons that the compiler does not write, which broke the snapshot on every commit.
- This restores the compile that #1125 removed. Please read the reviewer notes.

## How to Test

1. Run `pnpm install --ignore-scripts --filter @lynx-js/tailwind-preset`.
2. Run `pnpm vitest run` in `packages/tailwind-preset`. Both tests pass in about 400ms.
3. To check the staleness claim yourself, run `git checkout main -- src/__tests__/output.css`. Run the tests again. The snapshot test fails against the fixture that is on `main` today.

## Notes for Reviewer

#1125 removed this compile because it timed out on Windows with `Hook timed out in 10000ms`, and raising the ceiling appeared to do nothing. It did nothing because #1113 set `--hook-timeout=50000`, which vitest does not recognise. The flag was dropped, so hooks stayed at the 10s default. #1121 corrected it to `--hookTimeout=50000` at 11:24 UTC on 2025-06-20, and #1125 opened at 14:09 the same day, which means the failing runs behind it predated the fix.

That ceiling applies now. Running this project with `--hookTimeout=1` gives `Hook timed out in 1ms`, so the flag reaches the project instead of stopping at the root config. The compile takes about 200ms, and it calls the resolved CLI directly rather than going through `npx`.

Two points need your judgement:

- The Windows job is the real test here. I deliberately dropped a `describe.skipIf(process.platform === 'win32')` guard, because skipping Windows hides the only open question. Re-adding it is one line.
- The snapshot will also fire on a `tailwindcss` patch bump. That seems worth seeing for a preset whose job is constraining output, but I will drop it if it makes Renovate PRs noisy.

#2859 conflicts on the `vitest` import line only. `@rstest/core` exposes `toMatchFileSnapshot(filepath, message?)` with the same signature, so the assertion survives that migration.

## Checklist

## Summary by CodeRabbit

- **Tests**
  - Updated Tailwind preset validation to compile styles using the Tailwind CSS v3 command-line workflow.
  - Snapshot checks now compare freshly generated CSS against the expected output.

- **Chores**
  - Refreshed the generated CSS fixture to reflect current Tailwind output.
  - Excluded generated CSS from automatic formatting to preserve its expected structure.


Dugyu:
Blocking temporarily: I need to understand why the generated CSS snapshot omits trailing semicolons while the normal Tailwind build output includes them.

Dugyu (dismissed):
The generated CSS snapshot is missing trailing semicolons, unlike the normal Tailwind build output. We are investigating and align the generated snapshot with the compiler output before merge.

Dugyu (approved):

PR #3653, opened 2026-08-25
refactor(react): collect background init into a function

upupming:
First of a stack that lets several cards in one LynxGroup share the ReactLynx runtime.

The background runtime installs its Preact adapters, app-level callbacks and commit hooks as module side effects, so they run once per module evaluation. A runtime shared by a group is evaluated once for the whole group, which would leave every card after the first without them.

This moves the block into `initBackgroundRuntime()` and calls it from the same place. **No behavior change** — it only gives the later per-page work something to call.

**Stack**
1. **this PR** — extract the init block
2. `createRenderContext({ lynx })`
3. per-page module singletons + `useLynx`
4. event subscription replacing the property writes

## Summary by CodeRabbit

* **Refactor**
  * Improved the internal organization of background runtime initialization without changing user-visible behavior.

### DIFF
--- a/.changeset/widen-react-peer-range-127.md
+++ b/.changeset/widen-react-peer-range-127.md
@@ -1,0 +1,6 @@
+---
+"@lynx-js/react-rsbuild-plugin": patch
+"@lynx-js/react-signals": patch
+---
+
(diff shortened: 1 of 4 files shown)


upupming:
@coderabbitai review

PR #3654, opened 2026-08-25
feat(react): add createRenderContext({ lynx })

upupming:
Stacked on the init-extraction PR.

A runtime shared between the cards of a LynxGroup is evaluated once, so the app-level callbacks it installs at module scope belong to whichever card was evaluated first; every later card is left without them.

`createRenderContext` gives a page entry a call that runs that installation, so every card registers its own callbacks:

```tsx
createRenderContext({ lynx }).render(<App />)
```

The adapters and hooks that belong to the runtime rather than to a page are now installed only on the first call, so a second page does not add a second ctx-not-found listener.

Registration is synchronous and rendering is not, which is what lets a page defer `render()` and still receive the first-screen events the engine queued while `app-service.js` loaded.

Binding the callbacks to the `lynx` the page passes lands in the per-app-handler PR of this stack; `root` keeps working unchanged for single-page cards.

## Summary by CodeRabbit

- **New Features**
  - Added the public `createRenderContext` API for connecting a page’s runtime and obtaining its root for rendering.
  - Made the new API available through React compatibility and lazy runtime exports.
  - Added support for deferred rendering while preserving initial-screen event callbacks.

- **Documentation**
  - Documented the new `createRenderContext` API and updated package compatibility information.

### DIFF
--- a/.changeset/create-render-context.md
+++ b/.changeset/create-render-context.md
@@ -1,0 +1,15 @@
+---
+"@lynx-js/react": minor
+---
+
+Add `createRenderContext({ lynx })`, which installs the runtime's app-level
+callbacks for the calling page and returns that page's root. A runtime shared
+between the cards of a LynxGroup is evaluated once, so the callbacks it installs
+at module scope would otherwise belong to whichever card was evaluated first.
+
+The adapters and hooks that belong to the runtime rather than to a page are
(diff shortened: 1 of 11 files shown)


upupming:
@coderabbitai review

PR #3655, opened 2026-08-25
feat(react): add the useLynx hook

upupming:
Stacked on #3654.

Code in a chunk shared by several cards must not capture the module-scope `lynx`: that one belongs to whichever card evaluated the chunk, and stops working once that card is destroyed.

`useLynx()` resolves per render to the page actually rendering, which `createRenderContext` provides through a context:

```tsx
function Popup() {
  const lynx = useLynx()
  return <view style={{ height: lynx.__globalProps.screenHeight }} />
}
```

Falls back to the module-scope `lynx` when the tree has no render context, so single-page cards are unaffected.

## Summary by CodeRabbit

* **New Features**
  * Added the `useLynx()` hook to access the page context associated with the currently rendering component.
  * Exported `useLynx()` through the React runtime and compatibility APIs.
  * Preserved fallback behavior for components rendered without a page context.

* **Tests**
  * Added coverage for page-specific context resolution and fallback behavior.

### DIFF
--- a/.changeset/use-lynx-hook.md
+++ b/.changeset/use-lynx-hook.md
@@ -1,0 +1,11 @@
+---
+"@lynx-js/react": minor
+---
+
+Add the `useLynx()` hook, which reads the `lynx` of the page a component is
+rendering in. Code in a chunk shared by several cards must not capture the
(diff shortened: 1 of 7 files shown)


upupming:
@coderabbitai review

PR #3665, opened 2026-08-26
feat(react-rsbuild-plugin): own the ReactLynx profile switch

upupming:
`performance.profile` drives ReactLynx's `__PROFILE__`, and `pluginReactLynx` read it back out of the Rspeedy config through `Symbol.for('rspeedy.api')` — a DSL plugin depending on the Rspeedy layer, which leaves it unset under plain Rsbuild.

`pluginReactLynx({ profile })` is the plugin-owned option and takes precedence. `performance.profile` keeps working and stays the only source for the `dist/stats.json` byproduct.


upupming:
Closing: `profile` is dropped from `pluginReactLynx`. Lynx build engine options are configured by applying `pluginLynx({ ... })` directly, and DSL plugins do not take them.

PR #3668, opened 2026-08-26
feat(rsbuild-plugin): build named entries as background-only runtimes

upupming:
A host that runs a runtime outside of any card — a group-level app runtime, for instance — needs a plain background bundle: no main thread to render with, and no template to be packed into.

`experimental_backgroundOnlyEntries` names the entries to build that way. They emit `<name>-runtime.js` and skip the main-thread entry and the template plugin, while still going through the normal entry wiring, so they take part in chunk splitting and share chunks with the pages built alongside them.

## Summary by CodeRabbit

* **New Features**
  * Added the experimental `experimental_backgroundOnlyEntries` option for configuring named entries as background-only runtimes.
  * Background-only entries generate runtime bundles without main-thread or template output while continuing to participate in shared chunk splitting.
  * Entries not listed in the option continue to build with their existing main-thread behavior.

* **Tests**
  * Added coverage for default behavior, configured background-only entries, unlisted entries, and unknown entry names.

### DIFF
--- a/.changeset/background-only-entries.md
+++ b/.changeset/background-only-entries.md
@@ -1,0 +1,8 @@
+---
+"@lynx-js/react-rsbuild-plugin": minor
+---
+
+Add `experimental_backgroundOnlyEntries`, which builds the named entries as
+background-only runtimes: no main thread and no template, just the background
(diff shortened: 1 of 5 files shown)


upupming:
@coderabbitai review

PR #3671, opened 2026-08-27
perf(react): split MainThreadObject worklet runtime

f0rdream:
Stacked on #3477.

- Builds separate core and MainThreadObject-capable worklet runtimes.
- Selects the full runtime only when optimized export usage includes the MainThreadObject APIs, while preserving root imports from `@lynx-js/react`.
- Keeps indeterminate standalone lazy bundles conservative for compatibility.
- Separates handle recognition bookkeeping so ordinary worklets and `MainThreadRef` do not initialize MainThreadObject registries.


f0rdream:
This refactor is now included directly in #3477 at commit 9064dec7b, so the stacked PR is no longer needed.

PR #3682, opened 2026-08-28
feat(rsbuild-plugin)!: emit the intermediate files into .lynx

upupming:
The intermediate directory is written by the Lynx build engine rather than by Rspeedy, so it no longer carries Rspeedy's name: `dist/.rspeedy` becomes `dist/.lynx`.

It is also no longer configurable. `output.distPath.intermediate` has documented itself as never read since #1154, so nothing can depend on it; #3676 started reading it again, and this removes it instead. The engine now holds the only copy of the name.

`CssExtractRspackPlugin` still had its own hardcoded `.rspeedy/` for lazy-bundle CSS, which #3676 missed. Without that one, a build would have emitted into both directories.

BREAKING CHANGE: intermediate files move from `dist/.rspeedy` to `dist/.lynx`, and `output.distPath.intermediate` is removed.

## Summary by CodeRabbit

- **Breaking Changes**
  - Intermediate build files are now generated under `.lynx` instead of `.rspeedy`.
  - The intermediate output directory is no longer configurable; `output.distPath.intermediate` is no longer supported.
  - The deprecated `DistPath` public type has been removed.

- **Documentation**
  - Updated guides, examples, and debug metadata references to use the `.lynx` directory.

- **Compatibility**
  - Updated plugin version requirements to support the latest Rspeedy release.


.changeset/warm-donkeys-repeat.md
```diff
@@ -0,0 +1,8 @@
+---
+"@lynx-js/rsbuild-plugin": patch
+"@lynx-js/rspeedy": patch
+"@lynx-js/css-extract-webpack-plugin": patch
+"@lynx-js/template-webpack-plugin": patch
+---
+
+Emit the intermediate files into `.lynx` instead of `.rspeedy`, since the directory is written by the Lynx build engine rather than by Rspeedy. The directory is no longer configurable: `output.distPath.intermediate` was documented as never read, and nothing else reads it now either.
```
luhc228: Please update the `.gitignore` in create-rspeedy

luhc228 (approved):

PR #3692, opened 2026-08-29
refactor(template-webpack-plugin): assemble custom sections by a naming strategy

upupming:
A `FetchBundle` lazy bundle is assembled entirely from custom sections, under
the three names its runtime resolves. Nothing else could be assembled that way,
so a caller that wanted it — an external bundle, whose consuming application
loads a section by the name of the chunk it holds — had to repeat the logic.

Take the naming as a strategy, and the bundle kind and the section bytecode as
options, so the shape no longer depends on the flags of a lazy bundle.

### DIFF
--- a/packages/webpack/template-webpack-plugin/src/LynxTemplatePlugin.ts
+++ b/packages/webpack/template-webpack-plugin/src/LynxTemplatePlugin.ts
@@ -321,6 +321,34 @@ export interface LynxTemplatePluginOptions {
   lazyBundleFetcher?: 'FetchBundle' | 'QueryComponent';
 
   /**
+   * Assemble every chunk into a custom section, named by this strategy.
+   *
+   * @remarks
+   *
+   * A bundle assembled this way carries no `lepusCode`, and only the chunks
+   * left unnamed stay in the manifest. A `FetchBundle` lazy bundle is assembled
+   * this way by default, under the three section names its runtime resolves.
+   */
+  customSectionNaming?:
+    | ((compilation: Compilation) => CustomSectionNaming)
+    | undefined;
+
+  /**
+   * The kind of bundle to encode.
+   *
+   * @defaultValue `'DynamicComponent'` for a lazy bundle, `'card'` otherwise.
+   */
+  appType?: 'card' | 'DynamicComponent' | undefined;
+
+  /**
+   * Whether to compile the main thread sections to `JsBytecode`.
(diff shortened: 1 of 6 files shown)


upupming:
On the `mainThreadAssets: lepusCode.root ? [lepusCode.root] : []` finding: the observation is accurate but it is not something this PR changed. The method it replaces, `#buildLazyBundleFetchBundleSections(lepusCode.root, ...)`, took the same single asset, so `lepusCode.chunks` was already dropped in FetchBundle mode before this refactor. `LynxTemplatePlugin` also still carries a `TODO: support multiple lepus chunks` next to where `lepusChunk` is assembled, so multiple main-thread chunks are not supported on this path either way.

This PR is meant to be behaviour-preserving — the lazy bundle output is unchanged, which is what I verified before opening it. Widening the main-thread set would change what a FetchBundle lazy bundle encodes, so it belongs in its own change alongside the multi-lepus-chunk TODO rather than here.

packages/rspeedy/lynx-bundle-rslib-config/package.json
```diff
@@ -39,6 +39,7 @@
     "@lynx-js/css-serializer": "workspace:*",
     "@lynx-js/runtime-wrapper-webpack-plugin": "workspace:*",
     "@lynx-js/tasm": "0.0.49",
+    "@lynx-js/template-webpack-plugin": "workspace:*",
```
upupming: This should be in devDep

packages/webpack/template-webpack-plugin/src/LynxTemplatePlugin.ts
```diff
@@ -507,12 +507,122 @@ const SECTION_MAIN_THREAD = 'main-thread';
...
+    }
+    sections[name] = {
+      ...(enableBytecode ? { encoding: 'JsBytecode' as const } : {}),
+      content: asset.source.source().toString(),
+    };
+  });
+
+  const remainingManifest: Record<string, string> = {};
+  let backgroundIndex = 0;
+  for (const [chunkName, content] of Object.entries(manifest)) {
```
upupming: I do not think this is a chunkName here, it will be something like "/.rspeedy/lazy-bundle/src_utils_minus.ts/background.js" which starts with `/`, not does the chunkName

luhc228 (approved):

PR #3693, opened 2026-08-29
feat(rsbuild-plugin): configure an external bundle through pluginLynx

upupming:
An external bundle hardcodes its filename (`${libName}.${isWeb ? 'web' : 'lynx'}.bundle`) and has no way to read the engine options. Applying `pluginLynx` to it was not an option either: a library build brings its own pipeline, and the thirteen plugins that build a Lynx template rewrite what it emits — measured at 83788 bytes against a baseline of 83710.

Apply only the config of `pluginLynx` for the callers that do not build a template (`rslib`, `rstest`), which is the condition `pluginAutoLynx` already uses, and resolve the external bundle filename through it.

`pluginLynx` stays the only export. The output is unchanged: a test asserts it by comparing the sha256 of a build with the plugin against one without it, and a second test covers configuring the filename through it.

## Summary by CodeRabbit

- **New Features**
  - External Lynx bundles now support custom filenames configured through Lynx build settings.
  - Custom section naming is supported for main-thread, background, and CSS assets.
  - Build settings are applied consistently across supported build types.

- **Bug Fixes**
  - External bundle builds preserve existing output while correctly resolving configured filenames.
  - Template-specific behavior no longer affects library and test builds.

- **Tests**
  - Added coverage for consistent output and custom bundle filenames.
  - Added coverage for lazy-bundle section naming.

### DIFF
--- a/packages/rspeedy/lynx-bundle-rslib-config/src/externalBundleRslibConfig.ts
+++ b/packages/rspeedy/lynx-bundle-rslib-config/src/externalBundleRslibConfig.ts
@@ -5,10 +5,13 @@
 import { rsbuild } from '@rslib/core'
 import type { LibConfig, RslibConfig, Rspack } from '@rslib/core'
 
+import type { LynxConfig } from '@lynx-js/rsbuild-plugin'
 import { RuntimeWrapperWebpackPlugin as BackgroundRuntimeWrapperWebpackPlugin } from '@lynx-js/runtime-wrapper-webpack-plugin'
 
 import { ExternalBundleWebpackPlugin } from './webpack/ExternalBundleWebpackPlugin.js'
 import { MainThreadRuntimeWrapperWebpackPlugin } from './webpack/MainThreadRuntimeWrapperWebpackPlugin.js'
+
+const S_LYNX_CONFIG = Symbol.for('@lynx-js/rsbuild-plugin:config')
 
 /**
  * The options for encoding the external bundle.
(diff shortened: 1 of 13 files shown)


packages/rspeedy/lynx-bundle-rslib-config/src/externalBundleRslibConfig.ts
```diff
@@ -772,7 +775,14 @@ const externalBundleRsbuildPlugin = ({
...
             {
-              bundleFileName: `${libName}.${isWeb ? 'web' : 'lynx'}.bundle`,
+              // The engine config is only there when the user applied
+              // `pluginLynxConfig`, so the default stays the source of truth.
+              bundleFileName: api.useExposed<LynxConfig>(S_LYNX_CONFIG)
+                ?.resolveBundleFilename({
+                  entryName: libName,
+                  platform: isWeb ? 'web' : 'lynx',
+                })
+                ?? `${libName}.${isWeb ? 'web' : 'lynx'}.bundle`,
```
upupming, line 785: We should not use fallback; this will hide the edge case and errors. We should fail strictly when pluginLynx not exist.

packages/rspeedy/plugin-react/src/autoLynx.ts
```diff
@@ -19,8 +19,9 @@ export function pluginAutoLynx(): RsbuildPlugin {
   return {
     name: 'lynx:react:auto-lynx',
     async setup(api) {
-      // The callers that do not emit a Lynx template do not want the engine
-      // either. See the same condition in `applyEntry`.
+      // `rslib` and `rstest` drive the build themselves — an external bundle
+      // assembles its own `.lynx.bundle`, and a test run emits nothing — so
+      // the engine has nothing to add. See the same condition in `applyEntry`.
```
upupming, line 24: Just remove comments, not verbose comments

PR #3660, opened 2026-08-26
feat: external-bundle debug-metadata

luhc228:
## Summary by CodeRabbit

* **New Features**
  * Added debug metadata support for external bundles, including source maps and per-section bytecode debugging information.
  * Added support for matching debug data to emitted TASM custom sections.
  * Added a runnable React externals debug-build example.
* **Documentation**
  * Documented configuration, generated metadata, custom-section matching, and production-build behavior.
* **Bug Fixes**
  * Improved source-map URL rewriting for custom-section assets.
  * Preserved debug metadata generation for supported production component-library builds.

## Checklist

- [ ] Tests updated (or not required).
- [ ] Documentation updated (or not required).
- [ ] Changeset added, and when a BREAKING CHANGE occurs, it needs to be clearly marked (or not required).

PR #3694, opened 2026-08-29
feat(rsbuild-plugin): drive the template hooks for an external bundle

upupming:
Closes #3660.

An external bundle assembles itself, so the plugins that tap the template lifecycle never ran for it. `pluginLynxDebugMetadata` is the one that shows: an external bundle shipped no source maps and no release key, which is what made the debug metadata a separate, hand-wired path.

`pluginLynx` exposes the hooks whether or not it drives the build, and `ExternalBundleWebpackPlugin` drives them around its own encode. The payload handed to the hooks carries the shape `LynxTemplatePlugin` builds, since every tap reads that, while the encoder is still given only what an external bundle carries.

Verified end to end on `examples/react-externals`: decoding `comp-lib.lynx.bundle` shows the release key in the background section.

```
sections: ./App.js, ./App.js:CSS, ./App.js__main-thread
  ./App.js: debugMetadata=true
```

Without the plugin the bundle is byte for byte what it was, which a test asserts by comparing sha256.

## Summary by CodeRabbit

* **New Features**
  * External library bundles now honor configured bundle filenames.
  * Template lifecycle hooks are applied to externally assembled bundles.
  * External bundles now include source maps, release keys, and debug metadata.
  * Added support for configuring custom template section names.
  * Lazy-bundle sections now use consistent, runtime-compatible names.

* **Bug Fixes**
  * Improved handling of bundles without debug metadata configuration.
  * Prevented unrelated build settings from affecting externally managed builds.
  * Improved consistency between engine-managed and externally assembled bundles.

### DIFF
--- a/examples/react-externals/rslib-comp-lib.config.ts
+++ b/examples/react-externals/rslib-comp-lib.config.ts
@@ -1,5 +1,6 @@
 import { defineExternalBundleRslibConfig } from '@lynx-js/lynx-bundle-rslib-config';
 import { pluginReactLynx } from '@lynx-js/react-rsbuild-plugin';
+import { pluginLynx } from '@lynx-js/rsbuild-plugin';
 
 // REACTLYNX_ASYNC=true builds comp-lib against async (Promise) ReactLynx
 // externals to match an async host; output isolated in
@@ -14,6 +15,9 @@ export default defineExternalBundleRslibConfig({
     },
   },
   plugins: [
+    // Configures the bundle filename and brings the plugins that tap the
+    // template hooks, `pluginLynxDebugMetadata` among them.
+    ...pluginLynx(),
     pluginReactLynx(),
   ],
   // Sync and async share this config file, so rspack's persistent cache (keyed
--- a/packages/rspeedy/lynx-bundle-rslib-config/src/webpack/ExternalBundleWebpackPlugin.ts
+++ b/packages/rspeedy/lynx-bundle-rslib-config/src/webpack/ExternalBundleWebpackPlugin.ts
@@ -5,6 +5,18 @@ import type { Rspack } from '@rslib/core'
 
 import { cssChunksToMap } from '@lynx-js/css-serializer'
 import type { LynxStyleNode } from '@lynx-js/css-serializer'
+import type { LynxTemplatePlugin } from '@lynx-js/template-webpack-plugin'
+
+/**
+ * Provides the template lifecycle hooks, so the plugins that tap them run for
+ * a bundle assembled outside `LynxTemplatePlugin`.
+ *
+ * @public
+ */
+export interface LynxTemplatePluginHooksProvider {
+  getLynxTemplatePluginHooks:
+    typeof LynxTemplatePlugin.getLynxTemplatePluginHooks
+}
 
 /**
  * The options for {@link ExternalBundleWebpackPlugin}.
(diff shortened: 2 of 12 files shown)


upupming:
Went through the three findings.

**`LynxTemplatePluginHooksProvider` not exported from the entry point** — valid. The API report did carry the `ae-forgotten-export` warning for it. Fixed in #3703, which is where the rest of the export hygiene for this stack lives (it also exports `CustomSectionEntry` for the same reason); the report there now has no forgotten-export warnings. Doing it here would mean rebasing the eight PRs stacked on top for a warning that only matters once the stack lands.

**`intermediateAssets` from `beforeEncode` not cleaned up** — the premise does not hold. `pluginLynxDebugMetadata` installs `stripDebugMetadataFromOutput(compiler)` in `onAfterCreateCompiler`, unconditionally and independently of the `isLocalProductionBuild` gate below it, and that is what removes `debug-metadata.json` from the output unless `DEBUG=rspeedy`. It applies to an external bundle too, since this PR is what makes `pluginLynxDebugMetadata` run for one. The pre-hook snapshot cleanup in `#generateExternalBundle` is not the only thing standing between the metadata and the output.

**`debugInfo: ''` passed to `beforeEmit`** — accurate, and a real gap: an external bundle with `JsBytecode` carries no per-section bytecode debug info. It is not a regression though; before this PR an external bundle had no debug metadata at all. Closing it means widening `ExternalBundleWebpackPluginOptions.encode` from `(opts) => { buffer: Buffer }` to also return the encoder's debug info, which is a public API change and needs threading through both encode adapters. That is its own change, not part of driving the hooks, so I have left it out here.

examples/react-externals/rslib-comp-lib.config.ts
```diff
@@ -14,6 +15,9 @@ export default defineExternalBundleRslibConfig({
     },
   },
   plugins: [
+    // Configures the bundle filename and brings the plugins that tap the
+    // template hooks, `pluginLynxDebugMetadata` among them.
+    ...pluginLynx(),
```
upupming, line 20: This is not my expection, I think pluginLynx should be included in pluginReactLynx, user should not need to use it unless need specific lynx config such as `output.filename.bundle`

packages/rspeedy/lynx-bundle-rslib-config/src/webpack/ExternalBundleWebpackPlugin.ts
```diff
@@ -5,6 +5,18 @@ import type { Rspack } from '@rslib/core'
...
+/**
+ * Provides the template lifecycle hooks, so the plugins that tap them run for
+ * a bundle assembled outside `LynxTemplatePlugin`.
+ *
+ * @public
+ */
+export interface LynxTemplatePluginHooksProvider {
+  getLynxTemplatePluginHooks:
+    typeof LynxTemplatePlugin.getLynxTemplatePluginHooks
+}
```
upupming, line 19: This is totally redundant, just use LynxTemplatePlugin is okay like other places

packages/rspeedy/lynx-bundle-rslib-config/src/webpack/ExternalBundleWebpackPlugin.ts
```diff
@@ -116,16 +139,44 @@ export class ExternalBundleWebpackPlugin {
...
       assets,
       enableJsBytecode,
+      compilation,
+      hooks,
     )
 
     const { RawSource } = compiler.webpack.sources
+    const outputName = this.options.bundleFileName
+
+    const emitted = hooks
```
upupming, line 155: hooks should always exist please do not use fallback

packages/rspeedy/plugin-debug-metadata/src/LynxDebugMetadataPlugin.ts
```diff
@@ -346,7 +346,9 @@ export function rewriteSourceMappingURLs(
   args: Parameters<Parameters<TemplateHooks['beforeEncode']['tap']>[1]>[0],
   options?: RewriteSourceMappingURLsOptions,
 ): void {
-  const debugMetadataUrl = args.encodeData.sourceContent.config[
+  // A bundle assembled outside `LynxTemplatePlugin` carries no `config`, so
+  // the url is unset there in the same way an empty one is.
+  const debugMetadataUrl = args.encodeData.sourceContent.config?.[
```
upupming, line 351: please do not use fallback

PR #3695, opened 2026-08-31
fix(react-rsbuild-plugin): do not apply the engine twice per environment

upupming:
`isPluginExists` without an environment only reports the global plugins, so an
engine registered on an environment went unnoticed. `pluginAutoLynx` applied a
second one, and Rspeedy applied a third built from its own config, which
silently replaced the options the user had given theirs.

The engine is a global one wherever it is registered, so look it up in both
scopes from a single helper. Applying it twice is harmless now either way: the
SWC transforms it contributes are a set.

### DIFF
--- a/packages/rspeedy/plugin-react/src/autoLynx.ts
+++ b/packages/rspeedy/plugin-react/src/autoLynx.ts
@@ -4,17 +4,10 @@
 
 import type { RsbuildContext, RsbuildPlugin } from '@rsbuild/core'
 
-import { PLUGIN_LYNX_NAME, pluginLynx } from '@lynx-js/rsbuild-plugin'
-
-// Calling `setup` does not register the plugins, so `PLUGIN_LYNX_NAME` cannot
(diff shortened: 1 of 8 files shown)


packages/rspeedy/plugin-react/src/autoLynx.ts
```diff
@@ -27,16 +44,16 @@ export function pluginAutoLynx(): RsbuildPlugin {
         return
       }
 
+      const original = api.getRsbuildConfig('original')
+
       if (
-        api.isPluginExists(PLUGIN_LYNX_NAME)
+        isEngineRegistered(api, original)
         || Object.hasOwn(api.context, S_PLUGIN_LYNX_APPLIED)
```
upupming: How about just use `S_PLUGIN_LYNX_APPLIED`, not rely on the PLUGIN_LYNX_NAME?

luhc228 (approved):

PR #3696, opened 2026-08-31
feat(rsbuild-plugin): apply the build engine for rslib

upupming:
`rslib` was left out of the engine entirely, so a library build read none of the
Lynx config and had to mirror the parts it needed by hand.

Apply the engine there too, and skip the plugins that shape or serve a bundle:
`rslib` assembles its own, and `rstest` has none.


packages/rspeedy/plugin-react/src/autoLynx.ts
```diff
@@ -36,11 +36,11 @@ export function pluginAutoLynx(): RsbuildPlugin {
...
     async setup(api) {
-      // `rslib` and `rstest` drive the build themselves — an external bundle
-      // assembles its own `.lynx.bundle`, and a test run emits nothing — so
-      // the engine has nothing to add. See the same condition in `applyEntry`.
-      const { callerName } = api.context
-      if (callerName === 'rslib' || callerName === 'rstest') {
+      // A test run emits nothing, so the engine has nothing to add. `rslib`
+      // does emit a bundle and reads the engine config, so the engine is
+      // applied for it; `pluginLynx` leaves out the plugins that would rewrite
+      // what an external bundle assembles itself.
```
upupming: Just remove verbose comments

luhc228 (approved):

PR #3699, opened 2026-08-31
fix(lynx-bundle-rslib-config): encode the sections in a stable order

upupming:
The custom sections of an external bundle were written in whatever order `compilation.getAssets` returned. That order shifts with the plugins that took part in the compilation, so the same input could encode to different bytes.

Sort the assets by name before encoding. On the `css-lib` fixture the section order goes from `['index', 'index__main-thread', 'index:CSS']` to `['index__main-thread', 'index', 'index:CSS']` and stays there; the test asserts it instead of sorting the keys before comparing.

### DIFF
--- a/.changeset/external-bundle-section-order.md
+++ b/.changeset/external-bundle-section-order.md
@@ -1,0 +1,5 @@
+---
+"@lynx-js/lynx-bundle-rslib-config": patch
+---
+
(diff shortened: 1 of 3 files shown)


upupming:
Superseded: sorting the assets does not make the bundle byte-stable — the encoder reorders the custom sections regardless, so the change had no effect it could be held to and the order assertion it added was environment-dependent. Dropped from the stack; #3700 now sits on #3698.

PR #3700, opened 2026-08-31
feat(rsbuild-plugin): emit what a Lynx bundle can carry for an external bundle

upupming:
`pluginOutput` sat behind the caller gate, so an external bundle was built with the Rsbuild HTML plugin registered, with a `.LICENSE.txt` trailer it has nowhere to link to, and with `const`/`let` in the bundler runtime that QuickJS parses slower than `var`.

What the output may contain follows from the Lynx runtime rather than from who assembles the bundle, so apply it for every caller.

Two consequences:

- `HtmlRspackPlugin` and `RsbuildHtmlPlugin` are no longer registered, and the `.LICENSE.txt` files are gone from the fixture output.
- The engine emits CSS into a directory of its own (`[name]/[name].css`), which used to leak into the section name as `index/index:CSS`. A section is now named after the file rather than the path leading to it, and the sorted section order shifts accordingly.

### DIFF
--- a/packages/rspeedy/lynx-bundle-rslib-config/src/webpack/ExternalBundleWebpackPlugin.ts
+++ b/packages/rspeedy/lynx-bundle-rslib-config/src/webpack/ExternalBundleWebpackPlugin.ts
@@ -7,6 +7,10 @@ import { cssChunksToMap } from '@lynx-js/css-serializer'
 import type { LynxStyleNode } from '@lynx-js/css-serializer'
 import type { LynxTemplatePlugin } from '@lynx-js/template-webpack-plugin'
 
+function sectionName(assetName: string, extension: string): string {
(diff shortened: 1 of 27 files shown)


packages/rspeedy/lynx-bundle-rslib-config/src/webpack/ExternalBundleWebpackPlugin.ts
```diff
@@ -7,6 +7,10 @@ import { cssChunksToMap } from '@lynx-js/css-serializer'
 import type { LynxStyleNode } from '@lynx-js/css-serializer'
 import type { LynxTemplatePlugin } from '@lynx-js/template-webpack-plugin'
 
+function sectionName(assetName: string, extension: string): string {
+  return assetName.slice(assetName.lastIndexOf('/') + 1, -extension.length)
+}
```
upupming, line 12: Is this really necessary, why?

PR #3701, opened 2026-08-31
feat(rsbuild-plugin): minify an external bundle with the engine options

upupming:
`@lynx-js/lynx-bundle-rslib-config` carried its own minify defaults, down to the same `negate_iife: false` and `side_effects: false` that keep the module wrapper callable from `lynx_core.js` — its comment said as much, that it mirrors `pluginMinify`. The copy was a subset: an external bundle missed `join_vars`, `ecma`, `inline`, `comparisons`, `toplevel`, the `format` options and the whole CSS minifier config a template build gets.

Apply `pluginMinify` for every caller and drop the copy. Only whether to minify at all is still decided by the rslib config, since `rslib build` always compiles with rspack mode `production` and the development signal is `NODE_ENV`.

The `REACT_DEVTOOL` coverage moves to `pluginMinify`, which had none of its own.

### DIFF
--- a/.changeset/rslib-engine-minify.md
+++ b/.changeset/rslib-engine-minify.md
@@ -1,0 +1,6 @@
+---
+"@lynx-js/rsbuild-plugin": minor
+"@lynx-js/lynx-bundle-rslib-config": minor
+---
+
+Minify an external bundle with the engine options instead of a copy of them.
--- a/packages/rspeedy/lynx-bundle-rslib-config/src/externalBundleRslibConfig.ts
+++ b/packages/rspeedy/lynx-bundle-rslib-config/src/externalBundleRslibConfig.ts
@@ -56,38 +56,6 @@ export interface EncodeOptions {
   enableJsBytecode?: boolean
 }
 
-// When preact devtools is enabled (`REACT_DEVTOOL`), keep function and class
-// names. Devtools relies on them to resolve component names (`type.name`) and
-// to reconstruct the hook tree (it matches minified stack frames by function
-// name). Minification would otherwise mangle/inline those names away. Only
-// enabled with devtools since it slightly increases bundle size. Mirrors
-// `pluginMinify` in @lynx-js/rspeedy (lynx-family/lynx-stack#2880).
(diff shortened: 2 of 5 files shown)


.changeset/rslib-engine-minify.md
```diff
@@ -0,0 +1,6 @@
+---
+"@lynx-js/rsbuild-plugin": minor
+"@lynx-js/lynx-bundle-rslib-config": minor
+---
+
+Minify an external bundle with the engine options instead of a copy of them.
```
upupming, line 6: There are so many changesets after merge this stack, can you make the number of changesets minimal. We should append to the exist changeset or no changeset at all if it is already included in previous one

packages/rspeedy/plugin-lynx/src/index.ts
```diff
@@ -89,11 +89,13 @@ export function pluginLynx(
     // What the output may contain — no HTML, no license trailer, `var` in the
     // bundler runtime — follows from the Lynx runtime, not from the caller.
     pluginOutput(),
+    // How far the output may be compressed is a property of the Lynx runtime
+    // too: the module wrapper has to keep its IIFE and its return.
```
upupming, line 93: remove comments

packages/rspeedy/lynx-bundle-rslib-config/src/externalBundleRslibConfig.ts
```diff
@@ -107,9 +75,10 @@ export const DEFAULT_EXTERNAL_BUNDLE_LIB_CONFIG: LibConfig = {
       dependencies: false,
       peerDependencies: false,
     },
-    minify: process.env['NODE_ENV'] === 'development'
-      ? false
-      : DEFAULT_EXTERNAL_BUNDLE_MINIFY_CONFIG,
+    // `rslib build` always compiles with rspack mode `production`, so the
+    // development signal here is `NODE_ENV`. What minification may do is
+    // `pluginLynx`'s to say.
+    minify: process.env['NODE_ENV'] !== 'development',
```
upupming, line 81: is minify: true support REACT_DEVTOOL? why?

PR #3702, opened 2026-08-31
feat(rsbuild-plugin): minify the CSS and emit the source maps of an external bundle

upupming:
`pluginMinify` brought the Lynx CSS minifier options along, but the minimizer that reads them stayed behind the caller gate, so they were dead config for an external bundle. `pluginLynxDebugMetadata` likewise collects source maps that `pluginSourcemap` was not there to produce.

Apply both for every caller. What is left behind the gate — `pluginChunkLoading`, `pluginDev`, `pluginOptimization`, `pluginServer` — describes how a bundle is loaded and served, which an external bundle does not do for itself: it is loaded by `lynx_core.js` as sections of the template that embeds it. The gate is named for that now.

I measured the four that stay gated: applying them changes nothing in the `utils-lib` bundle and two bytes in the `css-lib` main-thread section. I could not attribute those two bytes to a specific plugin, and `pluginOptimization`'s `avoidEntryIife` reads as a direct conflict with the `negate_iife: false` that keeps the module wrapper callable, so I left them where they are rather than claim they are safe.

### DIFF
--- a/packages/rspeedy/plugin-lynx/src/index.ts
+++ b/packages/rspeedy/plugin-lynx/src/index.ts
@@ -39,12 +39,12 @@ export type {
   LynxPluginOptions,
 } from './config.js'
 
-// `rslib` and `rstest` drive the build themselves: an external bundle is
-// assembled by `@lynx-js/lynx-bundle-rslib-config`, which emits its own
-// `.lynx.bundle`, and a test run emits nothing. They read the config all the
-// same, so it is applied for them, but the plugins below assemble a bundle of
-// their own and would rewrite what those callers emit.
-function onlyWhenTheEngineOwnsTheBuild(
+// What the engine puts in a bundle applies wherever it is assembled. How a
+// bundle is loaded and served does not: an external bundle is loaded by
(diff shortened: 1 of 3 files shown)


packages/rspeedy/plugin-lynx/src/index.ts
```diff
@@ -39,12 +39,12 @@ export type {
...
-// `.lynx.bundle`, and a test run emits nothing. They read the config all the
-// same, so it is applied for them, but the plugins below assemble a bundle of
-// their own and would rewrite what those callers emit.
-function onlyWhenTheEngineOwnsTheBuild(
+// What the engine puts in a bundle applies wherever it is assembled. How a
+// bundle is loaded and served does not: an external bundle is loaded by
+// `lynx_core.js` as sections of the template that embeds it, and a test run
+// loads nothing. `rslib` and `rstest` are the callers that assemble their own,
+// so the plugins below would describe a way of loading that is not theirs.
+function onlyWhenTheEngineLoadsTheBundle(
```
upupming, line 47: This function name should be correct in the first pr that adds it

PR #3703, opened 2026-08-31
refactor(lynx-bundle-rslib-config): assemble the sections with the shared builder

upupming:
`ExternalBundleWebpackPlugin` built its custom sections by hand, repeating the CSS rule-list conversion and the `JsBytecode` encoding that `LynxTemplatePlugin` already does. The generalization in #3692 that made the template side pluggable stopped at an interface no other caller used.

Export `buildCustomSections` and drive it from both, with a naming strategy for the external bundle that keys the sections off the entry names. `CustomSectionEntry` is exported with it, which also clears an `ae-forgotten-export` warning from the API report.

The section order shifts with the change (the template side groups by role rather than by asset), so the test that locks the order is updated. It stays deterministic.

### DIFF
--- a/.changeset/external-bundle-custom-sections.md
+++ b/.changeset/external-bundle-custom-sections.md
@@ -1,0 +1,6 @@
+---
+"@lynx-js/template-webpack-plugin": minor
+"@lynx-js/lynx-bundle-rslib-config": patch
+---
+
+Export `buildCustomSections` and assemble an external bundle with it, so it shares the template path instead of a copy.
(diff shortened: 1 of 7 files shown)


.changeset/external-bundle-custom-sections.md
```diff
@@ -0,0 +1,6 @@
+---
+"@lynx-js/template-webpack-plugin": minor
+"@lynx-js/lynx-bundle-rslib-config": patch
+---
+
+Export `buildCustomSections` and assemble an external bundle with it, so it shares the template path instead of a copy.
```
upupming, line 6: This pr should be combined in the first one to be better reviewed

PR #3734, opened 2026-09-02
feat(web-core): support legacy XElement aliases

MoonfaceX:
## Summary

- add always-on legacy XElement aliases for Web components that are already implemented and loaded by `@lynx-js/web-elements/all`
- apply the aliases consistently to CSR and SSR element creation, binary CSS, and legacy JSON CSS
- make reverse tag lookup deterministic, preserve the upstream `textarea` identity, and align SSR `input` behavior with CSR
- fix the SSR shadow-template lookup key from `x-web-view` to the registered `x-webview`

No page config gate is introduced. The mapping follows the existing `view` to `x-view` runtime model; it does not register the legacy names as browser custom elements or rewrite `querySelector` strings.

## Test Plan

- `pnpm turbo build --concurrency=1` (73/73 tasks)
- `pnpm test` in `packages/web-platform/web-core` (437/437 tests)
- `cargo test --all-features` (120/120 tests)
- `cargo test --target wasm32-unknown-unknown --all-features` with `wasm-bindgen 0.2.126` (2/2 tests)
- `pnpm test:server` in `packages/web-platform/web-core-e2e` (17/17 tests)
- `pnpm turbo api-extractor` (25/25 tasks)
- `pnpm dprint check`
- staged ESLint, Biome, dprint, and cargo fmt checks

## Checklist

- [x] Tests updated.
- [x] Documentation updated.
- [x] Changeset added.

## Summary by CodeRabbit

- **New Features**
  - Added support for legacy XElement names, including input, textarea, viewpager, webview, overlay, refresh, blur-view, and scroll-coordinator elements.
  - Compatible names now resolve consistently to their corresponding HTML elements during client and server-side rendering.
  - CSS selectors using legacy Lynx element names, including legacy JSON styles, are translated correctly.
  - Webview elements now render with the expected webview template and iframe structure.
- **Bug Fixes**
  - Improved reverse tag resolution for aliases sharing HTML elements, ensuring stable canonical names.


packages/web-platform/web-core/ts/client/decodeWorker/cssLoader.ts
```diff
@@ -129,7 +130,8 @@ function parseAndPushSelector(selector: Selector, s: string) {
     if (tag === 'page') {
       selector.push_one_selector_section('AttributeSelector', 'part="page"');
     } else {
-      const typeName = tag.includes('-') ? tag : `x-${tag}`;
+      const typeName = LYNX_TAG_TO_HTML_TAG_MAP[tag]
```
PupilTong, line 133: duplicated logic? See packages/web-platform/web-core/src/template/template_sections/style_info/style_info_decoder.rs L151-153

PupilTong (approved):

packages/web-platform/web-core/ts/client/decodeWorker/cssLoader.ts
MoonfaceX: These are two different input paths. Binary StyleInfo reaches the Rust decoder as a raw TypeSelector, while legacy JSON reaches this TypeScript path as a `[lynx-tag="..."]` selector first. Without resolving the map here, the existing fallback would turn dashless aliases such as `viewpager`, `overlay`, and `refresh` into `x-viewpager`, `x-overlay`, and `x-refresh` before Rust sees them, so the Rust map could no longer match. Sharing one implementation would require a new helper across the Wasm boundary; this keeps the minimal runtime change, with synchronization documented and both paths covered by tests.

PR #3764, opened 2026-09-03
chore(deps): update @rslib/core to v1.0.0

upupming:
Upgrade the `rslib` catalog from `1.0.0-beta.3` to `1.0.0` (released 2026-09-03). Renovate will not open this itself for three days because of `minimumReleaseAge`, and pnpm enforces the same gate, so `@rslib/*` and `rsbuild-plugin-dts` join the Rstack entries in `minimumReleaseAgeExclude`.

Breaking changes between beta.3 and 1.0.0 (splitChunks moved to the Rsbuild config, `createRequire` parser on by default, declaration extension redirects on by default, `advancedEsm` and `native-preview` removed) touch nothing in this repository's rslib configs.

Rslib 1.0 depends on `@rsbuild/core ~2.2.2`, so the lockfile gains a nested `@rsbuild/core@2.2.3` (rspress dedupes onto it too). The repository catalog keeps `@rsbuild/core 2.1.10` and the `@rspack/core 2.1.7` override; the Rstack 2.2 line is tracked separately in #3415.

Verified locally: `pnpm dedupe --check` passes, the full `turbo build` (71 tasks) succeeds with no warnings, and the rstest suites of `lynx-bundle-rslib-config` (34), `genui/lynx-xml` (14) and `genui/server` (190) pass. `lynx-bundle-rslib-config` keeps its `^1.0.0-beta.1` peer range, which already covers 1.0.0, so no changeset is needed.

## Summary by CodeRabbit

* **Chores**
  * Updated release configuration to use the stable `@rslib/core` 1.0.0 version.
  * Adjusted package release-age exclusions for related build tooling.

### DIFF
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -32,8 +32,8 @@ catalogs:
       version: 2.1.10
   rslib:
     '@rslib/core':
-      specifier: 1.0.0-beta.3
-      version: 1.0.0-beta.3
+      specifier: 1.0.0
+      version: 1.0.0
   rspack:
     '@rspack/test-tools':
       specifier: 2.1.3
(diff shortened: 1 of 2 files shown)


upupming:
Superseded by #3793, which upgrades the whole Rstack family together.

PR #3773, opened 2026-09-04
perf(react): skip ET attribute diff before materialization

Yradex:
Attribute slot updates compare previous and next values even when hydration has not completed or the instance has not been materialized. Object-valued slots can be serialized just to discard the comparison result.

Return after preparing attributes, queuing eligible ref effects, and saving the latest shadow state when no update patch can be emitted. Later creation still uses the latest values, and materialized instances keep the existing attribute diff behavior.

## Checklist

### DIFF
--- a/packages/react/runtime/__test__/element-template/runtime/background/instance.test.ts
+++ b/packages/react/runtime/__test__/element-template/runtime/background/instance.test.ts
@@ -2024,6 +2024,43 @@ describe('BackgroundElementTemplateInstance Shadow State', () => {
     expect(instance.attributeSlots).toEqual(slots);
   });
 
+  it.each([false, true])('does not serialize attributes before materialization (hydrated=%s)', (isHydrated) => {
+    const instance = new BackgroundElementTemplateInstance('view', [{ id: 'before' }]);
+    if (isHydrated) {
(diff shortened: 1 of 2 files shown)


Yradex:
@coderabbitai full review

Yradex:
The initial Ubuntu Vitest failure is the existing QR-code interaction poll: `rspeedy/qrcode > custom schema object` expected two renders within 10 seconds and observed one. Its fixture only logs `Hello, world` and does not execute the ET runtime changed here. The plugin's [Vitest configuration](https://github.com/lynx-family/lynx-stack/blob/ef0c307233a7ab18844842fc5a5000348822972d/packages/rspeedy/plugin-qrcode/vitest.config.ts#L6) already documents this timing flake under CI load.

The [same Ubuntu check on the stacked child](https://github.com/lynx-family/lynx-stack/actions/runs/33842164981/job/100927463733), which includes this exact commit, passed. The [targeted rerun](https://github.com/lynx-family/lynx-stack/actions/runs/33842159171/job/100935924444) passed without source changes. The required `Done` and CLA checks are now green.

HuJean (approved):

PR #3774, opened 2026-09-04
perf(react): prune ET create checks for materialized subtrees

Yradex:
Moving an existing Element Template subtree recursively checks every descendant for creation even though its handles already exist. Each node only exits after the traversal reaches its individual create check.

Check whether the subtree root needs creation before descending. Newly inserted descendants are created at their own insertion boundary, while lifetime removal clears materialization throughout the subtree so reattachment still recreates children before parents. This removes one redundant traversal from moves; MTRef subtree collection, move payloads, and ref lifecycle ordering remain intact.

## Checklist

### DIFF
--- a/packages/react/runtime/__test__/element-template/runtime/background/instance.test.ts
+++ b/packages/react/runtime/__test__/element-template/runtime/background/instance.test.ts
@@ -883,6 +883,123 @@ describe('BackgroundElementTemplateInstance', () => {
       expect(ref).not.toHaveBeenCalled();
     });
 
+    it.each(['compiled', 'list'])('checks only the moved subtree root for creation under a %s parent', (parentKind) => {
+      __etAttrPlanMap._et_leaf = [0, adaptMTRefAttrSlot];
+      const root = new BackgroundPageRootInstance();
+      const parent = parentKind === 'list'
+        ? new BackgroundListElementTemplateInstance()
+        : new BackgroundElementTemplateInstance('_et_parent');
+      const anchor = new BackgroundElementTemplateInstance('_et_anchor');
+      const owner = new BackgroundElementTemplateInstance('_et_owner');
+      const nested = new BackgroundElementTemplateInstance('_et_nested');
+      const leaf = new BackgroundElementTemplateInstance('_et_leaf');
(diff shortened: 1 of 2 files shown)


Yradex:
@coderabbitai full review

Yradex:
The current performance report is not a comparison against this PR's stack base. It explicitly falls back to `main` (`ef0c307`) because no successful run for `wt/et-skip-unneeded-attr-diff` (`06a5703`) was available when the report was generated. The actual base of this PR is `06a5703`.

The only flagged regression is the transform benchmark `transform 1000 view elements` (11.9 ms → 13.7 ms). This PR changes only ET background instance creation and its tests; it does not change the transform benchmark or compiler. The report also warns that the WallTime runner environment is inconsistent. This isolated result against the fallback base is non-blocking for this PR; it does not establish a regression in the changed ET path.

Evidence: [current performance report](https://github.com/lynx-family/lynx-stack/pull/3774/checks?check_run_id=100929359080).

Yradex:
The initial CSR failure is the existing WebKit `api-dispose` timing test. It removes `lynx-view`, waits a fixed 50 ms, then synchronously expects the background cleanup log `fin`. This fixture uses the default Snapshot backend, with Element Template disabled, so it does not exercise the changed ET creation path.

The exact main baseline `ef0c307` already records the same `api-dispose` assertion failing in both WebKit and Chromium before passing on retry: [baseline CSR coverage job](https://github.com/lynx-family/lynx-stack/actions/runs/33816868546/job/100855069337). The [targeted rerun](https://github.com/lynx-family/lynx-stack/actions/runs/33842164981/job/100935959601) still failed the same WebKit assertion at line 1590, with no new failing test. This remains an admission blocker because the required [`Done` check](https://github.com/lynx-family/lynx-stack/actions/runs/33842164981/job/100937262443) is red. The PR stays Draft; no additional rerun or unrelated test change has been made.

The docstring coverage warning concerns the existing internal `emitMainThreadCreateIfNeeded` method. The change adds no public API, and its pruning invariant is documented beside the guard and covered by lifecycle tests. A docstring-only change is not needed for this diff.

HuJean (approved):

PR #3785, opened 2026-09-04
perf(react): cache element template ref plans

Yradex:
Element Template ref updates rescan every attribute adapter and repeat ref normalization even when a template's ref layout and the ref value are unchanged.

Cache an ordered ref-only plan by the compiled attribute plan's identity. The ref-effect pass visits only direct-ref and spread-ref slots, normalizes each ref once within that pass, and skips unchanged refs before allocating the selector token. Plan identity keeps registry replacement and teardown from reusing stale metadata. Ref ordering, hydration behavior, and the existing commit-time publication boundary remain unchanged.

This changes internal ET bookkeeping only, so no changeset is needed.

## Checklist

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved ref update handling to avoid unnecessary work when ref values remain unchanged.
  * Improved consistency when applying direct and spread ref updates, including clearing refs from empty or null-like values.
  * Ensured ref configurations are reused correctly across instances while honoring updated registry entries.
  * Added validation for invalid direct and spread ref values.

### DIFF
--- a/packages/react/runtime/__test__/element-template/runtime/prop-adapters/ref.test.ts
+++ b/packages/react/runtime/__test__/element-template/runtime/prop-adapters/ref.test.ts
@@ -1,5 +1,6 @@
 import { beforeEach, describe, expect, it, vi } from 'vitest';
 
+import { OrdinaryRefEffectQueue } from '../../../../src/core/ref.js';
 import { hydrationMap } from '../../../../src/element-template/hydration-map.js';
 import {
   clearRefState,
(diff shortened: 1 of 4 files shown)


Yradex:
@coderabbitai review

Yradex:
The CodSpeed regression report does not exercise the changed ET background ref path. The reported cases call [`executeTemplate`](https://github.com/lynx-family/lynx-stack/blob/4c9d273c4f85ff615d6b13851081bac687a5eaa7/packages/web-platform/web-core-e2e/bench/server.bench.vitest.spec.ts#L45) on Web first-screen bundles. Their [build configuration](https://github.com/lynx-family/lynx-stack/blob/4c9d273c4f85ff615d6b13851081bac687a5eaa7/packages/web-platform/web-core-e2e/tests/reactlynx/basic.config.ts#L20) uses the [default Snapshot backend](https://github.com/lynx-family/lynx-stack/blob/4c9d273c4f85ff615d6b13851081bac687a5eaa7/packages/rspeedy/plugin-react/src/pluginReactLynx.ts#L419), and the three reported fixtures contain no refs.

The report also falls back from the actual base `a15b0d9` to `5f54fb4` and warns about the WallTime runner environment. These measurements do not establish a regression caused by this PR. The specific source of the timing variation remains unverified; no timing improvement is claimed from this report either. The scoped regression tests verify plan reuse, ref ordering, normalization counts, and unchanged ref lifecycle behavior.

PR #3786, opened 2026-09-04
perf(react): reduce element template commit overhead

Yradex:
Element Template commits spend avoidable work deduplicating removed subtree roots and copying each create command's attribute slots. This groups two local optimizations in the existing commit and patch paths.

Use an insertion-ordered Set for pending teardown roots, then hand each delayed cleanup an independent array at dispatch. Deduplication no longer scans earlier roots, while cleanup order and timer ownership remain unchanged. The patch executor also reuses attribute arrays already normalized by the update event's JSON transport; MTRef slots still receive their separate native preparation when needed.

These are internal ET changes with no published API or protocol change, so no changeset is needed.

## Checklist

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved cleanup scheduling by preventing duplicate subtree teardown work and preserving independent cleanup batches.
  * Ensured sparse attribute updates are correctly normalized during transport and applied to native elements.
  * Updated runtime behavior to rely on transport-level attribute normalization without changing user-visible results.

* **Tests**
  * Expanded coverage for teardown batching, reset behavior, hydration, element removal, and sparse attribute updates.

### DIFF
--- a/packages/react/runtime/__test__/element-template/native/callDestroyLifetimeFun.test.ts
+++ b/packages/react/runtime/__test__/element-template/native/callDestroyLifetimeFun.test.ts
@@ -57,7 +57,7 @@ describe('callDestroyLifetimeFun', () => {
 
     callDestroyLifetimeFun();
 
-    expect(globalCommitContext.nonPayload.removedSubtreesAwaitingTeardown).toEqual([]);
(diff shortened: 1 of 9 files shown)


Yradex:
@coderabbitai review

HuJean (approved):

PR #3787, opened 2026-09-04
perf(react): combine element template removal walks

Yradex:
Removing a materialized Element Template subtree walks it separately to collect handles, clear materialization state, and queue ordinary ref cleanup. Removal inside a list item also walks the subtree to detect MTRefs.

Perform that work in one preorder traversal shared by ordinary and list-item removal. The walk fills the remove command's handle array, clears materialization state, queues ref cleanup, and checks for MTRefs when list notification needs it. Remove-command ordering, deferred ref publication, delayed manager cleanup, reattachment, and silent moves retain their existing behavior.

This is an internal ET lifecycle optimization with no published API or protocol change, so no changeset is needed.

## Checklist

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved cleanup when removing element subtrees, including nested and list-based content.
  * Ensured references are cleaned up in the correct order during removals.
  * Preserved accurate node tracking and update notifications when content is removed or re-added.

* **Tests**
  * Added coverage for multiple removal scenarios, including list descendants and missing references.
  * Expanded validation of cleanup behavior for sibling, child, and grandchild nodes.

### DIFF
--- a/packages/react/runtime/__test__/element-template/runtime/background/commit-context.test.ts
+++ b/packages/react/runtime/__test__/element-template/runtime/background/commit-context.test.ts
@@ -10,11 +10,7 @@ import {
   resetGlobalCommitContext,
   takeRemovedSubtreesForPostDispatchTeardown,
 } from '../../../../src/element-template/background/commit-context.js';
-import {
-  BackgroundElementTemplateInstance,
-  BUILTIN_RAW_TEXT_TEMPLATE_KEY,
-  collectElementTemplateSubtreeHandleIds,
-} from '../../../../src/element-template/background/instance.js';
+import { BackgroundElementTemplateInstance } from '../../../../src/element-template/background/instance.js';
 import { backgroundElementTemplateInstanceManager } from '../../../../src/element-template/background/manager.js';
 
 describe('ElementTemplate commit context', () => {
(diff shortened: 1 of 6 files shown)


Yradex:
@coderabbitai review

Yradex:
@coderabbitai review

Yradex:
The two reported `main-thread-transferRoot` changes are from the Snapshot first-screen benchmark, not the ET subtree-removal path changed here. The [benchmark configuration](https://github.com/lynx-family/lynx-stack/blob/326dbd7fab0a907f1cae08ff90bed8e7b777746a/benchmark/react/lynx.config.js#L101) leaves ET disabled, and the [`ReactLynx::transferRoot` profile](https://github.com/lynx-family/lynx-stack/blob/326dbd7fab0a907f1cae08ff90bed8e7b777746a/packages/react/runtime/src/snapshot/lifecycle/event/firstScreenSync.ts#L28) surrounds Snapshot's first-screen lifecycle event. This PR only changes ET background subtree removal.

The report uses fallback base `5f54fb4` because the actual base `a15b0d9` has no successful benchmark run, and also warns about the WallTime runner environment. It therefore does not establish a regression caused by this change; the specific source of the timing variation is unverified. This PR's performance evidence is the removal traversal-count test and its baseline negative control, alongside the lifecycle regression suite.

PR #3809, opened 2026-09-05
fix(react): provide the devtools lepus methods on the Element Template main thread

upupming:
Follow-up to #3808 (which makes the development build link); this makes the devtools work at runtime under Element Template.

## Problem

`@lynx-js/preact-devtools` maps each instance to the elements it owns by calling two lepus methods on the main thread, `getUniqueIdListBySnapshotId` and `getSnapshotIdByUniqueId`. The snapshot runtime injects both in `lynx.ts` whenever devtools are enabled; the Element Template runtime has its own main-thread bootstrap (`element-template/native`) and never did, so on a device every mapping call fails:

```
main-thread.js exception: TypeError: getUniqueIdListBySnapshotId is not a function
error_code:1101 sub_code:110100
```

`examples/react-et-lazy-bundle-standalone` in `rspeedy dev` shows ~30 of these on a single load (Pixel 4 XL, LynxExplorer built from lynx `develop` today). The devtools also derive the id they pass from `vnode.__e.__id`, which an Element Template instance did not expose.

## Fix

- Inject both methods on the Element Template main thread under the same `__DEV__ || __REACT_DEVTOOL__` gate as the snapshot runtime, resolving ids through the Element Template handle registry (`getElementTemplateTargetNativeRef` / a new `forEachElementTemplateNativeRef`).
- Expose `instanceId` as `__id` on `BackgroundElementTemplateInstance`, the name the devtools already use for a `SnapshotInstance`.

## Verification

- New unit tests cover the id → unique id mapping, the reverse lookup, null for unknown or missing ids, and `__id`.
- The native wiring test asserts the injection runs on the main thread only.
- `test:et`: 79 files, 806 tests pass.

### DIFF
--- a/.changeset/et-devtools-lepus-methods.md
+++ b/.changeset/et-devtools-lepus-methods.md
@@ -1,0 +1,5 @@
+---
+"@lynx-js/react": patch
+---
+
+Provide `getUniqueIdListBySnapshotId` and `getSnapshotIdByUniqueId` on the Element Template main thread when devtools are enabled, so `@lynx-js/preact-devtools` no longer fails with `getUniqueIdListBySnapshotId is not a function`.
--- a/packages/react/runtime/__test__/element-template/native/index.test.ts
+++ b/packages/react/runtime/__test__/element-template/native/index.test.ts
@@ -41,6 +41,7 @@ describe('element-template native index wiring', () => {
     globalThis.__ALOG_ELEMENT_API__ = true;
 
     const injectCalledByNative = vi.fn();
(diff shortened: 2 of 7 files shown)


upupming:
Closing: this bolts the snapshot runtime's id bridge onto Element Template, which has no snapshot concept. The devtools should tell the runtime backends apart (the runtime already publishes `__REACT_LYNX_RUNTIME_BACKEND__`) rather than the Element Template runtime imitating a `SnapshotInstance`. Follow-up goes to `@lynx-js/preact-devtools`.

PR #3893, opened 2026-09-10
fix(react): schedule Element Template effects as microtasks

Yradex:
Element Template leaves Preact's passive-effect scheduler on its frame/timer fallback, delaying `useEffect` callbacks and ordinary unmount cleanups compared with Snapshot.

Configure the scheduler during Element Template background initialization to use the existing `lynxQueueMicrotask` helper. Passive effects and ordinary unmount cleanups remain asynchronous, while page destruction continues to flush cleanups synchronously and restores the scheduler afterward. Regression tests render through the public Element Template hooks and drain only microtasks, so the frame/timer fallback cannot hide missing initialization.

## Checklist

No changeset is required for this unpublished Element Template implementation.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved background rendering so passive effects and unmount cleanup are processed promptly after microtasks complete.
  - Ensured cleanup runs synchronously when background content is destroyed, preventing delayed teardown behavior.
  - Preserved existing animation-frame scheduling behavior while improving effect handling.

### DIFF
--- a/packages/react/runtime/__test__/element-template/runtime/background/root-render.test.tsx
+++ b/packages/react/runtime/__test__/element-template/runtime/background/root-render.test.tsx
@@ -1,3 +1,4 @@
+import { options } from 'preact';
 import { beforeEach, describe, expect, it, vi } from 'vitest';
 
 import { resetElementTemplateCommitState } from '../../../../src/element-template/background/commit-hook.js';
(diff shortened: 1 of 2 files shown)


Yradex:
@coderabbitai review

HuJean (approved):

PR #3901, opened 2026-09-10
fix(rsbuild-plugin): respect selected environments when printing URLs

icecreamx10:
## Summary

Fix dev server startup when a configuration declares both `web` and `lynx`, but the CLI selects only one with `--environment`.

The URL printer previously captured all environments in `modifyRsbuildConfig`, before Rsbuild applied environment selection. At print time, it requested the normalized config for an excluded environment and threw `Cannot find normalized config by environment: web` (or `lynx`).

- Read the final normalized environments and merged entries at URL-print time.
- Cover both `--environment lynx` and `--environment web`, including shared and environment-specific entries and web preview URLs.
- Include a patch changeset for `@lynx-js/rsbuild-plugin`.

## Validation

- Baseline source with new regression tests: 2 failures, reproducing the missing normalized environment error; 56 existing tests pass.
- Fixed source: all 171 plugin tests across 16 files pass.
- Root TypeScript build through Turbo, package typecheck, targeted ESLint, Biome and dprint checks pass.
- Full repository build was attempted but blocked by `cargo: command not found` on this machine; no full native/e2e build claim.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed development server startup failures when using `--environment` to select specific environments.
  * Development server URLs now reflect the active environment and its resolved entry points.
  * Prevented URLs from displaying entries belonging to excluded environments.

### DIFF
--- a/packages/rspeedy/plugin-lynx/src/plugins/dev.plugin.ts
+++ b/packages/rspeedy/plugin-lynx/src/plugins/dev.plugin.ts
@@ -155,19 +155,6 @@ export function pluginDev(): RsbuildPlugin {
           config.server?.printUrls === undefined
           || config.server?.printUrls === true
         ) {
-          // A root entry is merged into every environment (the way Rsbuild
-          // itself resolves entries), not replaced by an environment's own.
-          const entriesByEnvironment = Object.entries(
-            config.environments ?? {},
-          ).map(([environmentName, environmentConfig]) =>
-            [
-              environmentName,
-              Object.keys({
-                ...config.source?.entry,
-                ...environmentConfig.source?.entry,
-              }),
-            ] as const
-          )
           return mergeRsbuildConfig(config, {
             server: {
               printUrls: (param) => {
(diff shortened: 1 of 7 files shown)


packages/rspeedy/plugin-lynx/src/plugins/dev.plugin.ts
```diff
@@ -155,19 +155,6 @@ export function pluginDev(): RsbuildPlugin {
           config.server?.printUrls === undefined
           || config.server?.printUrls === true
         ) {
-          // A root entry is merged into every environment (the way Rsbuild
-          // itself resolves entries), not replaced by an environment's own.
-          const entriesByEnvironment = Object.entries(
```
icecreamx10, line 160: 这里为啥删了

packages/rspeedy/plugin-lynx/test/dev.plugin.test.ts
```diff
@@ -1280,6 +1280,58 @@ describe('pluginDev', () => {
     })
   })
 
+  test.each(['lynx', 'web'])(
```
upupming, line 1283: We should also add a test for rspeedy in packages/rspeedy/core/test/plugins/dev.plugin.test.ts to ensure the same behavior of rspeedy and rsbuild.

icecreamx10: Added in fc987dd. The Rspeedy regression tests start the actual dev server with only lynx or web selected and assert the complete printed URL list (including the web preview URL). The dev plugin suite passes: 38 passed, 1 existing skipped. Rspeedy typecheck and targeted ESLint also pass.

packages/rspeedy/plugin-lynx/src/plugins/dev.plugin.ts
icecreamx10: This removes the pre-filter environment snapshot, not root-entry merging. The URL printer now reads getNormalizedConfig().environments at print time, after environment selection; Rsbuild has already merged root and environment entries there. Regression tests cover both selected environments and shared/environment-specific entries.

upupming (approved):

PR #3919, opened 2026-09-11
fix(react): preserve background component owner stacks

Yradex:
Element Template did not install the shared component-stack hooks in its background runtime, so development errors during attribute updates lacked component context. Enable these hooks for development background rendering.

Scheduled `setState` updates also clone a vnode before entering its render scope. Re-running the vnode hook overwrote its captured owner with `null`, losing the ancestor chain in both backends. Preserve existing owners, including explicit `null`, and initialize ownership only for new vnodes. A circular attribute update in a child component can now report both the child and its creating parent, while excluding components that only pass through `children`.

The regression covers a compiled Element Template tree through first-screen hydration, a successful scheduled update, and a circular attribute update. The existing Snapshot case now covers both root-driven and scheduled rendering. Main-thread owner tracking and source locations remain outside this change.

## Checklist

## Summary by CodeRabbit

- **Bug Fixes**
  - Preserved component ownership details in development error stacks after scheduled state updates.
  - Enabled component stack reporting for Element Template background rendering, including errors caused by circular attribute values.
- **Tests**
  - Added coverage for component ownership in background rendering errors.
  - Expanded circular-reference error tests to include scheduled rendering scenarios.

### DIFF
--- a/.changeset/preserve-component-stack-owners.md
+++ b/.changeset/preserve-component-stack-owners.md
@@ -1,0 +1,5 @@
+---
+"@lynx-js/react": patch
+---
+
+Preserve component owners in development error stacks after scheduled state updates, and enable background component stacks for Element Template.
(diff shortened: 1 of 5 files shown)


Yradex:
@coderabbitai review

HuJean (approved):

PR #3938, opened 2026-09-14
docs: add a generated API reference site

upupming:
Adds `docs/`, an API reference site generated at build time from the TSDoc of every public package with `@rspress/plugin-typedoc`, and publishes the pages as `@lynx-js/lynx-stack-docs` so lynxjs.org can sync them. The generated pages are not committed; only `docs/i18n/zh.json` is.

- `/api/react/` and `/api/genui/` are their own sections; every other package is under `/api/packages/`. Package introductions come from `@packageDocumentation`, ReactLynx groups from `@group`.
- `/api/config/` is generated from the Rspeedy `Config` and the `pluginLynx` options. Options with a changed default are tagged `@lynxDefaultChanged`.
- The build fails on any TypeDoc warning, on a dead link and on a string the Chinese dictionary does not translate.
- `docs/shown-packages.json` lists the packages lynxjs.org shows in its navigation.
- `docs/plugins/site.ts` holds what the reference takes from this repository — where its source and its packages are browsed, which packages get a section of their own, how the rest are grouped — so another repository can reuse the plugins with its own.

## What the sidebar lists

- A package lists what you call - classes, functions, variables, namespaces. The types they refer to have pages too, reached from the pages that use them, so a package of 40 exports is not 40 sidebar entries.
- A group lists its classes first, then interfaces, type aliases, functions and variables, the way the kind directories are ordered.
- The configuration options are grouped like Rsbuild (`Dev options`, `Output options`, ...), and a nested option gets its own directory, so the sidebar reads `template` instead of `output.filename.template`.

## Two things that used to pass silently

- [x] An unknown block tag of a dependency became a section of its own. `@lynx-js/type-config` marks each option with `@Android`, `@iOS`, `@Harmony` or `@Clay*`, which rendered as ~90 headings on the `config-rsbuild-plugin` page, half of them empty. The tags are dropped now and the `Since: LynxSDK x.y` line they carry stays in the description.
- [x] A Chinese page that could not look a string up kept the English text, and no check could see it: `i18n:check` only looks for empty values. Entries are keyed by the prose without its link targets now - a localized page spells a link its own way, with its own locale and its own directory names - and the build fails on any string the dictionary does not translate.

`@lynx-js/lynx-stack-docs` is a new package: `check-new-packages-published` fails until a `0.0.0-rc.0` placeholder (and `-canary`) is published and `npm trust github` is set up.

## Summary by CodeRabbit

* **New Features**
  * Added a Lynx Stack API reference site with English and Chinese documentation.
  * Added API references for React, GenUI, testing utilities, configuration options, and workspace packages.
  * Added localized navigation, package links, source links, and documentation manifests.
* **Documentation**
  * Added guidance for generating, translating, building, and publishing API documentation.
  * Expanded ReactLynx documentation covering directives, global events, import attributes, macros, and build-plugin usage.
* **Bug Fixes**
  * Improved generated documentation links and MDX rendering.
* **Tests**
  * Added validation to ensure generated translations remain synchronized.

## Checklist


upupming:
@coderabbitai review

AGENTS.md
```diff
@@ -82,11 +82,23 @@ pnpm run test --update
 
 ### 5. API Documentation
 
+The API reference in `docs/` (lynx-stack.dev, and the pages synced to lynxjs.org) is generated from the exports and TSDoc comments of the public packages. When you change a public API — an export, a type, an option, a default value or its TSDoc — update the documentation in the same change:
```
upupming: I think lynx-stack.dev will still be website, and we will not change the website in a long time untile `docs/` is working file and stable.

